PR #370, opened 2014-11-08
Added support for reserved variable expansion in URI templates

elishacook:
Moving this to a topic branch.


falcon/api.py
```diff
@@ -728,14 +728,22 @@ def _compile_uri_template(template):
         if template != '/' and template.endswith('/'):
             template = template[:-1]
 
-        expression_pattern = r'{([a-zA-Z][a-zA-Z_]*)}'
+        PCT_ENCODED = '%[0-9A-Fa-f]{2}'
+        UNRESERVED = '[^/]'
+        RESERVED = r'[:/\?#\[\]@\!\$&\'\(\)\*\+,;=]'
+        BASIC_VARIABLE = '{([a-zA-Z][a-zA-Z_]+)}'
+        RESERVED_VARIABLE = r'{\+([a-zA-Z][a-zA-Z_]+)}'
```
BenjamenMeyer: Please change the ```[a-zA-Z_]``` on L734-735 to ```\w``` as that allows 0-9 in the remainder of the variable names.
I did that in https://github.com/falconry/falcon/pull/454 for ```expression_pattern``` which you are replacing.

mcmasterathl:
It would be good to have some docs to go along with this, I can't tell just form looking what this is supposed to do.

kgriffs:
With the new routing engine, we are moving away from using RFC 6570 for the URI template schema, so  reserved variable expansion no longer applies.

PR #389, opened 2014-11-24
current API only allows one class to implement web methods according

litong01:
to its method names, this is limitting in some cases. With this new
API, a developer can use one class to implement a set of APIs.
The developer can simply add an instance of that class to the app
by calling add_route method. An example of how to do this has been
documented in the ResourceAPI class in resource_api.py module.

### DIFF
--- a/falcon/resource_api.py
+++ b/falcon/resource_api.py
@@ -1,0 +1,104 @@
+# Copyright 2013 IBM Corp
+#
+# Author: Tong Li <litong01@us.ibm.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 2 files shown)


kgriffs:
Apologies for taking so long to get back to you; I've been trying to decide what approach to take. In 0.3 we made it easier to plug in custom routers, so one approach might be to refactor this router to use  that [new interface][1]. I've also proposed an alternative strategy that may work for some use cases and I'd be happy to get your feedback on that as well (see issue #584).

[1]: https://falcon.readthedocs.org/en/stable/api/routing.html#routing

kgriffs:
All things considered, I think it would be better to implement this in an add-on library for now.

If you would like to map to arbitrary callables, while still taking advantage of the default `CompiledRouter`, you could to inherit from `falcon.API` and create a new method similar to `add_route()`, but one that takes a single responder callable and an HTTP method in lieu of a resource class. It would update a method map attribute (i.e., `self._method_maps`), keyed by URI template. For example, after calling the new `Api.my_add_route()` twice for the same URI template, you would have something like:

```python
{
    '/some_uri_template': {
        'GET': some_get_responder,
        'PUT': some_put_responder,
    }
}
```

Then you would pass that into the router's `add_route()` method with a dummy object (or `None`) for the resource, e.g.:

```python
self._router.add_route(uri_template, self._method_maps[uri_template], None)
```

Note that subsequent calls to `self._router.add_route()` will override the previous method map for a given `uri_template`. Anyway, this approach should work, but caveat emptor since this is just off the top of my head.

Issue #400, opened 2015-01-05
Middleware's process_request only executes before routing


rsyed83:
The behavior of the current "before" hook provides a handle to the "resource", the Handler object that is processing the request.

The way middleware is currently implemented (where process_request occurs only before routing) means that some before hook behavior cannot be translated into middleware.

There should be a change to the way middleware is executed that enables a middleware implementer to optionally signify that process_request should occur after routing has occurred, with access to the "resource" object.

kgriffs:
OK, so the two cases where people care whether process_request is called before or after the route is calculated are these:

1. I want to muck with the URI or HTTP method (rewrite rules) 
2. I need access to the resource instance that will be handling the request

Are there any other cases you can think of?

Something tells me (2) is going to be more common, since many people will likely do rewriting in their web server rather than in their app. Also, having access to the resource is a richer affordance, so people will naturally use it more.

What do you think? Assuming the above make sense, we have a couple of options. First, we could simply not support (1). Or we could support it but enable (2) as the default.

rsyed83:
Nothing comes to mind aside from these use cases for now at least.  

Do you know who was the original requester for middleware to occur before routing?  Before we change the default behavior, it might be important to understand the class of users who would be impacted by a behavior change?

Perhaps the usecase was a common enough scenario that could NOT be handled with webserver URI rewriting.

That being said, I personally don't have an objection to making (2) the default behavior, as you said access to the resource is a richer affordance.

rsyed83:
Another update, in working on tests for my proposed change, I realized that specifying middleware execution before/after routing results in process_request getting executed out-of-order as defined by the initialization in the API class's constructor.

That's not very clean in terms of understanding execution flow, right...

kgriffs:
The original patch was authored by @ealogar. Eduardo, can you comment?

ealogar:
IMO a process_request mw should not be coupled with any resource. Middlewares apply to request/response and the logic of what to do after routing and having a resource defined for a particular method is some kind of logic that vary on every resource.
This is exactly what I didn't like of current before/after, They behave as a global decorator for resources but what happen when routing not success (e.g. 404 not found, 405 invalid method)... The middleware is not executed ... So, I can not use before and after for all of my middlewares.

All the middlewares I use are of that kind:
- authentication
- tracing requests
- transactions id 
- Session mw (to define global parameters for all resources)

[...]

PR #401, opened 2015-01-05
Permit middleware process_request to occur after routing

rsyed83:
Change middleware behavior so that the presence of a middleware object field would enable process_request to happen after routing has occurred, while also providing access to the resource object.

Issue #400


rsyed83:
Closing until I address tests

PR #405, opened 2015-01-07
feat(Response) Add cookie helpers

### LINKED ISSUE #292: Add helpers for getting and setting cookies
Note: Cookies require special handling in responses because they must be set as distinct headers (cannot be folded using commas as you would other headers).

tbug:
Add the following new helpers for setting and getting cookies

- `set_cookie(name : str, value : str, **cookie_params) -> None`
- `get_cookie(name : str) -> Morsel`
- ` get_cookies() -> List[Morsel]`
- `unset_cookie(name : str) -> None`

Cookies will - if set - be rendered and merged with headers in
_wsgi_headers. For responses that don't use cookies, there will
no performance hit, except for a single "if cookies is not None:"
check in the _wsgi_headers call.
Implementation use the http.cookies.SimpleCookie (py3) or
Cookie.SimpleCookie (py2) from the stdlib.
Might find extra speed in doing our own implementation.

Fixes #292

### DIFF
--- a/doc/api/cookies.rst
+++ b/doc/api/cookies.rst
@@ -1,0 +1,85 @@
+
+.. _cookies:
+Cookies
+-------------
+
+Cookie support has been added to falcon from version 0.3.
+
+.. _getting-cookies:
+Getting Cookies
+~~~~~~~~~~~~~
+
+Cookie can be read from a request via the :py:attr:`~.Request.cookies` request attribute:
+
+
+.. code:: python
+
+    class Resource(object):
+        def on_get(self, req, resp):
+
+            cookies = req.cookies
+
+            if "my_cookie" in cookies:
+                my_cookie_value = cookies["my_cookie"]
+            # ....
+
+The :py:attr:`~.Request.cookies` attribute is a regular
+:py:class:`dict` object.
+
+.. tip :: :py:attr:`~.Request.cookies` returns a
+    copy of the response cookie dict. Assign it to a variable as in the above example
+    for better performance.
+
+.. _setting-cookies:
+Setting Cookies 
+~~~~~~~~~~~~~~~
+
+Setting cookies on a response is done via the :py:meth:`~.Response.set_cookie`.
+
+You should use :py:meth:`~.Response.set_cookie` instead of
+:py:meth:`~.Response.set_header` or :py:meth:`~.Response.append_header`.
+
+With :py:meth:`~.Response.set_header` you cannot set multiple headers
+with the same name (which is how multiple cookies are sent to the client).
+
+:py:meth:`~.Response.append_header` appends multiple values to the same
+header field, which is not compatible with the format used by `Set-Cookie`
+headers to send cookies to clients.
+
+
+
+Simple example:
+
+.. code:: python
+
+    class Resource(object):
+        def on_get(self, req, resp):
+            # Set the cookie "my_cookie" to the value "my cookie value"
+            resp.set_cookie("my_cookie", "my cookie value")
+
+
+You can of course also set the domain, path and lifetime of the cookie.
+
+
+
+
+.. code:: python
+
+    class Resource(object):
+        def on_get(self, req, resp):
+            # Set the 'max-age' of the cookie to 10 minutes (600 seconds)
+            # and the cookies domain to "example.com"
+            resp.set_cookie("my_cookie", "my cookie value",
+                            max_age=600, domain="example.com")
+
+
+If you set a cookie and want to get rid of it again, you can
+use the :py:meth:`~.Response.unset_cookie`:
+
+.. code:: python
+
+    class Resource(object):
+        def on_get(self, req, resp):
+            resp.set_cookie("bad_cookie", ":(")
+            # clear the bad cookie
+            resp.unset_cookie("bad_cookie")
--- a/doc/user/tutorial.rst
+++ b/doc/user/tutorial.rst
@@ -465,7 +465,6 @@ Query Strings
 
 *Coming soon...*
 
-
 Introducing Hooks
 -----------------
 
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1,5 +1,3 @@
-# Copyright 2013 by Rackspace Hosting, Inc.
-#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -33,6 +31,8 @@ from falcon.errors import *
 from falcon import util
 from falcon.util import uri
 from falcon import request_helpers as helpers
+
+from six.moves.http_cookies import SimpleCookie
 
 
 DEFAULT_ERROR_LOG_FORMAT = (u'{0:%Y-%m-%d %H:%M:%S} [FALCON] [ERROR]'
@@ -167,6 +167,11 @@ class Request(object):
             all the values in the order seen.
 
         options (dict): Set of global options passed from the API handler.
+
+        cookies (dict):
+            A dict of name/value cookie pairs.
+            See also: :ref:`Getting Cookies <getting-cookies>`
+
     """
 
     __slots__ = (
@@ -183,6 +188,7 @@ class Request(object):
         'context',
         '_wsgierrors',
         'options',
+        '_cookies',
     )
 
     # Allow child classes to override this
@@ -233,6 +239,8 @@ class Request(object):
 
         else:
             self.query_string = six.text_type()
+
+        self._cookies = None
 
         self._cached_headers = None
         self._cached_uri = None
@@ -477,6 +485,21 @@ class Request(object):
     @property
     def params(self):
         return self._params
+
+    @property
+    def cookies(self):
+        if self._cookies is None:
+            # NOTE(tbug): We might want to look into parsing
+            # cookies ourselves. The SimpleCookie is doing a
+            # lot if stuff only required to SEND cookies.
+            parser = SimpleCookie(self.get_header("Cookie"))
+            cookies = {}
+            for morsel in parser.values():
+                cookies[morsel.key] = morsel.value
+
+            self._cookies = cookies
+
+        return self._cookies.copy()
 
     # ------------------------------------------------------------------------
     # Methods
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -14,7 +14,8 @@
 
 import six
 
-from falcon.response_helpers import header_property, format_range
+from falcon.response_helpers import header_property, format_range,\
+    is_ascii_encodable
 from falcon.util import dt_to_http, uri, TimezoneGMT
 from six.moves.http_cookies import SimpleCookie, CookieError
 
@@ -193,15 +194,14 @@ class Response(object):
 
         """
 
+        if not is_ascii_encodable(name):
+            raise KeyError('"name" is not ascii encodable')
+        if not is_ascii_encodable(value):
+            raise ValueError('"value" is not ascii encodable')
+
         if six.PY2:  # pragma: no cover
-            # NOTE(tbug): In python 2 we want a str type name,
-            # as unicode will make the SimpleCookie implementation
-            # blow up.
-            if isinstance(name, unicode):
-                try:
-                    name = name.encode("ascii")
-                except UnicodeEncodeError as e:
-                    raise KeyError("'name' must consist of only ascii")
+            name = str(name)
+            value = str(value)
 
         if self._cookies is None:
             self._cookies = SimpleCookie()
@@ -209,6 +209,9 @@ class Response(object):
         try:
             self._cookies[name] = value
         except CookieError as e:  # pragma: no cover
+            # NOTE(tbug): we raise a KeyError here, to avoid leaking
+            # the CookieError to the user. SimpleCookie (well, BaseCookie)
+            # only throws CookieError on issues with the cookie key
             raise KeyError(str(e))
 
         if expires:
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -146,7 +146,7 @@ class Response(object):
         self.stream_len = stream_len
 
     def set_cookie(self, name, value, expires=None, max_age=None,
-                   domain=None, path=None, secure=True, httponly=True):
+                   domain=None, path=None, secure=True, http_only=True):
         """Set a response cookie.
 
         Note:
@@ -177,8 +177,8 @@ class Response(object):
                 server whenever it sends back this cookie.
                 Warning: You will also need to enforce HTTPS for the cookies
                 to be transfered securely.
-            httponly (bool) (default: True):
-                The attribute httponly specifies that the cookie
+            http_only (bool) (default: True):
+                The attribute http_only specifies that the cookie
                 is  only transferred in HTTP requests, and is not accessible
                 through JavaScript. This is intended to mitigate some forms
                 of cross-site scripting.
@@ -241,8 +241,8 @@ class Response(object):
         if secure:
             self._cookies[name]["secure"] = secure
 
-        if httponly:
-            self._cookies[name]["httponly"] = httponly
+        if http_only:
+            self._cookies[name]["httponly"] = http_only
 
     def unset_cookie(self, name):
         """Unset a cookie from the response
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -15,7 +15,10 @@
 import six
 
 from falcon.response_helpers import header_property, format_range
-from falcon.util import dt_to_http, uri
+from falcon.util import dt_to_http, uri, TimezoneGMT
+from six.moves.http_cookies import SimpleCookie, CookieError
+
+GMT_TIMEZONE = TimezoneGMT()
 
 
 class Response(object):
@@ -79,6 +82,7 @@ class Response(object):
         '_body_encoded',  # Stuff
         'data',
         '_headers',
+        '_cookies',
         'status',
         'stream',
         'stream_len'
@@ -87,6 +91,10 @@ class Response(object):
     def __init__(self):
         self.status = '200 OK'
         self._headers = {}
+
+        # NOTE(tbug): will be set to a SimpleCookie object
+        # when cookie is set via set_cookie
+        self._cookies = None
 
         self._body = None
         self._body_encoded = None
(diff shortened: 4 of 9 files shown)


tbug:
ping @kgriffs 

Would be cool if I could get a reject or merge on this one.

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
         self.stream = stream
         self.stream_len = stream_len
 
+    def set_cookie(self, key, value, **cookie_params):
+        """Set a response cookie.
+
+        Args:
+            key (str): Cookie key name
```
kgriffs: Including "key" in the description seems like it is exposing an implementation detail. What do you think about just saying "Cookie name"?

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
...
+            expires (str): Defines when the cookie should expire.
+                the default is to expire when the browser is closed.
+                The format given should be "Wdy, DD Mon YYYY HH:MM:SS GMT"
+            secure (bool): The Secure attribute directs the user
+                agent to use only secure means to contact the origin
+                server whenever it sends back this cookie.
+            version (str): The Version attribute, a decimal integer,
+                identifies to which version of the state management
+                specification the cookie conforms.
+            httponly (bool): The attribute httponly specifies that the cookie
```
kgriffs: What do you think about making this default to `True` in the interest of being "secure by default"?

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
...
+                An explicitly specified domain must always start with a dot.
+            max_age (int): The Max-Age attribute defines the lifetime of the
+                cookie, in seconds.  The delta-seconds value is a decimal non-
+                negative integer.  After delta-seconds seconds elapse,
+                the client should discard the cookie.
+                A value of 0 means the cookie should be discarded immediately.
+            expires (str): Defines when the cookie should expire.
+                the default is to expire when the browser is closed.
+                The format given should be "Wdy, DD Mon YYYY HH:MM:SS GMT"
+            secure (bool): The Secure attribute directs the user
```
kgriffs: Wondering if this should default to `True` as well (see comment below). Seems like the kinds of things Falcon is used for (apps, APIs) are usually running over HTTPS...

falcon/response.py
tbug: http://tools.ietf.org/html/rfc6265 refers to it as "cookie name" so i agree that would be the right naming.

falcon/response.py
tbug: I agree that would be the saner default.
Argument against it, is that everywhere else HTTP is default and HTTPS is opt-in.

falcon/response.py
tbug: Better.
Argument against is that - together with the `secure` flag - it is usually an opt-in.

I guess taking a stance and saying "secure by default" is better than "everyone else does it", so I agree.

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
...
+                session with this cookie.
+            domain (str): The Domain attribute specifies the domain for
+                which the cookie is valid.
+                An explicitly specified domain must always start with a dot.
+            max_age (int): The Max-Age attribute defines the lifetime of the
+                cookie, in seconds.  The delta-seconds value is a decimal non-
+                negative integer.  After delta-seconds seconds elapse,
+                the client should discard the cookie.
+                A value of 0 means the cookie should be discarded immediately.
+            expires (str): Defines when the cookie should expire.
```
kgriffs: Would it be easier for the caller to specify a datetime object rather than a string?

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
...
+            version (str): The Version attribute, a decimal integer,
+                identifies to which version of the state management
+                specification the cookie conforms.
+            httponly (bool): The attribute httponly specifies that the cookie
+                is  only transferred in HTTP requests, and is not accessible
+                through JavaScript. This is intended to mitigate some forms
+                of cross-site scripting.
+
+        Note:
+            Kwargs and there valid values are all
```
kgriffs: Typo: "there" ==> "their"

falcon/response.py
tbug: Or a timedelta. Yes, much easier.
The cookie arguments could use some work.
They are all well-define, so taking them as kwargs is also a bit sloppy.
I'll fix that.

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
...
+    def get_cookies(self):
+        """Return all cookies as a list of Morsel objects
+
+        Note:
+            You cannot set response cookies by changing the returned list.
+            Use response.set_cookie() for that instead.
+        """
+
+        # NOTE(tbug): Not sure i like this function, but we _do_ need
+        # a way to get all cookies, right?
```
kgriffs: Off the top of my head I can't think of a great use for this, but I could be missing something. `Request` does not currently implement a `get_headers` method. Nobody has asked for it (yet), but then again maybe people are just accessing `req._headers` directly and not telling anybody. :stuck_out_tongue:

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
...
+        return list(self._cookies.values())
+
+    def unset_cookie(self, name):
+        """Unset a cookie from the response
+
+        """
+        if self._cookies is not None:
```
kgriffs: There's a subtle inconsistency here. If no cookies were ever set, and I try to delete one, it's a noop. However, if one ore more cookies were set, but I try to unset a different cookie that doesn't exist, L231 will raise `KeyError`.

falcon/response.py
tbug: I'll leave it out. It leaks implementation in form of https://docs.python.org/2/library/cookie.html#Cookie.Morsel objects as return values.

falcon/response.py
```diff
@@ -417,6 +524,12 @@ def _wsgi_headers(self, media_type=None):
         if six.PY2:  # pragma: no cover
             # PERF(kgriffs): Don't create an extra list object if
             # it isn't needed.
-            return headers.items()
+            items = headers.items()
+        else:
+            items = list(headers.items())  # pragma: no cover
+
+        if self._cookies is not None:
+            items += [("set-cookie", c.OutputString())
```
kgriffs: Might be worth adding a comment here about how this approach executes more quickly than `self._cookies.output().split('\r\n')`.

falcon/response.py
```diff
@@ -17,6 +17,11 @@
 from falcon.response_helpers import header_property, format_range
 from falcon.util import dt_to_http, uri
 
+if six.PY2:  # pragma: no cover
```
kgriffs: Not a big deal, but you could save a few lines here by using six:

```python
from six.moves.http_cookies import SimpleCookie
```

tbug: Nice, didn't know that. Will do.

falcon/response.py
kgriffs: OK, maybe we can get some others from the community to chime in. It seems like the target audience for Falcon will usually want to enable `httponly` and `secure`, but that's just a guess.

falcon/response.py
```diff
@@ -123,6 +133,103 @@ def set_stream(self, stream, stream_len):
...
+                identifies to which version of the state management
+                specification the cookie conforms.
+            httponly (bool): The attribute httponly specifies that the cookie
+                is  only transferred in HTTP requests, and is not accessible
+                through JavaScript. This is intended to mitigate some forms
+                of cross-site scripting.
+
+        Note:
+            Kwargs and there valid values are all
+            specified in http://tools.ietf.org/html/rfc2109.html
```
kgriffs: Just noticed this is an old RFC... 2109 was superseded by 2965, which in turn has been superseded by http://tools.ietf.org/html/rfc6265.html

tbug: Yup. Already fixed locally. Will be in the next push.
Also removed the `version` and `comment` as they are no longer part of the RFC.

kgriffs:
P.S. - I just noticed that "Fixes #292" is referenced in the commit message. I realize now that the issue description wasn't very clear... Before closing #292 we also need helpers for reading cookies from the `Request` object. It could be as easy as adding a read-only `cookies` property that passes the value of
**HTTP_COOKIE** (if present) into the `SimpleCookie` initializer, and returning the instantiated object.

tbug:
Right. As cookies are sent purely by `name=value` I think the better solution would be to convert it into a dict and return that, to avoid exposing `SimpleCookie` objects.

I'll get hacking.

tbug:
That's the request part.
- I had to add a GMT implementation of `tzinfo` to avoid adding pytz as a dependency. We need  this if we want to accept a `datetime` object as argument to the `set_cookie` `expires` parameter.

sebasmagri:
The only thing I'm missing is a mention of this in the docs. Anyway, given the API docs are not yet ready, I wouldn't consider this a blocker.

LGTM.

:+1:

tbug:
Cool. Should i have a crack at the docs in this PR, or keep that seperate?

kgriffs:
@tbug I'd say include it in this PR so that if in writing the docs you notice some parts of the implementation that could use a little more polish (not sure about you, but when I put myself in the mindset of the "user" while writing docs, I often discover ways to improve the design). 

BTW, I'm heads-down trying to get 0.2.0rc1 out, but as soon as that's done we can open 0.3 development and then I will circle back to do a full review of your PR. Thanks!

tbug:
So, most of my time was spent learning the doc syntax (first time using sphinx), but I finally got around to finish this.

I had no idea where to place a section on cookies, so I placed it just under the "Query String" section in the tutorial.

(btw, Would have been nice with a few lines in the `CONTRIBUTING.md` file, about how to get started with the docs)

Could Someone:tm: have a look at the above commit and let me know if I need to add something?

doc/user/tutorial.rst
```diff
@@ -466,6 +466,76 @@ Query Strings
 *Coming soon...*
 
 
+Cookies
```
kgriffs: This looks more like reference material than something that would belong in the tutorial... Could you move this documentation to a new file, `doc/api/cookies.rst`? You can then refer to it from the relevant Request and Response docstrings.

falcon/request.py
```diff
@@ -478,6 +486,19 @@ def headers(self):
...
+    def cookies(self):
+        if self._cookies is None:
+            # NOTE(tbug): We might want to look into parsing
+            # cookies ourselves. The SimpleCookie is doing a
+            # lot if stuff only required to SEND cookies.
+            parser = SimpleCookie(self.get_header("Cookie"))
+            cookies = {}
+            for morsel in parser.values():
+                cookies[morsel.key] = morsel.value
+            self._cookies = cookies
```
kgriffs: Nitpick: Add a blank line above L499 to aid in readability

falcon/request.py
```diff
@@ -478,6 +486,19 @@ def headers(self):
...
+        return self._cookies.copy()
```
kgriffs: Nitpick: Add a blank line above L500 to aid in readability

falcon/response.py
```diff
@@ -136,6 +145,82 @@ def set_stream(self, stream, stream_len):
...
+
+        """
+
+        if self._cookies is None:
+            self._cookies = SimpleCookie()
+
+        self._cookies[name] = value
+
+        if expires:
+            # set Expires header. Format is Wdy, DD Mon YYYY HH:MM:SS GMT
```
kgriffs: This isn't actually setting an "expires" header, right? It's an attribute of the cookie...

falcon/response.py
```diff
@@ -136,6 +145,82 @@ def set_stream(self, stream, stream_len):
         self.stream = stream
         self.stream_len = stream_len
 
+    def set_cookie(self, name, value, expires=None, max_age=None,
+                   domain=None, path=None, secure=True, httponly=True):
+        """Set a response cookie.
+
```
kgriffs: What do you think about adding a note here talking about why this method should be used in lieu of set_header or append_header?

falcon/response.py
```diff
@@ -136,6 +145,82 @@ def set_stream(self, stream, stream_len):
...
+                cookie, in seconds.  The delta-seconds value is a decimal non-
+                negative integer.  After delta-seconds seconds elapse,
+                the client should discard the cookie.
+            domain (str): The Domain attribute specifies the domain for
+                which the cookie is valid.
+                An explicitly specified domain must always start with a dot.
+                A value of 0 means the cookie should be discarded immediately.
+            path (str): The Path attribute specifies the subset of URLs to
+                which this cookie applies.
+            secure (bool) (default: True): The Secure attribute directs the
```
kgriffs: Might be worth adding a warning here that HTTPS should be forced when using secure cookies, using a combination of HSTS, web server config, and checking at the app layer. Falcon could check for HTTPS and refuse to return a secure cookie if the transport is not secure, but then we'd have to provide a way for people to turn that off in dev environments since you typically don't bother to set up TLS in that case. Probably not worth it right now, but something to consider for the future...

See also: #460.

falcon/response_helpers.py
```diff
@@ -60,3 +62,18 @@ def format_range(value):
             str(value[0]) + '-' +
             str(value[1]) + '/' +
             str(value[2]))
+
+
+class TimezoneGMT(tzinfo):
```
kgriffs: I think this class might be better placed in `util/misc.py`. In the future, it might be useful outside of Response.

falcon/response.py
```diff
@@ -136,6 +145,82 @@ def set_stream(self, stream, stream_len):
...
+        Note:
+            Kwargs and their valid values are all
+            specified in http://tools.ietf.org/html/rfc6265
+
+        """
+
+        if self._cookies is None:
+            self._cookies = SimpleCookie()
+
+        self._cookies[name] = value
```
kgriffs: Under PY2 this fails:

```python
self._cookies[u'foo'] = 'bar'
```
This can also happen implicitly via:

```python
from __future__ import unicode_literals
```

If the user passes in a `unicode` string, should we try to encode it for them, or just raise an error? At a minimum this should be mentioned in the docstring.

P.S. - It would appear that you can set the value to a `unicode` string, however if it contains any non-ASCII characters, you will still get an error. I didn't test PY3 to see what it does there. We'll have to decide whether to raise an error in this case, or try to convert using UTF-8.

falcon/response.py
tbug: We could easily add some magic that refuses to send a response if it contains secure cookies, but is attempted over a non-secure connection.

I think this has to go just before `start_response`, as you can have custom response types and middlewares that modifies the response. So I think the most robust point for checking this is after all possible modifications of the response.

But I think this could lead to some very annoying situations for a developer, that would be horrible to debug.

It also means that the default `secure=True` suddenly becomes impractical.

If we want to protect the developer against a bad configuration, I think it would be better to do it for _all_ requests.
Yes, the cookie might contain some session priviledges, but who's to say that the response without a cookie isn't equally important (especially in falcon, where most things are probably REST, and API keys are flying around all over the place).
So, simply refusing to send _any_ response over an insecure connection if the magical "be secure, please"-flag is turned on seems like a better "be secure by default" solution.

(Personal opinion: I don't think it is the responsibility of Falcon (read: application) to ensure that the transport to a user is secure. Whoever sets up the LB, webserver, etc. should know the desired configuration, secure or not.)

falcon/response.py
tbug: I'll add some tests to catch this case to catch this in tests.

I'm not sure what the best aproach is. Simply refusing unicode strings seems like the cleanest solution,
but might be a bit unpractical. Checking for non-US-ASCII characters is probably the way to go.
Will probably come up with something when writing the tests.

tbug: In Python 2.7 I get an error from inside the `strings.py`, in Python 3.4 I correctly get `http.cookies.CookieError: Illegal key value` (although I should probably catch that and raise a falcon specific exception).

It seems that the `SimpleCookie` implementation is broken in python 2.7, but works correctly in Python 3.4.
In python 2.7 it tries to convert the unicode characters, but fails in the call inside `strings.py`

I guess we can do that ourselves before handing it to the cookie implementation.
Or just check, and raise error if the characters are not allowed.
That'll probably be the best solution.

doc/user/tutorial.rst
tbug: :+1:

falcon/response.py
tbug: :+1:

falcon/response.py
tbug: Not in the HTTP header sense. It's the cookies "expires property.
Confusing comment. Will fix.

falcon/response_helpers.py
tbug: :+1:

falcon/request.py
```diff
@@ -167,6 +167,13 @@ class Request(object):
             all the values in the order seen.
 
         options (dict): Set of global options passed from the API handler.
+
+        cookies (dict):
+            A dict of name/value cookie pairs.
+
+    See Also:
```
kgriffs: I think I'd rather just move this reference to L172, i.e.:

`A dict of name/value cookie pairs. See also: :ref:`Getting Cookies <getting-cookies>`

falcon/response.py
kgriffs: @tbug OK, I think it makes sense to simply refuse all non-HTTPS requests given a flag is enabled (#460). The developer is responsible for enabling that flag in production deployments. This acts as a last-ditch check in case HTTP is accidentally allowed throw at the LB/server layer.

Cookies won't try to do any additional checking. However, I would like to see some warning in the docstring for the `secure` arg that the developer should be sure to enforce HTTPS in production deployments when using secure cookies.

doc/api/cookies.rst
```diff
@@ -0,0 +1,71 @@
...
+:py:class:`dict` object.
+
+.. tip :: :py:attr:`~.Request.cookies` returns a
+    copy of the response cookie dict. Assign it to a variable as in the above example
+    for better performance.
+
+.. _setting-cookies:
+Setting Cookies 
+~~~~~~~~~~~~~~~
+
```
kgriffs: I think it would be helpful to reiterate here why it is necessary to set cookies using this method, as opposed to `set_header` and `append_header`. In the former case, it is because multiple instances of the same header cannot be set. In the laster case it is because it uses the list syntax which is incompatible with Set-Cookie (see also: http://tools.ietf.org/html/rfc7230#section-3.2.2)

kgriffs:
@tbug Once my last couple of comments are addressed, I think this PR is ready for final review. Could you also rebase and squash all the patches down to a single commit? Thanks!

tbug:
There you go.
Think that should be it.

tbug:
@kgriffs There you go.
Think that should be it.

tests/test_cookies.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertIn("_gh_sess", req.cookies)
+        self.assertIn("tz", req.cookies)
+        self.assertIn("_ga", req.cookies)
+        self.assertIn("_gat", req.cookies)
+        self.assertIn("_octo", req.cookies)
+
+    def test_unicode_ascii(self):
+        resp = falcon.Response()
+        # should be ok
+        resp.set_cookie(u"unicode_ascii", "foobar", max_age=60)
```
jmvrbanac: shouldn't "foobar" also be unicode?

tests/test_cookies.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertEqual("GH1.1.201722077.1422308165", req.cookies["_octo"])
+
+        self.assertIn("logged_in", req.cookies)
+        self.assertIn("_gh_sess", req.cookies)
+        self.assertIn("tz", req.cookies)
+        self.assertIn("_ga", req.cookies)
+        self.assertIn("_gat", req.cookies)
+        self.assertIn("_octo", req.cookies)
+
+    def test_unicode_ascii(self):
```
jmvrbanac: would this test be better named "test_unicode_in_ascii_range"?

falcon/response.py
```diff
@@ -136,12 +144,118 @@ def set_stream(self, stream, stream_len):
...
+                try:
+                    name = name.encode("ascii")
+                except UnicodeEncodeError as e:
+                    raise KeyError("'name' must consist of only ascii")
+
+        if self._cookies is None:
+            self._cookies = SimpleCookie()
+
+        try:
+            self._cookies[name] = value
```
jmvrbanac: If the value is unicode outside of the ascii range, it'll raise a UnicodeEncodeError just like the name does. As a result, it would probably be good to handle that and add a couple tests to cover that scenario.

tbug: Right.
How should we handle this?
My first thought was `ValueError`, but `UnicodeEncodeError` _is_ a `ValueError`.
But maybe just try to encode as ascii as with the cookie `name`, and throw a normal `ValueError`
with a more saying message if it fails?

tbug: It seems that the stdlib cookie module actually have some builtin way to handle non-ascii characters, but `SimpleCookie` is calling `str()` on the value before calling `_quote`

https://github.com/python-git/python/blob/master/Lib/Cookie.py#L671

We could try to fix this (implement a `FalconCookie` extending `BaseCookie`) so we can have the same cookie functionality in both Py2 and Py3

tbug: The Cookie module doesn't support the unicode type.
Fixing it requires some ugly hacks, or rewriting bits and pieces of the Cookie module.
We would probably be better off writing our own falcon cookie module if we want to support the rewrite
defined here https://github.com/python-git/python/blob/master/Lib/Cookie.py#L248-L305

Alternatively, simply refuse non-ascii in cookies.
It's the most sane practice anyway.

@jmvrbanac @kgriffs thoughts?

jmvrbanac: If I recall correctly, RFC 6265 mentions the cookie value being us-ascii. As a result, I'm inclined to refuse any non-ascii cookies. @kgriffs What do you think?

kgriffs: Makes sense to me.

falcon/response.py
```diff
@@ -136,12 +145,120 @@ def set_stream(self, stream, stream_len):
         self.stream = stream
         self.stream_len = stream_len
 
+    def set_cookie(self, name, value, expires=None, max_age=None,
+                   domain=None, path=None, secure=True, httponly=True):
```
jmvrbanac: Nitpic: httponly should follow the C style naming and be "http_only"

falcon/response_helpers.py
```diff
@@ -60,3 +61,18 @@ def format_range(value):
             str(value[0]) + '-' +
             str(value[1]) + '/' +
             str(value[2]))
+
+
+def is_ascii_encodable(s):  # pragma: no cover
+    """ check if argument encodes to ascii without error
```
kgriffs: Nitpick: No preceding whitespace, and capitalize the 'c' in 'check'. Also, include a period and remove the newline, i.e.:

```"""Check if argument encodes to ascii without error."""```

File: falcon/response_helpers.py [hunk 26]
```diff
@@ -60,3 +61,18 @@ def format_range(value):
...
+    """
+    if isinstance(s, six.text_type):
```
kgriffs: Suggestion for a minor performance optimization: You could first check if six.PY2, and if so, simply call str(s). If it raises an error, then it was a unicode type and could not be converted to ascii, since str() does that for you, and is actually slightly faster than doing it explicitly. E.g.:

```
try:
    if six.PY2:
        str(s)
    else:
        s.encode('ascii')

except UnicodeEncodeError:
    return False

return True
```

Note that this does not check whether the input is a string type, but that isn't really what we were originally guarding against, right?

tbug: Depends on what we want to protect against.
If you do the `str(s)` and `s` is a `str` created by say `s = "æøå"`, `str(s)` will do nothing. If you call `encode("ascii")` you will (correctly) get a `UnicodeDecodeError` (which most python2.7 folks working with non-ascii characters will be very familiar with).

I also did a tiny test of `is_ascii_encodable` with `str(s)` vs `s.encode("ascii")` and saw no difference.
if `str(s)` is faster, it is probably drowning in all the other things going on in that function.

@kgriffs Should I see a significant difference in performance between the two?
If so, how do we handle the case where the dev has an invalid `str` type?

if we removed the type checks, we could reduce the function to

```python
def is_ascii_encodable(s):  # pragma: no cover
    """Check if argument encodes to ascii without error."""
    try:
        s.encode("ascii")
    except UnicodeEncodeError:
        return False
    return True
```

Or

```python
def is_ascii_encodable(s):  # pragma: no cover
    """Check if argument encodes to ascii without error."""
    try:
        s.encode("ascii")
    except UnicodeEncodeError:
        return False
    except UnicodeDecodeError:
        return False
    return True
```

if we want to return False in the case that a py2 `str` contains non-ascii chars.

We could even catch `AttributeError` to handle the case where you accidentally hand it a NoneType, Dict or whatever.

If we use
```python
def is_ascii_encodable(s):  # pragma: no cover
    """Check if argument encodes to ascii without error."""
    try:
        s.encode("ascii")
    except UnicodeEncodeError:
        return False
    except UnicodeDecodeError:
        return False
    except AttributeError:
        return False
    return True
```

It will handle all sane uses for this function. You can still get a false-positive if you pass it  the _type_ `str`, (both py2 and py3), and i guess we could check for that.
_or_ we can accept that bad input returns bad output.

I think the last version is prob. the way to go.
We handle all valid input cases (str and unicode in py2 and str in py3) without doing tons of checking
to ensure the input is what we expected.

It _is_ a response helper, and not something that is supposed to face users.

falcon/response.py
tbug: Ok. :+1:

You sure though?
The spec defines the flag as `httponly` which is why i choose to reflect that.
In contrast, `max_age` is called `max-age` in the cookie spec.

kgriffs:
LGTM. Nice work, thanks!

PR #410, opened 2015-01-21
docs(API): Fix signature of process_request(...) in API docstring

b6d:
Hi! Stumbled over an inconsistency in a docstring. Tried to fix it.


kgriffs:
Nice catch, thanks!

PR #416, opened 2015-01-23
feat(middleware): Add middleware method "process_resource"

rsyed83:
Create a new middleware method process_resource that occurs after process_request and routing, and ensure that the resource object is accessible. Additionally, make the resource object available to process_response.

Issue #400

### DIFF
--- a/doc/api/middleware.rst
+++ b/doc/api/middleware.rst
@@ -1,0 +1,109 @@
+.. _middleware:
+
+Middleware
+==========
+
+Middleware components execute both before and after the framework
+routes the request.  Middleware is registered by passing components
+to the :ref:`API class <api>` initializer.
+
+The middleware interface is defined as follows:
+
+
+.. code:: python
+
+    class ExampleComponent(object):
+        def process_request(self, req, resp):
+            """Process the request before routing it.
+
+            Args:
+                req: Request object that will eventually be
+                    routed to an on_* responder method
+                resp: Response object that will be routed to
+                    the on_* responder
+            """
+
+        def process_resource(self, req, resp, resource):
+            """Process the request after routing.
+
+            Args:
+                req: Request object that will be passed to the
+                    routed responder
+                resp: Response object that will be passed to the
+                    responder
+                resource: Resource object to which the request was
+                    routed. May be None if no route was found for
+                    the request
+            """
+
+        def process_response(self, req, resp, resource)
+            """Post-processing of the response (after routing).
+
+            Args:
+                req: Request object
+                resp: Response object
+                resource: Resource object to which the request was
+                    routed. May be None if no route was found
+                    for the request
+            """
+
+Because middleware can execute before routing has occurred, if a
+component modifies ``req.uri`` in its *process_request* method,
+the framework will use the modified value to route the request.
+
+Each component's *process_request*, *process_resource*, and
+*process_response* methods are executed hierarchically, as a stack.
+For example, if a list of middleware objects are passed as
+``[mob1, mob2, mob3]``, the order of execution is as follows::
+
+    mob1.process_request
+        mob2.process_request
+            mob3.process_request
+                mob1.process_resource
+                    mob2.process_resource
+                        mob3.process_resource
+                <route to responder method>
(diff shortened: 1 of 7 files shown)


falcon/api.py
```diff
@@ -52,66 +52,27 @@ def process_request(self, req, resp):
                                 the on_* responder
                         \"""
 
-                    def process_response(self, req, resp)
-                        \"""Post-processing of the response (after routing).
+                    def process_resource(self, req, resp, resource):
+                        \"""Process the request after routing.
+
+                        Args:
+                            req: Request object that will eventually be
```
kgriffs: At this point, the routing has occurred, so maybe reword this to:

"req: Request object that will be passed to the routed responder"

falcon/api.py
```diff
@@ -52,66 +52,27 @@ def process_request(self, req, resp):
...
+                                routed to an on_* responder method
+                            resp: Response object that will be routed to
```
kgriffs: Similar comment to above. Suggested text:

"resp: Response object that will be passed to the responder"

falcon/api.py
```diff
@@ -52,66 +52,27 @@ def process_request(self, req, resp):
...
-            Finally, if one of the *process_response* methods raises an error,
-            or the routed on_* responder method itself raises an error, the
-            exception will be handled in a similar manner as above. Then,
-            the framework will execute any remaining middleware on the
-            stack.
+                    def process_response(self, req, resp, resource)
+                        \"""Post-processing of the response (after routing).
 
+                        Args:
+                            req: Request object that will eventually be
```
kgriffs: At this point, the request was routed to a responder, if one could be found. But even if no route is found, the middleware component's `process_response` method will still be called with the usual Request and Response objects. Rather than trying to explain all this, it may be best to keep things simple in the docstring, e.g.:

```
req: Request object
resp: Response object
```

falcon/api.py
```diff
@@ -52,66 +52,27 @@ def process_request(self, req, resp):
...
+                                routed to an on_* responder method
+                            resp: Response object that will be routed to
+                                the on_* responder
+                            resource: Resource object that owns the responder
```
kgriffs: I think it would be helpful to mention that this argument may be `None`:

"resource: Resource object to which the request was routed. May be *None* if no route was found for the request."

tests/test_middlewares.py
```diff
@@ -299,6 +321,38 @@ def handler(ex, req, resp, params):
...
+
+            def process_resource(self, req, resp, resource):
+                raise Exception("Always fail")
+
+        self.api = falcon.API(middleware=[ExecutedFirstMiddleware(),
+                                          RaiseErrorMiddleware(),
+                                          ExecutedLastMiddleware()])
+
+        def handler(ex, req, resp, params):
+            context['error_handler'] = True
```
kgriffs: Missing assertion? `context['error_handler']` is set here, but I can't find anywhere the value is being checked...

doc/api/middleware.rst
```diff
@@ -0,0 +1,108 @@
+.. _middleware:
+
+Middleware
+==========
+
+Middleware components execute both before and after the framework
+routes the request.  Middleware registration is achieved through
```
kgriffs: I think we can make the second sentence flow a little better, e.g.:

```rst
Middleware is registered by passing components to the :ref:`API class <api>` initializer.
```

doc/api/middleware.rst
```diff
@@ -0,0 +1,108 @@
...
+            Args:
+                req: Request object that will eventually be
+                    routed to an on_* responder method
+                resp: Response object that will be routed to
+                    the on_* responder
+                resource: Resource object that owns the responder
+            """
+
+
+Because middleware can execute before routing as occurred, if a
```
kgriffs: typo: "as" ==> "has"

ealogar:
Nice work!

kgriffs:
Nice job, LGTM.

rsyed83:
Thanks guys!  My first open source contribution ever (long overdue).

PR #429, opened 2015-02-04
doc(middleware): Fix middleware interface in quickstart

rsyed83:
The quickstart documentation has an incorrect example for process_resource middleware.


kgriffs:
Nice catch!

Issue #35, opened 2013-01-23
Support more URI template stuff


kgriffs:
http://tools.ietf.org/html/rfc6570

* {/list*} => /red/green/blue
* Path Segment Expansion: {/var}
* Mandatory query params (when literals are specified, will also need query continuation support)
* Max length? Maybe not, because would mean 404 if exceeded unless rearchitect how route matching works. Have to match and then throw a different exception if too long.
* Etc.

Depends on issue #114

bendemaree:
Another one for this list, brought in from IRC. This seems to be the parent issue for URI template things; if not, I'm happy to move this! :smile:

As far as I can tell, [the RFC](http://tools.ietf.org/html/rfc6570#section-2.3) allows leading underscores (`_`) in URI template variable names.

```
varname       =  varchar *( ["."] varchar )
varchar       =  ALPHA / DIGIT / "_" / pct-encoded
```

This should allow for `varname`s of the form `_ref` or `_link`, for example.

kgriffs:
Marking as invalid and closing this issue, since Falcon is moving away from the RFC altogether.

PR #454, opened 2015-02-19
(feat)(falcon/routing) Allow numbers in template fields

BenjamenMeyer:
Allow URI template fields to contain numbers so long as the field name does not start
with a numeral.

Example: "test123" is valid, while "123test" is not

Supports Issue #35 

Note: '\w' = [0-9A-Za-z_]


kgriffs:
#484 should address this. Can you give it a try?

BenjamenMeyer:
@kgriffs Sorry, I just noticed your comment.

I checked out PR #484 (as requested) and actually merged master into this PR, namely so that I could just check the tests more carefully. Since he already had the negative test (https://github.com/falconry/falcon/blob/4a5e5b1a54db42ecb537a43fd7ec2d9a5a030046/tests/test_uri_templates.py#L50) I did this primarily to check my positive test which passed.

The PR seems to cover my positive test in the complex test (https://github.com/falconry/falcon/blob/4a5e5b1a54db42ecb537a43fd7ec2d9a5a030046/tests/test_default_router.py#L64), just a little more indirect. The merge basically reduces this PR to adding the couple tests more explicitly.

https://github.com/falconry/falcon/blob/4a5e5b1a54db42ecb537a43fd7ec2d9a5a030046/falcon/routing/compiled.py#L183 seems to do what my regex changes does, with the negative prevented by https://github.com/falconry/falcon/blob/4a5e5b1a54db42ecb537a43fd7ec2d9a5a030046/falcon/routing/compiled.py#L67. I think the negative test has the potential to be bypassed, but if someone is doing that then the bug is in their code, not Falcon's.

In the end, what I reported looks to be fixed. I'll leave it to you whether to accept my extra couple tests.
If you want them, then merge. If not, you can close the PR. I'm happy with the result.

kgriffs:
Merged manually. Thanks @BenjamenMeyer !

PR #455, opened 2015-02-20
Update README to remove note about 2.0

lwcolton:
Now that 2.0 is released we don't need this note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,6 @@ Falcon is a [high-performance Python framework][home] for building cloud APIs. I
 
 ### Install ###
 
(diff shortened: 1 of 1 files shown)


kgriffs:
Nice catch. Could you also update README.rst ?

kgriffs:
I'll take care of the RST version.

lwcolton:
Thanks Kurt. Sorry for letting this go. Looks like we will be building our
next in house app using falcon and we appreciate your hard work so I will
try to support you as much as possible

kgriffs:
No worries, thanks for your help!

Issue #369, opened 2014-11-08
Decoding the query string before parsing it is problematic


elishacook:
Let's say I wanted to encode some JSON, like `{"foo":1,"bar":2}` in a query param. Usually I would encode this value and pass it in the URL like this, `/my/api?skidoo=%7B%22foo%22%3A1%2C%22bar%22%3A1%7D`. The way things currently work, the whole query string will be decoded and then parsed by splitting at commas. This results in a `skidoo` param with two values `'{"foo":1'` and `'"bar":1}'`.

If the query string were parsed and then each key and value decoded the result would be a single `skidoo` param with the value ``'{"foo":1, "bar":1}'`.

rsyed83:
That sounds reasonable.  Perhaps the existing behavior can be kept in a method should a user desire this behavior, but default behavior should decode the query string without assuming the values are comma-delimited.

kgriffs:
Yeah, so the issue isn't so much that the entire query string is being decoded in a single pass, but that each value is being split by comma. We could use RequestOptions to toggle splitting values on commas, and turn it off by default. That would be a breaking change, but it would be easy for anyone encoding lists into values to turn it back on and be back in business.

rogerbush8:
I don't think you can decode in one pass, and I've never seen it done that way.  The reason is that percent encoding is meant to take special characters with reserved meaning and turn that reserved meaning off.  It's a form of escaping, much like "\n" prevents the carriage return from going to the next line.  Therefore if I have an encoded "&" for example, decoding in one pass will mixup the special (non-escaped) & and escaped &.  You will split on all of the escaped &, which would be an error.

http://en.wikipedia.org/wiki/Percent-encoding

There is a nice side to all of this which is that if you slightly extended the special characters to include ",", then an escaped "," would not split (internal split to array, not param split).  This would allow things to mostly work as-is, with the small additional documentation that "," should not be percent encoded if their special split behavior is desired.

rogerbush8:
You may, of course, still have to have a flag on the off case that someone is percent encoding and their "," turned off would break.  However, you're at .3 so I'd think this type of thing can be squeezed through without the annoying flag.

rogerbush8:
I've seen lots of encoding/decoding errors in complex params (on both sides:  developers implementing, and frameworks).  With the niche falcon occupies, it would be great if parameter decoding/encoding just worked flawlessly.  I realize this is true spec-grunginess at it's finest, but...

[...]

PR #461, opened 2015-03-05
(fix)(utils/uri) change decode behavior for handle json/url values

steffgrez:
Fix problem when client send json/full url percent-encoded in a query string.
Now make decode in parse_query_string rather than in the requests processing.

ex:
url=http%3A%2F%2Ffalconframework.org%3Ftest%3D1
json=%7B%22test1%22%3A%20%22data1%22%2C%20%22test2%22%3A%20%22data2%22%7D

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -220,16 +220,13 @@ class Request(object):
 
         # PERF(kgriffs): if...in is faster than using env.get(...)
         if 'QUERY_STRING' in env:
-            query_str = env['QUERY_STRING']
+            self.query_string = env['QUERY_STRING']
 
-            if query_str:
-                self.query_string = uri.decode(query_str)
+            if self.query_string:
                 self._params = uri.parse_query_string(
                     self.query_string,
                     keep_blank_qs_values=self.options.keep_blank_qs_values,
                 )
-            else:
-                self.query_string = six.text_type()
 
         else:
             self.query_string = six.text_type()
@@ -915,7 +912,7 @@ class Request(object):
 
         if body:
             extra_params = uri.parse_query_string(
-                uri.decode(body),
+                body,
                 keep_blank_qs_values=self.options.keep_blank_qs_values,
             )
 
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -259,6 +259,10 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
         a more compact form in which the param may be given a single time
         but set to a ``list`` of comma-separated elements (e.g., 'foo=a,b,c').
 
+        When using this format, all commas uri-encoded will not be treated by
+        Falcon as a delimiter. If the client wants to send a value as a list,
+        it must not encode the commas with the values.
+
         The two different ways of specifying lists may not be mixed in
         a single query string for the same parameter.
 
@@ -287,14 +291,21 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
         if not (v or keep_blank_qs_values):
             continue
 
+        # Note(steffgrez): Falcon first decode name parameter for handle
+        # utf8 character.
+        k = decode(k)
+
+        # NOTE(steffgrez): Falcon decode value at the last moment. So query
+        # parser won't mix up between percent-encoded comma (as value) and
+        # comma-separated list (as reserved character for sub-delimiter)
         if k in params:
             # The key was present more than once in the POST data.  Convert to
             # a list, or append the next value to the list.
             old_value = params[k]
             if isinstance(old_value, list):
-                old_value.append(v)
+                old_value.append(decode(v))
             else:
-                params[k] = [old_value, v]
+                params[k] = [old_value, decode(v)]
 
         else:
             if ',' in v:
@@ -309,9 +320,11 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
                     # NOTE(kgriffs): Normalize the result in the case that
                     # some elements are empty strings, such that the result
                     # will be the same for 'foo=1,,3' as 'foo=1&foo=&foo=3'.
-                    v = [element for element in v if element]
-
-            params[k] = v
+                    params[k] = [decode(element) for element in v if element]
+                else:
+                    params[k] = [decode(element) for element in v]
+            else:
+                params[k] = decode(v)
 
     return params
 
--- a/tests/test_query_params.py
+++ b/tests/test_query_params.py
@@ -49,7 +49,7 @@ class _TestQueryParams(testing.TestBase):
         self.assertEqual(store['limit'], '25')
 
     def test_percent_encoded(self):
-        query_string = 'id=23%2c42&q=%e8%b1%86+%e7%93%a3'
+        query_string = 'id=23,42&q=%e8%b1%86+%e7%93%a3'
         self.simulate_request('/', query_string=query_string)
 
         req = self.resource.req
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -185,6 +185,38 @@ class TestFalconUtils(testtools.TestCase):
             actual = uri.decode(case)
             self.assertEqual(expect, actual)
 
+    def test_parse_query_string(self):
+        query_strinq = (
+            "a=http%3A%2F%2Ffalconframework.org%3Ftest%3D1"
+            "&b=%7B%22test1%22%3A%20%22data1%22%"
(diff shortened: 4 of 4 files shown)


falcon/request.py
```diff
@@ -220,10 +220,9 @@ def __init__(self, env, options=None):
 
         # PERF(kgriffs): if...in is faster than using env.get(...)
         if 'QUERY_STRING' in env:
-            query_str = env['QUERY_STRING']
+            self.query_string = env['QUERY_STRING']
```
kgriffs: With this change L230-L231 are no longer necessary. Could you remove those lines?

falcon/util/uri.py
```diff
@@ -292,9 +292,9 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
...
             old_value = params[k]
             if isinstance(old_value, list):
-                old_value.append(v)
+                old_value.append(decode(v))
             else:
-                params[k] = [old_value, v]
+                params[k] = [old_value, decode(v)]
 
         else:
             if ',' in v:
```
kgriffs: Seems like v must be decoded before checking if it contains commas, since the commas might be encoded. What do you think about factoring out the `decode(v)` and placing it above L290, e.g.:

```python
v = decode(v)

for k in params:
    # ...
```

tests/test_query_params.py
```diff
@@ -49,7 +49,7 @@ def test_simple(self):
         self.assertEqual(store['limit'], '25')
 
     def test_percent_encoded(self):
-        query_string = 'id=23%2c42&q=%e8%b1%86+%e7%93%a3'
+        query_string = 'id=23,42&q=%e8%b1%86+%e7%93%a3'
```
kgriffs: Could you please revert this change? See also my other note above re L300.

falcon/util/uri.py
steffgrez: In fact, making this factoring will don't fix original problem.

The only thing to determine if a comma is the reserved character or a simple character in your value is the percent-encoding. So, if i want to make a comma-separated list, i use the reserved character ','. Otherwise, if i want to send a value with a comma (like json, complex url, etc ...), i use  percent-encoding.
For me, falcon has to decode value at the last moment. So query parser will not handle all percent-encoded comma. Like that it won't confound with this sub-delimiter.

ex:
I want send the json '{"test1": "data1", "test2": "data2"}' by query parameter.
So we encode it : %7B%22test1%22%3A%20%22data1%22%
if we decode value before checking commas, the result will be a list: ['{"test1": "data1"', ' "test2": "data2"}'] while we wait a string.

As i understand URI rfc:
http://tools.ietf.org/html/rfc3986#section-2.4
"When a URI is dereferenced, the components and subcomponents significant to the scheme-specific dereferencing process (if any) must be parsed and separated before the percent-encoded octets within those components can be safely decoded, as otherwise the data may be mistaken for component delimiters"
http://tools.ietf.org/html/rfc3986#section-7.3
"Applications must split the URI into its components and subcomponents prior to decoding the octets, as otherwise the decoded octets might be mistaken for delimiters."

falcon/request.py
steffgrez: it makes sense, Request class don't use self.query_string elsewhere without make a test.

tests/test_query_params.py
steffgrez: See my response above.

falcon/util/uri.py
kgriffs: Ah, good point. Makes sense. Thanks for the clarification! Would you mind adding a `# NOTE(steffgrez): blah blah` comment to the code about this?

steffgrez: ok for me

falcon/util/uri.py
```diff
@@ -311,7 +314,7 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
                     # will be the same for 'foo=1,,3' as 'foo=1&foo=&foo=3'.
                     v = [element for element in v if element]
```
kgriffs: Technically, there's a breaking change here, although there isn't currently a test written for it. If individual elements are percent-encoded, they will no longer be decoded here. Do you think it would be worth doing the following? If so, it would be good to also add a test. I suspect few (if any) apps require this, but it may be better to be safe than sorry.

```python
if not keep_blank_qs_values:
    v = [decode(element) for element in v if element]
else:
    v = [decode(element) for element in v]
```

The alternative would be to add a "breaking change" notice to the commit message (and eventually the changelog when we release 0.3).

falcon/util/uri.py
rogerbush8: >> Seems like v must be decoded before checking if it contains commas, since the
>> commas might be encoded. What do you think about factoring out the decode(v)
>> and placing it above L290, e.g.:

This is incorrect.  The purpose of the % encoding is to "turn off the special nature" of the character.  It's just like backslashing a character in a C string.  So if we decode prior to splitting, we are in trouble.

The procedure should be:
1. Split on literal "&" to split the n=v parameter pairs.
2. For each of these parameter pairs, subsplit on ",", which is simply the next delimiter.
3. Now decode each fragment separately.  This means:
    a) decode the name of a parameter pair (e.g. a utf-8 funny char in an identifier...god help us)
    b) decode the value:
        i) Single value
       ii) Multiple values (split on ","), each of which must be decoded.

3.b.ii. also shows how someone needs to do encoding if they have multi-valued parameters separated by "," (they must encode each part separately, and NOT encode the ",").  Note that we are EXTENDING the meaning of "," to have a special meaning in the same spirit that the original spec had of "&".  "," special meaning is - turn this value into an array, and use "," as the subdelimiter.
   

v = decode(v)

if k in params:
    # ...

falcon/util/uri.py
rogerbush8: Individual elements in an array induced by "," should be decoded IMHO.  It's just recursive design logic.  Why would one type of value be encoded/decoded, whereas another would not?  This is counter-intuitive and makes no sense.  Values are encoded so that we can pass the funny characters without having issues.  Just go to each element in the array and decode it.

I know it sounds un-fast.  But first make it correct, then make it fast.

steffgrez: In fact, i didn't see this case, and for me, it is not a "breaking change" but an error. I think too that handle this case is the thing to do. So i fix it and i add also unit tests for later.

another question, do we accept a mix like that : a=1&a=2&a=3,4 ?
it's a little weird, but the case could exist ... Actually, falcon doesn't handle this case.

falcon/util/uri.py
kgriffs: At this point, I think we're just missing the decoding of the name (`k`), and improving the documentation per (3.b.ii).

@steffgrez Would you mind adding a line to decode `k` and a test for the same? Regarding (3.b.ii), we could update the note in the docstring: For example:

```
     Note:
         In addition to the standard HTML form-based method for specifying
         lists by repeating a given param multiple times, Falcon supports
         a more compact format in which the param may be given a single time
         but set to a ``list`` of comma-separated elements (e.g., 'foo=a,b,c').

         When using this format, if any values need to be uri-encoded, the 
         client should take care not to encode the commas used as 
         delimiters. Otherwise, Falcon will not treat the value as a list.
 
         The two different ways of specifying lists may not be mixed in
         a single query string for the same parameter.
```

EvanLyu:
In my case,
Request content-type is `application/x-www-form-urlencoded`
Raw data is "RESULT_CODE=000000&EXTEND=CARD_TYPE%3D20%26CARD_ID%3DMp8eLZ&CHECK_CODE=FEEFAF749"

If i use falcon version 0.2 ,
falcon parser result is:
{
    RESULT_CODE: 000000,
    EXTEND: CARD_TYPE=20,
    CARD_ID: Mp8eLZ,
    CHECK_CODE: FEEFAF749
}

But real parsed data is:
{
    RESULT_CODE: 000000,
    EXTEND: CARD_TYPE=20&CARD_ID=Mp8eLZ,
    CHECK_CODE: FEEFAF749
}

It deal my problem.

kgriffs:
@steffgrez Once my last two docstring comments have been addressed, I think this PR is ready for final review. Would you mind squashing all patches down to a single commit, in preparation for merging? Also, note that the commit message's first line should begin with `fix`, rather than `(fix)`. Please also mention the following at the bottom of the commit description on a line by itself:

`Fixes #369`

Thanks!

steffgrez:
done.  

Thanks for your feedback !

kgriffs:
Nice work! LGTM

PR #465, opened 2015-03-07
fix(dependencies): falcon needs at least six 1.4.0

cgtx:
six.PY2 is used several times in falcon, but that wasn't added to six until 1.4.0


kgriffs:
Verified that PY2 landed in 1.4. Thanks!

PR #477, opened 2015-03-31
feat(Request) Add default parameter to get_param

haandol:
I'm using falcon on my service and I'm loving it so much!

I added `default` parameter to request.get_param because I always need default value for getting parameters.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -570,7 +570,7 @@ class Request(object):
 
             raise HTTPMissingParam(name)
 
-    def get_param(self, name, required=False, store=None):
+    def get_param(self, name, required=False, store=None, default=None):
         """Return the value of a query string parameter as a string.
 
         Note:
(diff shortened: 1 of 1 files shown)


kgriffs:
@haandol Thanks! This idea is something I've wrestled with in the past, and I'd like to get your thoughts. One way to go about this is to use the following idiom:

```python
req.get_param('thing') or 'foo'
```

Adding a `default` kwarg would let you do this instead:

```python
req.get_param('thing', default='foo')
```

My personal preference is to use the first approach, but I wouldn't be opposed to merging this PR if others find the second approach more appealing.

rafiqrahim:
I am not sure if the first approach is more pythonic but the second is more natural to me.

mcmasterathl:
I'm not good enough with git/github to know why there are merge conflicts but I ported the changes forward in #496. The changes allow both options listed by @kgriffs since the default behavior remains unchanged.

PR #479, opened 2015-04-09
fix(uri, request): Fix parsing error when "&" character exists in form request

rafiqrahim:
This is an attempt to fix form encoded value being truncated after "&" character.
This fix passed all the tests but I need someone to test whether it has side effects (output and performance wise).

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -915,8 +915,9 @@ class Request(object):
 
         if body:
             extra_params = uri.parse_query_string(
-                uri.decode(body),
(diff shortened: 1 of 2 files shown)


kgriffs:
@rafiqrahim Thanks for working on this! A few other community members have ran into similar issues. We just merged PR #461 that should fix your problem. Could you give it a try and let me know?

rafiqrahim:
@kgriffs Yes, it fixed my problem! Thank you.

PR #482, opened 2015-04-13
feat(Request): Return date headers as datetime objects

### LINKED ISSUE #254: Return if-modified-since, if-unmodified-since as datetime.datetime

ginoledesma:
fix(falcon/request)
    
Change if-modified-since and if-unmidified-since headers to datetime
objects instead of strings.
    
Fixes #254

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -146,10 +146,10 @@ class Request(object):
             header is missing.
         if_none_match (str): Value of the If-None-Match header, or ``None``
             if the header is missing.
-        if_modified_since (str): Value of the If-Modified-Since header, or
-            ``None`` if the header is missing.
-        if_unmodified_since (str): Value of the If-Unmodified-Sinc header,
+        if_modified_since (datetime): Value of the If-Modified-Since header,
             or ``None`` if the header is missing.
+        if_unmodified_since (datetime): Value of the If-Unmodified-Since
+            header, or ``None`` if the header is missing.
         if_range (str): Value of the If-Range header, or ``None`` if the
             header is missing.
 
@@ -271,8 +271,6 @@ class Request(object):
 
     if_match = helpers.header_property('HTTP_IF_MATCH')
     if_none_match = helpers.header_property('HTTP_IF_NONE_MATCH')
-    if_modified_since = helpers.header_property('HTTP_IF_MODIFIED_SINCE')
-    if_unmodified_since = helpers.header_property('HTTP_IF_UNMODIFIED_SINCE')
     if_range = helpers.header_property('HTTP_IF_RANGE')
 
     @property
@@ -326,16 +324,15 @@ class Request(object):
 
     @property
     def date(self):
-        try:
-            http_date = self.env['HTTP_DATE']
-        except KeyError:
-            return None
+        return self.get_header_as_datetime('Date')
 
-        try:
-            return util.http_date_to_dt(http_date)
-        except ValueError:
-            msg = ('It must be formatted according to RFC 1123.')
-            raise HTTPInvalidHeader(msg, 'Date')
+    @property
+    def if_modified_since(self):
+        return self.get_header_as_datetime('If-Modified-Since')
+
+    @property
+    def if_unmodified_since(self):
+        return self.get_header_as_datetime('If-Unmodified-Since')
 
     @property
     def range(self):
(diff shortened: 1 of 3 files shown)


falcon/request.py
```diff
@@ -146,10 +146,10 @@ class Request(object):
             header is missing.
         if_none_match (str): Value of the If-None-Match header, or ``None``
             if the header is missing.
-        if_modified_since (str): Value of the If-Modified-Since header, or
-            ``None`` if the header is missing.
-        if_unmodified_since (str): Value of the If-Unmodified-Sinc header,
+        if_modified_since (datetime): Value of the If-Modified-Since header,
             or ``None`` if the header is missing.
+        if_unmodified_since (datetime): Value of the If-Unmodified-Sinc
```
kgriffs: Typo: missing 'e' on end of If-Unmodified-Since

falcon/request.py
```diff
@@ -571,6 +568,33 @@ def get_header(self, name, required=False):
...
+        Raises:
+            HTTPBadRequest: The header was not found in the request, but
+                it was required.
+            HttpInvalidHeader: The header contained a malformed/invalid value.
+        """
+
+        try:
+            http_date = self.get_header(header, required=required)
+            return util.http_date_to_dt(http_date) if http_date is not None \
+                else None
```
jmvrbanac: I believe you should be able to bring this down to:

```python
return util.http_date_to_dt(http_date) if http_date else None
```

Which should get rid of the backslash as well. :)

ginoledesma: That was my initial pass, but that caused some unit tests to fail. There are three cases we need to handle:
1. Date header is not set, expect None
2. Date header is set but invalid, expect HTTPInvalidHeader
3. Date header is set and valid, expect parsed datetime

The `get_header()` method does not distinguish between "header is not present" and "header is present but has a value of None or empty string" — both will return None. In the former case, we want to return None. For the latter, we want to raise a corresponding exception.

Here's another stab, but I'm not sure if it's any clearer:
```
        try:
            http_date = self.get_header(header, required=required)
            return util.http_date_to_dt(http_date)
        except TypeError:
            # When the header does not exist and isn't required
            return None
        except ValueError:
            msg = ('It must be formatted according to RFC 1123.')
            raise HTTPInvalidHeader(msg, header)
```

ginoledesma:
That was my initial pass, but that caused some unit tests to fail. There are three cases we need to handle:
1. Date header is not set, expect None
2. Date header is set but invalid, expect HTTPInvalidHeader
3. Date header is set and valid, expect parsed datetime

The get_header() method does not distinguish between "header is not present" and "header is present but has a value of None or empty string" — both will return None. In the former case, we want to return None. For the latter, we want to raise a corresponding exception.

Here's another stab, but I'm not sure if it's any clearer:

        try:
            http_date = self.get_header(header, required=required)
            return util.http_date_to_dt(http_date)
        except TypeError:
            # When the header does not exist and isn't required
            return None
        except ValueError:
            msg = ('It must be formatted according to RFC 1123.')
            raise HTTPInvalidHeader(msg, header)

jmvrbanac:
Could you update your commit message to conform to the documented standard? https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#commit-message-format

ginoledesma:
Git commit amended.

ginoledesma:
Restored coverage to 100%

falcon/request.py
```diff
@@ -571,6 +568,35 @@ def get_header(self, name, required=False):
...
+        Raises:
+            HTTPBadRequest: The header was not found in the request, but
+                it was required.
+            HttpInvalidHeader: The header contained a malformed/invalid value.
+        """
+
+        try:
+            http_date = self.get_header(header, required=required)
+            return util.http_date_to_dt(http_date)
+        except TypeError:
```
kgriffs: Catching an exception will be more expensive (in terms of CPU time) than explicitly checking whether `http_date is None`. If we expect the header to almost always exist in the request, the current model is ideal because it avoids the (usually) superfluous check against None. If the header will often not be found, relying on exceptions can be less performant.

That being said, much of the overhead of exception handling is incurred in the setup (when you wrap in a "try" block), and since we need to do that anyway to catch ValueError, I think the code should be fine as-is. :+1:

kgriffs:
LGTM

kgriffs:
@ginoledesma Oops, I just noticed the commit message is missing a few important pieces:

`fix(Request)`

Just needs a few words to help the reader when they run `git log --pretty=oneline`. Also, I think it is a feature, not necessarily a bug fix. E.g.,:

`feat(Request): Return date headers as datetime objects`

Finally, this is a breaking change since we changing the return type. Therefore, below the description but before the "Fixes" line, could you add a note describing what changed, and any suggestions for the caller on how they will need to modify their code? E.g.:

```
feat(Request): Return date headers as datetime objects

<description>

BREAKING CHANGE: <blah blah>

Fixes #254
```

Once that's done, I think we're ready to rock.

kgriffs:
Superseded by #495

PR #483, opened 2015-04-13
Enhancement 293

### LINKED ISSUE #293: Add a helper to the Request class for getting a param as a date

lwcolton:
Fix #293

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -838,6 +838,37 @@ class Request(object):
 
         raise HTTPMissingParam(name)
 
+    def get_param_as_date(self, name, format_string="%Y-%m-%d", required=False, store=None):
+        """
+        Args:
+            name (str): Parameter name, case-sensitive (e.g., 'ids').
+            format_string (str): String used to parse the param value into a date.
+                See https://docs.python.org/3.4/library/datetime.html#strftime-and-strptime-behavior
+            required (bool, optional): Set to ``True`` to raise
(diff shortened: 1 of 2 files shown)


falcon/request.py
```diff
@@ -838,6 +838,38 @@ def get_param_as_list(self, name,
...
+                ``ValueError``.
+        """
+        param_value = self.get_param(name, required=required)
+        try:
+            date = datetime.strptime(param_value, format_string).date()
+        except ValueError:
+            msg = "The date value does not match the required format"
+            raise HTTPInvalidParam(msg, name)
+            
+        if store != None:
```
jmvrbanac: I think the correct way of expressing this would be:

```python
if store is not None:
```

lwcolton: They will both work.  'Is' tests for identity meaning it works in this case
only because None is preinitialized as a constant by python and any
reference to it is a reference to the same object. We really want to test
for equality which is what the '!=' is doing, so I use that because its
more explicit about what I intend. Although they will both return the same
value in this case.

lwcolton:
My co-worker showed me the section of PEP8 that says "is" should be used here.  My apologies, PR has been updated.

jmvrbanac:
Thanks for the update. Sorry, I forgot to mention this before but could you squash down your changes and update the commit message to conform to the documented standard? https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#commit-message-format

Thanks!

lwcolton:
Now #487

Issue #406, opened 2015-01-13
Add a helper for redirects


rouge8:
Flask, Pyramid, Django, etc. all include helpers for redirects. It would be nice if falcon had something similar, especially if it could be used from hooks.

Currently, I'm implementing a redirect in the same before hook I do validation from (e.g. returning `404` if an object doesn't match a route), but it seems like the responder methods on the resource must check the status code after the hook so as not to overwrite the redirect.

kgriffs:
Makes sense. This is a pretty fundamental pattern, so I wouldn't be opposed to implementing it directly in the framework.

csojinb:
I'd like to take a stab at this.

aisola:
I've actually just completed it. Sorry forgot to comment to say I was going to try it.

PR #486, opened 2015-04-13
add redirect and tests for it

aisola:
Fixes https://github.com/falconry/falcon/issues/406

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -434,3 +434,10 @@ class Response(object):
             return headers.items()
 
         return list(headers.items())  # pragma: no cover
+
+    def redirect(self, uri, permanent=False):
+        if permanent:
+            self.status = "301 Moved Permanently"
+        else:
+            self.status = "303 See Other"
+        self.append_header("Location", uri)
--- a/tests/test_response_body.py
+++ b/tests/test_response_body.py
@@ -15,3 +15,16 @@ class TestResponseBody(testing.TestBase):
             resp.body += " "
 
         self.assertEqual(resp.body, text)
+
+    def test_redirect(self):
+        resp = falcon.Response()
+        resp.redirect('http://localhost/anything')
+        self.assertEqual(resp.status, "303 See Other")
+        self.assertEqual(resp._headers['Location'], 'http://localhost/anything')
+
+    def test_redirect_permanent(self):
+        resp = falcon.Response()
+        resp.redirect('http://localhost/anything', permanent=True)
+        self.assertEqual(resp.status, "301 Moved Permanently")
+        self.assertEqual(resp._headers['Location'], 'http://localhost/anything')
+


rouge8:
> Currently, I'm implementing a redirect in the same before hook I do validation from (e.g. returning 404 if an object doesn't match a route), but it seems like the responder methods on the resource must check the status code after the hook so as not to overwrite the redirect.

This looks like it would have the same problem I mentioned in #406. [Pyramid uses `raise` for redirects](http://pyramid-cookbook.readthedocs.org/en/latest/pylons/exceptions.html#issuing-redirects-and-http-errors), which is another possible approach.

I don't know if you discussed this at the sprints or IRC, just mentioning an alternative approach.

aisola:
If you use `return resp.redirect("http://example.com/endpoint")`, you won't run into the issue. (If I am understanding the issue correctly.)

rouge8:
If you return the `resp` from a hook (or similar) that, that won't skip the rest of the responder.

falcon/response.py
```diff
@@ -434,3 +434,10 @@ def _wsgi_headers(self, media_type=None):
             return headers.items()
 
         return list(headers.items())  # pragma: no cover
+
+    def redirect(self, uri, permanent=False):
+        if permanent:
+            self.status = "301 Moved Permanently"
```
kgriffs: Nitpick: This string could be replaced by `falcon.HTTP_301` or `falcon.HTTP_MOVED_PERMANENTLY`.

mcmasterathl:
I would agree with @rouge8 that raising an exception is probably a good way to do this, as it will let you skip the responder from a hook. Not sure if it makes sense to use the existing HTTPError stuff and pass a redirect through the error handlers or not. You could modify resp.redirect to raise this exception instead of modifying the resp object.

falcon/response.py
```diff
@@ -434,3 +434,10 @@ def _wsgi_headers(self, media_type=None):
...
+        else:
+            self.status = "303 See Other"
+        self.append_header("Location", uri)
```
kgriffs: I think we could use `self.set_header` instead, since we don't expect multiple Location headers.

falcon/response.py
```diff
@@ -434,3 +434,10 @@ def _wsgi_headers(self, media_type=None):
             return headers.items()
 
         return list(headers.items())  # pragma: no cover
+
+    def redirect(self, uri, permanent=False):
```
kgriffs: Since this method isn't a property, I'd prefer calling it `set_redirect`. Also, could you move the method to before `def _wsgi_headers(...)`? I'd like to keep public properties and methods grouped together.

kgriffs:
@mcmasterathl Actually, after hooks and middleware still run when an instance of HTTPError is raised. Couple of options if we want to raise an exception

#### Raise a redirect exception and do not skip hooks/middleware

We'd probably want to define a new base class for non-error exceptions like redirects, but we could still treat it the same as instances of HTTPError. 

This would not circumvent middleware used for timing requests or adding certain response headers, such as a transaction ID. The downside is that now there is the possibility that the status or location  might be overwritten. 

Questions:
* In what cases would an after hook or process_response middleware method want to modify the status line or location header?
* In each of the above cases, should a (potential) redirect still be allowed to proceed, or does it actually make sense to override it?

#### Raise a redirect exception and skip hooks/middleware

We could define a new base class for these types of exceptions, and modify `API.__call__` to skip hooks/middleware when it is raised.

The downside of this is that if people are using middleware for orthogonal functionality, we would break them.

kgriffs:
Depends on #502

kgriffs:
@aisola do you have some time to work on this in the next couple of weeks? I think it would make sense to raise an exception, perhaps creating a new base exception class (e.g., HTTPException) that HTTPError would also inherit from.

kgriffs:
I just came across the HTTPStatus exception (forgot about that!). I think we can use it for this.

kgriffs:
@aisola, @rouge8, @jmvrbanac: I'm working on an alternative design that would simply define a set of HTTPStatus subclasses. This should avoid the problem of hooks and responder methods possibly overriding the redirect. You can raise an instance of one of these classes and it will short-circuit processing similar to the way raising an instance of HTTPError does.

I.e., if raised in a *before* hook, it would skip  any remaining hooks and the responder method, but would not skip any `process_response` middleware methods. If raised within a responder, it would skip the rest of the responder and all *after* hooks. If raised in an *after* hook, it would skip remaining after hooks but not middleware methods. And finally, if raised within a middleware method, execution would perceive as described at the bottom of [(1)](https://falcon.readthedocs.org/en/stable/api/middleware.html).

(See also: [WIP](https://github.com/falconry/falcon/compare/master...kgriffs:issue_406?expand=1))

What does everyone think about this approach?

rouge8:
@kgriffs your design looks good to me, solves my original problem in #406. I haven't worked with middleware at all (our internal Falcon projects became roughly feature complete before the middleware release), so I don't have any opinions there.

jmvrbanac:
@kgriffs LGTM :+1:

kgriffs:
[superseded by #596]

PR #487, opened 2015-04-13
feat(request): Add function to get param as date

### LINKED ISSUE #293: Add a helper to the Request class for getting a param as a date

lwcolton:
A user requested a feature where a param could be automatically parsed into a date.

closes #293

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -838,6 +838,38 @@ class Request(object):
 
         raise HTTPMissingParam(name)
 
+    def get_param_as_date(self, name, format_string="%Y-%m-%d", required=False, store=None):
+        """
+        Args:
+            name (str): Parameter name, case-sensitive (e.g., 'ids').
+            format_string (str): String used to parse the param value into a date.
+                See https://docs.python.org/3.4/library/datetime.html#strftime-and-strptime-behavior
+                (default ``"%Y-%m-%d"``)
+            required (bool, optional): Set to ``True`` to raise
+                ``HTTPBadRequest`` instead of returning ``None`` when the
+                parameter is not found (default ``False``).
+            store (dict, optional): A ``dict``-like object in which to place
+                the value of the param, but only if the param is found (default
+                ``None``).
+        Returns:
+            datetime.date 
+
+        Raises:
+            HTTPBadRequest: A required param is missing from the request.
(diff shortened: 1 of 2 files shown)


lwcolton:
@jmvrbanac Hows this?

jmvrbanac:
Much better! Thank you! :+1:

mcmasterathl:
This is very likely going to clash with #493.

mcmasterathl:
This looks good to me, @kgriffs says we'll need to go through some revisions on #493 so we can rebase it on top of this.

falcon/request.py
```diff
@@ -838,6 +838,38 @@ def get_param_as_list(self, name,
 
         raise HTTPMissingParam(name)
 
+    def get_param_as_date(self, name, format_string="%Y-%m-%d", required=False, store=None):
```
kgriffs: Nitpick: Use single-quotes for strings

falcon/request.py
```diff
@@ -838,6 +838,38 @@ def get_param_as_list(self, name,
...
+                ``None``).
+        Returns:
+            datetime.date 
+
+        Raises:
+            HTTPBadRequest: A required param is missing from the request.
+            HTTPInvalidParam: A tranform function raised an instance of
+                ``ValueError``.
+        """
+        param_value = self.get_param(name, required=required)
```
kgriffs: If `required` is `False`, and the param is not found, `param_value` will be `None`. In that case, L864 will raise an instance of TypeError. We'll need to handle this case.

tests/test_query_params.py
```diff
@@ -414,6 +418,37 @@ def test_multiple_form_keys_as_list(self):
         # There are three 'cat' keys; order is preserved.
         self.assertEqual(req.get_param_as_list('cat'), ['6', '5', '4'])
 
+    def test_get_date_valid(self):
+        date_value = "20150420"
+        query_string = "thedate={0}".format(date_value)
+        format_string = "%Y%m%d"
```
kgriffs: Would you mind adding a test that doesn't pass a format string (uses the default)? That way, if we ever accidentally mess up the default string so it is different than what we've documented (and what the user expects), we'll catch that with a test?

mcmasterathl:
@kgriffs I fixed this up according to your comments in #498. We can close this one.

Issue #458, opened 2015-02-25
Add support for Jython


jimbaker:
Installed and tested directly with nose (tox doesn't yet support Jython due to missing virtualenv support). I built from trunk (hg.python.org/jython), but you should be able to replicate with the recent beta 4 (http://www.jython.org/). Use jython -m ensurepip to bootstrap, then install using bin/pip, bin/nosetests as usual.

```
======================================================================
ERROR: Failure: ImportError (No module named multiprocessing)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jbaker/jythondev/jython27/dist/Lib/site-packages/nose/loader.py", line 413, in loadTestsFromName
    module = self.importer.importFromPath(
  File "/Users/jbaker/jythondev/jython27/dist/Lib/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/jbaker/jythondev/jython27/dist/Lib/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/jbaker/opensource/falcon/tests/test_request_body.py", line 2, in <module>
    import multiprocessing
ImportError: No module named multiprocessing
```
No multiprocessing module on Jython, use threading module instead.

```
======================================================================
ERROR: Failure: SyntaxError (Unknown encoding: utf-8- (test_utils.py, line 1))
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jbaker/jythondev/jython27/dist/Lib/site-packages/nose/loader.py", line 413, in loadTestsFromName
    module = self.importer.importFromPath(
  File "/Users/jbaker/jythondev/jython27/dist/Lib/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/jbaker/jythondev/jython27/dist/Lib/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/jbaker/opensource/falcon/tests/test_utils.py", line 1

   ^
SyntaxError: Unknown encoding: utf-8-
```

This problem is due to a mangled encoding line; apparently this must be fine on other Python implementations, but still best to fix:

```python
# -*- coding: utf-8-*-
```

```
======================================================================
FAIL: tests.test_middlewares.TestRequestTimeMiddleware.test_log_get_request
[...]

PR #488, opened 2015-04-13
feat(general): support Jython 2.7rc2

csojinb:
~~Trying to see if I can get the tests to run with Jython on Travis. Haven't actually implemented Jython support fixes yet.~~ All tests run for me locally, now, except for coverage, which is not compatible with Jython.

Fixes https://github.com/falconry/falcon/issues/458

- [x] make sure Travis runs tests with Jython
-  ~~make Jython tests run with tox~~ looks like there are still problems with support for virtualenv
- skip coverage tests for Jython -- doesn't cause Travis to fail, though, so maybe okay to just let it be?
- [x] possibly add to docs somewhere?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 sudo: false
-install: pip install tox coveralls --use-mirrors
+install: travis_scripts/install.sh
 cache:
     directories:
         - $HOME/.cache/pip
@@ -18,8 +18,9 @@ env:
   - TOX_ENV=pylint
   - TOX_ENV=py27_smoke
   - TOX_ENV=py27_smoke_cython
+  - JYTHON=true
 
-script: tox -e $TOX_ENV
+script: travis_scripts/run_tests.sh
 after_success: coveralls
 notifications:
   email:
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 *.c
 *.so
 
+# Jython
+*$py.class
+
 # Packages
 *.egg
 *.egg-info
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Kurt Griffiths (kgriffs) is the creator and current maintainer of the Falcon framework. Pull requests are always welcome.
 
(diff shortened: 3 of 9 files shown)


csojinb:
Er, hmm, how do I actually get Travis to run?

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: python
 sudo: false
-install: pip install tox coveralls --use-mirrors
+before_install: if [ "$JYTHON" == "true" ]; then travis_scripts/install_jython2.7.sh; fi
```
csojinb: These single-line-ifs all over the place are so ugly. If someone has thoughts on how to do this better, would love to hear them!

I could pull the ifs and all into `travis_scripts` perhaps.

csojinb:
Is there a way to optionally skip the coverage tests? It would be nice to be able to skip them for Jython, without causing them to not run by default?

csojinb:
Argh. I think it didn't actually run with Jython, otherwise I think the coverage tests would have all failed.

csojinb:
Okay, here's the deal. I was successful in getting Jython tests to run on Travis here: https://travis-ci.org/falconry/falcon/jobs/58451945. Bad news is that this involved running the Jython tests separately from tox -- so, future changes that are jython-incompatible would break Travis build, but it breaks the uniformity of how version support is tested.

I dug into why tox won't run with Jython 2.7. The issue (obscured through many layers) is that, on startup, tox tries to get the version info for each python interpreter. It spins off a subprocess and basically runs `sys.version_info`. For some reason, with Jython 2.7, the message gets returned from the subprocess like `">>> ... ... ... >>> {'version_info': (2, 7, 0, 'candidate', 2)}\n>>> "`. It returns just a string of the dict with Jython 2.5, so it seems like maybe I should go talk to @jimbaker about it.

Concluding thoughts:
Depending on how badly we want Jython integration to get merged, I can clean up the non-tox solution and capture migration to testing Jython with tox as a separate issue. Or, this can just be put on hold until Jython issues are figured out? @kgriffs what do you think?

csojinb:
Looks like someone beat me to the bug by a few hours. http://bugs.jython.org/issue2325

csojinb:
@kgriffs I think this is ready to go, pending review

kgriffs:
LGTM

jmvrbanac:
:+1:

PR #489, opened 2015-04-13
fix(request.range): Require bytes prefix for range

### LINKED ISSUE #387: Make "bytes" prefix in Range header value required
Currently, range header values without a "bytes" prefix are allowed. Change the code to always require a "bytes" prefix.

lwcolton:
The range attribute of the Request object now requires that the value be prefix with 'bytes='

Fix #387
This will be incompatible with previous code that did not prefix the range value with 'bytes='

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -340,9 +340,12 @@ class Request(object):
     @property
     def range(self):
         try:
-            value = self.env['HTTP_RANGE']
+            value = self.env['HTTP_RANGE'].strip()
             if value.startswith('bytes='):
                 value = value[6:]
+            else:
+                msg = "The value must be prefixed with 'bytes='"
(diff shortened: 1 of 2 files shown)


kgriffs:
Going to try and close and reopen to trigger Travis.

mcmasterathl:
LGTM

falcon/request.py
```diff
@@ -340,9 +340,12 @@ def date(self):
     @property
     def range(self):
         try:
-            value = self.env['HTTP_RANGE']
+            value = self.env['HTTP_RANGE'].strip()
```
kgriffs: Web servers typically take care of stripping whitespace from header values, so I don't think this is necessary.

lwcolton:
@kgriffs removed strip(), see #505

PR #490, opened 2015-04-13
fix(falcon/request): HTTP headers as strings

### LINKED ISSUE #413: TypeError raised when passing unicode string literal to set_header (uWSGI)
The following raises an error:

```python
import falcon

class MyResource(object):
    def on_get(self, req, resp):
        resp.set_header('X-Name', u'value')
        resp.body = 'Hello world!'


application = api = falcon.API()

api.add_route('/', MyResource())
```

Traceback:

```
  File "/Users/kgriffs/Code/falcon/falcon/api.py", line 265, in __call__
TypeError: http header value must be a string
```
Work items:
* Add a type check and convert the incoming value to `str`. Fence this code based on python version (no need to add the overhead of type checking in py3k)
* Add a performance warning to the docstring for py2 users

ginoledesma:
Ensure HTTP headers and values are string types per PEP3333. This
fixes wsgi servers running under py2k.

Fixes #413
This doesn't enforce codepoint boundary yet (must be in \u00-\uFF)

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -152,9 +152,8 @@ class Response(object):
                 may be used on platforms that use wide characters.
 
         """
-
         # NOTE(kgriffs): normalize name by lowercasing it
-        self._headers[name.lower()] = value
+        self._headers[str(name).lower()] = str(value)
 
     def append_header(self, name, value):
         """Set or append a header for this response.
@@ -174,11 +173,11 @@ class Response(object):
                 may be used on platforms that use wide characters.
 
         """
-        name = name.lower()
+        name = str(name).lower()
         if name in self._headers:
             value = self._headers[name] + ',' + value
 
-        self._headers[name] = value
+        self._headers[name] = str(value)
 
     def set_headers(self, headers):
         """Set several headers at once.
@@ -209,7 +208,7 @@ class Response(object):
         # normalize the header names.
         _headers = self._headers
         for name, value in headers:
-            _headers[name.lower()] = value
+            _headers[str(name).lower()] = str(value)
 
     def add_link(self, target, rel, title=None, title_star=None,
                  anchor=None, hreflang=None, type_hint=None):
--- a/tests/test_headers.py
+++ b/tests/test_headers.py
@@ -117,6 +117,16 @@ class LocationHeaderUnicodeResource:
         resp.content_location = self.URL1
 
 
+class UnicodeHeaderResource:
+
+    def on_get(self, req, resp):
+        resp.set_headers([
+            (u'X-auTH-toKEN', 'toomanysecrets'),
+            ('Content-TYpE', u'application/json'),
+            (u'X-symBOl', u'\u0040'),
+        ])
+
+
 class VaryHeaderResource:
 
     def __init__(self, vary):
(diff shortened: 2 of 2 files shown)


mcmasterathl:
@kgriffs mentioned in #413 that this should be fenced to only occur on py2 so that py3k users performance isn't affected. However, the RFC says headers should be ASCII regardless of which version of Python you're running. The PR here looks good except for the stipulation about py2/py3k.

LGTM unless @kgriffs wants to limit to py3k. I'm not smart enough to know if not doing it there is OK.

kgriffs:
@mcmasterathl I chatted about this with @ginoledesma yesterday and we decided to just make it work consistently across py2 and py3.

The idea is that by passing through str we both catch strings with non-ascii chars and support setting values to anything that implements `__str__`.

mcmasterathl:
+1

tests/test_headers.py
```diff
@@ -117,6 +117,16 @@ def on_head(self, req, resp):
         resp.content_location = self.URL1
 
 
+class UnicodeHeaderResource:
+
+    def on_get(self, req, resp):
+        resp.set_headers([
+            (u'X-auTH-toKEN', 'toomanysecrets'),
+            ('Content-TYpE', u'application/json'),
+            (u'X-symBOl', u'\u0040'),
```
kgriffs: Would you mind adding a negative test that includes a non-ascii character, and asserting that it does indeed raise an error?

kgriffs:
FWIW, just benchmarked str vs explicity testing using isinstance, and just passing through str() is slightly faster.

falcon/response.py
```diff
@@ -152,9 +152,8 @@ def set_header(self, name, value):
                 may be used on platforms that use wide characters.
 
         """
-
         # NOTE(kgriffs): normalize name by lowercasing it
-        self._headers[name.lower()] = value
+        self._headers[str(name).lower()] = str(value)
```
kgriffs: Would you mind adding a comment here, e.g.:

`PERF(ginoledesma): Using str() is faster that explicitly checking blah blah`

ginoledesma:
I'll get the changes in and submit them in a jiff

kgriffs:
@ginoledesma Just following up. After those two minor comments are addressed, I think this is ready to merge.

tests/test_headers.py
ginoledesma: Ok, adding a negative test shows different behaviors for py2k and py3k, namely:

```python
str(u"\U0001F37A").lower()    # beer
```

This will raise UnicodeEncodeError in py2k but is handled fine in py3k.

I've been going over RFC2616, and the relevant section is Section 4.2 (Message Headers) which references RFC822 section 3.1 (General Description). Best as I can tell, HTTP headers can be encoded in ISO-8859-1/Latin-1.

Not sure how we'll want to proceed. I know we chose str() because it's more performant than type-checking, but I'm not sure how to get consistent behavior without resorting to that.

kgriffs: @ginoledesma  OK, on second thought, this test exposes an issue that is different from #413. I don't think we should try to solve for detecting what types of characters are in the string, because (a) most HTTP header fields only use a subset of US-ASCII and (b) it would require additional run-time cost to do so.

> Historically, HTTP has allowed field content with text in the ISO-8859-1 charset [ISO-8859-1], supporting other charsets only through use of [RFC2047] encoding.  In practice, most HTTP header field values use only a subset of the US-ASCII charset [USASCII]. Newly defined header fields SHOULD limit their field values to US-ASCII octets.  A recipient SHOULD treat other octets in field content (obs-text) as opaque data.

https://tools.ietf.org/html/rfc7230#section-3.2.4

So anyway, let's not add that negative test after all. If we need to add stronger checking later, we can always do that.

kgriffs:
@ginoledesma I'd like to get this merged in the next week if possible. Would you mind rebasing on master? Looks like there may be some conflicts. Once the commit is rebased, you can just push the branch with --force (no need to open another PR). Thanks!

ginoledesma:
Not at all. I'll sort it out by tomorrow.

- gino

kgriffs:
If you don't mind, I'll go ahead and finish this up and get it merged.

kgriffs:
[updated previous comment]

kgriffs:
[merged manually]

PR #493, opened 2015-04-14
fix(request): Decouple form and URL param parsing

### LINKED ISSUE #418: Decouple parsing of form fields and URL params
My application receives requests that have a field named "email" in the URL plus receives a field named "email" in the POST form data.  

Same field name, entirely different context, which presents an issue because Falcon integrates the two fields under one key into req.params.  How can I distinguish one from the other?  Might be worth considering Falcon doing it a different way  - form fields and URL params aren't the same thing so perhaps shouldn't be in the same data structure.

Leading on from this problem - I can't find a way to get the form POST data again because another call to req.stream.read() returns nothing and there's no stream.seek(0) to allow me to rewind to the start of the file so there's no way to manually read the form POST data myself.

Setting aside the issue described above, my application needs to forward the POST form fields onto another server anyway so I need to grab only the POST forms fields, so the automatic parsing and integration of fields into req.params means that there is no way for me to identify which fields in req.params came from POST data and which came from URL params.

I'm kinda stuck point in my development cause I can't implement my functionality cause the POST data is gone and req.params isn't usable for my purpose. 

Ideally Falcon would parse the POST form data fields to a separate data structure to req.params, and would also give the ability to get directly to the POST data even if the form data has been parsed.

thanks

necaris:
Separate form and URL parameter parsing into `request.param` and
`request.form_param`. Introduces new APIs for accessing the data.

Also, do not consume the POST stream unless the data is requested, 
and make the raw POST data available on the request.

Closes #418

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -175,7 +175,8 @@ class Request(object):
             values. Behaves exactly as the ``params`` attribute.
 
         raw_body (bytes): The raw POST data, if any. Note that Falcon will not
-            save the raw body unless the `save_raw_body` option is set.
+            save the raw body unless the `save_raw_body` option is set, in order
+            to minimize unnecessary memory usage.
 
         options (dict): Set of global options passed from the API handler.
 
@@ -556,8 +557,8 @@ class Request(object):
             # PERF(kgriffs): Technically, we should spend a few more
             # cycles and parse the content type for real, but
             # these heuristics will work virtually all the time.
-            content_type = self.content_type or ''
-            if 'application/x-www-form-urlencoded' in content_type:
+            if (self.content_type is not None and
+                    'application/x-www-form-urlencoded' in content_type):
                 self._parse_form_urlencoded()
             else:
                 self.log_error(
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -42,8 +42,6 @@ SimpleCookie = http_cookies.SimpleCookie
 DEFAULT_ERROR_LOG_FORMAT = (u'{0:%Y-%m-%d %H:%M:%S} [FALCON] [ERROR]'
                             u' {1} {2}{3} => ')
 
-TRUE_STRINGS = ('true', 'True', 'yes')
-FALSE_STRINGS = ('false', 'False', 'no')
 WSGI_CONTENT_HEADERS = ('CONTENT_TYPE', 'CONTENT_LENGTH')
 
 
@@ -120,11 +118,9 @@ class Request(object):
         stream: File-like object for reading the body of the request, if any.
 
             Note:
-                If an HTML form is POSTed to the API using the
-                *application/x-www-form-urlencoded* media type, Falcon
-                will consume `stream` in order to parse the parameters
-                and merge them into the query string parameters. In this
-                case, the stream will be left at EOF.
+                Falcon does *not* consume the request stream unless the
+                body data is actually read (see the ``raw_body`` property).
+                When it does, the stream is left at ``EOF``.
 
                 Note also that the character encoding for fields, before
                 percent-encoding non-ASCII bytes, is assumed to be
@@ -175,6 +171,12 @@ class Request(object):
             string, the value mapped to that parameter key will be a list of
             all the values in the order seen.
 
+        form_params (dict): The mapping of form data parameter names to their
+            values. Behaves exactly as the ``params`` attribute.
+
+        raw_body (bytes): The raw POST data, if any. Note that Falcon will not
+            save the raw body unless the `save_raw_body` option is set.
+
         options (dict): Set of global options passed from the API handler.
 
         cookies (dict):
@@ -191,6 +193,8 @@ class Request(object):
         'env',
         'method',
         '_params',
+        '_form_params',
+        '_raw_body',
         'path',
         'query_string',
         'stream',
@@ -198,6 +202,8 @@ class Request(object):
         '_wsgierrors',
         'options',
         '_cookies',
+        '_param',
+        '_form_param'
     )
 
     # Allow child classes to override this
@@ -235,6 +241,14 @@ class Request(object):
                 self.path = path
         else:
             self.path = '/'
+
+        self._params = {}
+        # NOTE(necaris): These are actually initialized and populated
+        # later, on-demand, through properties
+        self._param = None
+        self._form_param = None
+        self._form_params = None
+        self._raw_body = None
 
         # PERF(kgriffs): if...in is faster than using env.get(...)
         if 'QUERY_STRING' in env:
@@ -276,13 +290,6 @@ class Request(object):
                 # this time, it never needs to be wrapped since the server
                 # will continue using the same type for wsgi.input.
                 _maybe_wrap_wsgi_stream = False
-
-        # PERF(kgriffs): Technically, we should spend a few more
-        # cycles and parse the content type for real, but
-        # this heuristic will work virtually all the time.
-        if (self.content_type is not None and
-                'application/x-www-form-urlencoded' in self.content_type):
-            self._parse_form_urlencoded()
 
     # ------------------------------------------------------------------------
     # Properties
@@ -514,6 +521,58 @@ class Request(object):
 
         return self._cookies.copy()
 
+    def param(self):
+        if self._param is None:
+            self._param = helpers.ParamProxy(self, self.params)
+
+        return self._param
+
+    @property
+    def raw_body(self):
+        # NOTE(necaris): This will only store the raw body data if
+        # `self.options.store_raw_body` is True; otherwise it will consume
+        # the body data stream and discard data after returning
+        if self._raw_body is not None:
+            raw_body = self._raw_body
+        else:
+            # NOTE(kgriffs): This assumes self.stream has been patched
+            # above in the case of wsgiref, so that self.content_length
+            # is not needed. Normally we just avoid accessing
+            # self.content_length, because it is a little expensive
+            # to call. We could cache self.content_length, but the
+            # overhead to do that won't usually be helpful, since
+            # content length will only ever be read once per
+            # request in most cases.
+            raw_body = self.stream.read()
+
+            if self.options.store_raw_body:
+                self._raw_body = raw_body
+
+        return raw_body
+
+    @property
+    def form_params(self):
+        if self._form_params is None:
+            # PERF(kgriffs): Technically, we should spend a few more
+            # cycles and parse the content type for real, but
+            # these heuristics will work virtually all the time.
+            content_type = self.content_type or ''
+            if 'application/x-www-form-urlencoded' in content_type:
+                self._parse_form_urlencoded()
+            else:
+                self.log_error(
+                    'Content-Type not supported ({t}).'.format(t=content_type))
+                self._form_params = {}
+
+        return self._form_params
+
+    @property
+    def form_param(self):
+        if self._form_param is None:
+            self._form_param = helpers.ParamProxy(self, self.form_params)
+
+        return self._form_param
+
     # ------------------------------------------------------------------------
     # Methods
     # ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lwcolton:
This PR doubles memory usage in the case where the form data is parsed.  This happens due to storing the raw body and also the parsed form parameters, where previously the raw body was discarded when form parameters were parsed.  The behavior to store the raw body when parsing should be optional.

necaris:
@lwcolton Good point, I'll put that into a `RequestOption`

necaris:
Also, it's worth noting that `falcon-bench` on my machine is about 5% slower on this branch :-(

necaris:
Option added as above and rebased onto upstream master

necaris:
@falconry/owners Are multiple commits in a PR OK, or should I squash this into one commit?

lwcolton:
They will ask you to squash please and thank you

lwcolton:
Also make sure to follow contribution guidelines when putting together your
commit message it needs to have a specific format. There is a
contributing.MD file in the root of the repo with that info

necaris:
Thanks @lwcolton -- I think I've complied with most of the requirements, although I'm not sure about where to put the docstrings for the `get_param*` methods on the `Request`. Since the logic for these has been moved to `request_helpers.ParamProxy` it makes sense to keep the docstring close to the implementation, but since `Request.get_param*` forms part of the public API perhaps the docstring should be left on `Request`?

lwcolton:
Because it is part of the public API I personally would leave the docstring
on Request

On Sat, Apr 18, 2015 at 10:28 PM, Rami Chowdhury <notifications@github.com>
wrote:

> Thanks @lwcolton <https://github.com/lwcolton> -- I think I've complied
> with most of the requirements, although I'm not sure about where to put the
> docstrings for the get_param* methods on the Request. Since the logic for
> these has been moved to request_helpers.ParamProxy it makes sense to keep
> the docstring close to the implementation, but since Request.get_param*
> forms part of the public API perhaps the docstring should be left on
> Request?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/falconry/falcon/pull/493#issuecomment-94230050>.
>



-- 
Sincerely,

Colton Leekley-Winslow
651-242-9559
lwcolton@gmail.com

falcon/request.py
```diff
@@ -166,6 +162,12 @@ class Request(object):
             string, the value mapped to that parameter key will be a list of
             all the values in the order seen.
 
+        form_params (dict): The mapping of form data parameter names to their
+            values. Behaves exactly as the ``params`` attribute.
+
+        raw_body (bytes): The raw POST data, if any. Note that Falcon will not
+            save the raw body unless the `save_raw_body` option is set.
```
kgriffs: I think it would be helpful to explain why it defaults to not buffering/saving the body.

falcon/request.py
```diff
@@ -475,6 +480,59 @@ def headers(self):
...
+
+        return raw_body
+
+    @property
+    def form_params(self):
+        if self._form_params is None:
+            # PERF(kgriffs): Technically, we should spend a few more
+            # cycles and parse the content type for real, but
+            # these heuristics will work virtually all the time.
+            content_type = self.content_type or ''
```
kgriffs: Explicitly checking `is None` (as was done before this change) is slightly more performant.

necaris:
@kgriffs Thanks for the comments! I've made some changes to address your concerns but noticed that the tests are failing after I rebased onto the most recent master -- I'll address these and add a `get_as_date` method to update the PR to the current API.

falcon/request.py
```diff
@@ -475,6 +480,59 @@ def headers(self):
...
 
+    @property
+    def param(self):
+        if self._param is None:
+            self._param = helpers.ParamProxy(self, self.params)
+
+        return self._param
+
+    @property
+    def raw_body(self):
```
kgriffs: I think the most surprising thing with Falcon's form parsing has always been that it consumes req.stream and you can't go back to read it again. To mitigate this problem, we could say "hey, go and use this raw_body attribute over here in that case; otherwise maybe use req.stream as usual?" But I'm concerned about that leading to more confusion about when to use what, and generally making apps harder to reason about.

I understand that adding a more general `raw_body` feature may be useful in some cases, yet I still hesitate to do this without better understanding the use cases from the community (and how common they are). I want to be sure we don't encourage anti-patterns, and that we can provide something more elegant than using a middleware component or hook (which is how you would otherwise accomplish this.) All things considered, I think it would be better to split the general case out into a separate issue/PR from the one currently under discussion.

For now, as an alternative, what do you think about replacing req.stream with a file-like object that has been seeked back to the beginning, and simply wraps an instance of `bytes` that was read in the course of parsing the form body? This would only happen the first time a form param was accessed (thus triggering the consumption and parsing of the original req.stream). Since url-encoded form bodies are generally of reasonable size, this shouldn't cause a significant degradation in memory efficiency. We would have to make sure that the mock stream behaves similarly to the ones provided by, e.g., mod_wsgi, Gunicorn and uWSGI.

falcon/request.py
```diff
@@ -475,6 +480,59 @@ def headers(self):
     def params(self):
         return self._params
 
+    @property
+    def param(self):
```
kgriffs: IIRC, we decided to drop the 's' in order to avoid having to reconcile with the existing `params` property. However, I'm having second thoughts on this. It may be better to make a breaking change for Falcon 1.0 so that we don't carry this design cruft around for the foreseeable future. I suppose ParamProxy could implement the [mapping interface][1] to reduce the impact of this change (as well as a method to convert to a `dict`).

Thoughts?

[1]:https://docs.python.org/3.1/glossary.html#term-mapping

falcon/request.py
```diff
@@ -514,6 +522,58 @@ def cookies(self):
 
         return self._cookies.copy()
 
+    def param(self):
```
kgriffs: IIRC, we decided to drop the 's' in order to avoid having to reconcile with the existing params property. However, I'm having second thoughts on this. It may be better to make a breaking change for Falcon 1.0 so that we don't carry this design cruft around for the foreseeable future. I suppose ParamProxy could implement the mapping interface to reduce the impact of this change (as well as a method to convert to a dict).

Thoughts?

necaris: I'd be in favor of that -- adding the mapping interface to ParamProxy should make it a relatively small change, and definitely acceptable in a major version bump IMHO.

falcon/request_helpers.py
```diff
@@ -129,3 +134,271 @@ def readlines(self, hint=None):
...
+                store[name] = param
+
+            return param
+
+        if not required:
+            return None
+
+        raise HTTPMissingParam(name)
+
+    def get_as_int(self, name,
```
kgriffs: What do you think about dropping the `_as` from these names, i.e.:

```python
req.params.get_int('amount')

req.params.get_list('ids')
req.params.get('name')

req.params.get_bool('detailed')
```

falcon/request.py
necaris: That's an interesting alternative -- I think it could work pretty well. I agree that it's less than optimal to add more _stuff_ to the API surface if it doesn't serve a very clear use case; I wonder if it could be as simple as using a `BytesIO`?

falcon/request_helpers.py
```diff
@@ -129,3 +134,271 @@ def readlines(self, hint=None):
...
+
+
+class ParamProxy(object):
+    """
+    Provide `get_*` methods for both query-string parameters and
+    request body parameters.
+    """
+
+    def __init__(self, request, source):
+        self.req = request
```
kgriffs: Is `self.req` used anywhere?

falcon/request.py
kgriffs: OK, let's try it and see how it looks.

falcon/request.py
kgriffs: That might do it. I just tested on gunicorn under 2.7 and 3.4 and stream.read() returns `bytes`, and `b''` when the stream is at EOF. wsgiref is similar. You may want to check mod_wsgi and uWSGI as well.

kgriffs:
@necaris Will you have some time to work on this during the next week or two?

falcon/request.py
necaris: @kgriffs my aim is to get this done this upcoming weekend, if possible

--
Rami Chowdhury
"Be the change you wish to see in the world." -- Mahatma Gandhi
+1-408-372-6496 / +44-7581-430517 / +88-1778-986280

falcon/request_helpers.py
```diff
@@ -129,3 +134,271 @@ def readlines(self, hint=None):
...
+
+    def get(self, name, required=False, store=None):
+        """Return the raw value of a query string parameter as a string.
+
+        Note:
+            Similar to the way multiple keys in form data is handled,
+            if a query parameter is assigned a comma-separated list of
+            values (e.g., 'foo=a,b,c'), only one of those values will be
+            returned, and it is undefined which one. Use
+            `req.get_param_as_list()` to retrieve all the values.
```
yohanboniface: `req.params.get_as_list()` now :)

falcon/request_helpers.py
yohanboniface: I was about to suggest to drop the `get`, and have `req.params.as_int('amount')` but your suggestion is more consistent with the simple `get` method, so I'm +1.

falcon/request_helpers.py
```diff
@@ -129,3 +134,271 @@ def readlines(self, hint=None):
...
+    """
+    Provide `get_*` methods for both query-string parameters and
+    request body parameters.
+    """
+
+    def __init__(self, request, source):
+        self.req = request
+        self.source = source
+
+    def get(self, name, required=False, store=None):
```
yohanboniface: I suggest we take the opportunity to have a `default` parameter. This pattern is very common and useful:

```python
# Means: return 'fallback' only if 'key' is not defined, else return value even if value is an empty string.
value = req.params.get('key', 'fallback')
# Here 'fallback' will be set even if 'key' exists but is an empty string.
value = req.params.get('param') or 'fallback'
```

I think it's important to distinguish between `key` exist but it's empty and `key` is not given. A use case is resource patching, one could pass an empty string to reset some property.

Drawback here is that ideally `default` should be the first kwarg to make the common pattern shorter (`req.params.get('x', 'y')` rather than `req.params.get('x', default='y')`), but then this would be API breaking change.

falcon/request_helpers.py
```diff
@@ -11,11 +11,16 @@
...
+
+# Globals defined for the benefit of ParamProxy, below
+TRUE_STRINGS = ('true', 'True', 'yes')
+FALSE_STRINGS = ('false', 'False', 'no')
 
 
 def header_property(wsgi_name):
     """Creates a read-only header property.
-
+n
```
yohanboniface: Not sure what this 'n' is. Typo?

falcon/request_helpers.py
```diff
@@ -129,3 +134,271 @@ def readlines(self, hint=None):
...
+
+            if store is not None:
+                store[name] = items
+
+            return items
+
+        if not required:
+            return None
+
+        raise HTTPMissingParam(name)
```
yohanboniface: Suggestion: let's be pythonic and define `__getitem__` and `__contains__`. params being a dict like object, I'm expecting to be able to do:
```python
req.params['key']  # call req.params.get('key', required=True)
if 'key' in req.params:
   # do something
```

falcon/request_helpers.py
```diff
@@ -129,3 +134,271 @@ def readlines(self, hint=None):
...
+                falls outside the given interval, i.e., the value must be in
+                the interval: min <= value <= max to avoid triggering an error.
+
+        """
+
+        params = self.source
+
+        # PERF: Use if..in since it is a good all-around performer; we don't
+        #       know how likely params are to be specified by clients.
+        if name in params:
```
yohanboniface: This does not seem very DRY to me, given that it's done in `.get` already. Can't we call `.get` instead, then do int specific checks, and remove `param` from `store` if `store` was given and some check fails?

BTW, I'm not sure the `store` kwarg pattern is really strong, cf my comment about *defined* vs *empty*. Is it something we really want to keep?

falcon/request.py
```diff
@@ -514,6 +522,58 @@ def cookies(self):
...
+                self._parse_form_urlencoded()
+            else:
+                self.log_error(
+                    'Content-Type not supported ({t}).'.format(t=content_type))
+                self._form_params = {}
+
+        return self._form_params
+
+    @property
+    def form_param(self):
```
yohanboniface: I suggest to use `form` instead of `form_params`: it's shorted but meaningful.

yohanboniface:
As we are using lazy proxy objects, here is a suggestion:
- use `req.query` for query string parameters
- use `req.form` for form like parameters
- keep `req.params` as a merge of both: this will not break the current API plus keep the convenient shortcut.

And let's keep in mind that we'll need `req.files` at some point.

falcon/request_helpers.py
kgriffs: I think this makes sense; the only downside is that magic methods a slightly slower than the alternative. As long as we mention this in the docs, should be OK.

falcon/request.py
```diff
@@ -1021,12 +829,14 @@ def _parse_form_urlencoded(self):
                            'will be ignored.')
 
         if body:
-            extra_params = parse_query_string(
+            form_params = uri.parse_query_string(
                 body,
                 keep_blank_qs_values=self.options.keep_blank_qs_values,
             )
 
-            self._params.update(extra_params)
```
kgriffs: Since `self._params` no longer contains form params, we will need to modify `req.params` to avoid a breaking change.

falcon/request_helpers.py
kgriffs: > This does not seem very DRY to me

I'm also a big fan of DRY. However, we may want to keep this part of the code WET since it avoids an extra function call (and possibly a few additional operations).

> I'm not sure the store kwarg pattern is really strong

It can save some typing when composing options to pass to a different layer of the app, but that aside, I'm reluctant to make a breaking change since it would break some existing apps out there.

falcon/request_helpers.py
kgriffs: I like this idea. Good point about empty strings. We can get away with making default the first kwarg without introducing a breaking change, since it would be implemented only for the new attributes (i.e., `req.query` and `req.form`) while leaving the deprecated interface alone (i.e., `req.get_param()`)

falcon/request.py
philiptzou: The name `form_param` is confusion with `form_params`..

philiptzou:
Perhaps we should allow customization and extension of `ParamProxy` class. For example I may want to use [voluptuous](https://github.com/alecthomas/voluptuous) to validate the input body. Also the `content-type` can be others for example "multipart/form-data" or "application/json" but not only the "application/x-www-form-urlencoded".

```python

class Request(object):

    _param_proxy_class = helpers.URLEncodedParamProxy
    # so we can easily custom the param proxy by inherit the AbstractParamProxy class
    # which is neutral to any serialization formats and any validators

    ...

    @property
    def form(self):
        if self._form is None:
            self._form_param = self._param_proxy_class(self, self.raw_body)
        return self._form_param

    def set_param_proxy_class(self, klass):
        self._param_proxy_class = klass

```

So why not we decouple `_parse_form_urlencoded` from `Request` too?

philiptzou:
@necaris Would you still like to follow up the comments of this pull-request? Or I can take over of it since I am interested in and if you don't mind :)

necaris:
@philiptzou I don't have a clear plan to implement all the API changes requested recently so I'd be just fine with you taking over this branch :+1:

philiptzou:
@necaris I now understand why you don't have a clear plan... wow

kgriffs:
@philiptzou This may not not be the best place to create an extension method for supporting arbitrary body media types. It isn't really a "form" at that point in the sense of the HTML-centric intent of this interface. I'd like to see how this would compare and contrast with parsing body documents via middleware methods or hooks.

For now I like @yohanboniface's suggestion of adding `req.query` and `req.form` alongside `req.params`. `req.form` and `req.params` would only consume the request body stream the first time either is called. Once consumed, the body would be buffered into a BytesIO instance which would then be returned by `req.stream` in lieu of the original `req.env['wsgi.input']`.

@philiptzou Were you still interested in taking this on? In preparation for the imminent Falcon 1.0, I'm considering putting up a PR to just add the BytesIO buffer, since that could be considered a subtle breaking change. Meanwhile, you could work on adding the additional, non-breaking  `req.query` and `req.form` attributes in a 1.1 or something (assuming semantic versioning). Finally, in Falcon 2.0 we could consider removing `req.params`.

philiptzou:
@kgriffs: Actually I have an unfinished `BytesIO` reversion at now. I think I'm just too caution to minimize resources usage and performance impact. I mean if you want to use plain `BytesIO` it will always double the memory usage as @lwcolton worried (https://github.com/falconry/falcon/pull/493#issuecomment-92887955). And we'll also lose the advantages of stream reading (had to wait all request body arrived).

Alternatively, We can implement a compatible interface of `BytesIO` with proxy the original file-like object from WSGI. The advantages is we can still do stream reading, but the proxy will affect performance and there are a lot code to be written since we are going to implement an interface.

I have not decided yet which one should be implemented. And I don't think I have a clear idea. That's the reason why I need some more discussion but I'm still interested in taking this on.

kgriffs:
I can understand the trepidation, and I don't see a problem with making the buffering optional via `RequestOptions`. However, I suspect that most `application/x-www-form-urlencoded` strings are not very long, and so would not incur a significant memory overhead when buffered. Let's suppose that the lengths of form bodies in a given app average 256 bytes. Even if one of my web heads is serving an extremely large number of simultaneous requests/sec, say 10,000, the app would only use about 2 mb of RAM. 

Likewise, with such small bodies, I don't think streaming is going to be much of a concern. Since the body will usually fit in one or two TCP packets, all the data will already have arrived by the time the app attempts its first read.

Regardless, I've been thinking that we need to step back and re-examine our assumptions. I'm still not sure that I completely understand why people have been wanting to re-parse form bodies. Here are a few options:

* To deal with collisions between field names in the query string and in the form. But this doesn't make a lot of sense, since form params are parsed *after* the query string, meaning the app would need to re-parse the raw query string to get the clobbered value, not the request body.
* To work around some problem with the way Falcon parses `application/x-www-form-urlencoded`. Ideally this should be addressed by fixing the problem directly. 

Whatever the reason, if the request body were only parsed on-demand (i.e., when the hypothetical `req.form` is first called), then at least the application would have some control over whether or not (or when) the body is consumed.

With that in mind, it may be better to just go ahead with the plan to disentangle the interfaces for form and query string parsing. Parse params just-in-time (the first time requested), and don't replace req.stream with an instance of BytesIO or similar. If the community feedback still indicates a need for buffering the body, we can cross that bridge when we come to it.

kgriffs:
Another thought on this... perhaps we can provide robust HTML form handling via a standalone add-on (e.g., "falcon-forms"), vs. being part of the core framework. We could add an option to disable parsing form params as currently implemented, in the case that a developer would like to handle it in a different way.

kgriffs:
Update: I plan on discussing this during the PyCon sprints. I'd like to settle on a design direction and get this finished up.

kgriffs:
P.S. See also: https://github.com/yohanboniface/falcon-multipart/blob/master/falcon_multipart/middleware.py#L25

lwcolton:
@kgriffs You previously commented that you liked the idea of lazy-access request attributes for form data, so that on first access the request stream is consumed, used to populate the form data and files request attributes, and not saved or wrapped.  This is what I have seen in other frameworks and allows a developer to consume the stream themselves if they wish.  Would that be acceptable here?  Would love to help this finally get merged

kgriffs:
@lwcolton Cool, thanks for following up on this. I've also been thinking about this and finally sat down today to write up a proposal. I'd love to get your thoughts on what I just posted to #418. I'm going to close this PR since it is so old. If this is something you'd like to work on, let me know. Otherwise we'll see if anyone else has some bandwidth to take and run with it.

PR #494, opened 2015-04-14
fix(util) Add support for obsolete HTTP date formats

### LINKED ISSUE #255: Support obsolete HTTP date formats
In looking at [RFC 7231](http://tools.ietf.org/html/rfc7231#section-7.1.1.1), Falcon date parsing is actually restricted to IMF-fixdate and does not support the other two obsolete formats. Should we support those formats? Doing so may have minor performance implications for little benefit.

Otherwise, we should update the docstrings and such to reference IMF-fixdate instead of HTTP-Date.

mcm:
Update http_date_to_dt to support obsolete date formats per RFC 7231.

Fixes #255

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -617,12 +617,13 @@ class Request(object):
 
         try:
             http_date = self.get_header(header, required=required)
-            return util.http_date_to_dt(http_date)
+            return util.http_date_to_dt(http_date, obs_date=True)
         except TypeError:
             # When the header does not exist and isn't required
             return None
         except ValueError:
-            msg = ('It must be formatted according to RFC 1123.')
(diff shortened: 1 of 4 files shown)


mcmasterathl:
I will rebase this when #495 is merged, as I am pretty sure they conflict.

jmvrbanac:
@mcmasterathl when you have time, could you rebase your change? Thanks!

mcmasterathl:
Yup, was working on that right now actually.

mcmasterathl:
@jmvrbanac All set

jmvrbanac:
:+1: LGTM

kgriffs:
I think we could optimize this a little bit, but that can be done in another patch.

PR #495, opened 2015-04-14
fix(Request): Return date headers as datetime objects

### LINKED ISSUE #254: Return if-modified-since, if-unmodified-since as datetime.datetime

lwcolton:
Change if-modified-since and if-unmodified-since headers to datetime
objects instead of strings.

BREAKING CHANGE: The 'date', 'if_modified_since', and 'if_unmodified_since'
request parameters are now returned as datetime.datetime objects and not strings

Fixes #254

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -146,10 +146,10 @@ class Request(object):
             header is missing.
         if_none_match (str): Value of the If-None-Match header, or ``None``
             if the header is missing.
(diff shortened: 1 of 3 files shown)


kgriffs:
@lwcolton can you rebase?

lwcolton:
@kgriffs Done, see #504

PR #497, opened 2015-04-14
doc(install) Warn about python updates when using cython

### LINKED ISSUE #476: [docs] Warn about python updates when using cython
I just stumbled across a simple issue in falcon that I didn't expect and it took me half an hour to fix.

If falcon is compiled against cython to improve the speed, any updates in the system to python require a recompilation of the cython and falcon packages for precaution.

The exact problem that I had was this:

    File "/srv/sikr_live/sikre/app.py", line 15, in <module>
    
        import falcon

    File "/srv/sikr_live/venv/lib/python3.4/site-packages/falcon/__init__.py", line 31, in <module>
    
        from falcon.version import __version__  # NOQA
    
    ImportError: /srv/sikr_live/venv/lib/python3.4/site-packages/falcon/version.cpython-34m.so:
        undefined symbol: AnnotateIgnoreReadsEnd     

This was after an update to my system python (3.4.3) and uwsgi so I don't know which one to blame, but it seems more likely that the python update is responsible and somehow the cython compilation that I had failed for that reason.

My point is, maybe we should have a warning in the docs stating that a python update can trigger this.

UPDATE: Of course the soluction is a simple `pip install --force-reinstall --upgrade cython` and afterwards the same for falcon

mcmasterathl:
Closes #476

### DIFF
--- a/doc/user/install.rst
+++ b/doc/user/install.rst
@@ -21,6 +21,16 @@ type:
 
     $ pip install --upgrade falcon
 
+.. note::
+
+    When using Cython, you should always recompile Falcon after
+    upgrading Python. To do this, simply run:
+
+    .. code:: bash
+
+        $ pip install --force-reinstall --upgrade cython
+        $ pip install --force-reinstall --upgrade falcon
+
 Installing Cython on OS X
 -------------------------
 


doc/user/install.rst
```diff
@@ -21,6 +21,14 @@ type:
 
     $ pip install --upgrade falcon
 
+When using Cython, you should always recompile Falcon after upgrading Python.
```
jmvrbanac: I think it might be clearer if you put this change into a ```.. note::``` block

mcmasterathl:
I moved the comment into a ..note block for clarity and emphasis.

jmvrbanac:
:+1:

kgriffs:
LGTM, thanks!

PR #500, opened 2015-04-14
fix(errors): Reuse '100 Continue' string

kgriffs:
Alias HTTP_CONTINUE to HTTP_100 rather than setting it to separate string.


jmvrbanac:
:+1:

PR #502, opened 2015-04-14
feat(api) Add HTTPStatus exception for immediate response

### LINKED ISSUE #414: How do I force an immediate response?
In Falcon I can force an immediate error code response to return from anywhere in my code by raising a Falcon exception with the appropriate error code.

Is there a way to force an immediate return of the response to the client for VALID (i.e. not errors) responses?

mcmasterathl:
Add an HTTPStatus exception that is handled similarly to HTTPError in
terms of hooks and middleware, but does not pass through error handling.

Fixes #414

### DIFF
--- a/falcon/http_status.py
+++ b/falcon/http_status.py
@@ -1,0 +1,45 @@
+# Copyright 2015 by Hurricane Labs LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+class HTTPStatus(Exception):
+    """Represents a generic HTTP status.
+
+    Raise this class from a hook, middleware, or a responder to stop handling
+    the request and skip to the response handling.
+
+    Attributes:
+        status (str): HTTP status line, e.g. '748 Confounded by Ponies'.
+        headers (dict): Extra headers to add to the response.
+        body (str or unicode): String representing response content. If
+            Unicode, Falcon will encode as UTF-8 in the response.
+
+    Args:
+        status (str): HTTP status code and text, such as
+            '748 Confounded by Ponies'.
+        headers (dict): Extra headers to add to the response.
+        body (str or unicode): String representing response content. If
+            Unicode, Falcon will encode as UTF-8 in the response.
+    """
+
+    __slots__ = (
+        'status',
+        'headers',
+        'body'
+    )
+
+    def __init__(self, status, headers=None, body=None):
+        self.status = status
+        self.headers = headers
+        self.body = body
--- a/tests/test_httpstatus.py
+++ b/tests/test_httpstatus.py
@@ -1,0 +1,196 @@
+# -*- coding: utf-8
+
+import falcon.testing as testing
+import falcon
+from falcon.http_status import HTTPStatus
+
+
+def before_hook(req, resp, params):
+    raise HTTPStatus(falcon.HTTP_200,
+                     headers={"X-Failed": "False"},
+                     body="Pass")
+
+
+def after_hook(req, resp, resource):
+    resp.status = falcon.HTTP_200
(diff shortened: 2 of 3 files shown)


falcon/http_status.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2013 by Rackspace Hosting, Inc.
```
kgriffs: Since you created this file, you can put your own name or the name of your employer (depending on your company's IP policy) here instead of Rackspace.

falcon/http_status.py
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+class HTTPStatus(Exception):
+    """Represents a generic HTTP status.
+
+    Raise this class from a hook, middlware, or a responder to stop handling
```
kgriffs: Typo: Missing 'e' in 'middlware'

tests/test_httpstatus.py
```diff
@@ -0,0 +1,68 @@
...
+                         body="Pass")
+
+    @falcon.after(after_hook)
+    def on_put(self, req, resp):
+        raise HTTPStatus(falcon.HTTP_500,
+                         headers={"X-Failed": "True"},
+                         body="Fail")
+
+
+@ddt.ddt
```
jmvrbanac: Currently, it doesn't look like you're using ddt in this class. I would either use it to reduce the duplication between the 3 test methods or remove it.

falcon/http_status.py
```diff
@@ -0,0 +1,45 @@
...
+
+    __slots__ = (
+        'status',
+        'headers',
+        'body'
+    )
+
+    def __init__(self, status, headers=None, body=None):
+        self.status = status
+        self.headers = headers or {}
```
kgriffs: Since L489 checks for `None`, would it make sense to just directly assign without the conditional?

tests/test_httpstatus.py
```diff
@@ -0,0 +1,68 @@
...
+class TestHTTPError(testing.TestBase):
```
kgriffs: Would you mind adding a test that ensures a middleware process_response method does not get skipped when HTTPStatus is raised?

tests/test_httpstatus.py
```diff
@@ -0,0 +1,68 @@
...
+        body = self.simulate_request('/status', method='POST', decode='utf-8')
+        self.assertEqual(self.srmock.status, falcon.HTTP_200)
+        self.assertIn(('x-failed', 'False'), self.srmock.headers)
+        self.assertEqual(body, 'Pass')
+
+    def test_raise_status_runs_after_hooks(self):
+        body = self.simulate_request('/status', method='POST', decode='utf-8')
+        self.assertEqual(self.srmock.status, falcon.HTTP_200)
+        self.assertIn(('x-failed', 'False'), self.srmock.headers)
+        self.assertEqual(body, 'Pass')
```
kgriffs: I'd like to see one more test where the hook does not override the middleware, just to prove that the body set in HTTPStatus is returned as expected.

kgriffs: Finally, one more test to add. Let's check that an instance of HTTPStatus where body is `None` results in an empty body being returned to the client.

tests/test_httpstatus.py
```diff
@@ -0,0 +1,196 @@
...
+        self.api.add_route('/status', self.resource)
+
+        body = self.simulate_request('/status', method='GET', decode='utf-8')
+        self.assertEqual(self.srmock.status, falcon.HTTP_200)
+        self.assertIn(('x-failed', 'False'), self.srmock.headers)
+        self.assertEqual(body, 'Pass')
+
+    def test_raise_status_survives_after_hooks(self):
+        """ Make sure after hook doesn't overwrite our status """
+        self.api = falcon.API(after=[noop_after_hook])
```
kgriffs: The global after hook would be expected to overwrite the status. Therefore, it seems like this test should use an after hook that attempts to do so, and ensures it succeeds, as in `test_raise_status_runs_process_response`.

kgriffs:
Just one comment. Otherwise LGTM.

tests/test_httpstatus.py
mcmasterathl: L129 tests for this. The GET responder sets HTTP_500, and the after_hook overwrites it. L139 is testing to ensure that if responder raises HTTPStatus, using an after hook doesn't implicitly override the response.

kgriffs:
Thanks!

Issue #430, opened 2015-02-04
Add support for setting the realm to the HTTPUnauthorized class


kgriffs:
Currently the class recognizes a *scheme* kwarg that can include a realm string, but IMO it would be better to make this explicit. 

See also RFC 7235.

mcmasterathl:
I can work on this.

@kgriffs it looks like the RFC says you can have multiple comma-separated challenges in a single header, or multiple headers. Each challenge is a scheme and zero or more comma separated key=value pairs. Also, optionally, the key=value pairs could be base64 encoded.

In order to implement all of this, I suggest the following:
* Add a new kwarg for accepting a list of valid challenges. Each challenge would need a "scheme" item, and the other items would be used in comma-separate key=value format.
* Add a new kwarg to control whether to base64 encode the key=value options
* Add a new kwarg to control whether to set a single or multiple WWW-Authenticate parameters

I'm not sure whether this is the best way to implement though, so any feedback before I dive in would be appreciated. I'm also not sure what backwards-compatibility would look like, or if you want to drop it since this is a future milestone feature.

kgriffs:
@mcmasterathl  Sounds good. I think someone else started work on this, and was taking the approach of just taking a list of opaque challenge strings. So it just leaves the formatting of each challenge to the caller. This is probably OK for 0.3, but we may want to do something more sophisticated (as recommended above) for 0.4 or 0.5. 

Let's see if a PR comes in for this and then we may want to create an additional issue to look at doing more, as previously mentioned.

ueg1990:
hey Kurt, i will try and send a PR by today. After that you can let me know if you need anymore changes on that.

PR #503, opened 2015-04-14
Add support for setting the realm to the HTTPUnauthorized class

ueg1990:
Based on RFC 7235, add challenges to HTTP Unauthorized to indicate
authentication scheme. A server generating a 401 must send a WWW-Authenticate
header field containing at least one challenge

BREAKING CHANGE: Add new parameter to the __init__ method of class HTTPUnauthorized
to add determine authentication scheme in the WWW-Authenticate header field when a
401 (Unauthorized) is generated

### DIFF
--- a/falcon/errors.py
+++ b/falcon/errors.py
@@ -49,18 +49,18 @@ class HTTPUnauthorized(HTTPError):
         title (str): Error title (e.g., 'Authentication Required').
         description (str): Human-friendly description of the error, along with
             a helpful suggestion or two.
-        scheme (str): Authentication scheme to use as the value of the
-            WWW-Authenticate header in the response (default ``None``).
+        challenges (iterable of str): One or more authentication
+            challenges to use as the value of the WWW-Authenticate header in
+            the response.
         kwargs (optional): Same as for ``HTTPError``.
 
     """
 
-    def __init__(self, title, description, **kwargs):
+    def __init__(self, title, description, challenges, **kwargs):
         headers = kwargs.setdefault('headers', {})
 
-        scheme = kwargs.pop('scheme', None)
-        if scheme is not None:
-            headers['WWW-Authenticate'] = scheme
+        if challenges:
+            headers['WWW-Authenticate'] = ', '.join(challenges)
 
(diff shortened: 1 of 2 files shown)


kgriffs:
See also: Issue #430

kgriffs:
@ueg1990 Thanks for this! Would you mind ammending your commit message so the first line includes a  prefix ala:

`fix(HTTPUnauthorized): blah blah...`

Also, please reference the issue in the footer on a line by itself, as in:

`Fixes #430`

Thanks!

falcon/errors.py
```diff
@@ -49,18 +49,17 @@ class HTTPUnauthorized(HTTPError):
         title (str): Error title (e.g., 'Authentication Required').
         description (str): Human-friendly description of the error, along with
             a helpful suggestion or two.
-        scheme (str): Authentication scheme to use as the value of the
-            WWW-Authenticate header in the response (default ``None``).
+        challenges (iterable of strings): Authentication scheme to use as the
```
kgriffs: "Authentication scheme" => "One or more authentication challenges to use as the value of the..."

kgriffs:
@ueg1990 Looks like there's a minor pep8 error.

jmvrbanac:
:+1: LGTM

falcon/errors.py
```diff
@@ -49,18 +49,18 @@ class HTTPUnauthorized(HTTPError):
         title (str): Error title (e.g., 'Authentication Required').
         description (str): Human-friendly description of the error, along with
             a helpful suggestion or two.
-        scheme (str): Authentication scheme to use as the value of the
-            WWW-Authenticate header in the response (default ``None``).
+        challenges (iterable of strings): One or more authentication
```
kgriffs: "strings" => "str"

kgriffs:
@ueg1990  Just one nitpick to address, and then I think this is ready to merge.

kgriffs:
@ueg1990 In preparation for the merge, would you mind squashing all the patches down to a single commit? You cant `git push --force` to update the PR in place.

PR #504, opened 2015-04-15
fix(Request): Return date headers as datetime objects

### LINKED ISSUE #254: Return if-modified-since, if-unmodified-since as datetime.datetime

lwcolton:
Change if-modified-since and if-unmodified-since headers to datetime
objects instead of strings.

BREAKING CHANGE: The 'date', 'if_modified_since', and 'if_unmodified_since'
request parameters are now returned as datetime.datetime objects and not strings

Fixes #254

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1,4 +1,4 @@
-# Copyright 2013 by Rackspace Hosting, Inc.
+    # Copyright 2013 by Rackspace Hosting, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


falcon/request.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2013 by Rackspace Hosting, Inc.
+    # Copyright 2013 by Rackspace Hosting, Inc.
```
kgriffs: pep8 is complaining about this indent

lwcolton:
@kgriffs see #511

PR #505, opened 2015-04-15
fix(request.range): Require bytes prefix for range

### LINKED ISSUE #387: Make "bytes" prefix in Range header value required

lwcolton:
The range attribute of the Request object now requires that the value be prefix with 'bytes='

Fix #387
This will be incompatible with previous code that did not prefix the range value with 'bytes='

remove strip()


kgriffs:
LGTM, thanks!

PR #506, opened 2015-04-15
docs(api): Add info to docstring about resource middleware

lwcolton:
The docstring for the api class has info about middleware.
The part about resource middleware was a little non-descriptive.

Change to docstring


jmvrbanac:
:+1:

Issue #460, opened 2015-03-02
Force HTTPS


kgriffs:
Add support for enabling HSTS?

See also: http://tools.ietf.org/html/rfc6797

Perhaps this could simply be a flag or something that is enabled in API options, then applied to every response?

bryson:
Wouldn't this be better set by the webserver, rather than the framework? For example, in Nginx's config: `add_header  Strict-Transport-Security "max-age=31536000; includeSubdomains;";`

oscarcp:
I'm always of the opinion that HSTS has to be enforced by the web server instead (nginx, Apache...) but in the case of allowing Falcon to send the HSTS header, your approach would be the best, habing an extra parameter on the API class to activate it.

kgriffs:
@oscarcp Actually, now that you say that, I wonder if this would be considered a "hop-by-hop" header, and therefore WSGI apps shouldn't be touching it anyway...

oscarcp:
I've been digging through this for a couple hours. It seems that WSGI apps are not meant to handle it, but there is a bunch of apps for flask and django that handle HSTS on a URL basis, which is odd, because the premise of HSTS is to force HTTPS connection on the whole domain while the HSTS set is alive on the browser (default: 1 year), rendering those apps useless after the first request.

My bet is leaving this out of Falcon, let it be handled by the webserver unless we find a use case that requires it (in which case can be also implemented via middleware that sets the header)

In my personal use case enforce HTTPS at three points: first, a redirect in nginx from http to https, then HSTS also in nginx and last a middleware (I'm using falcon 0.2) that enforces HTTPS connection or the application doesn't respond.

kgriffs:
Ah, I wonder if it would be useful then to just have a simple "require HTTP" feature in the framework that does what your middleware does? We'd have to make it very clear that it is only a "last resort" and that doing the redirect and HSTS stuff is the first line of defense.

oscarcp:
That's actually a good idea, having an option that forces HTTPS communication is quite common in backends :)

kgriffs:
Note: We'll need to make sure our HTTPS detection logic takes into account `X-Forwarded-Proto` as well as the `Forwarded` header ([RFC 7239](https://tools.ietf.org/html/rfc7239)).

kgriffs:
[...]

PR #514, opened 2015-04-15
Add option to block non-https requests.

paul291:
This is for part of #460. 

Added enforce_https flag to API. If true, middleware is added that
checks whether https is the url scheme or the value of
X_FORWARDED_PROTO header. If it is not, an exception is raised
that can be caught be a user provided error handler.

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -88,6 +88,17 @@ class API(object):
             instead of Falcon's default class. (default
             ``falcon.response.Response``)
 
+        enforce_https (bool, False): Whether to allow only https requests.
+            If 'https' is not the url scheme, the value of the
+            X-Forwarded-Proto header, or the first proto value in the Forwarded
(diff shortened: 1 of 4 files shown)


paul291:
Oops left out new test file in commit. Just amended commit with it.

kgriffs:
Over the past several months there has been a lot of community discussion around where add-ons AKA plugins should live. I think we are coming to a rough consensus that the core framework should be as lean as possible, but that for additional functionality that is quite common/important, we could create standalone "official" packages with repos under the GH falconry org. Each of those would have a project technical lead.

@paul291 With that in mind, what do you think about creating a package names something like "falcon-appsec" that could have this middleware, and perhaps other middleware over time, such as for input validation or whatever?

kgriffs:
If someone wants to maintain this as a standalone package, please let me know.

kgriffs:
Superseded by https://github.com/falconry/falcon-require-https/pull/5

Issue #25, opened 2013-01-23
Development server


kgriffs:
(Copied by @vytas7 from the first comment of #1056 )

The updraft development server: https://github.com/falconry/updraft

**Desired Features**:
* Auto-reload
* Full pdb support
* Legible Tracebacks

**Notes:**
*  Should probably log out a note about this being a development server.

kevinpaladin:
+1

csojinb:
I'll take a look at this. I'll probably be slow at it though. :-)

vytas7:
Normally, it would be unclear if this provides much value over just using Gunicorn (autoreload is supported) or similar as a debugging server.

However, now that we are going to support both WSGI and ASGI, could it be worth revisiting this?
We could combine `wsgiref` and `asgiref` into one tool that would automatically recognize which flavour of Falcon (or a non-Falcon app, for that matter) we're dealing with, and would start serving accordingly.

vytas7:
We will rewrite [Falconry Updraft](https://github.com/falconry/updraft), started by @csojinb, to use a something similar stack to Hypercorn (`h11` + `wsproto` or similar) in order to support both WSGI and ASGI.

I will keep this issue open for documentation and eventual integration if Updraft is detected in the current environment, but I think we'll want to keep it as a separate package (for now at least).

PR #517, opened 2015-04-16
Debug server -- rough draft

csojinb:
I'm putting this up as a 30% draft for some design feedback. I pulled over the simple server with reloading from werkzeug (with no attention paid yet to having the correct attributions or docstrings or anything) and added the traceback to the response and to the debug log in the console. I also added a simple middleware to drop you into pdb on exception if you want.

The usage is Flask-like:
```
# start.py

from things import app

app.run_dev_server(use_debugger=True)  # use_debugger defaults to False
```

There are no tests yet, and the server code may be more than we need. I erred on the side of not removing things from the werkzeug server, as I haven't fully dug into what it's doing yet.

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -248,7 +248,8 @@ class API(object):
         start_response(resp.status, headers)
         return body
 
-    def run_dev_server(self, hostname=None, port=None, use_reloader=True):
+    def run_dev_server(self, hostname=None, port=None,
+                       use_reloader=True, use_debugger=False):
         from falcon.dev_server.serving import run_simple
 
         if hostname is None:
@@ -256,7 +257,8 @@ class API(object):
         if port is None:
             port = 5000
 
-        run_simple(hostname, port, self, use_reloader=use_reloader)
+        run_simple(hostname, port, self, use_reloader=use_reloader,
+                   use_debugger=use_debugger)
 
     def add_route(self, uri_template, resource):
         """Associates a templatized URI path with a resource.
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -248,6 +248,16 @@ class API(object):
         start_response(resp.status, headers)
         return body
 
+    def run_dev_server(self, hostname=None, port=None, use_reloader=True):
+        from falcon.dev_server.serving import run_simple
+
+        if hostname is None:
+            hostname = '127.0.0.1'
+        if port is None:
+            port = 5000
+
+        run_simple(hostname, port, self, use_reloader=use_reloader)
+
     def add_route(self, uri_template, resource):
         """Associates a templatized URI path with a resource.
 
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -27,6 +27,7 @@ import falcon.status_codes as status
 
 
 class API(object):
+
     """This class is the main entry point into a Falcon-based app.
 
     Each API instance provides a callable WSGI interface and a routing engine.
--- a/falcon/bench/bench.py
+++ b/falcon/bench/bench.py
@@ -192,7 +192,7 @@ def main():
         'falcon-ext',
         'flask',
         'pecan',
-        'werkzeug'
+        'falcon.dev_server'
     ]
 
     parser = argparse.ArgumentParser(description="Falcon benchmark runner")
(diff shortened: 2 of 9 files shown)


jmvrbanac:
@kgriffs should we create another repo (fork of werkzeug), or is putting this within the normal falcon code  base ok?

csojinb:
Depending on how sophisticated of a dev server we want to have, I can probably strip it down further.

@jmvrbanac is your thinking that the dev server stuff would be like an add-on kinda deal?

kgriffs:
@csojinb, @jmvrbanac: I think this is on the right track. Having the dev server in a standalone repo/package would allow us to take additional dependencies over time to add more features. Yes, it adds a little bit of additional administration overhead, but shouldn't be too bad. If you agree, I can go ahead and create the repo. Just need a catchy name. ;)

Regardless, the way I would personally use this tool is from the command line (via entry point), basically in place of a production server such as gunicorn, when I want:

* Automatically reload code when it changes
* Drop into pdb on the CLI (bonus points if we also support ipdb) when an unhandled error is raised
* Potentially also drop into a debugger when certain status codes are raised (e.g., 5xx)
* Other nifty things that we can add over time as we think of them. :D

```python
# mything.py
import falcon

api = falcon.API()
```

```bash
$ pip install falcon-trainer-thing
$ falcon-trainer-thing mything:api
```

Thoughts?

jmvrbanac:
@kgriffs I think this is a great idea. How about falcon-aviary for a package name?

csojinb:
@kgriffs @jmvrbanac are you thinking the separate package would still just have the reloader that's cut out of werkzeug? 

Thoughts on new repo name:

(in the sense of "running" falcon)
* a person who flies falcons is a falconer (this is probably too similar)
* a person who flies hawks is an austringer
* maybe updraft? falcons fly on updrafts?

(types of falcon -- on theme, but not relationally correct)
* peregrine
* kestrel

csojinb:
(I mean, have the reloader cut out of werkzeug as opposed to having a reloader lib dependency)

kgriffs:
> (I mean, have the reloader cut out of werkzeug as opposed to having a reloader lib dependency)

I guess the question is, can we use it as-is, or do we need to make some changes? I kind of like the idea of forking the code rather than requiring werkzeug just for the reloader, but it does mean we take on the maintenance burden. That being said, I can't imagine this part of the code needing many changes in the future.

If we do want to fork it, we'll want to acknowledge where it came from (werkzeug, which is MIT licensed, and copyrighted by Lukáš Rychtecký) in a comment in the module or whatever.

> Thoughts on new repo name:

`updraft` sounds kinda neat. It fits with the theme, and also makes sense since you are working on a "draft" of your project. @jmvrbanac what do you think?

csojinb:
I'm in favor of forking and cutting it down to what we need. The reloader
works out of the box, but werkzeug has that fancy in-browser debugger
that's basically useless if your app isn't actually serving webpages. The
reloader seems pretty solid, so hopefully not too much of a maintenance
burden. (And, yeah, doing the attribution is no problem. Didn't include it
here because this PR was meant to be a rough cut.)

I hadn't thought of the potential dual meaning of updraft. I like it!

(Mostly unrelatedly, in researching potential repo names, I learned that a
lot of common English idioms [come from falconry](
https://en.wikipedia.org/wiki/Falconry#English_language_words_and_idioms_derived_from_falconry)
)

jmvrbanac:
@kgriffs @csojinb I like it! :100:

kgriffs:
@csojinb, @jmvrbanac: created the repo. :+1:

kgriffs:
BTW, someone just mentioned on IRC they tried gunicorn --reload but it has a problem on OS X. Just another thing we might reference as prior art...

kgriffs:
See also #25

csojinb:
Cool. I will hopefully have some bandwidth to pick this up again in the next week or two.

holms:
are we there yet? restarting gunicorn each time a bit annoying I'd say

kevinisaac:
Just looked at https://github.com/falconry/updraft to see an empty repo there. Can't wait for the reloader, manual reloading everytime sucks.

csojinb:
Implementation of the dev server has moved to here https://github.com/falconry/updraft/pull/1

Closing this PR

PR #524, opened 2015-04-22
fix(Router): Fix some edge-case bugs in the new compiled router

kgriffs:
A few edge cases were not included in the test suite. Add additional tests
and fix any bugs that the additional tests uncover. Also test for the
detection and raising of error for conflicting routes to encourage
uniform behavior between routers in the future.


jmvrbanac:
LGTM :+1:

PR #536, opened 2015-04-30
Fix DefaultRouter conflict bug

xiaclo:
The DefaultRouter will now allow static routes to be placed before
simple variable ones.  Placing static routes after simple variable ones
still raises a ValueError since they would be unreachable.

### DIFF
--- a/falcon/routing/compiled.py
+++ b/falcon/routing/compiled.py
@@ -206,7 +206,7 @@ class CompiledRouter(object):
                     #   /foo/{id}/bar
                     #   /foo/{name}/bar
                     #
-                    assert len(nodes) == 1
+                    assert len([node for node in nodes if node.is_var]) == 1
                     found_simple = True
 
             else:
@@ -308,10 +308,6 @@ class CompiledRouterNode(object):
         else:
             self.is_var = False
 
-    @property
-    def is_simple_var(self):
-        return self.is_var and not self.is_complex
-
     def matches(self, segment):
         """Returns True if this node matches the supplied template segment."""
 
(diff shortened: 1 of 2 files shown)


falcon/routing/compiled.py
```diff
@@ -206,7 +206,7 @@ def line(text, indent_offset=0):
                     #   /foo/{id}/bar
                     #   /foo/{name}/bar
                     #
-                    assert len(nodes) == 1
+                    assert len([node for node in nodes if node.is_var]) == 1
```
xiaclo: This is a simple test, perhaps something more robust with order checking would be helpful.

falcon/routing/compiled.py
```diff
@@ -362,4 +358,4 @@ def conflicts_with(self, segment):
                 #   /foo/all
                 return True
```
xiaclo: This is the opposite case, and shouldn't be changed.  A variable followed by a static route would be unreachable.

kgriffs: I think we could support this by sorting the routing logic so it always tests static routes first, regardless of what order add_route was called in. I can take a stab at implementing that once your patch is merged.

kgriffs:
LGTM. Thanks!

falcon/routing/compiled.py
xiaclo: I'm not sure that's a good idea.  Personally I would rather have falcon raise an error when I do something stupid than have it try to work around it.  It's may be helpful when you first write it, but if you someone else comes back to debug that code, it's not clear if that path should be reachable or not and involves a deep dive into falcon to figure it out.

PR #538, opened 2015-05-01
fix(DefaultRouter): Handle additional edge cases

### LINKED ISSUE #535: URI template conflicts after router update
I have been abusing the route order a bit, and would like to know if there is a way to do this with the new router, or if this is actually a bug.

```python
application = falcon.API()
application.add_route('/action1', Action1Handler())
application.add_route('/action2', Action2Handler())
application.add_route('/action3', Action3Handler())
application.add_route('/{action}', ActionFallbackHandler())
```

I realize this is probably a strange use case, and I could move the top 3 handlers into the bottom one and just relay the requests to the various other handlers, but this way seems cleaner.

kgriffs:
1. The order routes are added should not matter
2. Ensure entire path is consumed before returning a matched route

Closes #535

### DIFF
--- a/falcon/routing/compiled.py
+++ b/falcon/routing/compiled.py
@@ -29,74 +29,6 @@ class CompiledRouter(object):
     tree for each look-up, it generates inlined, bespoke Python code to
     perform the search, then compiles that code. This makes the route
     processing quite fast.
-
-    The generated code looks something like this::
-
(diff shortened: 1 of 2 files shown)


jmvrbanac:
LGTM :+1:

sebasmagri:
This is a +1 from me.

PR #557, opened 2015-06-13
docs(tutorial): note depreciation of global hooks

hpincket:
- remove global hook code example
- add mention of middleware
- add reference link to middleware

### DIFF
--- a/doc/user/tutorial.rst
+++ b/doc/user/tutorial.rst
@@ -548,8 +548,7 @@ responders in a DRY way, e.g.,:
         params['project_id'] = req.get_header('X-PROJECT-ID')
 
 Now, you can imagine that such a hook should apply to all responders for
-a resource, or even globally to all resources. You can apply hooks to an
-entire resource like so:
+a resource. You can apply hooks to an entire resource like so:
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


jmvrbanac:
:+1:

doc/user/tutorial.rst
```diff
@@ -559,11 +559,8 @@ entire resource like so:
         # ...
 
 And you can apply hooks globally by passing them into the API class
-initializer:
-
-.. code:: python
-
-    falcon.API(before=[extract_project_id])
+initializer, though this has been depreciated in favor of middleware.
```
kgriffs: I think it may be even better to simply not mention how hooks can be applied globally, specifically:
 * In L551, remove the "or even globally" clause.
 * In L561-L563, say something about how logic can also be applied globally using middleware (with no mention of global hooks)

Typo: "depreciated" ==> "deprecated"

kgriffs:
@hpincket Just following up to get your thoughts on my inline comment?

hpincket:
@kgriffs Ah, silly mistakes. I've made the suggested changes.

kgriffs:
Looks good, thanks!

PR #559, opened 2015-06-19
Add additional HTTP status codes

### LINKED ISSUE #553: Missing HTTP status codes
There's a few status codes from here http://tools.ietf.org/html/rfc6585 that could be great including.

429 Too many requests for instance.

chripede:
Implement error codes from http://tools.ietf.org/html/rfc6585
Fixes #553


kgriffs:
@chripede This looks good, thanks! Not a big deal, but for next time it would be great if you could follow the commit message format as specified here:

https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#commit-message-format

falcon/status_codes.py
```diff
@@ -90,6 +90,12 @@
 HTTP_IM_A_TEAPOT = HTTP_418
 HTTP_426 = '426 Upgrade Required'
 HTTP_UPGRADE_REQUIRED = HTTP_426
+HTTP_428 = '428 Precondition Required'
```
kgriffs: It would be nice to add some subclasses of HTTPError that correspond to these...

kgriffs: Added #561

PR #563, opened 2015-07-11
Instruct browser to remove cookies

Freezerburn:
fix: Cookies not being removed from browser

The previous behavior of Response's unset_cookie function was just to delete the given name from the SimpleCookie object if it existed. However, this causes a problem in that the browser is not told to remove the cookie on its side, causing all future Request objects to still contain the cookie that was supposed to have been unset/removed. This change causes the Response object to actually set the expires token for the cookie that is being unset to a point in the past, causing the browser to immediately remove the now-expired cookie, which will remove the cookie from future Request objects.

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -246,6 +246,10 @@ class Response(object):
 
     def unset_cookie(self, name):
         """Unset a cookie from the response
+        
+        Note:
+            This will instruct the browser to expire its copy of the cookie
+            if it has previously been set.
         """
         if self._cookies is None:
             self._cookies = SimpleCookie()
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -247,8 +247,16 @@ class Response(object):
     def unset_cookie(self, name):
         """Unset a cookie from the response
         """
-        if self._cookies is not None and name in self._cookies:
-            del self._cookies[name]
+        if self._cookies is None:
+            self._cookies = SimpleCookie()
+        if name not in self._cookies:
+            self._cookies[name] = ""
+        # SimpleCookie apparently special cases the expires attribute to
+        # automatically use strftime and set the time as a delta from the
+        # current time. We use -1 here to basically tell the browser to
+        # immediately expire the cookie, thus removing it from future
+        # request objects.
+        self._cookies[name]["expires"] = -1
 
     def set_header(self, name, value):
         """Set a header for this response to a given value.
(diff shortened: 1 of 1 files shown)


falcon/response.py
```diff
@@ -247,8 +247,16 @@ def set_cookie(self, name, value, expires=None, max_age=None,
     def unset_cookie(self, name):
         """Unset a cookie from the response
```
kgriffs: Would you mind expanding the docstring with a note talking about how this method will also instruct the user agent to expire it's copy of the cookie, if previously set?

Freezerburn:
@kgriffs The docstring has been expanded. Let me know if the explanation is satisfactory.

falcon/response.py
```diff
@@ -246,9 +246,21 @@ def set_cookie(self, name, value, expires=None, max_age=None,
...
+        
+        Note:
+            This will instruct the browser to expire its copy of the cookie
+            if it has previously been set.
         """
-        if self._cookies is not None and name in self._cookies:
-            del self._cookies[name]
+        if self._cookies is None:
+            self._cookies = SimpleCookie()
+        if name not in self._cookies:
```
kgriffs: Would it make sense to always set the value to the empty string, even if it was set elsewhere in the app to some value?

falcon/response.py
```diff
@@ -246,9 +246,21 @@ def set_cookie(self, name, value, expires=None, max_age=None,
...
+            self._cookies[name] = ""
+        # SimpleCookie apparently special cases the expires attribute to
```
kgriffs: Nitpick: Add a blank line before the start of the comment to aid in readability. Also, start the comment with ` NOTE(Freezerburn): `.

kgriffs:
> The docstring has been expanded.

Looks good, thanks!

falcon/response.py
Freezerburn: That actually does make sense. I'll make that change.

kgriffs:
@Freezerburn Looks good. As a final step in preparation for merging, could you please rebase on master and squash down to a single commit, formatted according to our [style guide][1]? Thanks!

[1]: https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#commit-message-format

kgriffs:
@Freezerburn I think this is ready to go, but it needs to be rebased first.

kgriffs:
Manually rebased and resubmitted as #634

kgriffs:
[merged manually]

PR #570, opened 2015-07-16
feat(req+resp): Add support for Content-Range units other than 'bytes'

p3dda:
In RFC 2616 it says that "The only range unit defined by HTTP/1.1 is "bytes". HTTP/1.1 implementations MAY ignore ranges specified using other units.". Current falcon implementation only accepts 'bytes', which does not work with some client implementations using different range units. For example, dojo/store/JsonRest requires the server to support 'items' as range unit.

This pull request adds a request property 'range_unit', which contains the range-unit parsed from Range header. It enhances format_range to accept a range unit as optional parameter in range tuple. This range unit default is set to 'bytes' to not break current API.


tests/test_headers.py
```diff
@@ -336,6 +336,16 @@ def test_response_header_helpers_on_get(self):
...
                       self.srmock.headers)
 
+        resp.content_range = (0, 499, 10 * 1024, 'bytes')
+        self.assertEqual('bytes 0-499/10240', resp.content_range)
+        self.assertIn(('content-range', 'bytes 0-499/10240'),
+                      self.srmock.headers)
+
+        resp.content_range = (0, 25, 100, 'items')
+        self.assertEqual('items 0-25/100', resp.content_range)
+        self.assertIn(('content-range', 'bytes 0-499/10240'),
```
fxfitz: Shouldn't this test be checking for `'items 0-25/100'`?

fxfitz:
We don't appear to be breaking the API by adding this in (since we're defaulting to `bytes`), and range items are supposed to be extensible (https://svn.tools.ietf.org/svn/wg/httpbis/draft-ietf-httpbis/latest/p5-range.html#range.units.other).

Aside from that question above, assuming you update to master and all tests and such pass, I'm good with this.

tests/test_headers.py
p3dda: Thank you for your feedback. You are right, the second unit test should be testing for `items 0-25/100`. I have fixed this in unit test and also modified the mock response headers generator to return other range units than `bytes` if requested.

p3dda:
I have fixed the above question and updated the pull request to master.

fxfitz:
Well, LGTM then.

fxfitz:
Thanks for the quick turn around @p3dda; we'll get one other core reviewer to look at it and then we'll probably be good.

falcon/request.py
```diff
@@ -371,16 +373,16 @@ def if_unmodified_since(self):
     def range(self):
         try:
             value = self.env['HTTP_RANGE']
-            if value.startswith('bytes='):
-                value = value[6:]
+            if len(value.split('=', 1)) == 2:
+                value = value.split('=', 1)[1]
```
philiptzou: Try not run `value.split` twice, it wastes time.

falcon/request.py
```diff
@@ -400,11 +402,25 @@ def range(self):
...
             raise HTTPInvalidHeader(msg, 'Range', href=href,
                                     href_text=href_text)
 
     @property
+    def range_unit(self):
+        try:
+            value = self.env['HTTP_RANGE']
+
+            if len(value.split('=', 1)) == 2:
+                unit = value.split('=', 1)[0]
```
philiptzou: also try not run twice.

falcon/request.py
```diff
@@ -371,16 +373,16 @@ def if_unmodified_since(self):
     def range(self):
         try:
             value = self.env['HTTP_RANGE']
-            if value.startswith('bytes='):
-                value = value[6:]
+            if '=' in value:
+                value = value.split('=', 1)[1]
```
kgriffs: PERF: `str.partition` is a little faster than `str.split`

falcon/request.py
```diff
@@ -371,16 +373,16 @@ def if_unmodified_since(self):
...
             else:
-                msg = "The value must be prefixed with 'bytes='"
+                msg = "The value must be prefixed with range unit, e.g. 'bytes='"
```
kgriffs: "with range unit" ==> "with a range unit"

falcon/request.py
```diff
@@ -400,11 +402,25 @@ def range(self):
         except ValueError:
             href = 'http://goo.gl/zZ6Ey'
             href_text = 'HTTP/1.1 Range Requests'
-            msg = ('It must be a byte range formatted according to RFC 2616.')
+            msg = ('It must be a range formatted according to RFC 2616.')
```
kgriffs: While you're at it, would you mind changing "RFC 2616" to "RFC 7233"?

falcon/request.py
```diff
@@ -400,11 +402,25 @@ def range(self):
         except ValueError:
```
kgriffs: L397 still references "bytes" (GH won't let me comment on the line directly since it wasn't modified in the PR).

falcon/request.py
```diff
@@ -400,11 +402,25 @@ def range(self):
...
             raise HTTPInvalidHeader(msg, 'Range', href=href,
                                     href_text=href_text)
 
     @property
+    def range_unit(self):
+        try:
+            value = self.env['HTTP_RANGE']
+
+            if '=' in value:
+                unit = value.split('=', 1)[0]
```
kgriffs: PERF: Modify to use `str.partition`.

falcon/request.py
```diff
@@ -400,11 +402,25 @@ def range(self):
...
+                return unit
+            else:
+                msg = "The value must be prefixed with range unit, e.g. 'bytes='"
```
kgriffs: "with range unit" ==> "with a range unit"

falcon/response.py
```diff
@@ -499,10 +499,11 @@ def add_link(self, target, rel, title=None, title_star=None,
...
         """A tuple to use in constructing a value for the Content-Range header.
 
-        The tuple has the form (*start*, *end*, *length*), where *start* and
-        *end* designate the byte range (inclusive), and *length* is the
-        total number of bytes, or '\*' if unknown. You may pass ``int``'s for
-        these numbers (no need to convert to ``str`` beforehand).
+        The tuple has the form (*start*, *end*, *length*, [*unit*]), where *start* and
+        *end* designate the range (inclusive), and *length* is the
+        total length, or '\*' if unknown. You may pass ``int``'s for
+        these numbers (no need to convert to ``str`` beforehand). Optional value
```
kgriffs: "Optional value *unit* describes..." ==> "The optional *unit* value describes..."

falcon/response_helpers.py
```diff
@@ -51,12 +51,16 @@ def format_range(value):
...
-
     """
 
     # PERF: Concatenation is faster than % string formatting as well
     #       as ''.join() in this case.
-    return ('bytes ' +
+    if len(value) == 4:
+        unit = value[3]
+    else:
+        unit = 'bytes'
```
kgriffs: PERF: Probably not significant, but if you'd like to save an extra concatenation below, you could include the space in the literal string in L61 (i.e., `'bytes '`) and then append a space to the custom unit type in L59 (i.e., `value[3] + ' '`). That way, you only pay the penalty when using a custom unit type.

kgriffs:
OK, this is really close. Just a few minor issues (comments inline).

p3dda:
Thank you @kgriffs for reviewing the code and suggesting various improvements. I have fixed your issues and pushed.

falcon/request.py
```diff
@@ -371,20 +373,20 @@ def if_unmodified_since(self):
     def range(self):
         try:
             value = self.env['HTTP_RANGE']
-            if value.startswith('bytes='):
-                value = value[6:]
+            if '=' in value:
+                unit, sep, range = value.partition('=')
```
kgriffs: Ah, unfortunately this will end up masking the builtin `range` function, so that variable will need a different name.

kgriffs:
OK, just that last comment and then LGTM. Thanks!

Issue #562, opened 2015-07-09
Decoding router params.


rroslaniec:
Hello,
I found error on decoding params from url_template. This is examle of decoding GET param and params in url.

    import falcon

    from wsgiref import simple_server

    class ThingsParam:
        def on_get(self, req, resp):
            print(req.get_param('query', True))
            resp.status = falcon.HTTP_200

    class ThingsQuery:
        def on_get(self, req, resp, query):
            print(query)
            resp.status = falcon.HTTP_200

    app = falcon.API()

    app.add_route('/things', ThingsParam())
    app.add_route('/things/{query}', ThingsQuery())

    if __name__ == '__main__':
        httpd = simple_server.make_server('127.0.0.1', 8000, app)
        httpd.serve_forever()

Output will be:
    127.0.0.1 - - [09/Jul/2015 16:03:53] "GET / HTTP/1.1" 404 0
    127.0.0.1 - - [09/Jul/2015 16:03:53] "GET /favicon.ico HTTP/1.1" 404 0
    ąę
    127.0.0.1 - - [09/Jul/2015 16:04:10] "GET /things?query=%C4%85%C4%99 HTTP/1.1" 200 0
    ÄÄ
    127.0.0.1 - - [09/Jul/2015 16:04:17] "GET /things/%C4%85%C4%99 HTTP/1.1" 200 0

kgriffs:
Thanks for the report. I tried this under CPython 2.7.10 on OS X and it printed out the correct string, "ąę", each time. However, under CPython 3.4.3, the string is misinterpreted.

Digging into this further, I found that when constructing the WSGI environ dict under CPython 2.7.10, wsgiref and gunicorn both decode the escaped path segment to a UTF-8 encoded instance of `str`. On the other hand, they both leave the query string alone (the percent encoding is preserved).

Under CPython 3.4.3, the characters are copied directly to an instance of `str` as well, but under Py3k, `str` is a native Unicode type, and when you print it out, the code points are not assumed to be UTF-8.

```python
# Bytes posted from curl
b'/things/%C4%85%C4%99'

# Interpreted by the WSGI server
u'/things/\xC4\x85\xC4\x99' == u'/things/Ä\x85Ä\x99'

# What a correctly decoded string would look like
u'/things/\u0105\u0119' == u'/things/ąę'
```

@jmvrbanac Thoughts on what we should do about this?

kgriffs:
[fixed]

PR #571, opened 2015-07-17
ugly fix for "feature" of PEP3333 in py3k

bobuk:
As far as I notice this is a dummy fix for #562.


kgriffs:
@bobuk I think this makes sense. Can you add some tests to ensure this indeed fixes the issue, and rebase against master? Also, please amend the commit message to be formatted per our style guide: 

https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#commit-message-format

Thanks!

Issue #572, opened 2015-07-20
Unit tests for cookies fail with Python 3.4


desiderius:
Since Python 3.4.3 the [capitalization used by **http.cookies** has changed](https://hg.python.org/cpython/rev/0d8380c493ad) for these attributes:

   < 3.4.3 | 3.4.3 >
---------|---------
httponly | HttpOnly
secure   | Secure

The following tests are case sensitive and fails:
- tests.test_cookies.TestCookies.test_response_base_case
- tests.test_cookies.TestCookies.test_response_complex_case

Duplicate of #558 :bow:

desiderius:
Sorry about the duplicate. I was in a rush for a fix and missed the previous comments in #558.

PR #573, opened 2015-07-20
fix(test_cookies): Fix references to *httponly* and *secure*

desiderius:
Since Python 3.4 the attributes *httponly* and *secure* are capitalized like this:
- HttpOnly
- Secure

This quick fix adapt the attributes case to the Python version in use.

See issue #572

A better way to fix that would be to refactor the unit tests and use a unittest.skipIf(). If refactoring is welcome, I can do that when I have time.

### DIFF
--- a/tests/test_cookies.py
+++ b/tests/test_cookies.py
@@ -1,4 +1,4 @@
-
+import sys
 import falcon
 import falcon.testing as testing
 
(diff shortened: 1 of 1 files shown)


desiderius:
I am sorry, I cannot understand how I missed that the same issue was raised in #558 !

PR #577, opened 2015-07-29
fix(testing.base.TestBase): parameterize api and srmock

swizzard:
Make `api` and `srmock` arguments to `TestBase.setUp`

As currently implemented, `API` and `testing.srmock.StartResponseMock`
objects are instantiated during `setUp`. This means subclasses either
need to throw away those instances, or eschew the `super` call.

### DIFF
--- a/falcon/testing/base.py
+++ b/falcon/testing/base.py
@@ -46,14 +46,15 @@ class TestBase(unittest.TestCase):
         test_route (str): A simple, generated path that a test
             can use to add a route to the API.
     """
+    api_class = falcon.API
+    srmock_class = StartResponseMock
 
     def setUp(self):
         """Initializer, unittest-style"""
-
         super(TestBase, self).setUp()
         self._id = itertools.count(0)
-        self.api = falcon.API()
-        self.srmock = StartResponseMock()
+        self.api = self.api_class()
+        self.srmock = self.srmock_class()
         self.test_route = '/{0}'.format(next(self._id))
 
         # Reset to simulate "restarting" the WSGI container


TheSriram:
Looks good to me.

falcon/testing/base.py
```diff
@@ -47,13 +47,13 @@ class TestBase(unittest.TestCase):
...
 
-    def setUp(self):
+    def setUp(self, api=None, srmock=None):
         """Initializer, unittest-style"""
 
         super(TestBase, self).setUp()
         self._id = itertools.count(0)
-        self.api = falcon.API()
-        self.srmock = StartResponseMock()
+        self.api = api or falcon.API()
```
kgriffs: Nitpick: It's a good habit to explicitly test for None (e.g., `falcon.API() if api is None else api`) since the object being passed in may (unexpectedly) evaluate to False, but in this case I don't think it matters since we are expecting instances of API and StartResponseMock which are known to always evaluate to True.

falcon/testing/base.py
```diff
@@ -47,13 +47,13 @@ class TestBase(unittest.TestCase):
             can use to add a route to the API.
     """
 
-    def setUp(self):
+    def setUp(self, api=None, srmock=None):
```
kgriffs: Normally this function is just called by the testing framework (unittest/testtools) and doesn't take any parameters. How will this be used?

swizzard: my_api = falcon.API(before=[my_before_stuff], after=[my_after_stuff])
	my_srmock = MyStartResponseMockSubclass()
	super(TestBase, self).setUp(my_api, my_srmock)

kgriffs: Ah, OK. After re-reading the initial comment for the PR this makes sense.

falcon/testing/base.py
```diff
@@ -47,13 +47,19 @@ class TestBase(unittest.TestCase):
...
 
         super(TestBase, self).setUp()
         self._id = itertools.count(0)
-        self.api = falcon.API()
-        self.srmock = StartResponseMock()
+        if api is not None:
+            self.api = api
+        else:
+            self.api = falcon.API()
+        if srmock is not None:
```
kgriffs: Nitpick: Add a blank line above L59 to visually group the two if...else control structures.

falcon/testing/base.py
```diff
@@ -47,13 +47,19 @@ class TestBase(unittest.TestCase):
...
+            self.srmock = srmock
+        else:
+            self.srmock = StartResponseMock()
         self.test_route = '/{0}'.format(next(self._id))
```
kgriffs: Nitpick: Add a blank line above L63 to show it is not related to the manipulation of srmock in L59-62

kgriffs:
@swizzard Just a couple suggestions inline. Otherwise, could you please rebase onto master and squash down to a single commit? Thanks!

philiptzou:
I'd like to propose a new design here which I learned this trick first from Flask:

```python
class TestBase(unittest.TestCase):
    api_class = falcon.API
    srmock_class = StartResponseMock

    def setUp(self):
        # blah blah
        self.api = self.api_class()
        self.srmock = self.srmock_class()
```

I perceive that `super()` call is also ugly since you had to override `setUp` actually. But replace class attributes are much smooth and looks more "Pythonic" to me.

kgriffs:
@philiptzou's We're actually already using this strategy in `falcon.Request` to determine the type of the `context` attribute. I've seen it work well in a few other projects, assuming it's well-documented.

@swizzard What do you think?

swizzard:
I implemented @philiptzou's fix. I usually just let my agonized, flailing git logs speak for themselves, so if I've messed something up on the rebase, please let me know.

falcon/testing/base.py
```diff
@@ -46,14 +46,15 @@ class TestBase(unittest.TestCase):
...
 
-    def setUp(self):
+    def setUp(self, api=None, srmock=None):
         """Initializer, unittest-style"""
-
         super(TestBase, self).setUp()
         self._id = itertools.count(0)
-        self.api = falcon.API()
-        self.srmock = StartResponseMock()
+        self.api = self.api()
```
philiptzou: `self.api = self.api_class()`

falcon/testing/base.py
```diff
@@ -46,14 +46,15 @@ class TestBase(unittest.TestCase):
         test_route (str): A simple, generated path that a test
             can use to add a route to the API.
     """
+    api_class = falcon.API
+    srmock_class = StartResponseMock
 
-    def setUp(self):
+    def setUp(self, api=None, srmock=None):
```
kgriffs: These args are no longer needed, right?

kgriffs:
LGTM, thanks!

PR #585, opened 2015-08-12
add client_accepts_jsonapi property to Request

smcclstocks:
I’ll probably have a few PR’s as I develop my own JSONAPI serializer/deserializer. This first one is relatively simple.

For more info on JSON API see http://jsonapi.org/format/.

Thanks & I look forward to contributing!


Weyzu:
Adding `./.ropeproject` directory seems unnecessary.

smcclstocks:
Yes I agree & don't know why that is in there. I must have totally overlooked it in my review. I'm not great with git so how would you like me to proceed? Should I refork falcon & make my changes then pull request again? Thanks.

Weyzu:
I'd do `git reset HEAD~`, commit things properly and then `git push -f` (Usually you should avoid force pushing). 

Remember to follow the style guide for commit messages. Falcon uses AngularJS's style guide. More info [here](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit).

Don't forget to add tests of accepting JSON API format. A scenario can be added in `tests/test_req_vars.py::test_client_accepts()` although don't take my word for it, since I don't know if further changes should be made. :)

falcon/request.py
```diff
@@ -293,6 +295,10 @@ def client_accepts_json(self):
         return self.client_accepts('application/json')
 
     @property
+    def client_accepts_jsonapi(self):
```
kgriffs: I'm trying to decide how far we should go down the hypermedia path; there are a bunch of media types out there we would have to consider (HAL, Siren, Cj, JSON API, etc.)

smcclstocks: That's fine & I can understand it. I just passed in my own request object & that has plenty of flexibibility for what I need.

PR #589, opened 2015-08-21
perf(request): Optimize HTTP date parsing

kgriffs:
Don't check for obsolete date formats by default. When obsolete date
formats are not enabled, avoid unnecessary processing time and data
structure instantiation.


jmvrbanac:
:+1:

PR #591, opened 2015-08-25
perf(api): Optimize HTTPStatus handling

kgriffs:
Modify the _compose_status_response() function to no longer use getattr to get
the body for HTTPStatus. This will make the code more readable and more
performant.

As part of this work, rework _compose_error_response() to no longer call
_compose_status_response(), saving a few CPU cycles in that path as well. If
we ever end up needing to duplicate this code a third time, consider factoring
out into a separate helper function at that point.


jmvrbanac:
:+1:

PR #593, opened 2015-08-28
fix(HTTPError): add Vary header to default error serializer

dracos:
The default error serializer can return JSON or XML depending upon the Accept header, so include this in the response's Vary header so that proxies and caches do not potentially serve the wrong representation.

This commit adds a resp argument to the error serializer, which will break existing serializers.

(I did wonder about putting the header addition in api.py, if body is not None, so as not to affect the parameters to the serializer, but a custom serializer may not want/need to have any change to the Vary header.)

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 import re
 import six
 
@@ -413,7 +414,7 @@ class API(object):
         `to_json()` and `to_dict()`, that can be used from within
         custom serializers. For example::
 
-            def my_serializer(req, exception):
+            def my_serializer(req, resp, exception):
                 representation = None
 
                 preferred = req.client_prefers(('application/x-yaml',
@@ -425,8 +426,8 @@ class API(object):
                     else:
                         representation = yaml.dump(exception.to_dict(),
                                                    encoding=None)
-
-                return (preferred, representation)
+                    resp.body = representation
+                    resp.content_type = preferred
 
         Note:
             If a custom media type is used and the type includes a
@@ -437,16 +438,15 @@ class API(object):
 
         Args:
             serializer (callable): A function taking the form
-                ``func(req, exception)``, where `req` is the request
-                object that was passed to the responder method, and
-                `exception` is an instance of ``falcon.HTTPError``.
-                The function must return a ``tuple`` of the form
-                (*media_type*, *representation*), or (``None``, ``None``)
-                if the client does not support any of the
-                available media types.
+                ``func(req, resp, exception)``, where `req` is the request
+                object that was passed to the responder method, `resp` is
+                the response object, and `exception` is an instance of
+                ``falcon.HTTPError``.
 
         """
 
+        if len(inspect.getargspec(serializer).args) == 2:
+            serializer = helpers.wrap_old_error_serializer(serializer)
         self._serialize_error = serializer
 
     # ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


kgriffs:
Hi dracos, thanks for catching this and proposing a fix!

With this patch, the manipulation of the `Response` object is split between `API._compose_error_response` and the serializer itself; I wonder if it would make sense to either move the logic for setting the body and content type to the serializer, or conversely having the serializer return a flag indicating whether `API._compose_error_response` should set the Vary header.

What do you think?

dracos:
The first option sounds a bit clearer to me, in that it is the serializer deciding what the body/content-type should be, but on the other hand the second would be possible to implement without breaking existing custom serializers and also makes sense, keeping all the response stuff together as you say.

Another option I've just thought of, which is an expanded version of your second option: the serializer could return headers and body, rather than just "content type header" and body (I don't know if a custom serializer would ever want to change another header, but it seems possible?). `resp` wouldn't need to be passed in, and I think you could also do it safely without breaking any existing custom serializers, by maintaining the content type return parameter. I've implemented that as a second commit in this PR to see what you think.

falcon/api.py
```diff
@@ -418,14 +418,17 @@ def my_serializer(req, exception):
                 preferred = req.client_prefers(('application/x-yaml',
                                                 'application/json'))
 
+                headers = []
+
                 if preferred is not None:
+                    headers.append(('append', 'Vary', 'Accept'))
```
kgriffs: Having to specify 'append' seems a bit clunky to me and would require additional documentation to explain. I'm starting to wonder if it would be more straightforward, after all, to just manipulate `resp` directly.

What if we were to go back to passing in `resp` and move the logic for setting the body and content type into the serializer itself? In order to avoid breaking existing serializers, we could introspect the `serializer` passed into `set_error_serializer()` and wrap it in a shim if it does not have a `req` arg.

dracos:
"What if we were to go back to passing in resp and move the logic for setting the body and content type into the serializer itself? In order to avoid breaking existing serializers, we could introspect the serializer passed into set_error_serializer() and wrap it in a shim if it does not have a req arg." - I have rebased and hopefully implemented this, it appears to work successfully, hope that's helpful.

falcon/api_helpers.py
```diff
@@ -121,9 +116,29 @@ def default_serialize_error(req, exception):
             preferred = 'application/xml'
 
     if preferred is not None:
+        resp.append_header('Vary', 'Accept')
         if preferred == 'application/json':
             representation = exception.to_json()
         else:
             representation = exception.to_xml()
+        resp.body = representation
```
kgriffs: Nit: I think adding a blank line above L124 would aid in readability.

falcon/api_helpers.py
```diff
@@ -121,9 +116,29 @@ def default_serialize_error(req, exception):
...
+        resp.content_type = preferred
+
+
+def wrap_old_error_serializer(old_fn):
+    """Wraps an old-style error serializer to add body/content_type setting.
+
+    Args:
+        fn: Old style error serializer
```
kgriffs: "fn" ==> "old_fn"

kgriffs: "Old style" ==> "Old-style"

tests/test_httperror.py
```diff
@@ -317,6 +317,57 @@ def _check(media_type, deserializer):
         # _check('application/x-yaml', yaml.load)
```
kgriffs: TBH, I don't recall why this was commented out. It probably makes sense to uncomment it, and also add this check below.

falcon/api_helpers.py
```diff
@@ -121,9 +116,29 @@ def default_serialize_error(req, exception):
             preferred = 'application/xml'
 
     if preferred is not None:
+        resp.append_header('Vary', 'Accept')
```
kgriffs: Please add a test to ensure this header is set when the default serializer is used.

kgriffs:
@dracos This looks great! Just a few minor things to address (comments inline).

falcon/api_helpers.py
dracos: Have added an assertIn to the `test_base_class`. And dealt with all the other comments; have done it as a fixup commit so it’s easy to compare, let me know if you’d like me to rebase/squash it first.

kgriffs: Awesome, this looks great! As a final step would you mind rebasing on master and squashing down to a single commit?

dracos:
Rebased/squashed :)

PR #594, opened 2015-09-01
feat(falcon.testing): Testing infrastructure for Falcon Middleware

BenjamenMeyer:
Added a new testing object - `MiddlewareMock` - for developers to use
for testing their implementations of Falcon Middleware objects. The
`MiddlewareMock` allows a developer to unit test the specific parts
of the middleware in a uniform, reliable manner without having to
rebuild all the required infrastructure to do for every test. Further
it maintains the built infrastructure to be able to allow a developer
to test the stages in successive order.

As part of this work, an instance of the `MiddlewareMock` object is
added to the `falcon.testing.base.TestBase` class so it is easily
available, and clearly distinguishable from the `StartResponseMock`


BenjamenMeyer:
I'll add a fix for the code coverage. FYI - it passed ```tox``` locally. I wonder if there's a way to check it the same way as ```tox``` never really complained about code coverage...

falcon/testing/smmock.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # validate that self.response is an instance of falcon.Response
+        # or reset it
+        if self.response is not None:
+            if not isinstance(self.response, Response):
+                self.response = None
+        if self.response is None:
+            self.response = Response()
+
+    def simulate_process_request(self, **kwargs):
```
kgriffs: I'm a little leery of these methods, since they are a reimplementation of what falcon.API actually does, with the potential for them to get out of sync over time. It seems like it would be safer to test middleware directly against the falcon.API implementation.

BenjamenMeyer: How do isolate a middleware test to verify it is a bug in just the middleware?

Testing the middleware using the full falcon.API seems at least to me to be an integration test instead of a unit test which is the purpose of these additional functions.

kgriffs: I understand the value proposition, but I think we need to find a way to maybe factor out the middleware logic from the API class so that it could be driven directly from a test.

BenjamenMeyer: Sorry for the delayed response...

So, essentially you would prefer to have a new class in `falcon` proper that would handle the middleware functionality which is then used by the `falcon.API` and can also be used directly in the tests to do the validation. That shouldn't be too hard, let me think that through and I'll see what I can do.

BenjamenMeyer:
Opened https://github.com/falconry/falcon/issues/837 to discuss the matter at large. If the change is too drastic from this, then I'll close this PR and do the work in another one.

falcon/testing/smmock.py
kgriffs: Yeah, the middleware processing in falcon.API could use an overhaul anyway, since the way we handle exceptions isn't completely correct; we should be able to handle multiple exceptions raised, vs. only retrying once (see also: https://github.com/falconry/falcon/pull/745#discussion_r59302802).

I may try to tackle said overhaul unless you've got some bandwidth to tackle it over the next few weeks. Just let me know, thanks!

BenjamenMeyer: I've been busy with RCBU API v1 :arrow_right: v2 migration stuff; so not sure when I'll have bandwidth at the moment. I'll be happy to collaborate in some form though - reviews, discussions, etc.

kgriffs: For 1.1 I overhauled the processing logic. Do you want to take a look at the latest code to see if it would make sense and not significantly degrade performance to factor out the handling logic so it could be used in tests more directly?

kgriffs: I'm going to go ahead and close this PR for now while we work out how we want to proceed.

kgriffs:
I'm going to go ahead and close this PR for now while we work out how we want to proceed.

PR #596, opened 2015-09-04
feat: Add redirection support via raising subclasses of HTTPStatus

### LINKED ISSUE #406: Add a helper for redirects

kgriffs:
Define a set of HTTPStatus subclasses that can be raised to perform
various types of HTTP redirects. This should avoid the problem of
hooks and responder methods possibly overriding the redirect. Raising
an instance of one of these classes and will short-circuit request
processing similar to raising an instance of HTTPError.

Specifically, if raised in a before hook, it will skip any remaining
hooks and the responder method, but will not skip any process_response
middleware methods. If raised within a responder, it will skip the
rest of the responder and all after hooks. If raised in an after hook,
it would skip remaining after hooks but not middleware methods. And
finally, if raised within a middleware method, execution would perceive
as described at the bottom of [1].

The above behavior is inherited from HTTPStatus and so is not re-tested
in the subclasses.

[1]: https://falcon.readthedocs.org/en/stable/api/middleware.html

Closes #406
Supersedes #486


kgriffs:
Reviewer: Please pay particular attention to docstrings, class and variable names, etc. to ensure everything is correct.

falcon/redirects.py
```diff
@@ -0,0 +1,148 @@
...
+    """303 See Other.
+
+    The 303 (See Other) status code indicates that the server is
+    redirecting the user agent to a *different* resource, as indicated
+    by a URI in the Location header field, which is intended to provide
+    an indirect response to the original request.
+
+    A 303 response to a GET request indicates that the origin server
+    does not have a representation of the target resource that can be
+    transferred by the server over HTTP.  However, the Location field
```
jmvrbanac: You might clarify that this is a header field like you did earlier.

falcon/redirects.py
```diff
@@ -0,0 +1,148 @@
...
+    value refers to a resource that is descriptive of the target
+    resource, such that making a retrieval request on that other
+    resource might result in a representation that is useful to
+    recipients without implying that it represents the original target
+    resource.
```
jmvrbanac: That is one long sentence. I had to re-read it a couple times to fully understand the point. If people have a problem understanding this, you might consider rephrasing or breaking this sentence up.

jmvrbanac:
Other than my two comments, this LGTM.

kgriffs:
@jmvrbanac IIRC the text for those docstrings came from the RFCs. Let me see if I can improve the prose per your comments.

jmvrbanac:
:+1: LGTM

kgriffs:
@rouge8 We are getting ready to merge this. Does this approach look good?

rouge8:
Awesome! :+1: Ran my project's tests using this branch (and the minor code changes necessary). API worked as expected and the tests passed. Admittedly, I only used the 301, but this looks good!

guoyucheng:
when the redirectspy can merge to the master?  I need redirects :)

PR #597, opened 2015-09-06
Fixed #588: more robust when querystring contains malformed %

### LINKED ISSUE #588: Falcon crashes on decoding query string
When Falcon receives a request with query string that cannot be percent decoded (e.g. `x=%`), it crashes resulting in a rather unexpected 502 response.

	  File "/opt/ads/venv/local/lib/python2.7/site-packages/falcon/api.py", line 154, in __call__
		req = self._request_type(env, options=self.req_options)
	  File "/opt/ads/venv/local/lib/python2.7/site-packages/falcon/request.py", line 237, in __init__
	  File "/opt/ads/venv/local/lib/python2.7/site-packages/falcon/util/uri.py", line 327, in parse_query_string
		params[k] = decode(v)
	  File "/opt/ads/venv/local/lib/python2.7/site-packages/falcon/util/uri.py", line 181, in decode
		char, byte = _HEX_TO_BYTE[token[:2]]
	KeyError: ''

Would it make more sense to pass-through the original value in this case without trying to decode it? It's a rather small change so I could prepare a pull request for it, but I would like to confirm first that this was not intentional behavior.

philiptzou:
I tested malformed % on [Google Search](https://www.google.com/?q=%%20%25) and it works (interpret as "% %"). So I think for robust reason the best solution is just interpreting it like Google did since it's not ambiguous.

### DIFF
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -178,9 +178,12 @@ if six.PY2:
         tokens = decoded_uri.split('%')
         decoded_uri = tokens[0]
         for token in tokens[1:]:
-            char, byte = _HEX_TO_BYTE[token[:2]]
-            decoded_uri += char + token[2:]
-
+            token_partial = token[:2]
+            if token_partial in _HEX_TO_BYTE:
+                char, byte = _HEX_TO_BYTE[token_partial]
+            else:
+                char, byte = '%', 0
+            decoded_uri += char + (token[2:] if byte else token)
             only_ascii = only_ascii and (byte <= 127)
 
         # PERF(kgriffs): Only spend the time to do this if there
@@ -235,7 +238,12 @@ else:
         tokens = decoded_uri.split(b'%')
         decoded_uri = tokens[0]
         for token in tokens[1:]:
-            decoded_uri += _HEX_TO_BYTE[token[:2]] + token[2:]
+            token_partial = token[:2]
+            if token_partial in _HEX_TO_BYTE:
+                decoded_uri += _HEX_TO_BYTE[token_partial] + token[2:]
+            else:
+                # malformed percentage like "x=%" or "y=%+"
+                decoded_uri += b'%' + token
 
         # Convert back to str
         return decoded_uri.decode('utf-8', 'replace')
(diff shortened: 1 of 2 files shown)


philiptzou:
Rebased onto the latest version.

philiptzou:
Rebased onto the latest version. Again.

falcon/util/uri.py
```diff
@@ -178,9 +178,8 @@ def decode(encoded_uri):
         tokens = decoded_uri.split('%')
         decoded_uri = tokens[0]
         for token in tokens[1:]:
-            char, byte = _HEX_TO_BYTE[token[:2]]
-            decoded_uri += char + token[2:]
-
+            char, byte = _HEX_TO_BYTE.get(token[:2], ('%', 0))
+            decoded_uri += char + (token[2:] if byte else token)
```
philiptzou: Just test if a rebasing will overwrite this comment or not.

tests/test_query_params.py
```diff
@@ -64,6 +64,14 @@ def test_percent_encoded(self):
         self.assertEqual(req.get_param_as_list('id', int), [23, 42])
         self.assertEqual(req.get_param('q'), u'\u8c46 \u74e3')
 
+    def test_bad_percentage(self):
+        query_string = 'x=%%20%+%'
```
kgriffs: Please also test mixing in some non-escaped text (e.g., 'peregrine'), as well as some additional non-hex-pair, percent-escaped characters (e.g., '%a', '%z', '%zz').

falcon/util/uri.py
kgriffs: Out of curiosity, I tested an alternative approach of using `try...except KeyError`, and found that it was slightly more efficient under Python 2.7. Would you mind making this change?

kgriffs:
@philiptzou Nice work! Just a couple of nits to address and this should be ready to merge.

falcon/util/uri.py
philiptzou: This topic was kindly discussed at #637. I quote "A `try/except` block is extremely efficient if no exceptions are raised Actually catching an exception is expensive". Since there are chances that `token[:2]` can be anything, I conceive we can use `if ... in` here instead of `try...except KeyError`. The `if ... in` block, which is also more efficient than `dict.get()`, can balance the time cost better than `try/except` block.

kgriffs: Sounds good; makes sense.

philiptzou:
@kgriffs Sorry for the delay. Just fixed the issues you addressed.

kgriffs:
Nice work! LGTM.

Issue #598, opened 2015-09-09
Add client IP properties to the Request class


kgriffs:
Add helpers to get client's IP:

The user agent IP, which would check for:
* Forwarded (RFC 7239) 
* X-Forwarded-For
* X-Real-IP
* environ[‘REMOTE_ADDR’]

And one that always just returns REMOTE_ADDR in case you really do want to know the origin of the last hop.

philiptzou:
Cool. I could help coding (I know flask well).

kgriffs:
[implemented]

Issue #539, opened 2015-05-04
Add client_ip property to Request


kgriffs:
Add a client IP property to the Request class. A few relevant SO articles here:

* http://stackoverflow.com/questions/22868900/how-do-i-safely-get-the-users-real-ip-address-in-flask-using-mod-wsgi
* http://stackoverflow.com/questions/7835030/obtaining-client-ip-address-from-a-wsgi-app-using-eventlet

sebasmagri:
I'd like to see this implemented as a core middleware but not enabled by default to avoid the performance penalty if some more advanced logic such as the trusted proxies stuff is implemented.

pior:
What about in the core API, but with lazy evaluation?

kgriffs:
[duplicate of #598]

PR #599, opened 2015-09-09
Add `access_route` and `remote_addr` to Request object

philiptzou:
which is similar to Werkzeug but more powerful

Supported:

1. fetch from "Forwarded" header (defined by RFC7239)
2. fetch from "X-Forwarded-For" header
3. fetch from "X-Read-IP" header
4. fetch from wsgi "REMOTE_ADDR" header

Related to #539 and #598

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -29,7 +29,7 @@ import six
 
 from falcon.errors import *
 from falcon import util
-from falcon.util.uri import parse_query_string, parse_host
+from falcon.util.uri import parse_query_string, parse_host, unquote_string
 from falcon import request_helpers as helpers
 
 # NOTE(tbug): In some cases, http_cookies is not a module
@@ -198,6 +198,7 @@ class Request(object):
         '_wsgierrors',
         'options',
         '_cookies',
+        '_cached_access_route',
     )
 
     # Allow child classes to override this
@@ -258,6 +259,7 @@ class Request(object):
         self._cached_headers = None
         self._cached_uri = None
         self._cached_relative_uri = None
+        self._cached_access_route = None
 
         try:
             self.content_type = self.env['CONTENT_TYPE']
@@ -513,6 +515,35 @@ class Request(object):
             self._cookies = cookies
 
         return self._cookies.copy()
+
+    @property
+    def access_route(self):
+        """If a forwarded header exists this is a list of all ip addresses
+        from the client ip to the last proxy server.
+
+        Inspired by werkzeug's access_route
+        """
+        if self._cached_access_route is None:
+            access_route = []
+            if 'HTTP_FORWARDED' in self.env:
+                access_route = self._parse_rfc_forwarded()
+            if not access_route and 'HTTP_X_FORWARDED_FOR' in self.env:
+                # we don't handle X-Fowarded-By since it is not possible to
+                # know in which order the already existing fields were added
+                # just as rfc7239 stated
+                access_route = [ip.strip() for ip in
+                                self.env['HTTP_X_FORWARDED_FOR'].split(',')]
+            if not access_route and 'HTTP_X_REAL_IP' in self.env:
+                access_route = [self.env['HTTP_X_REAL_IP']]
+            if not access_route and 'REMOTE_ADDR' in self.env:
+                access_route = [self.env['REMOTE_ADDR']]
+            self._cached_access_route = access_route
+        return self._cached_access_route
+
+    @property
+    def remote_addr(self):
+        """The remote address of the client."""
+        return self.env.get('REMOTE_ADDR')
 
     # ------------------------------------------------------------------------
     # Methods
@@ -1027,6 +1058,22 @@ class Request(object):
             )
 
             self._params.update(extra_params)
+
+    def _parse_rfc_forwarded(self):
+        """rfc 7239"""
+        addr = []
+        for forwarded in self.env['HTTP_FORWARDED'].split(','):
+            for param in forwarded.split(';'):
+                param = param.strip().split('=', 1)
+                if len(param) == 1:
+                    continue
+                key, val = param
+                if key.lower() not in ('for', 'by'):
+                    # we only want for/by params
+                    continue
+                host, _ = parse_host(unquote_string(val))
+                addr.append(host)
+        return addr
 
 
 # PERF: To avoid typos and improve storage space and speed over a dict.
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -27,7 +27,7 @@ except AttributeError:  # pragma nocover
 import mimeparse
 import six
 
-from falcon.errors import *
+from falcon.errors import *  # NOQA
 from falcon import util
 from falcon.util.uri import parse_query_string, parse_host, unquote_string
 from falcon import request_helpers as helpers
@@ -518,19 +518,30 @@ class Request(object):
 
     @property
     def access_route(self):
-        """If a forwarded header exists this is a list of all ip addresses
-        from the client ip to the last proxy server.
+        """A list of all addresses from client to the last proxy server.
 
-        Inspired by werkzeug's access_route
+        Inspired by werkzeug's `access_route`. Notice this list may contain
+        string(s) other than IPv4 / IPv6 address. For example the "unknown"
+        identifier and obfuscated identifier defined by RFC 7239.
+
+        Warning: HTTP Forwarded headers can be forged by any client or proxy.
+        Use this property with caution and write your own verify function.
+        The best practice here is always using `.remote_addr` unless your
+        application is hosted behind some reverse proxy servers. Also only
+        trust the **last** N addresses provided by reverse proxy server(s).
+
+        This property will try to derive addresses sequentially from:
+            - `Forwarded`;
+            - `X-Forwarded-For`;
+            - `X-Real-IP`;
+            - **or** the IP address of the closest client/proxy
+
         """
         if self._cached_access_route is None:
             access_route = []
             if 'HTTP_FORWARDED' in self.env:
                 access_route = self._parse_rfc_forwarded()
             if not access_route and 'HTTP_X_FORWARDED_FOR' in self.env:
-                # we don't handle X-Fowarded-By since it is not possible to
-                # know in which order the already existing fields were added
-                # just as rfc7239 stated
                 access_route = [ip.strip() for ip in
                                 self.env['HTTP_X_FORWARDED_FOR'].split(',')]
             if not access_route and 'HTTP_X_REAL_IP' in self.env:
@@ -538,11 +549,20 @@ class Request(object):
             if not access_route and 'REMOTE_ADDR' in self.env:
                 access_route = [self.env['REMOTE_ADDR']]
             self._cached_access_route = access_route
+
         return self._cached_access_route
 
     @property
     def remote_addr(self):
-        """The remote address of the client."""
+        """String of the IP address of the closest client/proxy.
+
+        This property will only derive address directly from WSGI `REMOTE_ADDR`
+        header which can not be mofidied by any client or proxy.
+
+        If your application is behind one or more reverse proxies, you may use
+        `.access_remote` to retrieve the real IP address of client. Please See
+        the document of `.access_remote` for more detail.
+        """
         return self.env.get('REMOTE_ADDR')
 
     # ------------------------------------------------------------------------
@@ -1061,7 +1081,11 @@ class Request(object):
             self._params.update(extra_params)
 
     def _parse_rfc_forwarded(self):
-        """rfc 7239"""
+        """Parse RFC 7239 "Forwarded" header.
+
+        Returns:
+            list: addresses derived from "for" parameters.
+        """
         addr = []
         for forwarded in self.env['HTTP_FORWARDED'].split(','):
             for param in forwarded.split(';'):
@@ -1070,7 +1094,7 @@ class Request(object):
                     continue
                 key, val = param
                 if key.lower() != 'for':
-                    # we only want for/by params
+                    # we only want for params
                     continue
                 host, _ = parse_host(unquote_string(val))
                 addr.append(host)
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -538,10 +538,11 @@ class Request(object):
             proxy servers.
 
         This property will try to derive addresses sequentially from:
-            - ``Forwarded``;
-            - ``X-Forwarded-For``;
-            - ``X-Real-IP``;
-            - **or** the IP address of the closest client/proxy.
+
+            - ``Forwarded``
+            - ``X-Forwarded-For``
+            - ``X-Real-IP``
+            - **or** the IP address of the closest client/proxy
 
         """
         if self._cached_access_route is None:
@@ -564,7 +565,7 @@ class Request(object):
         """String of the IP address of the closest client/proxy.
 
         Address will only be derived from WSGI ``REMOTE_ADDR`` header, which
-        can not be mofidied by any client or proxy.
+        can not be modified by any client or proxy.
 
         Note:
             If your application is behind one or more reverse proxies, you may
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -520,21 +520,28 @@ class Request(object):
     def access_route(self):
         """A list of all addresses from client to the last proxy server.
 
-        Inspired by werkzeug's `access_route`. Notice this list may contain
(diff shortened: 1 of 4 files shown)


miguelramos:
When will have this feature?

philiptzou:
@miguelramos I don't know. I wish someone who has time could see this pull request and review it.

philiptzou:
Rebased onto the latest version.

philiptzou:
Rebased onto the latest version. Again.

falcon/request.py
```diff
@@ -1028,6 +1059,22 @@ def _parse_form_urlencoded(self):
...
+        for forwarded in self.env['HTTP_FORWARDED'].split(','):
+            for param in forwarded.split(';'):
+                param = param.strip().split('=', 1)
+                if len(param) == 1:
+                    continue
+                key, val = param
+                if key.lower() not in ('for', 'by'):
+                    # we only want for/by params
+                    continue
+                host, _ = parse_host(unquote_string(val))
```
kgriffs: Practically speaking, do we need to do anything beyond simply removing the enclosing double quotes?

Looking at RFC 7239, quoted strings are only required for IPv6 addresses, since those contain invalid token characters. Obfuscated identifiers and IPv4 nodes do not contain invalid token characters, and so do not need to be quoted (although the sender may choose to do so). Regardless, none of the allowed characters for the different node identifiers need to be escaped (with a backslash), and so it seems unlikely that a sender would ever do so.

Indeed, RFC 7230 states that senders should only use escaping when absolutely necessary. On the other hand, the RFC requires that the receiver unescape any `quoted-pair`'s, if present. But again, it seems highly unlikely that a Falcon app would ever come across any `quoted-pair`'s in the wild within the context of a Forwarded header.

Interestingly enough, Werkzeug doesn't strictly follow RFC 7320; `werkzeug.http.unquote_header_value` [only handles][1] escaped backslashes and double quotes. These are the two characters that RFC 7320 specifically calls out as the only cases in which a sender should generate a `quoted-pair`. But note that these two characters are actually disallowed by the ABNF for the Forwarded header.

[1]: https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/http.py#L198

philiptzou: Well, I didn't notice the "SHOULD NOT" part of RFC 7230 before. I just followed the requirement of ABNF:

```
quoted-pair    = "\" ( HTAB / SP / VCHAR / obs-text )
```

Since `VCHAR` stands for "any visible USASCII character", according to the ABNF (only), it seems legal to generate a `quoted-string` with unnecessary `quoted-pair`, for example `"127\.0\.0\.1"`. That's the reason I introduced the `unquote_string()` function at first.

So there's a balance to be made between the robust and performance. What do you think @kgriffs?

falcon/request.py
```diff
@@ -1028,6 +1059,22 @@ def _parse_form_urlencoded(self):
...
+    def _parse_rfc_forwarded(self):
+        """rfc 7239"""
+        addr = []
+        for forwarded in self.env['HTTP_FORWARDED'].split(','):
+            for param in forwarded.split(';'):
+                param = param.strip().split('=', 1)
+                if len(param) == 1:
+                    continue
+                key, val = param
+                if key.lower() not in ('for', 'by'):
```
kgriffs: It seems like we should only return "for" parameters to be consistent with the semantics of X-Forwarded-For, and to avoid ambiguity in the returned `addr` list. What do you think? As an alternative, we might compose a richer data structure that includes the "for" and "by" designations. To be honest, I don't have a good sense of how many apps would make use of the "by" parameter.

falcon/request.py
kgriffs: Indeed, it's not expressly forbidden by the RFC. It's interesting that werkzeug cheats a little bit here and doesn't actually try to unescape anything besides `"` and `\` characters. FWIW, the werkzeug code is about twice as fast as the list comprehension approach on my MBP.

Anyway, I guess we have three options:

(a) Don't try to unescape anything, only strip enclosing quotes. Simplest and most performant option. Some amount of risk, but likely minimal.

(b) Follow werkzeug's lead and only worry about the cases in which the RFC recommends senders SHOULD use escaping (for `"` and `\` chars). Fairly performant. But wouldn't actually be less risky than (a) in the case of the Forwarded header, since those two chars aren't used in addresses or obfuscated identifiers anyway.

(c) Strictly follow the RFC, and perhaps experiment with the implementation to see if we can make it a little bit more efficient.

Presumably, werkzeug has had pretty good luck with their strategy. It's not perfect, but is reasonably efficient. I'm guessing that senders tend not to escape characters unnecessarily, and that will carry over to their use of the new Forwarded header. Therefore, I would personally lean toward option (a). This is just my $0.02; happy to consider other opinions.

falcon/request.py
philiptzou: Good point, will do the modification (only return "for" parameters) with the `unquote_string`.

falcon/request.py
philiptzou: I think these are all the options we now have. I guess if we can find more efficient which can match the performance werkzeug has, we can stick to (c). Or we just switch to werkzeug code (b).

philiptzou: Well after I did an SNS optimization to unquote_string it becomes faster than werkzeug (in most time):

```python
from timeit import timeit
timeit('unquote(r\'"127.0.0.1"\')', 'from falcon.util.uri import unquote_string as unquote')
# 0.24731779098510742
timeit('unquote(r\'"127.0.0.1"\')', 'from werkzeug.http import unquote_header_value as unquote')
# 0.42057013511657715
timeit('unquote(r\'"127.0.0.1"\')', 'from falcon.util.uri import unquote_string_old as unquote')
# 0.6000280380249023
```

And even quoted `"`:

```python
timeit('unquote(r\'"\\"unknown\\""\')', 'from falcon.util.uri import unquote_string as unquote')
# 0.4324629306793213
timeit('unquote(r\'"\\"unknown\\""\')', 'from werkzeug.http import unquote_header_value as unquote')
# 0.5061039924621582
timeit('unquote(r\'"\\"unknown\\""\')', 'from falcon.util.uri import unquote_string_old as unquote')
# 0.6474518775939941
```

Well only the double backslash `\` are slower:

```python
timeit('unquote(r\'"bad\\\\"\')', 'from falcon.util.uri import unquote_string as unquote')
# 0.8454480171203613
timeit('unquote(r\'"bad\\\\"\')', 'from werkzeug.http import unquote_header_value as unquote')
# 0.47409915924072266
timeit('unquote(r\'"bad\\\\"\')', 'from falcon.util.uri import unquote_string_old as unquote')
# 0.7625319957733154
```

philiptzou:
@kgriffs let me know if you are good with my last two modifications. So I can rebase again.

falcon/util/uri.py
```diff
@@ -376,3 +376,31 @@ def parse_host(host, default_port=None):
...
+        return quoted
+    elif tmp_quoted[0] != '"' or tmp_quoted[-1] != '"':
+        # return original one, prevent side-effect
+        return quoted
+    tmp_quoted = tmp_quoted[1:-1]
+    if '\\' not in tmp_quoted:
+        return tmp_quoted
+    elif r'\\' not in tmp_quoted:
+        return tmp_quoted.replace('\\', '')
+    else:
```
kgriffs: :+1: I like this approach of short-circuiting the parsing based on a couple of quick string searches. Most strings will not contain `quoted-pair`'s

falcon/util/uri.py
```diff
@@ -376,3 +376,31 @@ def parse_host(host, default_port=None):
...
+        TypeError: `quoted` was not a ``str``.
+    """
+    tmp_quoted = quoted.strip()
+    if len(tmp_quoted) < 2:
+        return quoted
+    elif tmp_quoted[0] != '"' or tmp_quoted[-1] != '"':
+        # return original one, prevent side-effect
+        return quoted
+    tmp_quoted = tmp_quoted[1:-1]
+    if '\\' not in tmp_quoted:
```
kgriffs: Would you mind prefacing L400-406 with a comment explaining why this approach was taken? E.g.,:

```python
# PERF(philiptzou): blah blah blah...
```

falcon/util/uri.py
```diff
@@ -376,3 +376,31 @@ def parse_host(host, default_port=None):
...
+    Raises:
+        TypeError: `quoted` was not a ``str``.
+    """
+    tmp_quoted = quoted.strip()
+    if len(tmp_quoted) < 2:
+        return quoted
+    elif tmp_quoted[0] != '"' or tmp_quoted[-1] != '"':
+        # return original one, prevent side-effect
+        return quoted
+    tmp_quoted = tmp_quoted[1:-1]
```
kgriffs: Nitpick: Add a blank line before L399 to improve readability

falcon/util/uri.py
```diff
@@ -376,3 +376,31 @@ def parse_host(host, default_port=None):
     # or a domain name plus a port
     name, _, port = host.partition(':')
     return (name, int(port))
+
+
+def unquote_string(quoted):
+    """unquote a rfc7320 "quoted-string" into normal one
```
kgriffs: Style
* First word in the first line of a comment should be capitalized.
* First line of a comment should end in a period.
* `rfc7320` ==> `RFC 7320`

Grammar
* `a RFC 7320 "quoted-string"` ==> `an RFC 7320 "quoted-string"`
* `into normal one` is superfluous and can be removed from the sentence.

falcon/request.py
```diff
@@ -1028,6 +1060,22 @@ def _parse_form_urlencoded(self):
 
             self._params.update(extra_params)
 
+    def _parse_rfc_forwarded(self):
+        """rfc 7239"""
```
kgriffs: Style: 
* `rfc` ==> `RFC`
* The first line of a comment should end in a period.

falcon/request.py
```diff
@@ -514,6 +516,35 @@ def cookies(self):
...
+        from the client ip to the last proxy server.
+
+        Inspired by werkzeug's access_route
+        """
+        if self._cached_access_route is None:
+            access_route = []
+            if 'HTTP_FORWARDED' in self.env:
+                access_route = self._parse_rfc_forwarded()
+            if not access_route and 'HTTP_X_FORWARDED_FOR' in self.env:
+                # we don't handle X-Fowarded-By since it is not possible to
```
kgriffs: Can this comment be removed, since access_route is only concerned with `for` addresses now?

falcon/request.py
```diff
@@ -514,6 +516,35 @@ def cookies(self):
 
         return self._cookies.copy()
 
+    @property
+    def access_route(self):
+        """If a forwarded header exists this is a list of all ip addresses
```
kgriffs: Style: First line of a comment should be short enough to avoid wrapping. Additional documentation can be placed in subsequent paragraphs, separated by a blank line.

tests/test_access_route.py
```diff
@@ -0,0 +1,74 @@
...
+        self.assertEqual(req.remote_addr, '127.0.0.1')
+
+    def test_rfc_forwarded(self):
+        req = Request(testing.create_environ(
+            host='example.com',
+            path='/access_route',
+            headers={
+                'Forwarded': ('for=192.0.2.43,for=,'
+                              'for="[2001:db8:cafe::17]:555",'
+                              'for="unknown", by=_hidden,for="\\"\\\\",'
```
kgriffs: Would you mind adding an additional test that includes a `for` value with a mixture of regular chars, escaped backslashes, and  escaped ASCII chars? It might make sense to do this in `tests/test_utils.py` directly against `uri.unquote_string()`

falcon/request.py
```diff
@@ -514,6 +516,35 @@ def cookies(self):
...
+            if not access_route and 'HTTP_X_REAL_IP' in self.env:
+                access_route = [self.env['HTTP_X_REAL_IP']]
+            if not access_route and 'REMOTE_ADDR' in self.env:
+                access_route = [self.env['REMOTE_ADDR']]
+            self._cached_access_route = access_route
+        return self._cached_access_route
+
+    @property
+    def remote_addr(self):
+        """The remote address of the client."""
```
kgriffs: I think the reader may be confused about what this method does and when to use it vs. `access_route`. Would you mind updating the docstring to make this more clear?

falcon/request.py
```diff
@@ -514,6 +516,35 @@ def cookies(self):
...
+                # know in which order the already existing fields were added
+                # just as rfc7239 stated
+                access_route = [ip.strip() for ip in
+                                self.env['HTTP_X_FORWARDED_FOR'].split(',')]
+            if not access_route and 'HTTP_X_REAL_IP' in self.env:
+                access_route = [self.env['HTTP_X_REAL_IP']]
+            if not access_route and 'REMOTE_ADDR' in self.env:
+                access_route = [self.env['REMOTE_ADDR']]
+            self._cached_access_route = access_route
+        return self._cached_access_route
```
kgriffs: Nitpick: Add a blank line before L541 to improve readability.

kgriffs:
@philiptzou I think this approach is a good one. Comments inline.

tests/test_access_route.py
```diff
@@ -0,0 +1,74 @@
...
+                              'for="_don\\\"t_\\try_this\\\\at_home_\\42",'
```
philiptzou: @kgriffs: Do you mean like this?

philiptzou:
I opened another pull-request to put the "rebased" version of this modification. Is that okay to keep these two pull-requests open until #640 get merged, @kgriffs?

falcon/request.py
```diff
@@ -514,6 +516,55 @@ def cookies(self):
...
 
+    @property
+    def access_route(self):
+        """A list of all addresses from client to the last proxy server.
+
+        Inspired by werkzeug's `access_route`. Notice this list may contain
+        string(s) other than IPv4 / IPv6 address. For example the "unknown"
+        identifier and obfuscated identifier defined by RFC 7239.
+
+        Warning: HTTP Forwarded headers can be forged by any client or proxy.
```
kgriffs: So that sphinx will highlight this paragraph as a warning, can you indent the paragraph, e.g:

```
Warning:
    HTTP Forwarded headers can be forged by any client or proxy.
    Use this property with caution and...
```

https://gist.github.com/kgriffs/05667fb34fb3b69b8030

kgriffs: BTW, you can check how the docs are rendered by making them locally:

https://gist.github.com/kgriffs/05667fb34fb3b69b8030

tests/test_access_route.py
kgriffs: Looks good, thanks!

falcon/request.py
```diff
@@ -514,6 +516,63 @@ def cookies(self):
...
+            self._cached_access_route = access_route
+
+        return self._cached_access_route
+
+    @property
+    def remote_addr(self):
+        """String of the IP address of the closest client/proxy.
+
+        Address will only be derived from WSGI ``REMOTE_ADDR`` header, which
+        can not be mofidied by any client or proxy.
```
kgriffs: "mofidied" ==> "modified"

falcon/request.py
```diff
@@ -514,6 +516,63 @@ def cookies(self):
...
+
+        Warning:
+            HTTP Forwarded headers can be forged by any client or proxy.
+            Use this property with caution and write your own verify function.
+            The best practice is always using :py:attr:`~.remote_addr` unless
+            your application is hosted behind some reverse proxy server(s).
+            Also only trust the **last N** addresses provided by those reverse
+            proxy servers.
+
+        This property will try to derive addresses sequentially from:
```
kgriffs: I noticed that Sphinx was highlighting this line when building the docs. Adding a blank line before the list, directly after L540, resolves the problem.

falcon/request.py
```diff
@@ -514,6 +516,63 @@ def cookies(self):
...
+            - ``Forwarded``;
```
kgriffs: I think we can leave off the semicolons from these items in the list.

kgriffs:
@philiptzou Just a few minor issues and I think this is ready to merge. Thanks!

philiptzou:
@kgriffs Done. Please merge #640 and close this pull-request. Thanks!

kgriffs:
@philiptzou OK, I think we are ready to merge. Just one final step: Please squash all the patches down to a single commit and format the first line with a "feat(Request): Blah blah" prefix, per https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#commit-message-format.

PR #601, opened 2015-09-11
Document 'HTTPStatus' class. Fixes #590

### LINKED ISSUE #590: Document HTTPStatus
A new class, HTTPStatus, was added during the PyCon 2015 sprint, but I just discovered it wasn't documented. This class should at least be mentioned on the "Status Codes" page.

maxking:


doc/api/status.rst
```diff
@@ -22,6 +22,10 @@ Or, using the more verbose name:
 Using these constants helps avoid typos and cuts down on the number of
 string objects that must be created when preparing responses.
 
+Falcon also provides a generic 'HTTPStatus' class. Raise this class from a hook,
```
kgriffs: What do you think about expanding this to be similar to errors.rst, including pulling in the docstring?

maxking: Something like this?

Base Class
----------

.. autoclass:: falcon.http_status.HTTPStatus
    :members:

kgriffs: Right, although in this case we don't have any predefined child classes, so saying "Base Class" may be confusing. You  could simply use "HTTPStatus" as the heading instead.

kgriffs:
@maxking Just checking in to see if you have some time to get this over the finish line this week?

maxking:
@kgriffs I can maybe finish this off over the weekend? Maybe before but not very sure about that.

kgriffs:
Looks great, thanks for working on this!

PR #603, opened 2015-09-11
Request.client_accepts_msgpack now supports 'application/msgpack

### LINKED ISSUE #544: Request.client_accepts_msgpack only supports 'application/x-msgpack'
The use of the 'x-' prefix is now discouraged for media types. We should update this Request property to also return True for 'application/msgpack', and verify the change with additional tests.

maxking:
Fixes #544


kgriffs:
Merged via #676

PR #604, opened 2015-09-11
Refactor setup.py external module selection for Cython

### LINKED ISSUE #516: Refactor setup.py external module selection for Cython
Reduce complexity and duplication of the external module selection for Cython in setup.py.

At the time of this issue, this cruft was located in: 
https://github.com/kgriffs/falcon/blob/routing/setup.py#L35

maxking:
Fixes #516 
I am not very sure, but hopefully this is what the issue was about?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ if CYTHON:
 
     for each in dirs:
         ext_modules += [Extension(each + '.' + ext,
-                        [path.join(*each.split('.'), ext + '.py')])
+                        [path.join(*(each.split('.') + [ext + '.py']))])
                         for ext in list_modules(path.join(MYDIR, *each.split('.')))]
 
     cmdclass = {'build_ext': build_ext}
--- a/setup.py
+++ b/setup.py
@@ -52,21 +52,12 @@ if CYTHON:
 
         return module_names
 
-    ext_modules = [
-        Extension('falcon.' + ext, [path.join('falcon', ext + '.py')])
-        for ext in list_modules(path.join(MYDIR, 'falcon'))]
+    dirs = ['falcon', 'falcon.util', 'falcon.routing']
+    ext_modules = []
 
-    ext_modules += [
-        Extension('falcon.util.' + ext,
-                  [path.join('falcon', 'util', ext + '.py')])
-
-        for ext in list_modules(path.join(MYDIR, 'falcon', 'util'))]
-
-    ext_modules += [
-        Extension('falcon.routing.' + ext,
-                  [path.join('falcon', 'routing', ext + '.py')])
-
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -44,21 +44,12 @@ def list_modules(dirname):
...
-
-    ext_modules += [
-        Extension('falcon.routing.' + ext,
-                  [path.join('falcon', 'routing', ext + '.py')])
+    dirs = ['falcon', 'falcon.util', 'falcon.routing']
+    ext_modules = []
 
-        for ext in list_modules(path.join(MYDIR, 'falcon', 'routing'))]
+    for each in dirs:
+        ext_modules += [Extension(each + ext, [path.join(each, ext + '.py')])
```
kgriffs: `each` will need to be split on `.` and then re-joined into a path, e.g.:

```py
path.join(*each.split('.'))
```

Also, the module name is missing a `.` between `each` and `ext`.

setup.py
```diff
@@ -44,21 +44,12 @@ def list_modules(dirname):
...
+                            for ext in list_modules(path.join(MYDIR, each))]
```
kgriffs: Similar to the line above, `each` here is a dotted module name, and so would need to be split first.

kgriffs:
Hi @maxking, just following up to see where we are on this PR

maxking:
@kgriffs Oh, I missed your comments on this one. I will update this over the weekend.

setup.py
```diff
@@ -52,21 +52,13 @@ def list_modules(dirname):
...
-    ext_modules += [
-        Extension('falcon.routing.' + ext,
-                  [path.join('falcon', 'routing', ext + '.py')])
+    dirs = ['falcon', 'falcon.util', 'falcon.routing']
+    ext_modules = []
 
-        for ext in list_modules(path.join(MYDIR, 'falcon', 'routing'))]
+    for each in dirs:
+        ext_modules += [Extension(each + '.' + ext,
+                        [path.join(*each.split('.'), ext + '.py')])
```
kgriffs: This looks good, just needs a little modification. Currently, running this fails with `SyntaxError: only named arguments may follow *expression`. One way you might work around this is by using the extend operator (`+`), e.g.:

```py
path.join(*(each.split('.') + [ext + '.py'])) 
```

kgriffs:
Hi @maxking, just checking in to see if you had a chance to review my last comment?

maxking:
Hi! @kgriffs, I am afraid, I did not. Have been a little busy. I will update this over the weekend and let you know about it.

kgriffs:
Thanks! Merged as part of #741

PR #608, opened 2015-09-11
fix(response.py): Pylint/pep8 errors when default python is 3.X

### LINKED ISSUE #607: pep8 and pylint failing on master (response.py)
pep8 and pylint are failing on the master branch, due to an undefined reference to unicode in response.py  I am investigating this and will submit a PR shortly.

lwcolton:
When running tox on a machine with python3 as the default python,
pep8 and pylint would fail on a reference to unicode in response.py
The fix was simply to set the default python to 2.7 in tox.ini
for the pep8 and pylint environments

Fixes #607

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -69,6 +69,7 @@ commands = py3kwarn falcon
 
 [testenv:pep8]
 deps = flake8
+basepython = python2.7
 commands = flake8 \
              --max-complexity=15 \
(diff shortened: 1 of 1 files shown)


kgriffs:
[merged manually]

kgriffs:
Thanks! Nice catch.

PR #613, opened 2015-09-18
fix(request) Workarount for nonlatin1 path info (PEP3333 compliance)

bobuk:


bobuk:
Fix and test for #562, #571

falcon/request.py
```diff
@@ -220,6 +220,10 @@ def __init__(self, env, options=None):
         # Normalize path
         path = env['PATH_INFO']
         if path:
+            if six.PY3:
+                # PEP 3333 specifies that PATH_INFO variable are always
+                # "bytes tunneled as latin-1" and must be encoded back
+                path = path.encode('latin1').decode('utf-8', 'replace')
             if len(path) != 1 and path.endswith('/'):
```
kgriffs: Could you add a blank line here (just above L227) to improve readability?

tests/test_req_vars.py
```diff
@@ -107,6 +108,18 @@ def test_reconstruct_url(self):
 
         self.assertEqual(expected_uri, req.uri)
 
+    def test_nonlatin_path(self):
+        if not six.PY3:
```
kgriffs: Consider using the `testtools.skipUnless` decorator to make this a little cleaner, i.e.:

```python
@testtools.skipUnless(six.PY3, 'Test only applies to Python 3')
def test_nonlatin_path(self):
```

tests/test_req_vars.py
```diff
@@ -107,6 +108,18 @@ def test_reconstruct_url(self):
...
+            self.skipTest('Test only applies to Python 3')
+
+        cyrillic_path = six.u('/hello_\u043f\u0440\u0438\u0432\u0435\u0442' )
```
kgriffs: Nit: No need to use `six.u`, since Falcon only supports versions of Python that recognize the literal `u` prefix.

falcon/request.py
```diff
@@ -220,6 +220,10 @@ def __init__(self, env, options=None):
         # Normalize path
         path = env['PATH_INFO']
         if path:
+            if six.PY3:
```
kgriffs: Unfortunately, this needs a `# pragma no cover` comment since it is only covered for one of the test envs.

kgriffs:
Thanks, this is looking good, just a few minor suggestions. I manually tested this approach with gunicorn under python 3.3 and 3.4 and it looks good.

bobuk:
everything is changed with your recommendations

kgriffs:
Thanks! Looks like there are just two tiny pep8 errors to fix, and this is ready to merge.

bobuk:
how about now? and thank you for your patience

kgriffs:
Looks good, thanks again!

[merged manually]

PR #615, opened 2015-09-18
fix(hooks): Improve detection of requests arg

kgriffs:
In order to maintain backwards-compatibility with old hooks that do
not accept a requests argument, the framework performs introspection
when attaching hooks and inserts shims as needed. The introspection
raises an error for hooks that have been defined via
functools.partial().

Fix this by adding support for partial methods and generally making
the introspection more explicit. Modify tests to cover all the different
cases.

### DIFF
--- a/falcon/hooks.py
+++ b/falcon/hooks.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import functools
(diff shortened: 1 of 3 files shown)


fxfitz:
LGTM :+1:

painterjd:
:+1:

Issue #556, opened 2015-06-10
http_cookies importing error on python 2.7.6


haandol:
I just bumped up falcon from 0.2 to 0.3
importing http_cookies raises ImportError.

```python
at requests.py

from six.moves.http_cookies import SimpleCookie
```

while above occurs error, but below code on my terminal, that was fine.
```python

from six.moves import http_cookies
http_cookies.SimpleCookie
```

I solved this issue to bump python up from 2.7.6 to 2.7.10

kgriffs:
Did you update python in place, or was this a fresh environment with a new install of six?

tbug:
I can confirm this issue.
Python 2.7.6 with six==1.9.0
Six _kind of_ touches this in the docs here: http://pythonhosted.org/six/#module-six.moves
but it is not clear in what situations it won't work.

Easy fix would be to just import `http_cookies` as @haandol mentions, and
```py
SimpleCookie = http_cookies.SimpleCookie
```
after the import.

lucifersteph:
Im getting the same issue:

env: mac os yosemite 10.10.5 / python 2.7.10 / falcon 0.3.0

And I modified the source files as haandol said above, it worked.

tbug:
Aight, I'll do a PR to fix this.

kgriffs:
[fixed in #622]

PR #622, opened 2015-09-28
fix(cookies): Fixed issue #556 ImportError from six

tbug:
Fixes issue with ImportError in some rare cases.
IMO this is a six issue, but in this case it seems reasonable to work around it.

Applying the fix discussed in https://github.com/falconry/falcon/issues/556


kgriffs:
Thanks for taking care of this issue. Too bad six let this abstraction leak...

PR #625, opened 2015-10-02
fix(Request): Make wsgi.input wrapper more robust

kgriffs:
Fix two bugs with the wsgi.input wrapper that is used to normalize
behavior between wsgiref and production-class web servers:

1. Do not hang when reading with a size < Content-Length but the
   stream has already been consumed.
2. If Content-Length is invalid or missing, assume no content and
   still wrap the stream, rather than sometimes wrapping and sometimes
   not.

Also, improve the tests to more cleanly shut down the wsgiref server
between tests so that the above can be tested in a more modular
fashion.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1055,15 +1055,14 @@ class Request(object):
 
     def _wrap_stream(self):  # pragma nocover
         try:
-            # NOTE(kgriffs): We can only add the wrapper if the
-            # content-length header was provided.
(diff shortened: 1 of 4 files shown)


fxfitz:
LGTM, @kgriffs

fxfitz:
I'll merge if you rebase. ;-)

kgriffs:
@fxfitz Thanks, rebased.

PR #629, opened 2015-10-09
Do not display Cython note when installing under Jython

kgriffs:
Jython does not support Cython, so there is no need to bother users
with a note about it during the installation process. Detect Jython
and supress the message in that case. Also, refactor the detection
logic in setup.py to be more readable and explicit.


jmvrbanac:
:+1:

PR #631, opened 2015-10-12
doc(Request): Note what happends when stream_len is not set

kgriffs:
Expand the docstring for steam_len to provide information about what
the developer can expect will happen if they set stream but not
steam_len.


fxfitz:
Makes sense to me! :+1: Just needs to be rebased.

kgriffs:
@fxfitz Thanks! Rebased.

Issue #145, opened 2013-05-21
Resource serialization/deserialization to/from representation media types


kgriffs:
How to do serialization/deserialization is a common question from the community. How can we make this work out of the box?

Things to decide:
- Should this functionality be implemented in Talons as middleware for falcon>=0.2 ?
- Alternatively, should this be implemented directly in the framework? 
- What media types should be supported?
- In the case of JSON, should the implementation support other UTF encodings besides UTF-8?
- How will the solution support versioned / custom media types?

kgriffs:
Adding comment from @sebasmagri from duplicate issue #324:

> I'd like to think about a Content Type handlers approach which would allow devs to get and return native objects without worrying about the content type in responders, and implement generic deserializers and serializers for each content type they want to be able to handle.

sebasmagri:
I still believe the Content Type Handler approach would make sense as in:

* Providing a standard interface to implement and register content type handlers, as well as the elements they could access (params, files) and the parameters they could get (just as for middlewares)
* To provide at least two standard/reference implementations in the core. They could be JSON and Forms (form data, multipart) if they don't introduce new dependencies. Handlers with third party dependencies could be included in Talons.

Upon these two premises we could add more features like being able to use a content-type handler globally or for specific endpoints (like for image processing or custom chat protocols), probably by inheritance or reuse of some logic from middlewares.

kgriffs:
See also this discussion on the ML for more ideas: http://librelist.com/browser//falcon/2015/10/5/base-resource-class-to-implement-jsontranslate-middleware-functionality/

kgriffs:
I like the idea of being able to register `Content-Type` handlers (with JSON available out of the box). But I am reluctant to overload existing `Request` and `Response` properties since this may lead to some confusion in practice. One suggestion from @BenjamenMeyer on the mailing list was to add additional well-named properties, such as 'req.json' to expose the serialization mechanism. But how would we extend this model to arbitrary Internet media types per @sebasmagri's suggestion? Perhaps we might add a generic attribute to the `Request` and `Response` classes, e.g.: `req.media` and `resp.media`. 

One concern I have with this last approach is that `req.media` would consume the input stream. We would need to decide if that is OK or if it would lead to violating the Principle of Least Surprise (this was problematic in the past with URL-encoded form POSTs).

BenjamenMeyer:
[...]

PR #642, opened 2015-10-29
feat(request): add support to parse request body as json

jmvrbanac:
Adding in the ability to easily get the request body as a json dictionary. I tend to do this operation in nearly every Falcon project I create, so it seems appropriate to upstream this into the request itself.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 from datetime import datetime
+import json
 
 try:
     # NOTE(kgrifs): In Python 2.6 and 2.7, socket._fileobject is a
@@ -188,6 +189,9 @@ class Request(object):
             A dict of name/value cookie pairs.
             See also: :ref:`Getting Cookies <getting-cookies>`
 
+        json (dict): Attempts to parse the request stream as a Json string.
+            If called, it'll leave the request stream at EOF.
+
     """
 
     __slots__ = (
@@ -206,6 +210,7 @@ class Request(object):
         'options',
         '_cookies',
         '_cached_access_route',
+        '_json',
     )
 
     # Allow child classes to override this
@@ -292,6 +297,11 @@ class Request(object):
             # pylint will detect this as not-callable because it only sees the
             # declaration of None, not whatever type a subclass may have set.
             self.context = self.context_type()  # pylint: disable=not-callable
+
+        # NOTE(jmvrbanac): To preserve performance, we don't want to
+        # automatically parse the body as Json so we're just going to leave
+        # self._json empty until self.json is called
+        self._json = None
 
     # ------------------------------------------------------------------------
     # Properties
@@ -522,6 +532,12 @@ class Request(object):
             self._cookies = cookies
 
         return self._cookies.copy()
+
+    @property
+    def json(self):
+        if self._json is None:
+            self._parse_application_json()
+        return self._json
 
     @property
     def access_route(self):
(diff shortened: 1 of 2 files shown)


falcon/request.py
```diff
@@ -188,6 +189,9 @@ class Request(object):
             A dict of name/value cookie pairs.
             See also: :ref:`Getting Cookies <getting-cookies>`
 
+        json (dict): Attempts to parse the request stream as a Json string.
```
philiptzou: I think it should be called "JSON".<sup>[[1]](http://www.w3.org/TR/html-json-forms/)</sup>

falcon/request.py
```diff
@@ -1096,6 +1112,20 @@ def _parse_form_urlencoded(self):
 
             self._params.update(extra_params)
 
+    def _parse_application_json(self):
+        raw_body = self.stream.read()
+
+        body = None
+        try:
+            body = json.loads(raw_body.decode('utf-8'))
```
philiptzou: Instead of hard-coding as "utf-8", shouldn't this follows [HTTP charset parameter](http://www.w3.org/International/O-HTTP-charset)?

falcon/request.py
```diff
@@ -524,6 +534,12 @@ def cookies(self):
         return self._cookies.copy()
 
     @property
+    def json(self):
```
philiptzou: Consider not treat `.json` as property but as method.

The `python-requests` did the same thing before but [they changed it to method ultimately](https://github.com/kennethreitz/requests/commit/1451ba0c6d395c41f86da35036fa361c3a41bc90). I quote the reason here:

> A very specific one. It wasn't good design in the first place. Attributes shouldn't through errors on decoding JSON, but not throwing an error would be misleading and might make a user think there was no JSON. As a method, it is perfectly legal for it through an exception. And that's how it should be.

It's reasonable to me. And now we can avoid to repeat the same mistake they did before.

falcon/request.py
```diff
@@ -1096,6 +1112,20 @@ def _parse_form_urlencoded(self):
 
             self._params.update(extra_params)
 
+    def _parse_application_json(self):
+        raw_body = self.stream.read()
+
+        body = None
+        try:
+            body = json.loads(raw_body.decode('utf-8'))
+        except UnicodeDecodeError:
```
swistakm: I don't think that JSON decoding errors should be ignored like that. Note also that `'null'` string is a valid JSON. So there is no way you can distinguish deserialisation error from valid `'null'` JSON value.

falcon/request.py
```diff
@@ -1096,6 +1112,20 @@ def _parse_form_urlencoded(self):
...
+            self.log_error('Non-UTF8 characters found in the request body')
+        except ValueError as e:
+            msg = 'Could not parse the body as Json: {0}. Ignoring.'.format(e)
+            self.log_error(msg)
+
+        self._json = body
```
swistakm: This method should not set attribute but only return value and I think that we should avoid such side effects if there is no clue about that in method name or at least docstring (missing btw).

Where to store value it is a responsibility of the method caller caller.

falcon/request.py
swistakm: Also this is so trivial that there is no point in splitting implementation into two methods.

swistakm: Also when converting to method as @philiptzou suggested then there is no need for storing value in additional attribute for later reference (responsibility of a caller) and thus additional slot.

swistakm:
Despite I have added some comments to this PR I think this is generally not a good design decision. Here are the reasons:
* it favours JSON. What if someone needs to handle various content types like yaml, csv, xml, etc.? New method for every new supported content-type? 
* is misleading: `json` does not return a JSON but a `dict`. A **content** `dict`. Content that here was represented by a JSON string.

I think that when implementing API a resource class should not be interested in a content-type that was used to for content serialisation but only with a deserialised content (most likely a `dict`) - something that abstracts representation no matter what content-type was used as long as it is supported. This is what issue #145 is about and I think that work should be focused on deciding what do do with #145 instead of adding such sugar.

Also there are serious implementation flaws (opinionated encoding, exception suppressing, pointless property). I am afraid that if they will be solved then this will be no more than two-three lines of code (even counting function signature) with bunch of tests that do not solve any problem and even do not reduce amount of code that falcon user needs to write.

kgriffs:
In light of the discussion here and on #145, closing this PR since the rough consensus is to use a different strategy.

Issue #564, opened 2015-07-13
URI template conflict when separating parameters by a '.' (dot)


admiralobvious:
I am trying to do something like this:

```
class File:
    def on_get(self, req, resp, file_id):
        print(file_id)

class FileDetails:
    def on_get(self, req, resp, file_id, ext):
        print(file_id, ext)

api = falcon.API()
api.add_route('/files/{file_id}', File())
api.add_route('/files/{file_id}.{ext}', FileDetails())
```

but I am getting a: `ValueError: The URI template for this route conflicts with another route's template`.

I didn't see a test for this in the tests folder so I am guessing this is not supported? Is that really the case or is it simply an oversight?

kgriffs:
@admiralobvious It isn't supported yet, but I don't see any reason why we couldn't extend the default router to handle this.

maxking:
I am trying to attempt this isssue. I am wondering how could this be done, as my simple fix(mentioned below) breaks some tests.

```python
        path = re.split(r'[\/.]', uri_template.strip('/'))
```

So, I went through the codebase and I am of the opinion that, URI of the type `/home/file.jpg' can match either `/home/{file}' or '/home/{file}.(extension}' templates. Any pointers about how should I proceed?

philiptzou:
@admiralobvious #647 has been merged. Please check if this fixed your problem.

admiralobvious:
@philiptzou it did indeed. Cheers!

PR #647, opened 2015-11-03
fix(URI template): allow complex vars coexist with simple var

philiptzou:
Allow situation like this:

```python
api.add_route('/files/{file_id}', resource_1)
api.add_route('/files/{file_id}.{ext}', resource_2)
```

Which ValueError was raised due to only one var node was allowed whether
other nodes are complex or not. This fix changes the behaviour so now it
only applys to simple vars.

See #564

### DIFF
--- a/falcon/routing/compiled.py
+++ b/falcon/routing/compiled.py
@@ -112,16 +112,20 @@ class CompiledRouter(object):
         level_indent = indent
         found_simple = False
 
-        # NOTE(kgriffs): Sort static nodes before var nodes so that
-        # none of them get masked. False sorts before True.
-        nodes = sorted(nodes, key=lambda node: node.is_var)
+        # NOTE(kgriffs & philiptzou): Sort nodes in this sequence:
+        # static nodes(0), complex var nodes(1) and simple var nodes(2).
+        # so that none of them get masked.
+        nodes = sorted(
+            nodes, key=lambda node: node.is_var + (node.is_var and
+                                                   not node.is_complex))
 
         for node in nodes:
             if node.is_var:
                 if node.is_complex:
-                    # NOTE(richardolsson): Complex nodes are nodes which contain
-                    # anything more than a single literal or variable, and they
-                    # need to be checked using a pre-compiled regular expression.
+                    # NOTE(richardolsson): Complex nodes are nodes which
+                    # contain anything more than a single literal or variable,
+                    # and they need to be checked using a pre-compiled regular
+                    # expression.
                     expression_idx = len(self._expressions)
                     self._expressions.append(node.var_regex)
 
@@ -143,7 +147,8 @@ class CompiledRouter(object):
                     #   /foo/{id}/bar
                     #   /foo/{name}/bar
                     #
-                    assert len([node for node in nodes if node.is_var]) == 1
+                    assert len([_node for _node in nodes
+                                if _node.is_var and not _node.is_complex]) == 1
                     found_simple = True
 
             else:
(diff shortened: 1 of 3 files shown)


philiptzou:
So `{file_id}` is a simple var, but `{file_id}.{ext}` is a complex var (which is converted to compiled regexp).

yohanboniface:
LGTM :)

yohanboniface:
Humm, actually, two remarks:

- with this patch, we can register `/compare/{file_id}/` then `/compare/{file_id}.{ext}/`, but not the other way around; given that nodes are then reordered, I'm not sure the order we register the routes should have importance, but I may be wrong

- we can now register `/compare/{file_id}.{ext}/` then `/compare/{file_name}.{ext}/`which should conflict in reality IMHO

yohanboniface:
> we can now register /compare/{file_id}.{ext}/ then /compare/{file_name}.{ext}/which should conflict in reality IMHO

This was already the case before sorry. Is that on purpose, @kgriffs?

yohanboniface:
OK, complex means also `{usr0}:{branch0}...{usr1}:{branch1}` vs `{usr0}:{branch0}`, so it seems we don't want to enter in trying to detect complex nodes conflicts (and thus order they are registered is important).

philiptzou:
@yohanboniface Good point on the complex nodes conflicts! 👍 I think I have a simple way to detect such conflicts before we feed them to the compiled router.

falcon/routing/compiled.py
```diff
@@ -112,16 +112,20 @@ def line(text, indent_offset=0):
         level_indent = indent
         found_simple = False
 
-        # NOTE(kgriffs): Sort static nodes before var nodes so that
-        # none of them get masked. False sorts before True.
-        nodes = sorted(nodes, key=lambda node: node.is_var)
+        # NOTE(kgriffs & philiptzou): Sort nodes in this sequence:
+        # static nodes(0), complex var nodes(1) and simple var nodes(2).
+        # so that none of them get masked.
```
philiptzou: @yohanboniface The registering order is not important since we'll sort the nodes inside our code. That's here.

philiptzou:
Btw, I think we can simply detect conflicts between two complex nodes is like this:

```python
import re

def is_conflict(complex_nodes):
    pseudo_segments = [re.sub(r'({[-_a-zA-Z0-9]+})', '{v}', n.raw_segment) for n in complex_nodes]
    return len(set(pseudo_segments)) < len(pseudo_segments)
```

Perhaps another pull-request, @kgriffs?

falcon/routing/compiled.py
yohanboniface: humm, but then this should work equally, right?
```python
In [1]: from falcon.routing import DefaultRouter; router = DefaultRouter()

In [2]: router.add_route('/files/{file_id}', {}, 'xxxx2')

In [3]: router.add_route('/files/{file_id}.{ext}', {}, 'xxxx2')

```
```python
In [1]: from falcon.routing import DefaultRouter; router = DefaultRouter()

In [2]: router.add_route('/files/{file_id}.{ext}', {}, 'xxxx2')

In [3]: router.add_route('/files/{file_id}', {}, 'xxxx2')
ValueError: The URI template for this route conflicts with another route's template.

```

philiptzou:
@yohanboniface Good catch on the collision problem. See if the new revision I just submitted fix the issue you encountered.

yohanboniface:
@philiptzou excellent! Your last commit fixes both collision issue and order issue :)

@kgriffs LGTM!

yohanboniface:
(@philiptzou your should rebase the branch on master, github says there are conflicts ;) )

philiptzou:
@yohanboniface Well, I'll rebase after @kgriffs reviewed this. I have a feeling that he will find something too.

kgriffs:
@philiptzou Nice work, LGTM! Could you please rebase once more and then I think this is ready to merge.

philiptzou:
@kgriffs: done :)

kgriffs:
@philiptzou Thanks! And thanks @yohanboniface for reviewing!

PR #649, opened 2015-11-09
Reset stream after reading it while parsing urlencoded body

yohanboniface:
Other consumers may need it.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1075,6 +1075,7 @@ class Request(object):
         # content length will only ever be read once per
         # request in most cases.
         body = self.stream.read()
+        self.stream.seek(0)
 
         # NOTE(kgriffs): According to http://goo.gl/6rlcux the
         # body should be US-ASCII. Enforcing this also helps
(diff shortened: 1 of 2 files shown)


kgriffs:
Hi @yohanboniface, thanks for working on this. I'd love to get your thoughts on #493.

falcon/request.py
```diff
@@ -1075,6 +1075,7 @@ def _parse_form_urlencoded(self):
         # content length will only ever be read once per
         # request in most cases.
         body = self.stream.read()
+        self.stream.seek(0)
```
kgriffs: Unfortunately, this won't always work; the WSGI server may pass a non-seekable file-like object to Falcon. One option would be to replace the WSGI object with our own that buffers the request body (similar to [my suggestion here][1]). This could be used as a stepping stone to the broader redesign around form param handling that is addressed by #493.

[1]: https://github.com/falconry/falcon/pull/493#discussion_r40822962

yohanboniface:
Closing here as it should be addressed beforehand as suggested.

falcon/request.py
yohanboniface: OK, I see that `seek` is not part of the [spec](https://www.python.org/dev/peps/pep-3333/#input-and-error-streams). I think the BytesIO suggested in #493 is the best compromise to move forward at this point.

PR #651, opened 2015-11-13
feat(middleware) Pass params to process_resource

### LINKED ISSUE #612: Pass params to process_resource
This is needed for feature parity with global *before* hooks. We can use a shim to avoid breaking existing middleware.

yohanboniface:
This allows middlewares to interact with route parameters.

BREAKING CHANGE: `process_resource` must now accept `params`
argument. Fix #612

I've arbitrated in favour of an API breaking change, because I think it's better to break API now (before 1.0) than to support useless legacy tricks, but of course this is to be discussed :)

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -178,7 +178,8 @@ class API(object):
                 # e.g. a 404.
                 responder, params, resource = self._get_responder(req)
 
-                self._call_rsrc_mw(middleware_stack, req, resp, resource)
+                self._call_rsrc_mw(middleware_stack, req, resp, resource,
+                                   params)
 
                 responder(req, resp, **params)
                 self._call_resp_mw(middleware_stack, req, resp, resource)
@@ -537,13 +538,13 @@ class API(object):
             # Put executed component on the stack
             stack.append(component)  # keep track from outside
 
-    def _call_rsrc_mw(self, stack, req, resp, resource):
+    def _call_rsrc_mw(self, stack, req, resp, resource, params):
         """Run process_resource middleware methods."""
 
         for component in self._middleware:
             _, process_resource, _ = component
             if process_resource is not None:
-                process_resource(req, resp, resource)
+                process_resource(req, resp, resource, params)
 
     def _call_resp_mw(self, stack, req, resp, resource):
         """Run process_response middleware."""
--- a/tests/test_middlewares.py
+++ b/tests/test_middlewares.py
@@ -1,3 +1,5 @@
+import json
+
 import falcon
 import falcon.testing as testing
 from datetime import datetime
@@ -11,7 +13,7 @@ class RequestTimeMiddleware(object):
         global context
         context['start_time'] = datetime.utcnow()
 
-    def process_resource(self, req, resp, resource):
+    def process_resource(self, req, resp, resource, params):
         global context
         context['mid_time'] = datetime.utcnow()
 
@@ -34,7 +36,7 @@ class ExecutedFirstMiddleware(object):
         context['executed_methods'].append(
             '{0}.{1}'.format(self.__class__.__name__, 'process_request'))
 
-    def process_resource(self, req, resp, resource):
+    def process_resource(self, req, resp, resource, params):
         global context
         context['executed_methods'].append(
             '{0}.{1}'.format(self.__class__.__name__, 'process_resource'))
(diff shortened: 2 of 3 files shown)


tests/test_middlewares.py
```diff
@@ -55,9 +55,17 @@ def process_request(self, req, resp):
         req.path = req.path.replace('/base_path', '', 1)
 
 
+class AccessParamsMiddleware(object):
+
+    def process_resource(self, req, resp, resource, params):
```
kgriffs: I think we should also test to ensure that a parameter that is part of the path shows up here in params.

tests/test_middlewares.py
```diff
@@ -55,9 +55,17 @@ def process_request(self, req, resp):
...
+        global context
+        params['added'] = True
```
kgriffs: Would you mind adding a test to check that the modified parameter is getting propagated correctly to the responder method on the resource?

kgriffs:
I think this makes sense. It is a breaking change, but one that should be easy for developers to accommodate. A couple testing suggestions inline.

tests/test_middlewares.py
yohanboniface: Done :)

tests/test_middlewares.py
yohanboniface: Done :)

falcon/api.py
```diff
@@ -544,13 +545,13 @@ def _call_req_mw(self, stack, req, resp):
...
 
-    def _call_rsrc_mw(self, stack, req, resp, resource):
+    def _call_rsrc_mw(self, stack, req, resp, resource, params):
         """Run process_resource middleware methods."""
 
         for component in self._middleware:
             _, process_resource, _ = component
             if process_resource is not None:
-                process_resource(req, resp, resource)
+                process_resource(req, resp, resource, params)
```
philiptzou: I am not quite sure if detecting `process_resource` with `inspect.getargspec` is a good idea (a bit downside about the performance, huh?). But with that technique we can easily prevent the breaking change.

kgriffs: One way to mitigate the performance issue is to do the work up front in `prepare_middleware` to detect the number of args, and put a shim in place as needed for older middleware components. E.g. (pseudocode):

```
if is old-style method:
    @functools.wraps(orig_process_resource)
    def func(req, resp, resource, params)):
         orig_process_resource(req, resp, resource)

    process_resource = func
else:
    process_resource = orig_process_resource
```

Then the overhead is reduced to an extra function call. Eventually we can remove the shim as a breaking change once everyone has had a chance to migrate (e.g., 2.0).

What do you think?

tests/test_middlewares.py
```diff
@@ -368,3 +376,23 @@ def test_base_path_is_removed_before_routing(self):
...
+
+class TestResourceMiddleware(TestMiddleware):
+
+    def test_can_access_resource_params(self):
+        """Test that params can be accessed from within process_resource"""
+        global context
+
+        class Resource:
+            def on_get(self, req, resp, **params):
+                resp.body = params
```
kgriffs: Nitpick: The fact that this works without serializing to a string is an implementation detail that may or may not hold up in the future. Instead, you may want to pass `params` through `json.dumps()`, assigning the result to `resp.data`.

falcon/api.py
yohanboniface: Are we sure we want to take this path?
In my opinion, 0.x versions are made to keep maintainers out from 100% code backward compatibility. And specifically is this case we are only talking about adding a parameter to "process_resource" middlewares methods. It does not affect any logic or API behaviour. So instead, I would just add this breaking change in the release note, and let falcon users *just* go through every middleware and add an argument.
In other words, I think the balance here, in this specific case, is in favour of adding a bit of charge to falcon users instead of adding more code complexity and legacy on the shoulders of the maintainer.
Thoughts?

tests/test_middlewares.py
yohanboniface: Done :)

falcon/api.py
kgriffs: All things considered, I think it's OK to make the breaking change. There will be very few breaking changes in the next release as it is, and this one is easy enough to work around. Let's go ahead and land this for 0.4. That will be the final release before 1.0.

kgriffs:
Looks great, thanks!

PR #652, opened 2015-11-13
feat(routing) Pass by kwargs from api.add_route to router.add_route

yohanboniface:
This allows for more routing customisation. For example, to build
an indexed router so we can reverse build URLs (good for DRY).

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -247,7 +247,7 @@ class API(object):
         start_response(resp.status, headers)
         return body
 
-    def add_route(self, uri_template, resource):
+    def add_route(self, uri_template, resource, *args, **kwargs):
         """Associates a templatized URI path with a resource.
 
         A resource is an instance of a class that defines various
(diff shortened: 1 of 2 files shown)


falcon/api.py
```diff
@@ -247,7 +247,7 @@ def __call__(self, env, start_response):
         start_response(resp.status, headers)
         return body
 
-    def add_route(self, uri_template, resource):
+    def add_route(self, uri_template, resource, **kwargs):
```
philiptzou: how about add `*args` also?

```python
def add_route(self, uri_template, resource, *args, **kwargs):
```

Don't see any harm.

yohanboniface: Yeah, good point. I've thought about this, and arbitrated without `args` because I thought it was better separation between internals and custom params to keep named params only for custom uses. But honestly that's not a strong argument, just the way I've made the choice. So I'm OK to add `args` if that's the preference. :)

philiptzou: ok, let's ask @kgriffs :P

kgriffs: Adding `*args` should be OK if we think it will make custom routers easier to use. To be honest, I can't imagine a custom router wanting to add more than one or two additional positional args. Anyway, the mechanics will be invisible to most people, since the docs won’t mention them (and presumably docs for custom routers would only mention the additional params they are using).

yohanboniface: Added `*args` :)

philiptzou:
LGTM :+1:

kgriffs:
LGTM, thanks!

PR #655, opened 2015-11-17
feat: change API._get_responder() 404 behavior to raise falcon.HTTPNotFound()

michaelgugino:
feat: change API._get_responder() 404 behavior to raise falcon.HTTPNotFound()

Class API() currenlty accepts optional method add_error_handler() to override raised exceptions.
By default, API._get_responder() method returns a tuple including
responder = falcon.responders.path_not_found. Change made to API._get_responder() to raise
falcon.HTTPNotFound() in place of returning falcon.responders.path_not_found
This change allows easy override of 404 behavior in the application.

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -494,8 +494,7 @@ class API(object):
 
                     break
             else:
-                responder = falcon.responders.path_not_found
-
+                raise falcon.HTTPNotFound()
         return (responder, params, resource)
 
     def _compose_status_response(self, req, resp, http_status):
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -466,8 +466,7 @@ class API(object):
         Note:
             If a responder was matched to the given URI, but the HTTP
             method was not found in the method_map for the responder,
-            the responder callable element of the returned tuple will be
-            `falcon.responder.bad_request`.
+            falcon.HTTPBadRequest() error will be raised.
         """
 
         path = req.path
@@ -490,7 +489,8 @@ class API(object):
 
                     break
             else:
-                raise falcon.HTTPNotFound()
+                responder = falcon.responders.path_not_found
+
         return (responder, params, resource)
 
     def _compose_status_response(self, req, resp, http_status):
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -453,20 +453,21 @@ class API(object):
 
     def _get_responder(self, req):
         """Searches routes for a matching responder.
-
         Args:
             req: The request object.
-
         Returns:
             A 3-member tuple consisting of a responder callable,
             a ``dict`` containing parsed path fields (if any were specified in
             the matching route's URI template), and a reference to the
             responder's resource instance.
-
         Note:
             If a responder was matched to the given URI, but the HTTP
             method was not found in the method_map for the responder,
-            falcon.HTTPBadRequest() error will be raised.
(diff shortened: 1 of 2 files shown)


philiptzou:
Will raising an exception instead of returning a response object in `_get_responder` affects some middlewares?

michaelgugino:
I'm not sure if middleware will be affected or not.  I'm not an expert in this code base, this pull is more of a suggestion/request for new behavior.  This pull should only affect default responder behavior, as best I can tell.

kgriffs:
@philiptzou This patch will skip execution of `process_resource()` middleware methods, although the other middleware methods will still run. This would be a breaking change (currently, `process_resource()` receives `None` for its `resource` arg when no route was found.

falcon/api.py
```diff
@@ -469,10 +469,6 @@ def _get_responder(self, req):
             method was not found in the method_map for the responder,
             the responder callable element of the returned tuple will be
             `falcon.responder.bad_request`.
```
kgriffs: Seems like if we are going to do this for 404, we should also do it for 400. What do you think?

falcon/responders.py
```diff
@@ -18,11 +18,6 @@
 from falcon.status_codes import HTTP_405
 
 
-def path_not_found(req, resp, **kwargs):
-    """Simply sets responseto "404 Not Found", no body."""
-    resp.status = HTTP_404
```
kgriffs: I would prefer keeping this function and modifying it to raise falcon.HTTPNotFound(). That way, we can avoid any unintended side-effects; `API.__call__()` is currently written in way that assumes `API._get_responder()` does not raise an error.

michaelgugino:
@kgriffs I implemented your suggested changes, including similar handling of 400, and as a plus the build is now passing.

yohanboniface:
@michaelgugino I think you should rebase :)

falcon/api.py
```diff
@@ -465,12 +465,7 @@ def _get_responder(self, req):
         Note:
             If a responder was matched to the given URI, but the HTTP
             method was not found in the method_map for the responder,
-            the responder callable element of the returned tuple will be
```
kgriffs: Would you please revert this change to the docstring? `_get_responder` does not call the responder itself, but just returns it.

falcon/responders.py
```diff
@@ -13,19 +13,19 @@
 # limitations under the License.
 
 from falcon.status_codes import HTTP_204
-from falcon.status_codes import HTTP_400
-from falcon.status_codes import HTTP_404
 from falcon.status_codes import HTTP_405
+from falcon.errors import HTTPNotFound
```
kgriffs: Nitpick: Not a big deal, but it can be nice to order imports alphabetically.

yohanboniface: https://github.com/timothycrosley/isort is a nice tool to handle this :)

kgriffs:
@michaelgugino I'd like to get this finished up and merged soon. Do you have some time to take a look over the next couple of days?

michaelgugino:
Sure, I can make these changes by the end of the week.  I'm a little busy today and tomorrow, possibly Thursday or Friday I'll send up the changes rebased.

kgriffs:
Awesome, thanks!

kgriffs:
Rebased and merged as #696

PR #656, opened 2015-11-19
Remove mention of the Talons project from the tutorial, update FAQ

kgriffs:
The Talons project has not become a general repository for
community add-ons, as originally envisioned. Remove from the
the tutorial to avoid confusion (based on community feedback.)

Also update the FAQ as follows:

    * Remove mention of talons to allow for diversity in
      the Falcon ecosystem
    * Remove mention of specific "choices" that Falcon lets
      you make regarding complementary libraries, since in
      each case there are actually many more choices than
      given, and the list only seems to belabor the point.

### DIFF
--- a/doc/community/faq.rst
+++ b/doc/community/faq.rst
@@ -18,39 +18,21 @@ simply wrap your api instance with a middleware app. For example:
 
 See also the `WSGI middleware example <http://legacy.python.org/dev/peps/pep-3333/#middleware-components-that-play-both-sides>`_ given in PEP-3333.
 
-
-Why doesn't Falcon include X?
------------------------------
-The Python ecosytem offers a bunch of great libraries that you
-are welcome to use from within your responder, hooks, and middleware.
-Falcon doesn't try to dictate what you should use, since that would take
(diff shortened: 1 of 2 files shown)


doc/community/faq.rst
```diff
@@ -18,39 +18,21 @@ simply wrap your api instance with a middleware app. For example:
 
 See also the `WSGI middleware example <http://legacy.python.org/dev/peps/pep-3333/#middleware-components-that-play-both-sides>`_ given in PEP-3333.
 
-
 Why doesn't Falcon include X?
 -----------------------------
-The Python ecosytem offers a bunch of great libraries that you
```
fxfitz: Nit: Maybe rephrase to 'Why doesn't Falcon include <X>?'

fxfitz: Oh, well, looks like markdown took out my "< X >"

kgriffs: I rewrote this. Let me know what you think.

fxfitz:
LGTM

PR #665, opened 2015-12-01
chore: Optimize Travis test order

kgriffs:
Start the slow Jython test first since it takes the longest to
complete; other tests can run in parallel. Also, run the relatively
quick pep8 and pylint jobs in parallel with the slower Jython and
py26 jobs.


painterjd:
:+1:

PR #679, opened 2016-01-01
fix(Request): Improve access_route perf and error handling

kgriffs:
Tune performance of the Request.access_route property, while also
making error handling more consistent, in the case that a header value
is malformed. Also, add a test to ensure that wsgiref passes
REMOTE_ADDR in the environment dict, as has been observed in production
WSGI servers. This is necessary to ensure that Request.access_route's
fallback to REMOTE_ADDR makes sense.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -580,17 +580,28 @@ class Request(object):
     @property
     def access_route(self):
         if self._cached_access_route is None:
-            access_route = []
+            # NOTE(kgriffs): Try different headers in order of
+            # preference; if none are found, fall back to REMOTE_ADDR.
+            #
+            # If one of these headers is present, but its value is
+            # malformed such that we end up with an empty list, or
+            # a non-empty list containing malformed values, go ahead
+            # and return the results as-is. The alternative would be
+            # to fall back to another header or to REMOTE_ADDR, but
+            # that only masks the problem; the operator needs to be
+            # aware that an upstream proxy is malfunctioning.
+
             if 'HTTP_FORWARDED' in self.env:
-                access_route = self._parse_rfc_forwarded()
-            if not access_route and 'HTTP_X_FORWARDED_FOR' in self.env:
-                access_route = [ip.strip() for ip in
-                                self.env['HTTP_X_FORWARDED_FOR'].split(',')]
-            if not access_route and 'HTTP_X_REAL_IP' in self.env:
-                access_route = [self.env['HTTP_X_REAL_IP']]
-            if not access_route and 'REMOTE_ADDR' in self.env:
-                access_route = [self.env['REMOTE_ADDR']]
-            self._cached_access_route = access_route
+                self._cached_access_route = self._parse_rfc_forwarded()
+            elif 'HTTP_X_FORWARDED_FOR' in self.env:
+                addresses = self.env['HTTP_X_FORWARDED_FOR'].split(',')
+                self._cached_access_route = [ip.strip() for ip in addresses]
+            elif 'HTTP_X_REAL_IP' in self.env:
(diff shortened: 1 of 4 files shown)


falcon/request.py
```diff
@@ -580,17 +580,26 @@ def cookies(self):
     @property
     def access_route(self):
         if self._cached_access_route is None:
-            access_route = []
```
kgriffs: No need to use CPU cycles to initialize a new variable.

falcon/request.py
```diff
@@ -580,17 +580,26 @@ def cookies(self):
...
-            if not access_route and 'REMOTE_ADDR' in self.env:
-                access_route = [self.env['REMOTE_ADDR']]
-            self._cached_access_route = access_route
+                self._cached_access_route = self._parse_rfc_forwarded()
+            elif 'HTTP_X_FORWARDED_FOR' in self.env:
+                addresses = self.env['HTTP_X_FORWARDED_FOR'].split(',')
+                self._cached_access_route = [ip.strip() for ip in addresses]
+            elif 'HTTP_X_REAL_IP' in self.env:
+                self._cached_access_route = [self.env['HTTP_X_REAL_IP']]
+            else:
```
kgriffs: This should be safe to do; wsgiref and production servers seem to always include REMOTE_ADDR in the environ.

falcon/request.py
```diff
@@ -1119,18 +1128,25 @@ def _parse_rfc_forwarded(self):
...
+
         for forwarded in self.env['HTTP_FORWARDED'].split(','):
             for param in forwarded.split(';'):
-                param = param.strip().split('=', 1)
-                if len(param) == 1:
+                # PERF(kgriffs): Partition() is faster than split().
+                key, _, val = param.strip().partition('=')
+                if not val:
+                    # NOTE(kgriffs): The '=' separator was not found or
+                    # it was, but the value was missing.
```
kgriffs: This is slightly different from previous behavior, which would return an empty string in the case of an empty value. In this case, the field is malformed and behavior is ostensibly undefined, so let's just be consistent and always skip it when there is a problem.

falcon/util/uri.py
```diff
@@ -390,14 +390,15 @@ def unquote_string(quoted):
     Raises:
         TypeError: `quoted` was not a ``str``.
     """
-    tmp_quoted = quoted.strip()
```
kgriffs: This was inconsistent with how other helper functions work, which is to require the caller to provide a pre-stripped string, therefore allowing the caller to skip this call (and creating of an additional string instance) when it is unnecessary.

tests/test_access_route.py
```diff
@@ -23,12 +23,13 @@ def test_rfc_forwarded(self):
             headers={
                 'Forwarded': ('for=192.0.2.43,for=,'
                               'for="[2001:db8:cafe::17]:555",'
+                              'for=x,'
```
kgriffs: Test coverage for the length check in `uri.unquote_string()`

tests/test_access_route.py
```diff
@@ -42,9 +43,9 @@ def test_malformed_rfc_forwarded(self):
             headers={
                 'Forwarded': 'for'
             }))
-        self.assertEqual(req.access_route, ['127.0.0.1'])
+        self.assertEqual(req.access_route, [])
```
kgriffs: Don't fall back to REMOTE_ADDR when the field is malformed; expose the problem with an upstream proxy to the app (and the operator of said app).

falcon/request.py
jmvrbanac: Good plan!

tests/test_access_route.py
philiptzou: Since the header can be malformed from client, we'd better add a document note for this.

PR #686, opened 2016-01-15
perf(Response): Optimize response body processing

kgriffs:
**Please review carefully** before merging. This is a significant change to the way the framework processes the response body.

---

Remove overhead from response body processing in API and Response.

BREAKING CHANGE: Response.body_encoded has been removed. The property was only intended to be used by the framework itself, so there should not be many apps using it.

To migrate the code, follow the example below:

Before:

```py
    body = resp.body_encoded
```

After:

```py
    if resp.body:
        body = resp.body.encode('utf-8')
    else:
        body = b''
```

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -226,8 +226,9 @@ class API(object):
         if req.method == 'HEAD' or resp.status in self._BODILESS_STATUS_CODES:
             body = []
         else:
-            self._set_content_length(resp)
-            body = self._get_body(resp, env.get('wsgi.file_wrapper'))
+            body, length = self._get_body(resp, env.get('wsgi.file_wrapper'))
+            if length is not None:
+                resp._headers['content-length'] = str(length)
 
         # Set content type if needed
         use_content_type = (body or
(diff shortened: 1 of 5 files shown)


fxfitz:
LGTM

falcon/api.py
```diff
@@ -615,26 +578,37 @@ def _get_body(self, resp, wsgi_file_wrapper=None):
...
+                * If resp.data is not ``None``, returns [resp.data]
+                * If resp.stream is not ``None``, returns resp.stream
+                  iterable using wsgi.file_wrapper, if possible.
+                * Otherwise, returns []
 
         """
 
-        body = resp.body_encoded
-
+        body = resp._get_encoded_body()
```
kgriffs: Perhaps I should just get rid of `_get_encoded_boy()` altogether and put the logic here instead?

kgriffs:
@painterjd Nice catch! In fact, I can just return the value of `resp.stream_len` directly.

kgriffs:
I amended the commit, let me know what you think.

painterjd:
:+1:

PR #696, opened 2016-01-29
feat(API): Change error responders to raise instances of HTTPError

kgriffs:
When an invalid HTTP method is requested, or a path is requested for which there is no route configured, raise instances of HTTPError instead of directly manipulating the Response object. This makes it easier for applications to customize behavior in these cases.


kgriffs:
This is a rebased version of #655

PR #697, opened 2016-01-29
test(tox): Add debug envs and organize into sections

kgriffs:
Add debug environments that accept pos args for specifying individual tests to run, and for passing options such as -s for use with pdb, ipdb, etc.

Also organize envs into sections with headers, and clean up the default envlist.


jmvrbanac:
:+1:

painterjd:
:+1:

PR #698, opened 2016-01-29
feat(testing): New test case base, result class, and test resource

### LINKED ISSUE #689: Missing Content-Type on 404 Response
Greetings,

I've started [testing my webapps with wsgiref.validate](http://blog.dscpl.com.au/2015/05/returning-string-as-iterable-from-wsgi.html), and discovered that Falcon's 404 responses (and possible others) don't conform to the WSGI spec.  The `Content-Type` header is required but not present.

This script demonstrates:

    import falcon
    from webtest import TestApp
    import wsgiref.validate

    # the app -- note that it has no routes so all GETs will return 404
    app = falcon.API()

    # wrap in validator middleware
    app = wsgiref.validate.validator(app)

    # wrap in test middleware
    app = TestApp(app)

    # provoke a 404
    resp = app.get('/doesntexist', status=[404])

Output is:

        <...snip...>
      File "/usr/local/lib/python2.7/dist-packages/falcon/api.py", line 247, in __call__
      File "/usr/lib/python2.7/wsgiref/validate.py", line 167, in start_response_wrapper
        check_content_type(status, headers)
      File "/usr/lib/python2.7/wsgiref/validate.py", line 419, in check_content_type
        assert_(0, "No Content-Type header found in headers (%s)" % headers)
      File "/usr/lib/python2.7/wsgiref/validate.py", line 129, in assert_
        raise AssertionError(*args)
    AssertionError: No Content-Type header found in headers ([('content-length', '0')])

FYI, as a sanity check, I ran the same test using Bottle in place of Falcon, and it conforms (sends Content-type with the 404 response.)

I'm assuming that we want to be compliant with the WSGI spec, in which case this is a bug. Happy to submit a PR if you point me in a recommended direction to patch.  (Set Content-type in `falcon/responders.py:path_not_found` ?)

kgriffs:
Add new testing classes for unittest-style test cases, based on lessons-learned using the original base class. Port several Falcon tests to use the new classes (the remainder to be ported in a subsequent release.) Include wsgiref valdiation by default.

DEPRECATION: The old base class and test resource have been deprecated and will be removed in a future release.

Fixes #689

### DIFF
--- a/falcon/testing/base.py
+++ b/falcon/testing/base.py
@@ -26,14 +26,22 @@ from falcon.testing.helpers import create_environ
 
 
 class TestBase(unittest.TestCase):
-    """Extends ``testtools.TestCase`` to support WSGI integration testing.
+    """Extends :py:mod:`unittest` to support WSGI functional testing.
 
-    ``TestBase`` provides a base class that provides some extra plumbing to
-    help simulate WSGI calls without having to actually host your API
-    in a server.
+    Warning:
+        This class has been deprecated and will be removed in a future
+        release. Please use :py:class:`~.TestCase`
+        instead.
 
     Note:
-        If ``testtools`` is not available, ``unittest`` is used instead.
+        If available, uses :py:mod:`testtools` in lieu of
+        :py:mod:`unittest`.
+
+    This base class provides some extra plumbing for unittest-style
+    test cases, to help simulate WSGI calls without having to spin up
+    an actual web server. Simply inherit from this class in your test
+    case classes instead of :py:class:`unittest.TestCase` or
+    :py:class:`testtools.TestCase`.
 
     Attributes:
         api (falcon.API): An API instance to target when simulating
@@ -46,6 +54,7 @@ class TestBase(unittest.TestCase):
         test_route (str): A simple, generated path that a test
             can use to add a route to the API.
     """
+
     api_class = falcon.API
     srmock_class = StartResponseMock
 
--- a/falcon/testing/helpers.py
+++ b/falcon/testing/helpers.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import cgi
 import random
 import io
 import sys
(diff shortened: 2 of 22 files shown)


falcon/testing/helpers.py
```diff
@@ -40,7 +68,7 @@ def rand_string(min, max):
     int_gen = random.randint
     string_length = int_gen(min, max)
     return ''.join([chr(int_gen(ord(' '), ord('~')))
-                    for i in range(string_length)])
+                    for __ in range(string_length)])
```
jmvrbanac: was the double underscore intended here?

falcon/testing/base.py
```diff
@@ -26,14 +26,16 @@
 
 
 class TestBase(unittest.TestCase):
-    """Extends ``testtools.TestCase`` to support WSGI integration testing.
+    """Extends ``testtools.TestCase`` to support WSGI functional testing.
```
jmvrbanac: I know this is a little out of scope of discussion, but isn't this extends testtools.TestCase or unittest.TestCase? As a side note, I do appreciate us not forcing people to use testtools. I believe if we're staying with a unittest compatible approach, being neutral and allowing the implementer to choose is important.

tests/test_after_hooks.py
```diff
@@ -2,7 +2,12 @@
 import json
 
 import falcon
-import falcon.testing as testing
+from falcon import testing
+
+
+# --------------------------------------------------------------------
+# Hookos
```
jmvrbanac: Hookos? Hooks? ;)

falcon/testing/helpers.py
kgriffs: TBH, I changed this on a whim while reviewing this function. I've started following this recommendation in my work:

http://docs.python-guide.org/en/latest/writing/style/#create-an-ignored-variable

tests/test_after_hooks.py
kgriffs: Fixed

falcon/testing/base.py
kgriffs: I went ahead and reconciled the docstrings between the old and new classes. Let me know what you think!

jmvrbanac:
lgtm :+1:

PR #700, opened 2016-02-02
test(req.client_accepts): Fix invalid assertion based on q=0.0

kgriffs:
A bug was fixed in python-mimeparse that caused an assertion to fail. A quality setting of "0.0" is now (correctly) treated as indicating that the associated media type is "not acceptable" by the client.
Adjust the tests accordingly.


painterjd:
:+1:

PR #705, opened 2016-02-15
Adding status 451 Unavailable For Legal Reasons

davidbgk:
See https://en.wikipedia.org/wiki/HTTP_451 for details.


davidbgk:
Reopened as https://github.com/falconry/falcon/pull/706

PR #708, opened 2016-02-18
doc(contrib-snip): Clarified mailing list subscription information

fxfitz:
This PR simply clarifies what happens when you send your first e-mail to the mailing list.

(oh, and I also snuck a typo fix in there)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -415,13 +415,14 @@ Community
 ---------
 
 The Falcon community maintains a mailing list that you can use to share
-your ideas and ask questions about the framework. We use the
-appropriately minimalistic `Librelist <http://librelist.com/>`__ to host
-the discussions. Subscribing is super easy and doesn't require any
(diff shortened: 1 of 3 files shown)


kgriffs:
Thanks, this looks great! Would you also mind updating the relevant section in the README?

fxfitz:
Will do. 

Is there any reason you wouldn't just want the README to `.. include::` the other contributing restructured text?

kgriffs:
TBH, I don't know how well GH would handle that when rendering the markup. An alternative would be to just link to the relevant page on RTD. I suppose the same could be said for the Contributing section, although that may be better modified in a separate PR, since I think there needs to be some additional work to reconcile CONTRIBUTING.md and the RTD page regarding the same topic.

fxfitz:
Relevant portion of the README has been updated! :-)

kgriffs:
Looks good, thanks!

Issue #704, opened 2016-02-11
Passing max_age as float instead of int to set_cookie(), no cookie will be set


jgehrcke:
When using `response.set_cookie()` and passing the `max_age` parameter as a `float` instead of an `int`, the cookie will not be set in the response (that's Python 3.4 and Falcon 0.3.0).

It's clearly documented in the docstring of the `set_cookie()` method that this parameter is expected to be an `int`, so I spotted the problem rather quickly after re-reading the docs. However, either a raising an exception or implicitly converting to int would be much less painful, IMO. What do you think?

jgehrcke:
I'd also be happy to provide a PR if we decide to change this behavior.

kgriffs:
I don't think it would harm anything to simply convert it to an int. Happy to take a PR for this change.

jgehrcke:
See PR #711. After more investigation, the cookie actually was set in the response, with the floating point value to max-age. This violates RFC 6265, so that the cookie parser I used (the `requests` lib) simply ignored the entire cookie.

PR #711, opened 2016-02-18
feat(set_cookie): convert max-age value to int implicitly

jgehrcke:
This tackles issue #704.

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -154,7 +154,8 @@ class Response(object):
                 default, cookies expire when the user agent exits.
             max_age (int): Defines the lifetime of the cookie in seconds.
                 After the specified number of seconds elapse, the client
-                should discard the cookie.
+                should discard the cookie. Coercion to `int` is attempted
+                if provided with `float` or `str`.
             domain (str): Specifies the domain for which the cookie is valid.
                 An explicitly specified domain must always start with a dot.
                 A value of 0 means the cookie should be discarded immediately.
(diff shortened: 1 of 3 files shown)


falcon/response.py
```diff
@@ -218,7 +218,13 @@ def set_cookie(self, name, value, expires=None, max_age=None,
...
 
         if max_age:
-            self._cookies[name]["max-age"] = max_age
+            # RFC 6265 section 5.2.2 says about the max-age value:
+            #   "If the remainder of attribute-value contains a non-DIGIT
+            #    character, ignore the cookie-av."
+            # That is, RFC-compliant response parsers will ignore the max-age
+            # attribute if the value contains a dot, as in floating point
+            # numbers. Therefore, attempt to convert the value to an integer.
+            self._cookies[name]["max-age"] = int(max_age)
```
kgriffs: Could you also add a note to the `set_cookie()` docstring, mentioning that a float will be coerced to an int? You might also mention that a string value is acceptable (e.g., `'60'` vs. `60`).

tests/test_cookies.py
```diff
@@ -50,6 +50,13 @@ def on_put(self, req, resp):
         resp.unset_cookie("bad")
 
 
+class CookieResourceMaxAgeFloat:
+
+    def on_get(self, req, resp):
```
kgriffs: Could you also add a test for setting max_age to a string value? Since that now works, apps may come to rely on it and it would be helpful to know if the implementation ever breaks that contract (implicit though it is).

kgriffs:
@jgehrcke  Just checking in to see if you saw my earlier comments?

jgehrcke:
Thanks for the reminder. I very much agree with your suggestions. Will take care of both, this week.

jgehrcke:
@kgriffs sorry for the delay, please have another look, thanks!

falcon/response.py
jgehrcke: done

kgriffs:
@jgehrcke This looks good, just needs to be rebased!

jgehrcke:
@kgriffs Thanks for notification. Rebased on latest master, resolved merge conflicts, and made CI happy (made another change to use single quotes)!

jgehrcke:
@kgriffs anything else required?

kgriffs:
Looks great, thanks!

PR #721, opened 2016-03-03
test: Force pep8 to run under py27

kgriffs:
Run pep8 with py27 since some code branches assume the unicode type is defined, and pep8 complains in those cases when running under py3.


jmvrbanac:
:+1:

PR #723, opened 2016-03-03
fix(falcon.routing.DefaultRouter): Do not mask variable path segments

### LINKED ISSUE #702: Path segment in one route's URI template masks the field expression in another route
The following test demonstrates the issue. The assertion fails since the resulting status code is 404, rather than 200. "/v2.0/thing" should route to `/{version}/thing` instead of `/v2.0`.

```py
def test_string_vs_var(self):
    self.api.add_route('/v2.0', self.resource)
    self.simulate_request('/v2.0')

    self.api.add_route('/{version}/thing', testing.TestResource())
    self.simulate_request('/v2.0/thing')
    self.assertEqual(self.srmock.status, falcon.HTTP_200)
```

```
/{version}/foo/bar
/v1.0
```

kgriffs:
fix(falcon.routing.DefaultRouter): Do not mask variable path segments

When a single segment in a requested path can match more than one node at that branch in the routing tree, and the first branch taken happens to be the wrong one (i.e., the subsequent nodes do not match, but they would have under a different branch), do not mask the other branches that could result in a successful resolution of the requested path.

I chose to implement this by not immediately returning None when a branch prediction turns out to be false. Alternatively, I could have introduced functions but that would likely incur more overhead vs. performing a few additional node comparisons on the way back up the routing tree. I did, however, include a simple optimization that allows a "fast return" when only simple nodes are involved in the branching, since in that case there is only one possible route.

More advanced optimizations are possible, and may be attempted in the future. For example:

* Graft the alternate branches directly into each point where the
  routing tree would otherwise 'return None'. This would balloon the
  size of the generated code, but would be akin to inline expansion
  of a routing representing the traversal of each branch where
  branch mispredictions are possible.
* Introspect competing regular expressions to determine whether it
  is actually possible for them to match the same string given in
  the path.
* Set a flag that can be referenced when returning from a branch
  misprediction in order to short-circuit checks. This may or may
  not be faster that some of the checks (i.e., those that check
  path length would not benefit from this flag, while those that
  index into the path list and perform a string comparison may be).

Fixes #702

### DIFF
--- a/falcon/routing/compiled.py
+++ b/falcon/routing/compiled.py
@@ -95,7 +95,7 @@ class CompiledRouter(object):
         else:
             return None, None, None
 
-    def _compile_tree(self, nodes, indent=1, level=0):
+    def _compile_tree(self, nodes, indent=1, level=0, fast_return=True):
         """Generates Python code for a routing tree or subtree."""
 
         def line(text, indent_offset=0):
@@ -118,6 +118,18 @@ class CompiledRouter(object):
         nodes = sorted(
             nodes, key=lambda node: node.is_var + (node.is_var and
                                                    not node.is_complex))
+
+        # NOTE(kgriffs): Down to this branch in the tree, we can do a
+        # fast 'return None'. See if the nodes at this branch are
+        # all still simple, meaning there is only one possible path.
+        if fast_return:
+            if len(nodes) > 1:
+                # NOTE(kgriffs): There's the possibility of more than
+                # one path.
+                var_nodes = [node for node in nodes if node.is_var]
+                found_var_nodes = bool(var_nodes)
+
+                fast_return = not found_var_nodes
 
         for node in nodes:
             if node.is_var:
@@ -162,10 +174,11 @@ class CompiledRouter(object):
                 resource_idx = len(self._return_values)
                 self._return_values.append(node)
 
-            self._compile_tree(node.children, indent, level + 1)
+            self._compile_tree(node.children, indent, level + 1, fast_return)
 
             if node.resource is None:
-                line('return None')
+                if fast_return:
+                    line('return None')
             else:
                 # NOTE(kgriffs): Make sure that we have consumed all of
                 # the segments for the requested route; otherwise we could
@@ -173,11 +186,12 @@ class CompiledRouter(object):
                 line('if path_len == %d:' % (level + 1))
                 line('return return_values[%d]' % resource_idx, 1)
 
-                line('return None')
+                if fast_return:
+                    line('return None')
 
             indent = level_indent
 
-        if not found_simple:
+        if not found_simple and fast_return:
             line('return None')
 
     def _compile(self):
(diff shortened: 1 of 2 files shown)


kgriffs:
Routing tree from the test: https://github.com/falconry/falcon/pull/723

jmvrbanac:
Interesting. Seems to look good to me.

kgriffs:
Sorry, I had the wrong link for the routing tree. Updated.

falcon/routing/compiled.py
```diff
@@ -119,6 +119,18 @@ def line(text, indent_offset=0):
...
+        # fast 'return None'. See if the nodes at this branch are
+        # all still simple, meaning there is only one possible path.
+        if fast_return:
+            if len(nodes) > 1:
+                # NOTE(kgriffs): There's the possibility of more than
+                # one path.
+                var_nodes = [node for node in nodes if node.is_var]
+                found_var_nodes = bool(var_nodes)
+
+                fast_return = not found_var_nodes
```
fxfitz: This section all looks fine to me.

falcon/routing/compiled.py
```diff
@@ -162,10 +174,11 @@ def line(text, indent_offset=0):
                 resource_idx = len(self._return_values)
                 self._return_values.append(node)
 
-            self._compile_tree(node.children, indent, level + 1)
+            self._compile_tree(node.children, indent, level + 1, fast_return)
 
             if node.resource is None:
-                line('return None')
+                if fast_return:
+                    line('return None')
```
fxfitz: One thing that isn't overly clear to me is why we're adding these lines only if fast-return is True?  What happens If fast-return is false, such as the case where there are variable nodes?

(this is for all the `line('return None')` lines)

fxfitz: Hmm. I think I understand it now after going through it once more.

:+1:

tests/test_default_router.py
```diff
@@ -167,6 +178,32 @@ def test_variable(self):
...
+        self.assertEqual(resource.resource_id, 20)
+
+        resource, method_map, params = self.router.find('/gists/first/raw')
+        self.assertEqual(resource.resource_id, 18)
+
+        resource, method_map, params = self.router.find('/gists/first/pdf')
+        self.assertEqual(resource.resource_id, 21)
+
+        resource, method_map, params = self.router.find('/gists/1776/pdf')
+        self.assertEqual(resource.resource_id, 21)
```
fxfitz: Can we add test assertions for when a router *can't* find a route, instead of just for ones where there is a clear route defined?

jmvrbanac: That would probably be a good idea

jmvrbanac:
Outside of of @fxfitz's suggestion, :+1:

kgriffs:
@fxfitz, @jmvrbanac Thanks for the feedback! I rebased and added several negative tests, as well as adding tests for a couple of the specific cases that were reported in #702.

jmvrbanac:
:+1:

fxfitz:
LGTM too!

PR #725, opened 2016-03-03
fix(API): Skip process_resource() when no route is found

### LINKED ISSUE #691: `api.__call__()` will call `process_resource()` even if routing fails
Run into a strange bug that a middleware was failing, and traced it back to this unusual anomaly which we found in comments on `api.py`:

```
                                resource: Resource object to which the request was
                                routed. May be None if no route was found for
                                the request.
```

It's not documented in the middleware documentation that this is the behavior, but it also seems odd since every `process_resource()` would then have to start with `if resource is not None`. If routing fails, why not immediately raise `HTTPNotFound`?

kgriffs:
Avoid the problem of having to include a `if resource is not None` check at the top of every process_resource() method body.

Closes #691

### DIFF
--- a/doc/api/middleware.rst
+++ b/doc/api/middleware.rst
@@ -30,14 +30,17 @@ Falcon's middleware interface is defined as follows:
         def process_resource(self, req, resp, resource):
             """Process the request after routing.
 
+            Note:
+                This method is only called when the request matches
+                a route to a resource.
+
             Args:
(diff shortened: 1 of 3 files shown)


fxfitz:
LGTM :+1:

painterjd:
:+1:

foresmac:
Awesome. Do you know what the next release will be that would contain this fix?

PR #727, opened 2016-03-03
feat(Request): Add an option to toggle parsing of form params

### LINKED ISSUE #719: Add an option to RequestOptions to toggle parsing of form params
Add an option to RequestOptions to enable/disable parsing form params as currently implemented. It would be disabled by default (breaking change, but easy to work around). This would mitigate the problems people have experienced re the request stream being automatically slurped when the POSTed content type is `application/x-www-form-urlencoded`. 

This would not preclude working on a longer-term strategy for handling forms in a manner distinct from query parameters.

See also:
- #418
- #493

kgriffs:
BREAKING CHANGE: 

Falcon will no longer automatically parse requests that have the content type `application/x-www-form-urlencoded`. This was done to avoid unintended side-effects that may arise from consuming the request stream. It also makes it more straightforward for applications to customize and extend the handling of form submissions.

Applications that require this functionality must re-enable it explicitly, by setting a new request option that was added for that purpose, e.g.:

```py
app = falcon.API()
app.req_options.auto_parse_form_urlencoded = True
```

Closes #719

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -167,18 +167,12 @@ class Request(object):
 
             Note:
                 If an HTML form is POSTed to the API using the
-                *application/x-www-form-urlencoded* media type, Falcon
+                *application/x-www-form-urlencoded* media type, and
+                the :py:attr:`~.RequestOptions.auto_parse_form_urlencoded`
+                option is set, the framework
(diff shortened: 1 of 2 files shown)


jmvrbanac:
:+1:

PR #728, opened 2016-03-05
refactor(hooks): Remove deprecated global hooks feature

### LINKED ISSUE #384: Remove (deprecated) global hooks feature
Before doing this, carefully consider any pros for leaving it around...

kgriffs:
Middleware components superseded global hooks in version 0.3 of the framework. By now, developers should have had enough time to migrate away from the deprecated feature. If not, they can continue using 0.3 until they are able to migrate.

BREAKING CHANGE

falcon.API no longer accepts `before` and `after` kwargs to specify global hooks. Applications should migrate any logic contained in global hooks to reside in middleware components instead.

Old:
```py
    def do_something(req, resp, resource, params):
        pass

    app = falcon.API(before=[do_something])
```
New:
```py
    class ExampleMiddlewareComponent(object):
        def process_resource(self, req, resp, resource):
            pass
```

Closes #384

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -32,11 +32,6 @@ class API(object):
 
     Each API instance provides a callable WSGI interface and a routing engine.
 
-    Warning:
-        Global hooks (configured using the `before` and `after` kwargs) are
-        deprecated in favor of middleware, and may be removed in a future
-        version of the framework.
-
     Args:
         media_type (str, optional): Default media type to use as the value for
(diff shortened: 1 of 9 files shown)


fxfitz:
Okay, so this took me awhile to go over because I'm not as familiar with these hooks... but, I _think_ it LGTM.  I would really like it if someone else reviewed though. :-/

jmvrbanac:
:+1:

PR #729, opened 2016-03-12
Add options to opt out of the automatic splitting parameter containing commas

qwesda:
While the feature to automatically treat commas in parameters as indicating list values can be useful, for some it might not the ideal behaviour. Thus this pull requests introduces these new options:

- `auto_parse_query_string_lists` in `RequestOptions`
- `auto_parse_query_string_lists` in `parse_query_string()` 
- `encode_lists_as_comma_delimited_string` in `to_query_str()`

Setting `auto_parse_query_string_lists` to `False` `thing=1,2,3&thing=4` will be parsed to `{'thing': ['1,2,3','4']}`

Setting `encode_lists_as_comma_delimited_string ` to `False`will encode `{'thing': [1,2,3,4]}` to `thing=1&thing=2&thing=3&thing=4` 

The options default to the previous behaviour so that they don't introduce breaking changes.


tests/test_utils.py
```diff
@@ -128,6 +128,11 @@ def test_pack_query_params_one(self):
             falcon.to_query_str({'things': ['a', 'b']}),
             '?things=a,b')
 
+        self.assertEqual(
+            falcon.to_query_str({'things': ['a', 'b', True, False]},
```
kgriffs: It may also be helpful to test values that evaluate to `False`, such as `0`, to ensure that in the future someone doesn't inadvertently change the processing logic from the explicit identity checks of `list_value is True` and `list_value is False`, as currently implemented.

falcon/request.py
```diff
@@ -1160,6 +1162,8 @@ class RequestOptions(object):
             request's content type is
             *application/x-www-form-urlencoded* (default ``False``). In
             this case, the request's content stream will be left at EOF.
+        auto_parse_query_string_lists: Set to ``False`` in order to disable
```
kgriffs: Perhaps it would be helpful here to provide a short example of when this option would be useful.

falcon/util/misc.py
```diff
@@ -145,7 +145,7 @@ def http_date_to_dt(http_date, obs_date=False):
     raise ValueError('time data %r does not match known formats' % http_date)
 
 
-def to_query_str(params):
+def to_query_str(params, encode_lists_as_comma_delimited_string=True):
```
kgriffs: Style/nitpick: This is quite a long parameter name. Is there any way it could be shortened a bit?

falcon/util/uri.py
```diff
@@ -246,7 +246,8 @@ def decode(encoded_uri):
         return decoded_uri.decode('utf-8', 'replace')
 
 
-def parse_query_string(query_string, keep_blank_qs_values=False):
+def parse_query_string(query_string, keep_blank_qs_values=False,
+                       auto_parse_query_string_lists=True):
```
kgriffs: This param name could be shortened a little bit by omitting "query_string", since that is already implied by the function name.

falcon/util/uri.py
```diff
@@ -271,6 +272,8 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
         query_string (str): The query string to parse.
         keep_blank_qs_values (bool): If set to ``True``, preserves boolean
             fields and fields with no content as blank strings.
+        auto_parse_query_string_lists: Set to ``False`` in order to disable
+            splitting query parameters on ``,`` (default ``True``)
```
kgriffs: Again, a short blurb about when this might be useful could help the reader.

kgriffs:
Thanks for this! A few suggestions inline. I think it would be a good addition to the framework, but others may chime in if they disagree.

falcon/request.py
qwesda: I can not think of a good and short description of a use case except "If you need it for some reason".
My use case was that it was just easier to pass the values this way in the JS framework I use. It also seemed more "standard".

falcon/util/misc.py
qwesda: I could change it to just `encode_lists` or `comma_delimited_lists`, tell me which you  prefer.

falcon/util/uri.py
qwesda: done

tests/test_utils.py
qwesda: I changed the test to: 

```python
self.assertEqual(
  falcon.to_query_str({'things': ['a', 'b', '', None, True, False, 0]},
    encode_lists_as_comma_delimited_string=False),
 '?things=a&things=b&things=&things=None&things=true&things=false&things=0')
```

I also noticed that `None` is not is converted to `'None'` since it if not special cased like `True` and `False`. Was there already a discussion about how to handle "NULL" values? I would prefer to translate `None` to the empty string.

falcon/request.py
kgriffs: That makes sense. Perhaps you could just say something about how, for example, it can be used to make Falcon compatible with the style used by some front-end JS frameworks.

falcon/util/misc.py
kgriffs: `comma_delimited_lists` sounds good, thanks!

qwesda: done

falcon/request.py
qwesda: How about: `This might be helpful if comma encoding lists is not easy in the language/framework talking to the api.`?

tests/test_query_params.py
```diff
@@ -64,6 +64,28 @@ def test_percent_encoded(self):
         self.assertEqual(req.get_param_as_list('id', int), [23, 42])
         self.assertEqual(req.get_param('q'), u'\u8c46 \u74e3')
 
+    def test_option_auto_parse_lists(self):
```
stevelle: nit: I would prefer to have two test functions instead.

stevelle:
I would like to see this merge, as it makes falcon play nicely with more client-side frameworks and third-party libs without breaking backwards compat.

tests/test_query_params.py
qwesda: `test_option_auto_parse_lists_true` and `test_option_auto_parse_lists_false`?

kgriffs: Yeah, those names make sense.

falcon/util/misc.py
```diff
@@ -154,6 +154,10 @@ def to_query_str(params):
             something that can be converted into a ``str``. If `params`
             is a ``list``, it will be converted to a comma-delimited string
             of values (e.g., 'thing=1,2,3')
+        comma_delimited_lists (bool, default ``True``):
+            If set to false encode lists by specifying multiple instances
```
kgriffs: style/nit: false ==> \`\`False\`\`

tests/test_query_params.py
qwesda: done

falcon/util/misc.py
qwesda: done

qwesda:
I think I addressed all the comments – please review again.

kgriffs:
@qwesda Thanks for your patience. I'm closing up the 1.0 milestone. Once 1.1 opens, this will be one of the first things to land. I plan on doing smaller releases, more often, going forward.

qwesda:
No problem. Will use my branch until 1.1 comes around. Any approximate timeframe?

kgriffs:
Please see #820

kgriffs:
@qwesda can you please review #820?

PR #736, opened 2016-03-22
chore: Remove pylint test

kgriffs:
pylint has not been catching anything that flake8 wouldn't also catch, and flake8 continues to improve, yielding pylint somewhat obsolete.


fxfitz:
LGTM!

Issue #635, opened 2015-10-18
Raising errors from resource methods doesn't change status for middleware


karmux:
If raising HTTPNotFound or any other error that inherits from HTTPError or HTTPStatus from resource method or hook, then `process_response` method of middleware still has `200 OK` status. Response status is changed after running `process_response` of each middleware.

```
    def on_get(self, req, resp, pk):
        raise falcon.HTTPNotFound()
```

```
    # resp.status is '200 OK' but actual status in browser is '404 NOT FOUND'
    def process_response(self, req, resp, resource):
        logging.info(resp.status)
```

I think there should be correct status already in `process_response` of middleware if error is raised.

Currently I'm using following workaround:
```api.add_error_handler(falcon.HTTPError, handler=lambda e, *_: exec('raise e'))```

Second option would be using error handler that calls `_compose_error_response`.

**Comment by kgriffs** [comment]
> I think there should be correct status already in process_response of middleware if error is raised.

Thanks for the report! Seems like a reasonable change.

**Comment by joarleymoraes** [comment]
I was able to verify this issue. Looking through the code, I noticed the reason is that  `process_response` is called before `_compose_error_response` in case of an HTTPError, so status hasn't been set by then:

```
    self._call_resp_mw(middleware_stack, req, resp, resource)
    raise

(...)
except HTTPError as ex:
    self._compose_error_response(req, resp, ex)
    self._call_resp_mw(middleware_stack, req, resp, resource)
```

To fix, I thought of checking the error type like this:

```
if not isinstance(ex, HTTPError):
    self._call_resp_mw(middleware_stack, req, resp, resource)
raise
```
Here's the pull request #745

PR #745, opened 2016-03-27
fix(API): Handle HTTPError before calling process_response from middleware 

joarleymoraes:
In response to issue #635

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -206,7 +206,11 @@ class API(object):
                     # process_response when no error_handler is given
                     # and for whatever exception. If an HTTPError is raised
                     # remaining process_response will be executed later.
-                    self._call_resp_mw(middleware_stack, req, resp, resource)
+
+                    # NOTE(joarley): The condition will prevent that process_reponse
+                    # is called before self._compose_error_response on HTTPErrors
+                    if not isinstance(ex, HTTPError):
+                        self._call_resp_mw(middleware_stack, req, resp, resource)
                     raise
 
         except HTTPStatus as ex:
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -382,6 +382,28 @@ class TestSeveralMiddlewares(TestMiddleware):
         ]
         self.assertEqual(expectedExecutedMethods, context['executed_methods'])
 
+    def test_status_at_proc_resp_mw_on_http_error(self):
+        """Test that http status is correctly set by the time process_response is called"""
+        context = {}
+
+        class CheckStatusMiddleware(object):
+
+            def process_response(self, req, resp, resource):
+                context['status'] = resp.status
+
+        class RaiseHTTPErrorResource(object):
+
+            def onget(self, req, resp):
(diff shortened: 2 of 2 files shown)


tests/test_middleware.py
```diff
@@ -382,6 +382,28 @@ def handler(ex, req, resp, params):
...
+        context = {}
+
+        class CheckStatusMiddleware(object):
+
+            def process_response(self, req, resp, resource):
+                context['status'] = resp.status
+
+        class RaiseHTTPErrorResource(object):
+
+            def onget(self, req, resp):
```
kgriffs: `onget` ==> `on_get`

tests/test_middleware.py
```diff
@@ -382,6 +382,28 @@ def handler(ex, req, resp, params):
...
+                raise falcon.HTTPBadRequest(title='bad request', description='description')
+
+        self.api = falcon.API(middleware=[CheckStatusMiddleware()])
+
+        self.api.add_route(self.test_route, RaiseHTTPErrorResource())
+
+        self.simulate_request('GET', self.test_route)
```
kgriffs: The path comes first, i.e.:

```py
self.simulate_request(self.test_route, method='GET')
```

However, note that 'GET' is the default method, so you can simply write:

```py
self.simulate_request(self.test_route)
```

falcon/api.py
```diff
@@ -206,7 +206,11 @@ def __call__(self, env, start_response):
                     # process_response when no error_handler is given
                     # and for whatever exception. If an HTTPError is raised
                     # remaining process_response will be executed later.
-                    self._call_resp_mw(middleware_stack, req, resp, resource)
+
+                    # NOTE(joarley): The condition will prevent that process_reponse
+                    # is called before self._compose_error_response on HTTPErrors
+                    if not isinstance(ex, HTTPError):
```
kgriffs: Doing this causes a subtle breaking change in the case that `ex` is an instance of `HTTPError`. Before the change, if one of the response middleware methods (RMMs), itself, raised an error of type `HTTPStatus` or `HTTPError`, that will be caught and one more attempt will be made to execute the remaining RMMs. After this change, if the resource's responder raises an error, then one of the RMMs raises an error, it will simply go unhandled.

Admittedly, this highlights a weakness of the current design; IMO the framework should be able to handle any or all of the RMM raising instances of `HTTPStatus` or `HTTPError`, after which the exception is addressed and the remaining middleware methods executed. This would be more consistent and predictable.

joarleymoraes: You're right. I see the problem. Wouldn't be better if each `_call_req_mw`, `_call_rsrc_mw` and `_call_resp_mw` had their own try/except block ?

kgriffs: Yeah, we may have to just bite the bullet and do that.

kgriffs:
Thanks for taking a stab at this! As suspected, the solution ended up being quite involved. It could have been hacked to work but would have been pretty ugly, so I went ahead and reworked the entire section of this code (see #877).

Issue #747, opened 2016-04-01
Submitting anything as the body of a `simulate_request` leads to empty AssertionErrors


MasterKale:
I'm trying to write a test for a resource but for some reason whenever I set a string for the `body` parameter in my `simulate_request` I see AssertionErrors from `wsgiref.validate`:

	  File "/home/me/Repos/falconproject/middleware.py", line 30, in process_request
	    body = req.stream.read()
	  File "/usr/lib/python3.4/wsgiref/validate.py", line 197, in read
	    assert_(len(args) == 1)
	  File "/usr/lib/python3.4/wsgiref/validate.py", line 128, in assert_
	    raise AssertionError(*args)
	nose.proxy.AssertionError: 

There's no associated exception text so I have no idea why it's dying there. Line 30 in my `middleware` module is part of the `JSONTranslator` middleware I pulled from the quickstart:

    def process_request(self, req, resp):
        # req.stream corresponds to the WSGI wsgi.input environ variable,
        # and allows you to read bytes from the request body.
        #
        # See also: PEP 3333
        if req.content_length in (None, 0):
            # Nothing to do
            return

        # V This is line 30 V
        body = req.stream.read()

If I set `body` to an empty string then the request goes through because the JSONTranslator just returns out, but what good is request testing if you can't include anything in the request?

The docs are frustratingly vague, especially on this:

> body (str) – A string to send as the body of the request. Accepts both byte strings and Unicode strings (default: None). If a Unicode string is provided, it will be encoded as UTF-8 in the request.

What am I doing wrong here?

MasterKale:
I forgot to include my `TestCase`:

	import json

	import falcon
	from falcon import testing

	import app	


	class TestLoginResource(testing.TestCase):
	    def setUp(self):
	        super(TestLoginResource, self).setUp()

	        self.api = app.api

	    def test_bad_credentials_get_401(self):
	        response = self.simulate_post(
	            path='/login',
	            body=json.dumps({'username': 'foo', 'password': 'bar'})
	        )

	        self.assertEqual(response.status_code, falcon.HTTP_401)

MasterKale:
I figured out what the issue is, I'll issue a PR with fixes once I finish documenting it and writing tests.

kgriffs:
Fixed in #748

PR #748, opened 2016-04-02
fix(resource): prevent req.stream.read() from throwing assertionerror during tests

MasterKale:
Alright, please forgive the rather wordy PR but this was, in my opinion, a really obscure bug.

## Problem

The following TestCase:

```python
class TestLoginResource(testing.TestCase):
    def test_bad_credentials_get_401(self):
        result = self.simulate_post(
            path='/login',
            body=json.dumps({'username': 'foo', 'password': 'bar'})
        )

        self.assertEqual(result.status, falcon.HTTP_401)
```

Produces an `AssertionError` with the following stack trace:

	  File "/wherever/falconproject/middleware.py", line 30, in process_request
	    body = req.stream.read()
	  File "/usr/lib/python3.4/wsgiref/validate.py", line 197, in read
	    assert_(len(args) == 1)
	  File "/usr/lib/python3.4/wsgiref/validate.py", line 128, in assert_
	    raise AssertionError(*args)
	nose.proxy.AssertionError: 

## Cause

Under normal conditions, `req.stream` is an instance of `<class gunicorn.http.body.Body>`. However, during testing it is a `<class wsgiref.validate.InputWrapper>`. Both have `read()` methods, but they differ as such:

`gunicorn.http.body.Body` ([source](https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/body.py#L199)):

```python
def read(self, size=None):
    size = self.getsize(size)
    if size == 0:
        return b""
    # ...snip...
```

`wsgiref.validate.InputWrapper` ([source](https://github.com/python/cpython/blob/1fe0fd9feb6a4472a9a1b186502eb9c0b2366326/Lib/wsgiref/validate.py#L196)):

```python
def read(self, *args):
    assert_(len(args) == 1)
    v = self.input.read(*args)
    assert_(type(v) is bytes)
    return v
```

`InputWrapper` throws an AssertionError because **the call to `req.stream.read()` in my middleware doesn't declare a value for `size`!** Normally it's not needed because gunicorn sets the size automatically, but `InputWrapper` only asserts that it needs to be set.

When `size` is None (as in case of a typical call to `req.stream.read()`), `gunicorn.http.body.Body`'s [`getsize()` method](https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/body.py#L190) returns a value called `six.MAXSIZE`. [Going into `gunicorn.six`](https://github.com/benoitc/gunicorn/blob/master/gunicorn/six.py#L38), we see that, in a Python 3 environment, `MAXSIZE` is simply set to `sys.maxsize` (in Python 2, gunicorn runs several checks to help determine what `MAXSIZE` should be but I wasn't sure if I could just copy-paste it or not. `sys.maxsize` returned the same value locally in Python 2 and Python 3 so I left it as-is in my code).

To get my test to run, I just needed to update my middleware from this:

```python
class JSONTranslator(object):
	def process_request(self, req, resp):
	    if req.content_length in (None, 0):
	        return

	    body = req.stream.read()
```

To this:

```python
class JSONTranslator(object):
	def process_request(self, req, resp):
	    if req.content_length in (None, 0):
	        return
		
		# You'll need to add 'import sys' above
	    body = req.stream.read(sys.maxsize)
```

And calls to `TestCase.simulate_request()` in my tests started to pass. Even better, this appears to have no impact on my server when I run it as a gunicorn instance with `gunicorn app:api`.

Unfortunately this isn't much of a fix. Falcon users would need to pass in `sys.maxsize` every time they call `req.stream.read()` for the sole purpose of allowing tests to pass. That'd get pretty annoying pretty quickly.

## Solution

Fortunately it appears possible to fix this issue by subclassing `wsgiref.validate.InputWrapper` and overriding `read()` to include a size. I created the following `WsgiRefInputWrapperWithSize` class in `falcon.request.request_helpers`:

```python
class WsgiRefInputWrapperWithSize(InputWrapper):
    def read(self, *args):
        return super(WsgiRefInputWrapperWithSize, self).read(sys.maxsize)
```

Then, wrap `self.stream` in `falcon.request.Request.__init__()` if it's an instance of `InputWrapper`:

```python
def __init__(self, env, options=None):
    # ...snip...
    if type(self.stream) is wsgiref.validate.InputWrapper:
        self.stream = helpers.WsgiRefInputWrapperWithSize(self.stream)
    # ...snip...
```

This addresses issue #747

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -26,6 +26,7 @@ except AttributeError:
 
 import mimeparse
 import six
+from wsgiref.validate import InputWrapper
 
 from falcon.errors import *  # NOQA
 from falcon import util
@@ -303,7 +304,7 @@ class Request(object):
         # NOTE(kgriffs): Wrap wsgi.input if needed to make read() more robust,
         # normalizing semantics between, e.g., gunicorn and wsgiref.
         if _maybe_wrap_wsgi_stream:
-            if isinstance(self.stream, NativeStream):
+            if isinstance(self.stream, (NativeStream, InputWrapper,)):
                 self._wrap_stream()
             else:
                 # PERF(kgriffs): If self.stream does not need to be wrapped
--- a/tests/test_wsgiref_inputwrapper_with_size.py
+++ b/tests/test_wsgiref_inputwrapper_with_size.py
@@ -1,0 +1,32 @@
+import json
+
+import falcon
+from falcon import testing
+
+
+class TypeResource(testing.SimpleTestResource):
+    """ A simple resource to return the posted request body """
+    @falcon.before(testing.capture_responder_args)
+    def on_post(self, req, resp, **kwargs):
+        resp.status = falcon.HTTP_200
+        resp.body = json.dumps({'data': req.stream.read().decode('utf-8')})
+
+
+class TestWsgiRefInputWrapper(testing.TestCase):
+    def setUp(self):
+        super(TestWsgiRefInputWrapper, self).setUp()
+
+        # Set up a route to our TypeResoure
+        self.type_route = '/type'
+        self.api.add_route(self.type_route, TypeResource())
+
+    def test_resources_can_read_request_stream_during_tests(self):
+        """ Make sure we can perform a simple request during testing
+
+        Note: Normally, testing would fail after performing a request because no size is specified
+              when calling `wsgiref.validate.InputWrapper.read()` via `req.stream.read()`
+        """
+        result = self.simulate_post(path=self.type_route, body='hello')
+
+        self.assertEqual(result.status, falcon.HTTP_200)
+        self.assertEqual(result.json, {'data': 'hello'})
--- a/tests/test_wsgiref_inputwrapper_with_size.py
+++ b/tests/test_wsgiref_inputwrapper_with_size.py
@@ -5,10 +5,14 @@ from falcon import testing
 
 
 class TypeResource(testing.SimpleTestResource):
-    """ A simple resource to return the posted request body """
+    """A simple resource to return the posted request body."""
     @falcon.before(testing.capture_responder_args)
     def on_post(self, req, resp, **kwargs):
         resp.status = falcon.HTTP_200
+        # NOTE(masterkale): No size needs to be specified here because we're
+        # emulating a stream read in production. The request should be wrapped
+        # well enough to automatically specify a size when calling `read()`
+        # during either production or when running tests
         resp.body = json.dumps({'data': req.stream.read().decode('utf-8')})
 
 
@@ -21,11 +25,11 @@ class TestWsgiRefInputWrapper(testing.TestCase):
         self.api.add_route(self.type_route, TypeResource())
 
     def test_resources_can_read_request_stream_during_tests(self):
-        """ Make sure we can perform a simple request during testing
+        """Make sure we can perform a simple request during testing.
 
-        Note: Normally, testing would fail after performing a request because no size is specified
-              when calling `wsgiref.validate.InputWrapper.read()` via `req.stream.read()`
-        """
+        Originally, testing would fail after performing a request because no
+        size was specified when calling `wsgiref.validate.InputWrapper.read()`
+        via `req.stream.read()`"""
         result = self.simulate_post(path=self.type_route, body='hello')
 
         self.assertEqual(result.status, falcon.HTTP_200)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,144 @@
+1.0.0
+=====
+
+Breaking Changes
+----------------
+- The deprecated global hooks feature has been removed. ``API`` no
+  longer accepts ``before`` and ``after`` kwargs. Applications can work
+  around this by migrating any logic contained in global hooks to reside
+  in middleware components instead.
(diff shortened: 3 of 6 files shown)


MasterKale:
Whoa, this PR went pretty poorly...I just installed 2.6 locally, I'll keep tweaking things until tox returns green across the board. Should I close this until I get everything working?

kgriffs:
Hi @MasterKale, nice catch! Thanks for the comprehensive writeup; makes sense.

falcon/request.py
```diff
@@ -256,6 +257,9 @@ def __init__(self, env, options=None):
         self.stream = env['wsgi.input']
         self.method = env['REQUEST_METHOD']
 
+        if self.stream.__class__.__name__ == InputWrapper.__name__:
```
kgriffs: Alternatively, would adding `InputWrapper` to the `isinstance()` check on [L306](https://github.com/falconry/falcon/blob/master/falcon/request.py#L306) do the trick?

MasterKale:
@kgriffs I just confirmed, an instance of `Request` isn't created during testing until some time after `wsgiref.validate.validator()` is called ([L327](https://github.com/falconry/falcon/blob/master/falcon/testing/test_case.py#L327) in `testing/test_case.py`). Since L327 is the one that errors out during a `simulate_request()` there'd be no chance for that wrap check in L306 to work its magic.

falcon/request.py
MasterKale: @kgriffs I just confirmed, an instance of `Request` isn't created during testing until some time after `wsgiref.validate.validator()` is called ([L327](https://github.com/falconry/falcon/blob/master/falcon/testing/test_case.py#L327) in `testing/test_case.py`). Since L327 is the one that errors out during a `simulate_request()` there'd be no chance for that wrap check in L306 to work its magic.

MasterKale: Okay, I had to delete the Cython-compiled files before my debug `print()`s started showing. It turns out, yes, adding `or isinstance(self.stream, InputWrapper):` to to L306 is enough to get testing working. So much for this PR :cry:

kgriffs: In this case, I'd recommend just replacing your changes to `request.py` and `request_helpers.py` with an addition of `InputWrapper` to L306. You can rebase and squash your commits to keep things clean. The additional tests you added in `test_wsgiref_inputwrapper_with_size.py` are still helpful, methinks.

By the way, [isinstance()](https://docs.python.org/2/library/functions.html#isinstance) supports passing a tuple of types to check, so that may be the most elegant way to go.

Thanks for working on this!

MasterKale:
Okay, as requested I've squashed down to that single-line change to L306 and accompanying tests, and recommitted.

tests/test_wsgiref_inputwrapper_with_size.py
```diff
@@ -0,0 +1,32 @@
+import json
+
+import falcon
+from falcon import testing
+
+
+class TypeResource(testing.SimpleTestResource):
+    """ A simple resource to return the posted request body """
```
kgriffs: Style/nitpick: Remove leading and trailing whitespace of the tripple-quoted string, and end the first line with a period.

tests/test_wsgiref_inputwrapper_with_size.py
```diff
@@ -0,0 +1,32 @@
...
+class TestWsgiRefInputWrapper(testing.TestCase):
+    def setUp(self):
+        super(TestWsgiRefInputWrapper, self).setUp()
+
+        # Set up a route to our TypeResoure
+        self.type_route = '/type'
+        self.api.add_route(self.type_route, TypeResource())
+
+    def test_resources_can_read_request_stream_during_tests(self):
+        """ Make sure we can perform a simple request during testing
```
kgriffs: See above comment.

tests/test_wsgiref_inputwrapper_with_size.py
```diff
@@ -0,0 +1,32 @@
...
+
+        Note: Normally, testing would fail after performing a request because no size is specified
```
kgriffs: Style/nit: Although Falcon's pep8 config allows lines over 80 characters long, my rule of thumb is to wrap when you can and only go over for that occasional line of code that would otherwise look really ugly if wrapped.

tests/test_wsgiref_inputwrapper_with_size.py
```diff
@@ -0,0 +1,32 @@
...
+import falcon
+from falcon import testing
+
+
+class TypeResource(testing.SimpleTestResource):
+    """ A simple resource to return the posted request body """
+    @falcon.before(testing.capture_responder_args)
+    def on_post(self, req, resp, **kwargs):
+        resp.status = falcon.HTTP_200
+        resp.body = json.dumps({'data': req.stream.read().decode('utf-8')})
```
kgriffs: Might be worth adding an additional `#NOTE(masterkale): ...` comment here, to point out that no size is passed to `read()`, even though it is also mentioned in a comment later on.

kgriffs:
OK, this is almost ready to merge. Just a few final nitpicks to look at. Thanks!

kgriffs:
LGTM. Nice work!

PR #759, opened 2016-04-19
fix(doc): Include RTD theme and force a full rebuild of docs

kgriffs:
Render docs using the RTD theme. Also, force a full rebuild of the docs
each time since sphinx sometimes doesn't detect docstring changes and
refuses to re-render them.


fxfitz:
LGTM

PR #762, opened 2016-04-25
feat(Request): Allow 1 and 0 as valid boolean values in query string parsing

### LINKED ISSUE #684: Also convert '1' and '0' as booleans from query string
I see that [tests](https://github.com/falconry/falcon/blob/master/tests/test_query_params.py#L194) specifically exclude `0` and `1` from valid values to be converted to boolean by `Request.get_param_as_bool` method.

I'd like to discuss this situation :)

I see no reason not to convert those values as they seem non ambiguous to me and there are a bunch of situations where 1 and 0 are de facto boolean (database outputs…), but we only deals with string in query string.

What is the rationale behind this?

Thanks! :)

yohanboniface:
Fix #684

Thanks :)


kgriffs:
This looks good, thanks! Would you mind rebasing? Once that's done, this is ready to merge.

yohanboniface:
Done :)

PR #765, opened 2016-04-27
doc(install): Update PyPy and CPython instructions

kgriffs:
Edit installation instructions to match how 1.0.0 will be published on PyPI.


fxfitz:
Reads good to me!

PR #766, opened 2016-04-27
chore(publish): Split publish script into build and publish steps

kgriffs:
Split up the publish script into separate build and publish scrips. Use twine and build wheels on OS X. Use pyenv to keep everything tidy. Move everything under the tools directory to keep scripts in one place.


jmvrbanac:
:+1:

Issue #769, opened 2016-04-28
Add test for hyphens and other non-arg-friendly chars in URI template field names


kgriffs:
Currently we only check up-front in `add_route()` for field names that start with a digit or contain whitespace. This can lead to cryptic errors later on when requests are routed to responders.

kgriffs:
See also: `test_whitespace_not_allowed()` in `tests.test_uri_template` and `add_route()` in `routing.compiled`.

enzzc:
Is this template supposed to be valid? '/repos/{org}/{repo}/compare/{simple-vs-complex}' (from /tests/test_default_router.py:140). `simple-vs-complex` is not a valid Python identifier.

kgriffs:
Good catch! That will blow up if it were actually passed as kwargs to a resource responder method.

PR #771, opened 2016-04-29
in response to issue #769

enzzc:
Make sure fields are valid Python identifiers

### DIFF
--- a/falcon/routing/compiled.py
+++ b/falcon/routing/compiled.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import re
+import keyword
 
 
 TAB_STR = ' ' * 4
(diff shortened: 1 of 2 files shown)


tests/test_default_router.py
```diff
@@ -137,7 +137,6 @@ def test_collision(self, template):
         )
 
     @ddt.data(
-        '/repos/{org}/{repo}/compare/{simple-vs-complex}',
```
kgriffs: Can we keep this, but just use underscores instead of hyphens?

falcon/routing/compiled.py
```diff
@@ -43,8 +44,12 @@ def add_route(self, uri_template, method_map, resource):
         """Adds a route between URI path template and resource."""
         # Can't start with a number, since these eventually get passed as
         # args to on_* responders
-        if re.search('{\d', uri_template):
-            raise ValueError('Field names may not start with a digit.')
+
+        fields = re.findall('{(.+?)}', uri_template)
```
kgriffs: Should we also look for the empty string? E.g., given a path template of `'/some/{}/template'`we would raise an error?

tests/test_default_router.py
kgriffs: Separately, it would be good to create some new tests that expect (and assert) that ValueError is raised when an invalid field name is specified in the template.

kgriffs:
@enzzc have you had a chance to review my comments?

kgriffs:
Addressed comments and merged as #843

Issue #789, opened 2016-05-21
Implement get_param_as_dict()


kgriffs:
This helper would deserialize JSON into a `dict`. For example:

`thing=%7B%22name%22%3A%20%22value%22%2C%20%22another%22%3A%20%22field%22%2C%20%22some%22%3A%20%22thing%22%7D`

...would be interpreted by req.get_param_as_dict as:

```json
{"name": "value", "another": "field", "some": "thing"}
```

yashmehrotra:
Hi @kgriffs, I would like to implement this feature, can you assign this issue to me.
Thanks.

PR #791, opened 2016-05-21
feat(Request): Implemented get_param_as_dict()

yashmehrotra:
Implemented `get_param_as_dict()` function to allow users to deserialize JSON data from query string parameters as discussed in #789

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 from datetime import datetime
+import json
 
 try:
     # NOTE(kgrifs): In Python 2.6 and 2.7, socket._fileobject is a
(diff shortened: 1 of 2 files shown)


fxfitz:
:+1:

falcon/request.py
```diff
@@ -1039,6 +1040,47 @@ def get_param_as_date(self, name, format_string='%Y-%m-%d',
...
+                the value of the param, but only if the param is found (default
+                ``None``).
+
+        Returns:
+            dict: The value of the param if it is found. Otherwise, returns
+                ``None`` unless required is True.
+
+        Raises:
+            HTTPBadRequest: A required param is missing from the request.
+            HTTPInvalidParam: A transform function raised an instance of
```
kgriffs: It looks like `HTTPInvalidParam` is only raised when the value of the parameter can't be parsed as JSON...

kgriffs:
This looks good, just needs a tweak for the docstring.

falcon/request.py
yashmehrotra: @kgriffs So, should I change the docstring to `HTTPInvalidParam: A transform function raised an instance of Value Error due to invalid JSON formatting` ?

kgriffs: In this case there's no transform function involved, so perhaps something more along the lines of "The parameter's value could not be parsed as JSON"

yashmehrotra: Ok.

yashmehrotra:
@kgriffs I have fixed the docstring as suggested.

yashmehrotra:
@kgriffs Any updates on this ?

kgriffs:
@yashmehrotra sorry for the delay. I think this is ready to go. Would you mind rebasing onto master so we can get the CI checks to pass before merging? Thanks!

kgriffs:
Rebased and merged as #839. Thanks!

PR #801, opened 2016-05-26
fix(Response): Default OPTIONS responder sets Content-Length to 0

### LINKED ISSUE #796: Default OPTIONS responder does not set Content-Length to "0"
Per RFC 7231:

> A server MUST generate a Content-Length field with a value of "0" if no payload body is to be sent in the response.

yashmehrotra:
Closes #796


painterjd:
This looks good. 

It would be nice to see a test for this, however.

yashmehrotra:
@painterjd Hi, I have added the test for this. Please Review.
Thanks.

kgriffs:
LGTM

jmvrbanac:
:+1:

PR #804, opened 2016-06-02
chore: adds py35 environment to tox.ini

painterjd:
This patch adds Python 3.5 to the list of environments in tox.ini.


fxfitz:
:+1:

PR #805, opened 2016-06-02
perf(Request): use try/catch instead of if..in for query string

### LINKED ISSUE #637: Using env.get instead of if...in
I had a question about the following piece of code in falcon.request.Request:

```python
# PERF(kgriffs): if...in is faster than using env.get(...)
        if 'QUERY_STRING' in env:
            self.query_string = env['QUERY_STRING']

            if self.query_string:
                self._params = parse_query_string(
                    self.query_string,
                )

            else:
                self._params = {}

        else:
            self.query_string = ''
            self._params = {}
```

For the part,
```python
 if 'QUERY_STRING' in env:
     self.query_string = env['QUERY_STRING']
```
Arent you performing two hash operations (one for checking and one for retrieving)? Based on that, env.get should be better?

ueg1990:
use try-catch instead of if..in for query string. try-catch is cheaper
and faster in Python

Fixes #637

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -272,10 +272,13 @@ class Request(object):
         else:
             self.path = '/'
 
-        # PERF(kgriffs): if...in is faster than using env.get(...)
-        if 'QUERY_STRING' in env:
+        # # PERF(ueg1990): try/catch cheaper and faster (and more Pythonic)
+        try:
             self.query_string = env['QUERY_STRING']
-
+        except KeyError:
+            self.query_string = ''
+            self._params = {}
+        else:
             if self.query_string:
                 self._params = parse_query_string(
                     self.query_string,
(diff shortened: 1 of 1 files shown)


falcon/request.py
```diff
@@ -272,10 +272,13 @@ def __init__(self, env, options=None):
         else:
             self.path = '/'
 
-        # PERF(kgriffs): if...in is faster than using env.get(...)
-        if 'QUERY_STRING' in env:
+        # # PERF(ueg1990): try/catch cheaper and faster (and more Pythonic)
```
fxfitz: Nit: double #

fxfitz:
LGTM

falcon/request.py
ueg1990: oh didn't see that....after updating this, should i squash my commits and update pull request?

fxfitz: Sure! We've been using the new fancy "squash-and-merge" button for Github anyway, so if you don't squash, no worries. :-)

PR #809, opened 2016-06-02
feat(HTTPError): set title to status message if title is None

ueg1990:
set HTTPError title attribute to status attribute if title is None

Fixes 778


falcon/http_error.py
```diff
@@ -51,7 +52,7 @@ class HTTPError(Exception):
         status (str): HTTP status code and text, such as "400 Bad Request"
 
     Keyword Args:
-        title (str): Human-friendly error title (default ``None``).
+        title (str): Human-friendly error title.
```
kgriffs: We should explain that this arg is optional and if not passed, will default to the status.

kgriffs:
Looks like codecov is complaining...

ueg1990:
hey @kgriffs im not sure why the codecov is giving less than 100% even though I changed only one line of code...do you know what i am missing?

kgriffs:
Looks like title can never be `None`, so this branch never executes: 

https://codecov.io/gh/falconry/falcon/src/17c18029de7d663d968df69bd71b51d4d2a367e6/falcon/http_error.py#L135

ueg1990:
should i remove the line 'if self.title is not None' from lines 135 and 173?

kgriffs:
That makes sense. Now that it can never be `None`, there's no need to check.

kgriffs:
@ueg1990  Hmm, looks like at some point the history of this patch diverged from master. Would you mind rebasing just your commits onto master and doing a force push to clean up this branch (ueg1990:ueg1990_issue_778)?

ueg1990:
How can I rebase only my commits? It looks like other commits are now included in this branch when i rebased before (Which i had not intend to do)

kgriffs:
If all else fails you can create a new branch and then cherry-pick just the commits you want into that new branch.

kgriffs:
Alternatively, you should be able to do an interactive rebase onto master (`git rebase -i master`) and then you can select which commits to keep.

kgriffs:
[Rebased and merged as #809]

Issue #749, opened 2016-04-02
Add option to opt-out from comma separated value parsing


rr-:
I'm porting a project to Falcon and I stumbled upon an issue regarding its parsing of CSV values inside URIs. Let's say I have filtering engine that accepts queries such as this:

    http://great.dude/api/cars?query=added:yesterday,today+spoilers:red

I obviously want to make `req.get_param('query')` return `'added:yesterday,today spoilers:red'`, and not `['added:yesterday', 'today spoilers:red']`.

Right now this [isn't really configurable](https://github.com/falconry/falcon/blob/35987b2be85456f431bbda509e884a8b0b20ed11/falcon/util/uri.py#L312-L328) and I need to check if `get_param()` returns a `list` and then join it back if needed, which looks sort of silly. Fortunately, the ability to use custom request classes alleviates the issue to some extent.

I see a few ways to improve things upstream:

1. Offer explicit `get_param_as_string` that will possibly do `','.join(...)` under the hood.
2. Add an option to disable this mechanism as an additional option to `Api`.
3. Add an option to disable this mechanism as an additional option to `add_route()`.
4. Change `get_param` to always return string.

Option 4 makes the most sense to me, but it breaks BC. If option 4 is not feasible, I'd went with option 1.

kgriffs:
I think we can start by adding an option to `RequestOptions` to enable/disable parsing of comma-separated query string values. It can default to `True` for the sake of BC. For Falcon 2.0 we can consider changing the default to `False`. 

For more fine-grained control I like the idea of passing options to `API.add_route()`. But perhaps we should wait to see if there is demand for that from the community (unless you already have a need for it, of course).

rr-:
FWIW for time being, I ended up rolling own solution like this:

- Input parsing facilities within custom context class: https://github.com/rr-/szurubooru/blob/master/server/szurubooru/api/context.py 
- Filling context class with user input inside middleware: https://github.com/rr-/szurubooru/blob/master/server/szurubooru/middleware/context_adapter.py#L54-L55

I don't think I currently need options within `add_route()`. The idea sounds pretty good, as it'll add some flexibility, though.

[...]

PR #811, opened 2016-06-02
feat(RequestOptions): Add option to opt-out from comma separated value

orcsly:
parsing #749

Add parse_csv_qs_values_as_list to make get_param return a list for csv.
Defaults to True.


tests/test_query_params.py
```diff
@@ -384,6 +384,48 @@ def test_list_transformer(self):
                              'The value is not formatted correctly.')
             self.assertEqual(ex.description, expected_desc)
 
+    def test_list_as_string_option(self):
```
molund: the test name doesn't match the name of the option being tested

kgriffs:
Please see #820

Issue #744, opened 2016-03-26
Generate HTTP status string from status code


momoneko:
Does falcon have any way of generating the status string from the status code? 

I have to pass the status from a HTTP call to the requester, and I get only the HTTP status code, not the entire string. Having a way to generate the string from the status would make it way easier.

kgriffs:
This isn't in the framework, but I think it would be worth adding. We'll need to note that there's a slight performance penalty for doing the conversion, but 

I would suggest adding a utility function for this as well as a new property on the Response class that accepts just a code.

sherwu:
I can take this!

kgriffs:
[Merged]

PR #812, opened 2016-06-02
Issue #744 - added a utility function for looking up status messages

molund:
Added a utility function for looking up status messages by status code.

Additional work is still required to add either one or two new attributes to the response object.  There should be 3 altogether for consistency with the python requests library and the JavaScript XmlHttpRequest:

status  (e.g.  "404 Not Found")  <= already exists
status_code  (e.g. 404)  <= NEW (read / write)
reason  (e.g. "Not Found")  <= NEW  (read only)

### DIFF
--- a/falcon/util/misc.py
+++ b/falcon/util/misc.py
@@ -19,6 +19,8 @@ import warnings
 
 import six
 
+from falcon import status_codes
+
 __all__ = (
     'deprecated',
     'http_now',
@@ -26,6 +28,7 @@ __all__ = (
     'http_date_to_dt',
     'to_query_str',
     'get_bound_method',
+    'get_http_status'
 )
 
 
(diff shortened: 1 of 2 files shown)


falcon/util/misc.py
```diff
@@ -210,3 +212,25 @@ def get_bound_method(obj, method_name):
...
+        ValueError: the value entered could not be converted to an integer
+
+    """
+    code = int(status_code)  # converts to integer or raises ValueError
+
+    try:
+        return getattr(falcon.status_codes, 'HTTP_' + str(code))
+    except AttributeError:
+        # not found
+        return str(code) + ' Unknown Status Code'
```
ueg1990: 1) why use str(code) when you can use status_code, which is already a string
2) why do you want to raise ValueError in this function?

molund: 1)  converting to int and then back to string handles status codes like 401.1 and 401.2 that are not available for lookup in the status_codes.py file
2) a ValueError would indicate that somebody passed a non-numeric value into the function

ueg1990: w.r.t. to your 1), i meant to say that after doing an int, you do not need to convert it back to string, you can just re-use 'status_code'
And for '401.2', if you do int('401.2'), you would get:
>>> int('401.2')
ValueError: invalid literal for int() with base 10: '401.2'

falcon/util/misc.py
```diff
@@ -210,3 +214,33 @@ def get_bound_method(obj, method_name):
...
+    Returns:
+        str: status code e.g. "404 Not Found"
+
+    Raises:
+        ValueError: the value entered could not be converted to an integer
+
+    """
+    # sanitize inputs
+    try:
+        code = decimal.Decimal(str(status_code))  # Decimal can validate values like "401.1"
```
kgriffs: If we are expecting either a number or a string, do we need to pass this through `str()`? Also, can we just use `float()` here instead of `Decimal()`, since the value is going to be truncated anyway by `int()`?

falcon/util/misc.py
```diff
@@ -210,3 +214,33 @@ def get_bound_method(obj, method_name):
...
+            raise ValueError
+    except (decimal.InvalidOperation, ValueError):
+        raise ValueError('get_http_status "%s" is not a valid status code', status_code)
+
+    # lookup the status code
+    try:
+        return getattr(status_codes, 'HTTP_' + str(code))
+    except AttributeError:
+        # not found
+        return str(code) + ' Unknown Status Code'
```
kgriffs: Should the function raise a ValueError instead?

molund: Possibly.  
However, according to the RFC, http status codes are extensible.  The current behaviour of the function differentiates between unknown and invalid.

kgriffs: That makes sense, but if it is intentional, wouldn't the caller want to supply their own description string?

kgriffs:
Thanks, LGTM!

jmvrbanac:
:+1:

PR #814, opened 2016-06-02
docs: updates docstrings to explain uri encoding

### LINKED ISSUE #687: resp.location double-encodes urlencoded strings
I've got a URL which contains a url-encoded URL as a parameter, and when setting resp.location, that results in it being double-encoded:

resp.location = "http://something?redirect_uri=http%3A%2F%2Fsite"

This should result in:
Location: http://something?redirect_uri=http%3A%2F%2Fsite...

But what actually happens is:
Location: http://something?redirect_uri=http%253A%252F%252Fsite

I worked around by raising a HTTPStatus() object, but this doesn't seem like the ideal situation.

### LINKED ISSUE #782: Response href headers are double escaped
#135 makes an attempt to simplify things for developers, but unfortunately it actually breaks well-formed URIs by double escaping them. 

I may never have noticed this were I not trying to pass a signed URI where the double escaping broke the signature. Signed S3 URLs that are double escaped definitely don't work!

Of course, the workaround is easy enough — just use `Response.set_header()` directly — I just wanted to flag this as behaviour that was quite surprising and may not be desirable in the long run.

painterjd:
This patch updates docstring to explain that parameters set to Location and Content-Location headers shall be URI encoded per RFC 3986.

closes #687, closes #782

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -474,7 +474,12 @@ class Response(object):
 
     content_location = header_property(
         'Content-Location',
-        'Sets the Content-Location header.',
+        """Sets the Content-Location header.
+
+        This value will be URI encoded per RFC 3986. If the value that is
+        being set is already URI encoded it should be decoded first or the
+        header should be set manually using the set_header method.
+        """,
         uri_encode)
 
     content_range = header_property(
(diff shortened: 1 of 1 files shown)


kgriffs:
LGTM, thanks!

PR #815, opened 2016-06-02
feat(Response): add support to Response for the Accept-Ranges header

### LINKED ISSUE #646: Add support to Response for the Accept-Ranges header
Should be fairly straightforward (see also: https://tools.ietf.org/html/rfc7233#section-2.3)

ueg1990:
add support to Response class for Accept-Ranges header
(https://tools.ietf.org/html/rfc7233#section-2.3)

Fixes #646


falcon/response.py
```diff
@@ -554,6 +554,22 @@ def add_link(self, target, rel, title=None, title_star=None,
         """,
         lambda v: ', '.join(v))
 
+    accept_ranges = header_property(
+        'Accept-Ranges',
+        """
+        Sets the Accept-Ranges headers.
```
kgriffs: Typo: "headers" ===> "header"

falcon/response.py
```diff
@@ -554,6 +554,22 @@ def add_link(self, target, rel, title=None, title_star=None,
...
+
+        The "Accept-Ranges" header field allows a server to indicate that it
+        supports range requests for the target resource.
```
kgriffs: Perhaps make it a little clearer what the value should be (e.g., 'bytes')

kgriffs: The RFC calls it a "range unit"

ueg1990: So,
'The "Accept-Ranges" header field allows a server to indicate that it supports range requests for the target resource. The value of this header indicate what range units are supported (e.g. bytes)'

kgriffs: That works. Just a couple of nitpicks though: remove the quotes around the header name, but do quote the "bytes" range unit.

falcon/response.py
```diff
@@ -554,6 +554,22 @@ def add_link(self, target, rel, title=None, title_star=None,
...
+
+        If the server does not support any kind of range request for the
+        target resource, then set the header to ``none``
```
kgriffs: Nitpick: I think it would read better to simple quote "none" rather than formatting it as an identifier.

falcon/response.py
```diff
@@ -554,6 +554,22 @@ def add_link(self, target, rel, title=None, title_star=None,
...
+
+        Note:
+            ``none`` is the literal string, not Python's built-in type ``None``
```
kgriffs: See above. On the other hand, keep \`\`None\`\`

ueg1990:
@kgriffs i did a git rebase to latest master before adding your review comments and pep8 is failing but the issue is not coming from my changes. do you know what might have gone wrong?

ueg1990:
There are two main errors:
1) F405....i am not sure how to handle that
2) F401....This comes from line 4 in utils/__init__.py. Add '# NOQA' at the end of the line will fix the issue

kgriffs:
@ueg1990 Were you going to create a new PR?

Issue #783, opened 2016-05-18
Add package and subpackage docstrings


kgriffs:
Add docstrings for `falcon`, `falcon.util`, etc. for use in `help()`.

PR #816, opened 2016-06-02
Add docstrings for falcon, falcon.util, etc. for use in help().

t20:
Closes : #783


falcon/util/__init__.py
```diff
@@ -1,3 +1,8 @@
+"""
+Utility functions for the Falcon framework.
+These modules are already imported by the falcon module and can be accessed directly.
```
kgriffs: Except for the uri module... Also, we should add a docstring for `uri.py.

kgriffs:
[Augmented and merged as #850]

PR #818, opened 2016-06-03
feat(request,response): add __dict__ to __slots__

### LINKED ISSUE #785: Make it easier to add custom attributes to slotted classes
Users of the framework sometimes want to assign additional variables to the `Request`, `Response`, or other "slotted" classes. This can be allowed without hurting efficiency by adding a `'__dict__'` string to the `__slots__` declarations.

From the Python [data model docs](https://docs.python.org/2/reference/datamodel.html#slots): 

> Without a __dict__ variable, instances cannot be assigned new variables not listed in the __slots__ definition. Attempts to assign to an unlisted variable name raises AttributeError. If dynamic assignment of new variables is desired, then add '__dict__' to the sequence of strings in the __slots__ declaration.

I did a quick test and adding '__dict__' does not decrease performance for accessing the other slotted variables in 2.7 and 3.5. 

Question: Does this enable an anti-pattern? Would we rather that apps use `req.context`? Would it be better to have a context type that supports attribute-based access? (e.g., `req.context.foo` == `req.context['foo']`)?

autoferrit:
adds `__dict__` to the `__slots__` methods to the Request and Response classes
to make it more extensible via subclassing and adding custom attributes. I have
also added tests to make sure adding custom attributes do not break and raise
an `AttributeError`.

fixes #785

### DIFF
--- a/tests/test_slots.py
+++ b/tests/test_slots.py
@@ -1,0 +1,26 @@
+from falcon import Request, Response
+import falcon.testing as testing
+
+
+class TestSlots(testing.TestBase):
+
+    def __init__(self):
+        self.failed = False
+        super(TestSlots, self).__init__()
+
+    def test_slots_request(self):
+        env = testing.create_environ()
+        req = Request(env)
+        try:
+            req.doesnt = 'exist'
+        except AttributeError:
+            self.failed = True
+        self.assertFalse(self.failed, 'Unable to add to __slots__ dynamically')
+
+    def test_slots_response(self):
+        resp = Response()
+        try:
+            resp.doesnt = 'exist'
+        except AttributeError:
+            self.failed = True
+        self.assertFalse(self.failed, 'Unable to add to __slots__ dynamically')
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -241,6 +241,7 @@ class Request(object):
         'options',
         '_cookies',
         '_cached_access_route',
+        '__dict__',
     )
 
     # Allow child classes to override this
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -98,7 +98,8 @@ class Response(object):
         '_cookies',
         'status',
         'stream',
-        'stream_len'
+        'stream_len',
+        '__dict__',
     )
 
     def __init__(self):


tests/test_slots.py
```diff
@@ -0,0 +1,25 @@
...
+        raised = False
+        try:
+            req.doesnt = 'exist'
+        except AttributeError:
+            raised = True
+        self.assertFalse(raised, 'Unable to add to __slots__ dynamically')
+
+    def test_slots_response(self):
+        resp = Response()
+        resp.doesnt = 'exist'
```
kgriffs: I think you can just remove L19 since this it is the same as L22.

tests/test_slots.py
```diff
@@ -0,0 +1,25 @@
...
+        raised = False
+        try:
+            resp.doesnt = 'exist'
+        except AttributeError:
+            raised = True
```
kgriffs: What do you think about just calling `self.fail()` here instead of using the `raised` variable?

kgriffs:
Thanks, this looks good. Just a couple inline suggestions re the tests.

tests/test_slots.py
SkiftCreative: Yea, i like that idea. Ill commit those changes.

tests/test_slots.py
```diff
@@ -0,0 +1,26 @@
...
+        self.failed = False
+        super(TestSlots, self).__init__()
+
+    def test_slots_request(self):
+        env = testing.create_environ()
+        req = Request(env)
+        try:
+            req.doesnt = 'exist'
+        except AttributeError:
+            self.failed = True
```
kgriffs: I think you can just use `self.fail()` here, as in:

```py
self.fail('Unable to add to __slots__ dynamically')
```

tests/test_slots.py
```diff
@@ -0,0 +1,26 @@
...
+        self.assertFalse(self.failed, 'Unable to add to __slots__ dynamically')
+
+    def test_slots_response(self):
+        resp = Response()
+        try:
+            resp.doesnt = 'exist'
+        except AttributeError:
+            self.failed = True
```
kgriffs: See above.

tests/test_slots.py
```diff
@@ -0,0 +1,26 @@
+from falcon import Request, Response
+import falcon.testing as testing
+
+
+class TestSlots(testing.TestBase):
+
+    def __init__(self):
```
kgriffs: If `self.failed` is not necessary (see below), the `__init__()` method can be removed.

tests/test_slots.py
SkiftCreative: if I change that up, then how should the test be updated as it is using self.fail

```
self.assertFalse(self.failed, 'Unable to add to __slots__ dynamically')
```

what about:

```
class TestSlots(testing.TestBase):

    def __init__(self):
        self.failed = False
        super(TestSlots, self).__init__()

    def test_slots_request(self):
        env = testing.create_environ()
        req = Request(env)
        try:
            req.doesnt = 'exist'
        except AttributeError:
            self.failed = True
        self.assertFalse(self.failed, self.fail('Unable to add to __slots__ dynamically'))

    def test_slots_response(self):
        resp = Response()
        try:
            resp.doesnt = 'exist'
        except AttributeError:
            self.failed = True
        self.assertFalse(self.failed, self.fail('Unable to add to __slots__ dynamically'))
```

kgriffs: My thought was that you could remove the `self.failed` variable altogether:

```py
def test_slots_response(self):
        resp = Response()
        try:
            resp.doesnt = 'exist'
        except AttributeError:
            self.fail('Unable to add to __slots__ dynamically')
```

Even better (now that I think of it) would be to use [ExpectedException](http://testtools.readthedocs.io/en/latest/for-test-authors.html#expectedexception) (both `falcon.testing.TestBase` and `falcon.testing.TestCase` inherit from `testtools.TestCase).

tests/test_slots.py
```diff
@@ -0,0 +1,26 @@
+from falcon import Request, Response
+import falcon.testing as testing
+
+
+class TestSlots(testing.TestBase):
```
kgriffs: The `TestBase` class is deprecated; please use `falcon.testing.TestCase` instead.

kgriffs:
@SkiftCreative just following to see if you've had a chance to review my most recent comments?

kgriffs:
[merged with minor edits as #868]

PR #820, opened 2016-06-07
feat(util,Request): Add options to disable csv parsing of query params

### LINKED ISSUE #749: Add option to opt-out from comma separated value parsing

kgriffs:
In all cases, the options default to the previous behavior in order to avoid introducing any breaking changes.

* Add option to to_query_str to encode lists as multiple occurences of the same parameter.
* Add option to parse_query_string to not split comma-delimited values.
* Add an option to RequestOptions for toggling csv parsing.

Co-Authored-By: Daniel Schwarz <qwesda@me.com>
Co-Authored-By: Steven Ly <sly@openroad.ca>
Co-Authored-By: Kurt Griffiths <inbox@kgriffs.com>

Closes #749

### DIFF
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -246,11 +246,12 @@ else:
         return decoded_uri.decode('utf-8', 'replace')
 
 
-def parse_query_string(query_string, keep_blank_qs_values=False):
+def parse_query_string(query_string, keep_blank_qs_values=False,
+                       parse_qs_csv=True):
     """Parse a query string into a dict.
 
     Query string parameters are assumed to use standard form-encoding. Only
-    parameters with values are parsed. for example, given 'foo=bar&flag',
+    parameters with values are returned. For example, given 'foo=bar&flag',
     this function would ignore 'flag' unless the `keep_blank_qs_values` option
     is set.
 
@@ -269,8 +270,16 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
 
     Args:
         query_string (str): The query string to parse.
-        keep_blank_qs_values (bool): If set to ``True``, preserves boolean
-            fields and fields with no content as blank strings.
+        keep_blank_qs_values (bool): Set to ``True`` to return fields even if
+            they do not have a value (default ``False``). For comma-separated
+            values, this option also determines whether or not empty elements
+            in the parsed list are retained.
+        parse_qs_csv: Set to ``False`` in order to disable splitting query
+            parameters on ``,`` (default ``True``). Depending on the user agent,
+            encoding lists as multiple occurrences of the same parameter might
+            be preferable. In this case, setting `parse_qs_csv` to ``False``
+            will cause the framework to treat commas as literal characters in
+            each occurring parameter value.
 
     Returns:
         dict: A dictionary of (*name*, *value*) pairs, one per query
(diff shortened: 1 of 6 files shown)


kgriffs:
Looks like we accidentally had some overlap in effort between #729 and #811. I attempted to combine the two into a new PR with a few tweeks to param naming and docstrings. Everyone please take a look and provide feedback. Thanks!

orcsly:
Looks good, thanks!

jmvrbanac:
:+1:

falcon/util/uri.py
```diff
@@ -309,7 +317,7 @@ def parse_query_string(query_string, keep_blank_qs_values=False):
                 params[k] = [old_value, decode(v)]
 
         else:
-            if ',' in v:
+            if ',' in v and csv:
```
qwesda: could be `csv and ',' in v` so the evaluation can short-circuit

qwesda:
The parameter name `csv` for the `parse_query_string` could have a more descriptive name. `parse_qs_csv` sounds logical to me, since the option general option is called `auto_parse_qs_csv`.

One evaluation in `parse_query_string` can be short-circuited:
[https://github.com/falconry/falcon/pull/820/files#diff-7d2a078ae72702ba816f18a9aa1c48b9R319](https://github.com/falconry/falcon/pull/820/files#diff-7d2a078ae72702ba816f18a9aa1c48b9R319)

Otherwise it looks good.

falcon/util/uri.py
kgriffs: fixed

kgriffs:
@qwesda since the method name already implies we are working with parsing query strings, is it necessary to also include that in the name of the kwarg?

falcon/util/uri.py
qwesda: it's in the wrong order, `',' in v` is evaluated before `csv`. If it should short-circuit it has to be `csv and ',' in v`.

qwesda:
@kgriffs: I just thought something more explicit would be more in line with `keep_blank_qs_values`, which is pretty verbose. Having one verbose param and one very non-verbose seemed weird.

kgriffs:
@qwesda, ah, good point! I'll switch to `parse_qs_csv` and we can see how that looks.

kgriffs:
@qwesda @jmvrbanac @orcsly I think this is ready for final review.

qwesda:
@kgriffs looks ok

orcsly:
Yup looks good.  Thanks!

PR #821, opened 2016-06-09
feat(Response): Add context property to the Response class

kgriffs:
Add a context property to the Response class, to mirror the context property that is already available on the Request class.

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -83,12 +83,26 @@ class Response(object):
             blocks as byte strings. Falcon will use *wsgi.file_wrapper*, if
             provided by the WSGI server, in order to efficiently serve
             file-like objects.
-
         stream_len (int): Expected length of `stream`. If `stream` is set,
             but `stream_len` is not, Falcon will not supply a
             Content-Length header to the WSGI server. Consequently, the
             server may choose to use chunked encoding or one of the
             other strategies suggested by PEP-3333.
+        context (dict): Dictionary to hold any data about the response which is
+            specific to your app. Falcon itself will not interact with this
+            attribute after it has been initialized.
+        context_type (class): Class variable that determines the factory or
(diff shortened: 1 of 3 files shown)


falcon/response.py
```diff
@@ -83,12 +83,26 @@ class Response(object):
...
             Content-Length header to the WSGI server. Consequently, the
             server may choose to use chunked encoding or one of the
             other strategies suggested by PEP-3333.
+        context (dict): Dictionary to hold any data about the response which is
+            specific to your app. Falcon itself will not interact with this
+            attribute after it has been initialized.
+        context_type (class): Class variable that determines the factory or
+            type to use for initializing the `context` attribute. By default,
+            the framework will instantiate standard ``dict`` objects. However,
+            You may override this behavior by creating a custom child class of
```
ueg1990: *you

falcon/response.py
```diff
@@ -114,6 +132,12 @@ def __init__(self):
         self.stream = None
         self.stream_len = None
 
+        if self.context_type is None:
+            # PERF(kgriffs): The literal syntax is more efficient than dict().
+            self.context = {}
+        else:
+            self.context = self.context_type()
+
```
ueg1990: self.context = {} if self.context_type is None or self.context_type() (more compact?)

falcon/response.py
kgriffs: fixed

falcon/response.py
kgriffs: More compact, but still that is a pretty long line and I think it may actually be a little less readable than the mult-line version.

fxfitz: I personally am a fan of using ternary operators, but I've met so many people who are against it, so now I avoid them. +1 for keeping it as is since it is easier to read.

fxfitz:
LGTM

PR #823, opened 2016-06-10
doc(Request.client_prefers): Note significance of media_types ordering

kgriffs:


falcon/request.py
```diff
@@ -636,7 +636,8 @@ def client_prefers(self, media_types):
         Args:
             media_types (iterable of str): One or more Internet media types
                 from which to choose the client's preferred type. This value
-                **must** be an iterable collection of strings.
+                **must** be an iterable collection of strings. The iterable
+                should be sorted in the order of increasing desirability.
```
jmvrbanac: order of increasing or decreasing desirability? I would think decreasing as you would want the preferred at the front of the iterable. Thoughts?

kgriffs: That would be more intuitive... :) The problem is, this is a leaky abstraction based on python-mimeparse. I suppose we could reverse the list before passing to `best_match()`. What do you think?

jmvrbanac: That would be cool

kgriffs: OK, I can do that. And modify the docstring to suit. I suppose this would qualify as a breaking change though. We'll have to hold this one out for the next major version bump.

PR #824, opened 2016-06-10
feat(testing.Result): Default to UTF-8 when charset is missing

kgriffs:
The requests library attempts to auto-detect the charset using chardet, but we just assume UTF-8 to avoid having to pull in another dependency.

### DIFF
--- a/falcon/testing/test_case.py
+++ b/falcon/testing/test_case.py
@@ -55,8 +55,8 @@ class Result(object):
             response body was empty.
         text (str): Decoded response body of type ``unicode``
             under Python 2.6 and 2.7, and of type ``str`` otherwise.
-            Raises an error if the response encoding can not be
-            determined.
+            If the content type does not specify an encoding, UTF-8 is
+            assumed.
         json (dict): Deserialized JSON body. Raises an error if the
             response is not JSON.
     """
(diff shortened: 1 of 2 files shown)


falcon/testing/test_case.py
```diff
@@ -95,10 +95,11 @@ def text(self):
...
             else:
                 if self.encoding is None:
-                    msg = 'Response did not specify a content encoding'
-                    raise RuntimeError(msg)
+                    encoding = 'UTF-8'
+                else:
+                    encoding = self.encoding
 
-                self._text = self.content.decode(self.encoding)
+                self._text = self.content.decode(encoding)
```
ueg1990: you can replace your changes with:
self._text = self.content.decode(self.encoding or 'UTF-8')

jmvrbanac:
This looks fine... I'm guessing from the changes, if someone specifies a blank string for encoding then we should attempt to try that. If that's not the preferred behavior, we might get a slight performance bump by looking into @ueg1990's suggestion. Thoughts?

kgriffs:
@jmvrbanac Practically speaking, we shouldn't ever get a blank string since that would require the app to specify `charset=""` or similar in `Content-Type`, which doesn't seem very likely. Re performance, I'm not too concerned in this case since it is for testing, and any perf benefit would be negligible anyway. Practically speaking, I think this just boils down to a question of style. I can change it if everyone likes the inline style, just let me know what you think.

PR #829, opened 2016-06-19
Add error class for HTTP 414  URI Too Long

Vesuvium:
As mentioned in #561, some status codes do not have an error class. This PR adds a class for HTTP 414.

PR #834, opened 2016-06-26
feat: add missing HTTP status codes

wssbck:


ueg1990:
Assuming all these status codes are accepted for falcon by @kgriffs and @jmvrbanac, maybe you can define there classes and add tests for them too like here:
https://github.com/falconry/falcon/pull/829/files

wssbck:
@ueg1990 Sure, I just need to know which status codes will be accepted.

falcon/status_codes.py
```diff
@@ -90,8 +96,14 @@
 HTTP_EXPECTATION_FAILED = HTTP_417
 HTTP_418 = "418 I'm a teapot"
 HTTP_IM_A_TEAPOT = HTTP_418
+HTTP_421 = '421 Misderected Request'
```
jeremylow: Typo: should be `Misd*i*rected` rather than `Misderected`

falcon/status_codes.py
```diff
@@ -115,8 +131,18 @@
...
+HTTP_VARIANT_ALSO_NEGOTIATES = HTTP_506
+HTTP_507 = '507 Insufficient Storage'
+HTTP_INSUFFICIENT_STORAGE = HTTP_507
+HTTP_508 = '508 Loop Detected'
+HTTP_LOOP_DETECTED = HTTP_508
+HTTP_510 = '510 Not Extended'
+HTTP_NOT_EXTENDED = HTTP_510
 HTTP_511 = '511 Network Authentication Required'
 HTTP_NETWORK_AUTHENTICATION_REQUIRED = HTTP_511
+HTTP_599 = '599 Network Connect Timeout Error'
```
ashcrow: nit: I don't think this is part of the official spec.

falcon/status_codes.py
```diff
@@ -100,8 +112,12 @@
 HTTP_TOO_MANY_REQUESTS = HTTP_429
 HTTP_431 = '431 Request Header Fields Too Large'
 HTTP_REQUEST_HEADER_FIELDS_TOO_LARGE = HTTP_431
+HTTP_444 = '444 Connection Closed Without Response'
+HTTP_CONNECTION_CLOSED_WITHOUT_RESPONSE = HTTP_444
 HTTP_451 = '451 Unavailable For Legal Reasons'
 HTTP_UNAVAILABLE_FOR_LEGAL_REASONS = HTTP_451
+HTTP_499 = '499 Client Closed Request'
```
ashcrow: nit: I don't think this is part of the official spec.

falcon/status_codes.py
```diff
@@ -100,8 +112,12 @@
 HTTP_TOO_MANY_REQUESTS = HTTP_429
 HTTP_431 = '431 Request Header Fields Too Large'
 HTTP_REQUEST_HEADER_FIELDS_TOO_LARGE = HTTP_431
+HTTP_444 = '444 Connection Closed Without Response'
```
ashcrow: nit: I don't think this is part of the official spec.

jmvrbanac:
Thanks for the PR @wssbck! I'm good with adding these; however, we really should have tests covering the added definitions to help prevent breaking our contract with users later down the road. We already have a bunch of tests similar to this in `test_httperror.py` and `test_httpstatus.py`.

wssbck:
@jmvrbanac I will add classes and tests, just give me some time :-)

jmvrbanac:
Swapping out the labels. Take as much time as you need. We'll be here :-)

falcon/status_codes.py
```diff
@@ -90,8 +96,14 @@
 HTTP_EXPECTATION_FAILED = HTTP_417
 HTTP_418 = "418 I'm a teapot"
 HTTP_IM_A_TEAPOT = HTTP_418
+HTTP_421 = '421 Misdirected Request'
```
kgriffs: I don't think we need this code, since it deals with connection reuse in HTTP/2, which is a web sever (not application) concern.

falcon/status_codes.py
```diff
@@ -115,6 +127,14 @@
 HTTP_GATEWAY_TIMEOUT = HTTP_504
 HTTP_505 = '505 HTTP Version not supported'
 HTTP_HTTP_VERSION_NOT_SUPPORTED = HTTP_505
+HTTP_506 = '506 Variant Also Negotiates'
```
kgriffs: 506 is from a very old experimental RFC. Are there any examples that demonstrate modern usage of this status? Otherwise I'm inclined to say it doesn't need to be included in the framework.

falcon/status_codes.py
```diff
@@ -115,6 +127,14 @@
...
+HTTP_VARIANT_ALSO_NEGOTIATES = HTTP_506
+HTTP_507 = '507 Insufficient Storage'
+HTTP_INSUFFICIENT_STORAGE = HTTP_507
+HTTP_508 = '508 Loop Detected'
+HTTP_LOOP_DETECTED = HTTP_508
+HTTP_510 = '510 Not Extended'
```
kgriffs: I'm hesitant to include 510, since it is also from a very old experimental RFC to define an extension mechanism for the HTTP protocol itself.

kgriffs:
I'm OK with adding the WebDAV codes. But I'm not so sure about the other three (see inline comments).

wssbck:
@kgriffs Will it be OK if, for testing, I run:

`tox -e py26,py27,py35`

instead of:

`tox -e py26,py27,py34`

(Testing against Python 3.5 instead of 3.4)

kgriffs:
@wssbck that's fine, although you won't be able to check coverage. But codecov should catch any issues.

wssbck:
@kgriffs I have finally managed to push additional error classes and tests. Would this be enough? The PR is also rebased on recent master. Additionally, I could implement more error classes for status codes that were already there, but I think it should be a separate PR.

wssbck:
@jmvrbanac @ueg1990 Please have a look as well :) Thanks!

wssbck:
@kgriffs Done, I did a bit of mindless copypaste from 429 :-)

PR #813, opened 2016-06-02
refactor: Use DDT for uri_templates tests. (#178)

sherwu:
I thought adding ddt to some tests would be a good starter project to get me acquainted to this codebase.

PR #848, opened 2016-07-19
refactor: Use DDT for uri_templates tests. 

kgriffs:


kgriffs:
[rebased #813 and applied pep8 fix]

PR #849, opened 2016-07-20
chore(adds print-api-routes): new utility to print routes

ashcrow:
While doing some documentation work I realized it would be nice to have a quick script that dumps out route information and wrote this. I think it may be helpful for other folks as well.

### DIFF
--- a/falcon/cmd/print_routes.py
+++ b/falcon/cmd/print_routes.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python
+# Copyright 2013 by Rackspace Hosting, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


kgriffs:
@ashcrow Good idea! What do you think about moving this under falcon/cmd and adding it as a `'console_script'` entry point? Up to this point, the tools directory has only been used for things related to building and packaging of the framework.

Regardless, it would be great to have some tests for this as well. Thanks!

ashcrow:
@kgriffs Sounds good to me. Being part of the codebase proper I agree tests make sense and will add some :smile:

yohanboniface:
I'd say it's a good candidate for a plugin.

ashcrow:
:arrow_up: Moved to a command and added a few tests.

jmvrbanac:
:+1: LGTM

kgriffs:
Nice work, thanks!

Issue #16, opened 2013-01-23
Support pytest as an option in falcon.testing


kgriffs:
Magic assertions, oh my!

yohanboniface:
I may work on this, if it's still on track.
Two questions:
- do you agree on keeping pytest facilities as a third-party plugin (as in https://github.com/yohanboniface/pytest-falcon) ?
- do you feel comfortable with this kind of function based tests: https://github.com/etalab/ban/blob/master/ban/tests/http/test_api.py?

Basically, I think such move should take two steps:
- switch to pytest as test runner; but keeping tests as they are
- moving to function based tests and `assert` based assertions little by little

Thoughts?

butla:
I totally think that testing without test classes (so essentially easy support for pytest) should be supported. @yohanboniface pytest-falcon looks cool, but I went with something really minimalistic and unopinionated with my project (with the thought of merging it into Falcon):

```
from falcon.testing import StartResponseMock, create_environ

def simulate_falcon_request(api, path='/', encoding=None, **kwargs):
    """Simulates a request to a `falcon.API`.

    Args:
        api (falcon.API): API object to send the request to.
        path (str, optional): The path to request.
        encoding (str, optional): If this is set to a character encoding,
            such as 'utf-8', `simulate_falcon_request` will assume the
            response is a single byte string, and will decode it as the
            result of the request, rather than simply returning the
            standard WSGI iterable.
        kwargs (optional): Same as those defined for
            `falcon.testing.create_environ`.
    """
    resp_headers = StartResponseMock()
    result = api(
        create_environ(path=path, **kwargs),
        resp_headers)

    final_result = result

    if encoding is not None:
        if result:
            final_result = result[0].decode(encoding)
        else:
            final_result = ''

    return final_result, resp_headers
```
With the function above you can create helpers or partial functions that have some common defaults (encoding='utf-8', some Authorization header, etc.) and decrease the amount of code in tests. Those helpers could also go upstream.

I would also want to create a documentation page about testing including a few useful examples, because the current docs on readthedocs left me scratching my head for a while and going through the code when I was starting with Falcon. It would showcase both the old style of testing and the one I'd introduce.

But I need to know if there'll be interest in merging my changes. @kgriffs What do you say?

kgriffs:
Circling back on this. Something along the lines of what @butla proposed would be cool, albeit using the new requests-like testing framework. Should be pretty straightforward to make some free functions and then have class methods that wrap them for backwards compat.

[...]

PR #859, opened 2016-08-11
feat(testing): Add free functions for simulating WSGI requests

kgriffs:
Add free simulate_request, etc. functions that can be used without
requiring a unittest-style OO/inheritance model. Most notably, this
makes it easier to use pytest's more functional style when testing
falcon-based apps.

As part of this work, refactor the existing WSGI simulation logic
to layer on top of the free functions. Don't break backwards
compatibility.

Partially implements #16

### DIFF
--- a/doc/api/testing.rst
+++ b/doc/api/testing.rst
@@ -3,20 +3,11 @@
 Testing
 =======
 
-.. autoclass:: falcon.testing.TestCase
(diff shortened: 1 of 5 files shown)


jmvrbanac:
:+1:

fxfitz:
LGTM

PR #862, opened 2016-08-12
feat(testing): Add TestClient class and a pytest example

kgriffs:
Add a contextual wrapper for the simulate_* methods and a pytest
example demonstrating its use.

Also deprecate "api" in favor of "app" in TestCase, since these utils
can work with any WSGI app (not just those that express HTTP APIs). This
terminology makes TestCase more consistent with the new simulate_*
functions and with TestClient.

Finally, deprecate "api_class" in TestCase and update the docstring
for the testing module to demonstrate a less hacky alternative for
controlling the api instance type. As a bonus, this avoids having to
try and alias api_class (a class variable).

Partially implements #16

### DIFF
--- a/falcon/testing/__init__.py
+++ b/falcon/testing/__init__.py
@@ -24,13 +24,17 @@ supported::
     # -----------------------------------------------------------------
 
     from falcon import testing
-    from myapp import api
+    import myapp
 
 
     class MyTestCase(testing.TestCase):
         def setUp(self):
             super(MyTestCase, self).setUp()
-            self.app = api.create()
+
+            # Assume the hypothetical `myapp` package has a
+            # function called `create()` to initialize and
+            # return a `falcon.API` instance.
+            self.app = myapp.create()
 
 
     class TestMyApp(MyTestCase):
@@ -48,12 +52,15 @@ supported::
     from falcon import testing
     import pytest
 
-    from myapp import api
+    import myapp
 
 
     @pytest.fixture(scope='module')
     def client():
-        return testing.TestClient(api.create())
+        # Assume the hypothetical `myapp` package has a
+        # function called `create()` to initialize and
+        # return a `falcon.API` instance.
+        return testing.TestClient(myapp.create())
 
 
     def test_get_message(client):
--- a/falcon/testing/__init__.py
+++ b/falcon/testing/__init__.py
@@ -16,23 +16,51 @@
 
 This package contains various test classes and utility functions to
 support functional testing for both Falcon-based apps and the Falcon
-framework itself::
+framework itself. Both unittest-style and pytest-style tests are
+supported::
+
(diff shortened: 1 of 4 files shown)


falcon/testing/__init__.py
```diff
@@ -16,23 +16,51 @@
 
 This package contains various test classes and utility functions to
 support functional testing for both Falcon-based apps and the Falcon
-framework itself::
+framework itself. Both unittest-style and pytest-style tests are
+supported::
```
kgriffs: Please carefully review the example code below for correctness. Thanks!

falcon/testing/test_case.py
```diff
@@ -38,30 +38,57 @@ class TestCase(unittest.TestCase):
...
-            :py:class:`falcon.API`).
-        api (object): An API instance to target when simulating requests
-            (default: ``self.api_class()``). When testing your
-            application, you will need to overwrite this with your own
-            instance of ``falcon.API``, or use `api_class` to specify a
-            factory method for your application.
+        app (object): A WSGI application to target when simulating
+            requests (default: ``falcon.API()``). When testing your
+            application, you will need to set this to your own instance
+            of ``falcon.API``. For example::
```
kgriffs: Please carefully review the example code below for correctness. Thanks!

falcon/testing/test_case.py
```diff
@@ -38,30 +40,61 @@ class TestCase(unittest.TestCase):
...
+
+                from falcon import testing
+                from myapp import api
+
+
+                class MyTestCase(testing.TestCase):
+                    def setUp(self):
+                        super(MyTestCase, self).setUp()
+                        self.app = api.create()
```
jmvrbanac: Perhaps I'm just missing something, but where is ``create()`` defined? is that just hypothetical for the example?

kgriffs: Just hypothetical... let me try to add a comment or something to avoid confusion. 😄

kgriffs: OK, I tweaked this. Let me know if it makes more sense!

falcon/testing/__init__.py
```diff
@@ -16,23 +16,51 @@
...
+    from myapp import api
+
 
-    class TestMyApp(testing.TestCase):
+    class MyTestCase(testing.TestCase):
         def setUp(self):
-            super(TestMyApp, self).setUp()
-            self.api = app.create_api()
+            super(MyTestCase, self).setUp()
+            self.app = api.create()
```
jmvrbanac: You probably want to add that comment about ``create()`` here as well.

kgriffs: Ah, nice catch! Will do.

kgriffs: Updated

jmvrbanac:
:+1:

PR #863, opened 2016-08-12
fix(HTTPError): Remove asserts for self.has_representation

kgriffs:
Even if an error should not result in a representation being returned
in the body of the HTTP response, it may still be useful to serialize
the error for purposes of logging, without having to check the
representation flag first.


jmvrbanac:
:+1:

PR #869, opened 2016-08-16
fix(util): Don't double-encode percent-encoded URIs

### LINKED ISSUE #687: resp.location double-encodes urlencoded strings

kgriffs:
Detect a URI that is already encoded, and don't re-encode in that case.

Fixes #687

### DIFF
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -60,6 +60,7 @@ def _create_char_encoder(allowed_chars):
 def _create_str_encoder(is_value):
 
     allowed_chars = _UNRESERVED if is_value else _ALL_ALLOWED
+    allowed_chars_plus_percent = allowed_chars + '%'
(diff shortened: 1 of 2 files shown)


kgriffs:
[rebased]

kgriffs:
[rebased]

fxfitz:
LGTM

jmvrbanac:
:+1:

PR #871, opened 2016-08-23
test(mimeparse): Test for the correct version of mimeparse

kgriffs:
Document the difference between python-mimeparse and mimeparse, and
add a test to ensure Linux package maintainers, et al., are pulling
python-mimeparse, not mimeparse.


jmvrbanac:
:+1:

PR #873, opened 2016-08-24
feat(testing): Specify query parameters as a dict in simulate_request()

kgriffs:
As an alternative to specifying a raw query string, also allow for specifying parameters as a dict.

### DIFF
--- a/falcon/testing/client.py
+++ b/falcon/testing/client.py
@@ -37,7 +37,7 @@ import wsgiref.validate
 
 from falcon.testing import helpers
 from falcon.testing.srmock import StartResponseMock
-from falcon.util import CaseInsensitiveDict
+from falcon.util import CaseInsensitiveDict, to_query_str
 
 
 class Result(object):
@@ -122,7 +122,8 @@ class Result(object):
 
 
 def simulate_request(app, method='GET', path='/', query_string=None,
-                     headers=None, body=None, file_wrapper=None):
+                     headers=None, body=None, file_wrapper=None,
+                     params=None, params_csv=True):
         """Simulates a request to a WSGI application.
 
         Performs a request against a WSGI application. Uses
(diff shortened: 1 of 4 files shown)


falcon/testing/client.py
```diff
@@ -134,8 +135,20 @@ def simulate_request(app, method='GET', path='/', query_string=None,
...
+                where each key is a parameter name, and each value is
+                either a ``str`` or something that can be converted
+                into a ``str``, or a list of such values. If a ``list``,
+                the value will be converted to a comma-delimited string
+                of values (e.g., 'thing=1,2,3').
+            params_csv (bool): Set to ``False`` to encode list values
+                in query string params by specifying multiple instances
+                of the parameter (e.g., 'thing=1&thing=2&thing=3').
+                Otherwise, parameters will be encoded as comma-separated
+                values (e.g., 'thing=1,2,3'). Defaults to ``True``.
```
kgriffs: At some point we may want to go throughout the framework and switch the defaults to use form-based encoding instead of csv encoding. Keeping the default to `True` for now for the sake of consistency.

falcon/testing/helpers.py
```diff
@@ -120,6 +120,9 @@ def create_environ(path='/', query_string='', protocol='HTTP/1.1',
 
     """
 
+    if query_string and query_string.startswith('?'):
```
kgriffs: Got burned by this while testing, so added the check for posterity's sake. :)

falcon/testing/client.py
```diff
@@ -167,6 +180,13 @@ def simulate_request(app, method='GET', path='/', query_string=None,
                 'query_string parameter instead.'
             )
 
+        if query_string is None:
+            query_string = to_query_str(
+                params,
+                comma_delimited_lists=params_csv,
+                prefix=False,
```
kgriffs: `prefix` is a new parameter also implemented in this PR.

fxfitz:
LGTM

jmvrbanac:
:+1:

PR #874, opened 2016-08-25
chore(tox): Switch test runner from nose to pytest

kgriffs:
Use pytest instead of nose since it has a better ecosystem and will
allow us to write pytest-style tests in the future.


tox.ini
```diff
@@ -59,17 +62,6 @@ deps = {[with-debug-tools]deps}
 [testenv:py34_debug]
 deps = {[with-debug-tools]deps}
 
-[testenv:py27_profile]
```
kgriffs: Since pytest-profiling is buggy, and since this testenv is rarely used, I just removed it.

PR #877, opened 2016-08-26
fix(API): Improve and optimize execution of middleware methods

### LINKED ISSUE #635: Raising errors from resource methods doesn't change status for middleware

kgriffs:
Rework the execution of middleware, responders, and related error
handling in API. This was necessary in order to fix a couple of
bugs (raising errors from resource methods doesn't change status for
middleware, middleware processing isn't always continued when an
instance of HTTPError or HTTPStatus is raised). Along the way,
minor optimizations were made to reduce CPU time in this portion
of the code (offsets additional try..except setup overhead).

Fixes #635

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -138,7 +138,11 @@ class API(object):
         self._serialize_error = helpers.default_serialize_error
         self.req_options = RequestOptions()
 
-    def __call__(self, env, start_response):
+        # NOTE(kgriffs): Add default error handlers
+        self.add_error_handler(falcon.HTTPError, self._http_error_handler)
+        self.add_error_handler(falcon.HTTPStatus, self._http_status_handler)
+
+    def __call__(self, env, start_response):  # noqa: C901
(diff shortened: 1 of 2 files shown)


jmvrbanac:
:+1:

PR #881, opened 2016-08-30
chore: Remove portion of build.sh that deals with binary Darwin wheels

kgriffs:
Don't build binary wheels for Darwin (OS X) since it isn't really
necessary to have cythonized code in a development environment, and
using the pyenv-installed Pythons results in compatibility issues
that just aren't worth it in any case.


jmvrbanac:
:+1:

PR #882, opened 2016-08-30
Doctor the docs

kgriffs:
Usually these would be broken out into individual PRs, but in the interest of getting 1.1 out the door, I combined several patches into an omnibus PR:

doc(install): Reconcile install.rst and README.rst
doc: Bump version string to 1.1
doc: Add "Who's Using Falcon" section to index.rst, README.rst
doc: Update mailing list, IRC info
doc: Correct docs path in CONTRIBUTING.md


README.rst
```diff
@@ -122,31 +145,70 @@ these issues by setting additional Clang C compiler flags as follows:
...
+ones out there, but anything that can load a WSGI app will do.
 
 .. code:: bash
 
+    $ pip install [gunicorn|uwsgi]
+
+Source Code
+-----------
+
+Falcon `lives on GitHub <https://github.com/racker/falcon>`_, making the
```
jmvrbanac: The address should change

docs/conf.py
```diff
@@ -62,9 +62,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '1.0'
+version = '1.1'
 # The full version, including alpha/beta/rc tags.
-release = '1.0.0'
+release = '1.1.0'
```
jmvrbanac: We should take a look at using ``bumpversion``. I started using it for a couple projects, it's quite nice and doesn't add any deps to your package.

docs/user/install.rst
```diff
@@ -65,26 +65,33 @@ these issues by setting additional Clang C compiler flags as follows:
...
 .. code:: bash
 
-    $ pip install gevent [gunicorn|uwsgi]
-
+    $ pip install [gunicorn|uwsgi]
 
 Source Code
 -----------
 
 Falcon `lives on GitHub <https://github.com/racker/falcon>`_, making the
```
jmvrbanac: I know this isn't under changes in review, but we should probably change this one as well.

docs/conf.py
kgriffs: Yeah, that could be helpful for sure. We really should be able to get to a fully scripted release process. All we need is someone to volunteer to be the official Falcon release manager. 😉

AndrewBurdyug: Hi! 

I use in my project the 'VERSION' file and when I want make a new release,
I just increment version number in 'VERSION' file and run [make newrelease](https://github.com/AndrewBurdyug/makechat/blob/master/Makefile#L127-L151), that's all! 

All rest(coverage, run tests, update docs) doing related services(TravisCI, CircleCI, ReadTheDocs, Codecov) which have been connected to github. 

Maybe this approach help you implement your own release management scenario ;)

jmvrbanac:
:+1:

docs/conf.py
kgriffs: @AndrewBurdyug oh, cool, thanks for sharing!

PR #883, opened 2016-08-31
doc(testing.client): Document file_wrapper kwarg for simulate_get

kgriffs:
A developer wishing to test high-performance file transmission may wish
to pass a mock for wsgi.file_wrapper to simulate_get(). This kwarg is
not documented for other simulate_*() functions since they are unlikely
to use it.


jmvrbanac:
:+1:

PR #889, opened 2016-09-06
feat(Request): Expose the URI template that was matched for the request

### LINKED ISSUE #619: Expose routing metadata
Per a [discussion][1] on the mailing list, it can be useful to introspect the URI template that is associated with a given resource. One way to implement this would be to check for an attribute and fill it in as part of the `add_route` logic. For example:

```python
class Thing(object):
    def __init__(self):
        self.uri_templates = []

    def on_get(self, req, resp):
        pass

# ...

def add_route(self, uri_template, resource):

    # ...

    if hasattr(resource, 'uri_templates'):
        resource.uri_templates.append(uri_template)

    # ...
``` 

We could also do this:

```python
    if hasattr(resource, 'uri_templates'):
        resource.uri_templates.append(uri_template)
    else:
        resource.uri_templates = [uri_template]
```

But that would break in the case that `Thing` defined `__slots__`.

Another option would be to store the URI template in the routing tree and then set an attribute on each Request instance:

```python
def __call__(self, env, start_response):

    # ...

    responder, params, resource, uri_template = self._get_responder(req)
    
    req = self._request_type(env, options=self.req_options, 
                             uri_template=uri_template)

    # ...
```

Either way, this would provide a way for middleware and hooks to key off the URI template to perform additional actions.

[1]: http://librelist.com/browser//falcon/2015/7/28/getting-the-url-template-when-servicing-a-request/

kgriffs:
This has been requested by several members of the community who would
like a way to access the URI template in middleware, etc. (e.g., for
logging).

Fixes #619

### DIFF
--- a/docs/api/routing.rst
+++ b/docs/api/routing.rst
@@ -24,14 +24,15 @@ A custom router is any class that implements the following interface:
             """
 
         def find(self, uri):
-            """Search for a route that matches the given URI.
+            """Search for a route that matches the given partial URI.
 
             Args:
-                uri (str): Request URI to match to a route.
+                uri(str): The requested path to route
 
             Returns:
-                tuple: A 3-member tuple composed of (resource, method_map, params)
-                    or ``None`` if no route is found.
+                tuple: A 4-member tuple composed of (resource, method_map,
+                    params, uri_template), or ``None`` if no route matches
+                    the requested path
             """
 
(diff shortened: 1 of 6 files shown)


jmvrbanac:
:+1:

fxfitz:
LGTM2!!!

PR #891, opened 2016-09-08
feat(util): Make it easier to reference uri module and TimezoneGMT class

kgriffs:
Previously "from falcon import uri" worked but "import falcon.uri" did
not. TimezoneGMT is already available from the front-door falcon
module, but the documentation referenced falcon.util.TimezoneGTM.

Fix this up and update the docs to suit.


jmvrbanac:
:+1: LGTM

PR #892, opened 2016-09-08
doc: Unindent additional Return lines to work around napolean bug

kgriffs:
In a google-style docstring, this is problematic:

    Returns:
        str: Some really long description that requires more
            than one line

Since the additional text (i.e., "than one line") is indented,
napolean gets confused and incorrectly formats the description. The
fix is to align subsequent lines with the first:

    Returns:
        str: Some really long description that requires more
        than one line.

Along the way I also added missing return types as needed.

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -482,7 +482,7 @@ class API(object):
             req: The request object.
 
         Returns:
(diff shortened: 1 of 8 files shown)


jmvrbanac:
LGTM :+1:

Issue #895, opened 2016-09-12
405 errros are not passed through custom error handlers


ksonj:
I tried writing a custom error handler for `falcon.HTTPError` (to control the representation of failures) and realized that 405 errors were never passed to the handler.

This is probably due to [`method_not_allowed`](https://github.com/falconry/falcon/blob/master/falcon/responders.py#L43) not raising an exception (like e.g. [`path_not_found`](https://github.com/falconry/falcon/blob/master/falcon/responders.py#L23)) but modifying the response directly.

Is that intended behavior and if so, what is the motivation?

My fix is simply raising `HTTPMethodNotAllowed`:

    def create_method_not_allowed(allowed_methods):
        """Creates a responder for "405 Method Not Allowed"
    
        Args:
            allowed_methods: A list of HTTP methods (uppercase) that should be
                returned in the Allow header.
    
        """
        def method_not_allowed(req, resp, **kwargs):
            """Raise 405 HTTPNotAllowed error"""
            raise HTTPMethodNotAllowed(allowed_methods)
    
        return method_not_allowed

 Everything looks good so far.

PR #896, opened 2016-09-12
Raise HTTPMethodNotAllowed instead of modifying response directly

ksonj:
Fixes this issue: https://github.com/falconry/falcon/issues/895

405 errors are never passed to registered error handlers, because the na-responder does not raise an exception, but modifies the response directly.


kgriffs:
Thanks, this looks good! Would you mind rebasing in preparation for a merge?

PR #898, opened 2016-09-14
fix(Request): Improve various related aspects of stream handling

### LINKED ISSUE #407: Request hangs if method is GET and content-type is application/x-www-form-urlencoded
For some reason I have some software that sends content-type=application/x-www-form-urlencoded with GET requests.  It shouldn't, but that's not important.

Such requests hang, whereas the content-type should just be ignored.

The reason appears to be in request.py line 258

        if (self.content_type is not None and
                'application/x-www-form-urlencoded' in self.content_type):
            self._parse_form_urlencoded()

I think it probably should make the above conditional upon either the request METHOD being PUT or POST or PATCH maybe.  Or some other solution.

kgriffs:
* Add a new bounded_stream property that can be used for more
  predictable behavior vs. stream, albeit with a slight performance
  overhead (the app developer is free to decide whether or not to
  use it).
* Only automatically consume the incoming stream on POST requests,
  since that is the only time form-encoded params should be included
  in the body (vs. the query string). This guards against unexpected
  side-effects caused by misbehaving or even malicious clients.
* Check Content-Length to ensure a body is expected, before attempting
  to parse form-encoded POSTs. Also pass the Content-Length to
  stream.read as an extra safety measure to guard against differences
  in WSGI input read() behavior.
* Improve the documentation surrounding all of these behaviors.

Fixes #407

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -50,10 +50,6 @@ TRUE_STRINGS = ('true', 'True', 'yes', '1')
 FALSE_STRINGS = ('false', 'False', 'no', '0')
 WSGI_CONTENT_HEADERS = ('CONTENT_TYPE', 'CONTENT_LENGTH')
 
-
-_maybe_wrap_wsgi_stream = True
-
-
 # PERF(kgriffs): Avoid an extra namespace lookup when using these functions
 strptime = datetime.strptime
 now = datetime.now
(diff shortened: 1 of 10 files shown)


falcon/request.py
```diff
@@ -305,21 +350,30 @@ def __init__(self, env, options=None):
...
         # normalizing semantics between, e.g., gunicorn and wsgiref.
-        if _maybe_wrap_wsgi_stream:
-            if isinstance(self.stream, (NativeStream, InputWrapper,)):
-                self._wrap_stream()
-            else:
-                # PERF(kgriffs): If self.stream does not need to be wrapped
-                # this time, it never needs to be wrapped since the server
-                # will continue using the same type for wsgi.input.
-                _maybe_wrap_wsgi_stream = False
+        if not Request._wsgi_input_type_known:
```
kgriffs: A bit of an optimization here, so we only ever have to check once.

falcon/testing/resource.py
```diff
@@ -108,11 +109,20 @@ def __init__(self, status=None, body=None, json=None, headers=None):
...
+        self.captured_kwargs = None
+
     @falcon.before(capture_responder_args)
     @falcon.before(set_resp_defaults)
     def on_get(self, req, resp, **kwargs):
         pass
 
+    @falcon.before(capture_responder_args)
+    @falcon.before(set_resp_defaults)
+    def on_post(self, req, resp, **kwargs):
```
kgriffs: Needed this since the PostQueryParams tests are now required to use POST instead of GET.

falcon/request.py
```diff
@@ -305,21 +350,30 @@ def __init__(self, env, options=None):
...
+            self._bounded_stream = None  # Lazy wrapping
 
         # PERF(kgriffs): Technically, we should spend a few more
         # cycles and parse the content type for real, but
         # this heuristic will work virtually all the time.
         if (self.options.auto_parse_form_urlencoded and
                 self.content_type is not None and
-                'application/x-www-form-urlencoded' in self.content_type):
+                'application/x-www-form-urlencoded' in self.content_type and
+                self.method == 'POST'):
```
jmvrbanac: ehhh... If you did that, it would also apply for DELETE and HEAD as well, so I think it's fine.

PR #903, opened 2016-09-19
doc(Request): Expand and correct documentation for set_cookie()

kgriffs:
Correct outdated information in the set_cookie() docstring, as well
as expound on certain key points and provide references to relevant
sections of RFC 6265.

### DIFF
--- a/docs/ext/rfc.py
+++ b/docs/ext/rfc.py
@@ -24,7 +24,7 @@ formatted like this::
 import re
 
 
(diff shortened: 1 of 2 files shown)


docs/ext/rfc.py
```diff
@@ -24,7 +24,7 @@
 import re
 
 
-RFC_PATTERN = re.compile('\RFC (\d{4}), Section ([\d\.]+)')
```
kgriffs: Slightly off-topic, but found this typo while working on these docs, and went ahead and fixed it.

PR #904, opened 2016-09-20
fix(util): Fix falcon.uri import

kgriffs:
The previous attempt at fixing this failed, and so do something a bit more evil. We'll remove this later in 2.0 once util goes away and everything is just implemented under the falcon package directly.

### DIFF
--- a/docs/api/util.rst
+++ b/docs/api/util.rst
@@ -6,12 +6,6 @@ Utilities
 URI Functions
 -------------
 
-.. NOTE(kgriffs): Due to the way we have been hoisting uri into falcon,
-.. sphinx can't import just falcon.uri (root cause TBD). Whether or not
-.. the way we are hoisting things is legit (hint: probably not, this
-.. is something to address in Falcon 2.0), sphinx can't handle it
-.. by default so we have a custom extension to fix things up.
-
 .. automodule:: falcon.uri
     :members: encode, encode_value, decode, parse_host,
         parse_query_string, unquote_string
--- a/falcon/uri.py
+++ b/falcon/uri.py
@@ -1,0 +1,16 @@
+"""URI utilities.
+
+This module provides utility functions to parse, encode, decode, and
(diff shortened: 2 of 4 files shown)


docs/api/util.rst
```diff
@@ -6,12 +6,6 @@ Utilities
 URI Functions
 -------------
 
-.. NOTE(kgriffs): Due to the way we have been hoisting uri into falcon,
-.. sphinx can't import just falcon.uri (root cause TBD). Whether or not
-.. the way we are hoisting things is legit (hint: probably not, this
-.. is something to address in Falcon 2.0), sphinx can't handle it
-.. by default so we have a custom extension to fix things up.
-
 .. automodule:: falcon.uri
```
kgriffs: This was not working before this patch (the section in the docs was blank due to "cannot import falcon.uri").

tox.ini
```diff
@@ -187,4 +187,4 @@ commands = falcon-bench []
 deps = -r{toxinidir}/tools/doc-requires
 whitelist_externals = rm
 commands =
-    sphinx-build -j4 -E -b html docs docs/_build/html
+    sphinx-build -j4 -E -b html docs docs/_build/html []
```
kgriffs: This came in handy while diagnosing the problem

falcon/uri.py
```diff
@@ -0,0 +1,16 @@
...
+    from falcon import uri
+
+    name, port = uri.parse_host('example.org:8080')
+"""
+
+# NOTE(kgriffs): This module exists to make "import falcon.uri"
+# work. Eventually we will remove the util module and flatten the
+# falcon namespace, but in the meantime...
+
+from falcon.util.uri import *  # NOQA
```
jmvrbanac: All the wildcard imports!

docs/api/util.rst
jmvrbanac: :+1:

PR #905, opened 2016-09-21
perf(uri): Replace if...else with try...except in the decode function

kgriffs:
Most of the time the token will be found in the dict, so let's optimize
for that and avoid an extra membership check (if token in _HEX_TO_BYTE).


kgriffs:
Merging based on fxfitz' approval

PR #908, opened 2016-09-22
fix(Request): Also allow PUT and PATCH parsing of x-www-form-urlencoded

kgriffs:


abompard:
Could you also allow the `DELETE` method?

Issue #835, opened 2016-06-26
Better way to get the port from the request?


EmilStenstrom:
I'm in the process of porting an old bottle app to falcon. The first thing my app does is generate some api documentation and returns it as HTML. In the documentation I have clickable links that makes it easy to explore the api. Whether on localhost or on production I want the links to work, so I generate a "site" parameter that I prefix all links with. 

In bottle this was: 

``` python
site = "%s://%s" % (request.urlparts.scheme, request.urlparts.netloc)
```

In falcon the easiest way I can find to do this is

``` python
site = "%s://%s" % (request.protocol, request.headers["HOST"])
```

**Having to rely on manually reading the HOST header feels very inelegant to me**. Note that this is not the same as `request.host` which returns the host without the port number.

I see a couple of ways this can be improved:
1. Make `request.host` include the port in the response. Backwards incompatible, even though everyone that previously relied on port numbers probably didn't use request.host anyway.
2. Add new property `request.host_with_port`.
3. Add new property `request.port` and let people that need the host + port combination write their own utility that combine it with the host.

I'm willing to write a pull request for this, but would love some feedback on which of the versions you would prefer.

kgriffs:
Hi @EmilStenstrom, thanks for bringing this up. To improve parity with other libraries/frameworks, I propose making all three of the following changes:

* Add a new property, `Request.port`.
* Add a new property, `Request.netloc` that will contain the host as well as the port (if present). Note that "netloc" is consistent with the standard library's `urlparse()`. The value can be taken directly from the WSGI variables 'HTTP_HOST', or 'SERVER_NAME' and 'SERVER_PORT' (use the latter variables when 'HTTP_HOST' is missing. See also: https://www.python.org/dev/peps/pep-3333/#url-reconstruction).
* Add a new property, `Request.scheme` and convert `Request.protocol` into a deprecated alias of the new property (honestly, I don't remember why "protocol" was used in the first place, since that isn't the standard term). 

Does that sound reasonable?

EmilStenstrom:
Hi! All your suggestions sound good to me. Thanks for listening!

EmilStenstrom:
Thanks! ❤️

PR #917, opened 2016-09-29
attributes on request

navyad:
Related to issue [https://github.com/falconry/falcon/issues/835](https://github.com/falconry/falcon/issues/835)


falcon/request.py
```diff
@@ -683,6 +692,14 @@ def access_route(self):
     def remote_addr(self):
         return self.env.get('REMOTE_ADDR')
 
+    @property
+    def port(self):
+        return self.env.get('SERVER_PORT')
+
+    @property
+    def netloc(self):
+        return self.host + ':' + self.port
```
kgriffs: Usually the port will be derived from `HTTP_HOST` and may be equal to `None` if the client did not specify one explicitly (depending on how we handle that case).

falcon/request.py
```diff
@@ -69,9 +69,11 @@ class Request(object):
         options (dict): Set of global options passed from the API handler.
 
     Attributes:
-        protocol (str): Either 'http' or 'https'.
+        protocol (str): Either 'http' or 'https' Depreciated.
```
navyad: Hi @kgriffs  
description changed [https://github.com/falconry/falcon/pull/917/commits/9c8d34188d147482f44a484a7287d593221f5441](https://github.com/falconry/falcon/pull/917/commits/9c8d34188d147482f44a484a7287d593221f5441)

falcon/request.py
```diff
@@ -69,9 +69,11 @@ class Request(object):
...
+        scheme (str): Either 'http' or 'https'.
         method (str): HTTP method requested (e.g., 'GET', 'POST', etc.)
         host (str): Hostname requested by the client
+        port (str): Port used for request
```
navyad: updated [https://github.com/falconry/falcon/pull/917/commits/47c73ee0223fa376d48ccc32445ab0405a7bdf4b](https://github.com/falconry/falcon/pull/917/commits/47c73ee0223fa376d48ccc32445ab0405a7bdf4b)

falcon/request.py
```diff
@@ -537,10 +539,17 @@ def app(self):
         return self.env.get('SCRIPT_NAME', '')
 
     @property
-    def protocol(self):
+    def scheme(self):
         return self.env['wsgi.url_scheme']
 
     @property
+    def protocol(self):
```
navyad: Didn't know that earlier, nice one. updated [https://github.com/falconry/falcon/pull/917/commits/d3ce2c40eb2276183e845d9c384b28bfcf694586#diff-8608936f256c7b35254ee64f862001b9R545](https://github.com/falconry/falcon/pull/917/commits/d3ce2c40eb2276183e845d9c384b28bfcf694586#diff-8608936f256c7b35254ee64f862001b9R545)

falcon/request.py
```diff
@@ -537,10 +539,17 @@ def app(self):
...
+        """
+        To be depreciated in favour of scheme
```
navyad: that have been removed, for alias

falcon/request.py
```diff
@@ -683,6 +692,14 @@ def access_route(self):
     def remote_addr(self):
         return self.env.get('REMOTE_ADDR')
 
+    @property
+    def port(self):
```
navyad: In uri attribute, if port is not coming from HTTP_HOST port is taken from SERVER_PORT.
[https://github.com/falconry/falcon/blob/master/falcon/request.py#L553](https://github.com/falconry/falcon/blob/master/falcon/request.py#L553)

falcon/request.py
```diff
@@ -683,6 +692,14 @@ def access_route(self):
...
+        return self.env.get('SERVER_PORT')
+
+    @property
+    def netloc(self):
```
navyad: Thats been done [https://github.com/falconry/falcon/pull/917/commits/106529e5c0b2ca5d26f1470b3e495b85af2e3e62#diff-8608936f256c7b35254ee64f862001b9R696](https://github.com/falconry/falcon/pull/917/commits/106529e5c0b2ca5d26f1470b3e495b85af2e3e62#diff-8608936f256c7b35254ee64f862001b9R696)

falcon/request.py
```diff
@@ -683,6 +687,22 @@ def access_route(self):
...
+        except KeyError:
+            port = self.env.get('SERVER_PORT')
+        return port
+
+    @property
+    def netloc(self):
+        try:
+            return self.env['HTTP_HOST']
+        except KeyError:
+            return self.host + ':' + self.port
```
kgriffs: Should this behave like [request.py#L560](https://github.com/falconry/falcon/blob/master/falcon/request.py#L560)?

falcon/request.py
```diff
@@ -683,6 +687,22 @@ def access_route(self):
     def remote_addr(self):
         return self.env.get('REMOTE_ADDR')
 
+    @property
+    def port(self):
+        try:
+            host_header = self.env['HTTP_HOST']
+            host, port = parse_host(host_header)
+        except KeyError:
+            port = self.env.get('SERVER_PORT')
```
kgriffs: SERVER_PORT will always be present, so `self.env.get('SERVER_PORT')` ===> `self.env['SERVER_PORT']`

tests/test_req_vars.py
```diff
@@ -647,3 +647,68 @@ def _test_error_details(self, headers, attr_name,
         except error_type as ex:
             self.assertEqual(ex.title, title)
             self.assertEqual(ex.description, description)
+
```
kgriffs: It would be cool if you could also add a few HTTP/1.1 tests for the new properties (to ensure those code paths return what we expect).

falcon/request.py
```diff
@@ -683,6 +687,22 @@ def access_route(self):
     def remote_addr(self):
         return self.env.get('REMOTE_ADDR')
 
+    @property
+    def port(self):
+        try:
+            host_header = self.env['HTTP_HOST']
+            host, port = parse_host(host_header)
```
kgriffs: If port isn't present, we should probably return the default port based on the schema, in oder to normalize what would be returned with the `SERVER_PORT` branch. What do you think?

falcon/request.py
navyad: fixed [here](https://github.com/falconry/falcon/pull/917/commits/68c3015eed25cf8a048b3f7a931c34406f6ad492#diff-8608936f256c7b35254ee64f862001b9R696)

falcon/request.py
kgriffs: Since we already know the `HTTP_HOST` key does not exist, the redundant checks in self.host and self.port seem redundant. As an alternative, what do you think about factoring out [this section](https://github.com/falconry/falcon/blob/master/falcon/request.py#L554-L569) from the `uri` property to use as the implementation of `netloc`, then use `netloc` from `uri()` as per below?

```py
@property
def uri(self):
    if self._cached_uri is None:
        env = self.env
        protocol = env['wsgi.url_scheme']

        # PERF: For small numbers of items, '+' is faster
        # than ''.join(...). Concatenation is also generally
        # faster than formatting.
        value = (protocol + '://' +
                 self.netloc +
                 self.app +
                 self.path)

        if self.query_string:
            value = value + '?' + self.query_string

        self._cached_uri = value

    return self._cached_uri
```

tests/test_req_vars.py
kgriffs: Looks like all the new tests have `protocol='HTTP/1.0'`... I'm not seeing any for checking with the default `'HTTP/1.1'` protocol?

falcon/request.py
kgriffs: We can always refactor this in another PR, since it is just an optimization.

tests/test_req_vars.py
kgriffs: I can add this in a followup PR.

PR #924, opened 2016-10-04
chore: Use pytest runner instead of nose collector for "setup.py test"

kgriffs:
See also: http://docs.pytest.org/en/latest/goodpractices.html?highlight=setup#integrating-with-setuptools-python-setup-py-test-pytest-runner

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 dist
 build
 eggs
+.eggs
 parts
 var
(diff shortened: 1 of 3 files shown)


kgriffs:
[Note: merge before cutting 1.1rc1]

fxfitz:
(TIL pytest-runner)

PR #926, opened 2016-10-04
feat(middleware) optional, independent execution of request and response middleware

mattgiles:
This PR implements one way falcon might introduce the option to execute request and response middleware independently.

The current stack of **response** middleware is request-specific, as it is dependent on whether or not **request** middleware raises. This can be obstructive if request middleware raises, e.g. an `HTTPError`, before needed response middleware has been added to the ephemeral response middleware stack.

Solution proposed here preserves current behavior, enabling new behavior with a flag. New behavior is: all request middleware executed in order (current behavior), all resource middleware executed in order if request is successfully routed (current behavior), all response middleware is executed in reverse order (new behavior). In other words: the stacks for all three kinds of middleware (request, resource, and response) are immutable, prepared during api creation, and the same for every request.


mattgiles:
rebased on 1.1.0

PR #937, opened 2016-10-31
fix(router): Allow single character route fields

cag:


tests/test_default_router.py
```diff
@@ -159,6 +159,10 @@ def test_invalid_field_name(self, template):
             ValueError,
             self.router.add_route, template, {}, ResourceWithId(-1))
 
+    def test_single_character_field_name(self):
+        self.router.add_route(
+            '/search?q={q}', {}, ResourceWithId(22))
```
jmvrbanac: Perhaps I'm just overlooking something, but I think we need a assertion here to verify this works as expected.

cag: This changeset is simply to prevent this sort of add_route call with a single character parameter from raising an exception.

There is a longer form of this sort of test, but this is what I reached for first. If it is desired, it can be changed: http://stackoverflow.com/a/20275035/1796894

jmvrbanac: Ahh! Sorry, what I meant was that I think we need to add an assertion that verifies that the route actually works with the single character, not just that it doesn't raise an exception. So something like:
```
self.router.add_route('/item/{q}', {}, ResourceWithId(22))
__, __, params, __ = self.router.find('/item/1234')
self.assertEqual(params, {'q': '1234'})
```
This should make sure the routing actually works with your change.

jmvrbanac: A quick question that just popped in my head. Was it intentional that you're using the field on a query param? Shouldn't the route just be `/search` and then access the query params via `req.params.get('q')`? The change is still relevant, but I just thought I would mention it in case you were looking for a way to get query parameters instead of route parameters.

cag:
Ah, thanks, I see. I'll make the appropriate changes soon. Yes, that is
probably closer to my intention, but I am actually not directly familiar
with falcon, so I plead ignorance here :P

jmvrbanac:
No worries! We all start somewhere. Regardless of the the original reason for the change, the change is still good. Anyhow, we appreciate your contribution!

kgriffs:
Fixes #959

PR #944, opened 2016-11-11
Remove unnecessary content-length

w1z2g3:
The 204 response MUST NOT include a message-body, thus should be no content-length.


w1z2g3:
Turned out I was wrong. See: https://github.com/falconry/falcon/issues/796

Close it. Sorry for the noise.

PR #952, opened 2016-11-22
fix(API): Always set the Vary header when serializing errors

kgriffs:


jmvrbanac:
@kgriffs I was going to merge this, but you originally added the ``needs-revision`` label. Were you looking to add something or was it just related to CI?

kgriffs:
@jmvrbanac I think it was related to CI. Should be good to merge. Thanks!

PR #954, opened 2016-12-02
perf: Json dump speedup

masellfoodpanda:
Import ujson if available. Stop using pretty printing in dumps for a 14x speedup in dumping json

With pretty printing:
```bash
$ python -m timeit -s 'import perf' 'perf.dumps()'
json
10000 loops, best of 3: 22.9 usec per loop
```

without pretty printing (~2x performance)

```bash
$ python -m timeit -s 'import perf' 'perf.dumps()'
json
100000 loops, best of 3: 12.2 usec per loop
```

With ujson Without pretty printing (~14.5x performance)
```bash
$ python -m timeit -s 'import perf' 'perf.dumps()'
ujson
1000000 loops, best of 3: 1.57 usec per loop
```

Loads has the same impact.

json:
```bash
n$ python -m timeit -s 'import perf' 'perf.loads()'
json
100000 loops, best of 3: 4.45 usec per loop
```

ujson (~6.66x performance):
```bash
$ python -m timeit -s 'import perf' 'perf.loads()'
ujson
1000000 loops, best of 3: 0.668 usec per loop
```

perf.py:

```python
#!/usr/bin/env python
# -*- coding: utf8 -*-

try:
    import ujson as json
except ImportError:
    import json


data = {
    'a': False,
    'b': 0.001,
    'c': None
}

input = '{"a": false, "b": 0.001, "c": null}'

print(json.__name__)

def dumps():
#    json.dumps(data, indent=4, separators=(',', ': '), ensure_ascii=False)
    json.dumps(data, ensure_ascii=False)

def loads():
    json.loads(input)
```


falcon/http_error.py
```diff
@@ -170,8 +173,7 @@ def to_json(self):
         """
 
         obj = self.to_dict(OrderedDict)
-        return json.dumps(obj, indent=4, separators=(',', ': '),
-                          ensure_ascii=False)
+        return json.dumps(obj, ensure_ascii=False)
```
masellfoodpanda: @kgriffs Developers probably pretty print the json on their own. (Via browser plugin or json_pp in bash). There is no need for it in a api.

PR #957, opened 2016-12-12
fix: strip whitespaces when parsing cookie in tests

houqp:
Currently, ' foo=bar' is parsed as ' foo' being the cookie_name in the tests.


houqp:
The pep8 error does not seem to be related to my change, do you want me update those too?

jmvrbanac:
Don't worry about it, I tossed up a PR (#958) to fix those issues. It should merge shortly and make travis happy again. Thanks!

PR #965, opened 2016-12-25
feat: Make it easier to set content-type

### LINKED ISSUE #618: Make it easier to set content type
The response content type must be set to a raw string. Make it easier to set the content type to common media types (including charset, where appropriate). One simple thing we might do is just define some constants. How else might we make this easier and less error-prone?

rjrhaverkamp:
hi @kgriffs 

this is some basic code to fix #618 

Could you please review it and let me know what you think about it?

thanks and happy holidays! 🎉


ofek:
Great addition! I'd also recommend documenting it here [https://github.com/falconry/falcon/blob/master/falcon/response.py#L608](url)

rjrhaverkamp:
@ofek thanks for the feedback, I will. Not sure how to format it though. Should I also mention the options? So like falcon.HTML, falcon.XML etc?

ofek:
LGTM, just add test.

falcon/__init__.py
```diff
@@ -37,7 +37,13 @@
 )
 
 DEFAULT_MEDIA_TYPE = 'application/json; charset=UTF-8'
-
+HTML = 'text/html; charset=utf-8'
+JSON = DEFAULT_MEDIA_TYPE
+JS = 'text/javascript; charset=utf-8'
+XML = 'text/javascript; charset=utf-8'
```
kgriffs: The RFCs generally recommend not specifying a charset for XML unless the XML encoding declaration is known (i.e., `<?xml version="1.0" encoding="UTF-8"?>`) and matches the charset parameter. Therefore, in this case, it is probably best to leave off the `charset` parameter.

falcon/response.py
```diff
@@ -678,6 +684,18 @@ def add_link(self, target, rel, title=None, title_star=None,
 
         """)
 
+    def set_media_type(self, media_type):
```
kgriffs: It seems like this is a sort of alias for the `content_type` property. Instead of defining an additional setter method, what do you think about doing this instead:

```py
media_type = content_type
```

And then adding `media_type` under the "Attributes" section in the `Response`class docstring to note that it is simply an alias of `content_type`?

You can check how the docs looks per https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#documentation .

kgriffs:
Re how to organize these constants, here are some options for the sake of discussion:

```py
# import falcon
resp.content_type = falcon.TYPE_PNG
resp.content_type = falcon.PNG
resp.content_type = falcon.media_type.PNG
resp.content_type = falcon.MEDIA_TYPE.PNG

# from...import
resp.content_type = media_type.PNG
resp.content_type = MEDIA_TYPE.PNG
```

falcon/api.py
```diff
@@ -36,6 +36,7 @@ class API(object):
     Args:
         media_type (str, optional): Default media type to use as the value for
             the Content-Type header on responses (default 'application/json').
+            It is possible to use default types like falcon.HTML, falcom.XML etc.
```
isaacm: typo: `falcom.XML` should be `falcon.XML`

jmvrbanac:
The changes discussed are being worked on as a part of the PyCon Sprint and are going into #1058. The commit history is preserved on that one, so we're going to close this one for now.

Issue #967, opened 2017-01-04
Proposal: pass request object to router.find


sseg:
It can be useful to make decisions about headers in the routing layer (e.g. for API versioning or media types). Having the full request available to the router allows the application to route requests based on any value of the WSGI request env.

To implement this kind of functionality currently requires putting routing logic into response handlers (after HTTP method resolution) or manipulating the request URI using middlewares. Neither of these solve the problem in a straightforward way, and both overload another layer of the application with values computed from headers (respectively the response handler and the request path).

`falcon.api.API`'s `_get_responder` method should pass the request object to the router during route-finding to provide greater control in the routing process:
```python
route = self._router.find(path, request=req)
```

This would constitute a breaking change in the routing API, where the `find` methods of routers would need to accept a new keyword argument, but I believe the power this affords makes it worth the change. I expect the biggest win from this change would be in allowing applications to solve API versioning (https://github.com/falconry/falcon/issues/645, https://github.com/falconry/falcon/issues/932) using a custom router or a router wrapper.

What do you think?

kgriffs:
Hmm, interesting idea. Since `path` is simply `req.path`, you might just pass in `req` instead, e.g.:

```py
route = self._router.find(req)
```

If we make this a breaking change, it will have to wait for 2.0. Alternatively, we could introduce a new method that, if present, is called instead of find(), e.g.:

```py
if self._router_implements_search:
    route = self._router.search(req)
else:
    route = self._router.find(path)
```

Alternatively, we could wrap `find()` with a shim when `search()` is not available, or use it directly when it is. This would incur the overhead of an extra function call in the former case:

```py
    route = self._router_search(req)
```

Finally, we could avoid introducing a new method by simply adding a kwarg as originally proposed. We could check the argspec to determine whether or not the router supports the new parameter. But those two args would be somewhat redundant as I mentioned earlier.

sseg:
I'm in favour of branching within `_get_responder` as it reads simply and introduces very little overhead for both the default router and those implementing the new method. 

There's a risk that whichever name is chosen for the new method could already be used in the namespace of a user's custom router. We could introduce a new keyword argument to the API class that allows users to specify which method on their router accepts a Request object (if any), defaulting to `"search"`.

sseg:
[...]

PR #995, opened 2017-01-29
feat(API): add support for routing on full request

sseg:
One way of resolving https://github.com/falconry/falcon/issues/967

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -103,6 +103,10 @@ class API(object):
             to use in lieu of the default engine.
             See also: :ref:`Routing <routing>`.
 
+        router_search_method (str, optional): The name of a custom router method
+            which searches for a route given a ``Request`` object. (default
+            ``"search"``)
+
         independent_middleware (bool): Set to ``True`` if response
             middleware should be executed independently of whether or
             not request middleware raises an exception.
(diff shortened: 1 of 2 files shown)


sseg:
@kgriffs: let me know if you want to go with this approach and then I'll update the custom routing docs to match.

falcon/api.py
```diff
@@ -103,6 +103,10 @@ def process_response(self, req, resp, resource, req_succeeded)
             to use in lieu of the default engine.
             See also: :ref:`Routing <routing>`.
 
+        router_search_method (str, optional): The name of a custom router method
```
kgriffs: How does this compare to simply defining a new `search()` method for the router interface and deprecating `find()`? Does the method name need to be configurable?

kgriffs: > There's the possibility that existing custom router implementations will already use whatever method name we choose for some other functionality.

Good point. I don't think there is any way around potentially breaking a custom router if we introduce another method name. Making that name configurable provides a way to mitigate the break, but doesn't avoid it in the first place. At this point a think we have the following options:

**a)** Introduce a new method in the next minor point release, which may break some custom routers out there.

**b)** Don't introduce a new method, but wait until 2.0 to change the `find()` method to take a `Request` object instead of a path

**c)** Don't introduce a new method, but add a `req` param to `find()` for the next minor point release. Then remove the path param in 2.0 as a breaking change. When a custom router is passed in, check the argspec and call in one of two ways, accordingly, to be backwards-compatible.

I'm keen to not do (a) because that goes against semver. (c) means passing redundant params, but it gets this functionality in people's hands sooner than (b), and can be cleaned up with the 2.0 release anyway. 

What do you think? I would be happy with either (b) or (c).

sseg: Up to you, really. 

I'm happy to write (c) and a ticket to track (b) if you think that's a good approach. 

Otherwise we can update https://github.com/falconry/falcon/issues/967 to track the 2.0 plan and leave this out for now.

kgriffs: OK, let's move forward with (c) then. Thanks!

kgriffs:
Hi @sseg, it would be nice to get this in 1.2 if we can get this over the finish line in the next couple days. Do you think you'll have some time to work on this some more today or tomorrow?

sseg:
I can get some time to write the implementation tomorrow morning, but I
don't think I'll have cycles to follow-up on feedback until the weekend.

What's your release schedule?

sseg:
Closing in favour of https://github.com/falconry/falcon/pull/1001

PR #1001, opened 2017-02-09
feat(API): add support for routing on full request

sseg:
This change allows the `find` method of a custom router to accept a `req` keyword argument to which a request instance is passed.

### DIFF
--- a/docs/api/routing.rst
+++ b/docs/api/routing.rst
@@ -23,16 +23,24 @@ A custom router is any class that implements the following interface:
                     will handle requests for the given URI.
             """
 
-        def find(self, uri):
+        def find(self, uri, req=None):
             """Search for a route that matches the given partial URI.
 
             Args:
                 uri(str): The requested path to route
 
+            Keyword Args:
+                req(Request): The Request object that will be passed to the
+                    routed responder
+
             Returns:
                 tuple: A 4-member tuple composed of (resource, method_map,
                     params, uri_template), or ``None`` if no route matches
                     the requested path
+
+            Note:
+                The `req` keyword argument was added in version 1.2, but routers
+                which do not implement the interface are still supported.
             """
 
 A custom routing engine may be specified when instantiating
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -28,6 +28,22 @@ import falcon.status_codes as status
 from falcon.util.misc import get_argnames
 
 
+def _make_router_search(router):
+    """Given a router instance returns a function that accepts a request
+    and invokes the router's find method.
+    """
+    arg_names = get_argnames(router.find)
+    supports_req = 'req' in arg_names and len(arg_names) > 1
+
+    if supports_req:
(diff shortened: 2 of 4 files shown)


sseg:
@kgriffs I'm not sure how you'd like to describe this change in the docs. The second commit https://github.com/falconry/falcon/pull/1001/commits/a54f401a2627cda289498315aa95ecb4fc1a22ef has a very short treatment of it.

sseg:
@kgriffs do you have any tips for using inspect on Cython funcs in 2.7?

kgriffs:
Looking into this, I found that Cython uses duck typing, which is only supported starting with Python 3.4. You can work around this by circumventing the type check. Replace this:

```py
spec = inspect.getargspec(func)
```

with this:

```py
# NOTE: Don't use inspect.getargspec() since it does not support Cython duck typing.
args, varargs, varkw = inspect.getargs(func.func_code)
spec = inspect.ArgSpec(args, varargs, varkw, func.func_defaults)
```

docs/api/routing.rst
```diff
@@ -35,6 +35,16 @@ A custom router is any class that implements the following interface:
                     the requested path
             """
 
+The `find` method of a custom router can optionally accept a `req` keyword argument
```
kgriffs: I'd recommend instead of adding this extra section, just add a kwarg to the original interface definition above. If you like, you may also add a note about how this kwarg was added in 1.2 but routers that do not define it are still supported.

falcon/api.py
```diff
@@ -28,6 +28,21 @@
...
+    """
+    arg_names = get_argnames(router.find)
+    supports_req = 'req' in arg_names and len(arg_names) > 1
+
+    def search(request, pass_req=supports_req):
+        if pass_req:
+            return router.find(request.path, req=request)
+        return router.find(request.path)
+
+    return search
```
kgriffs: In the case that the router supports the new interface, you can just `return router.find` directly; no need to shim it.

tests/test_custom_router.py
```diff
@@ -81,3 +87,31 @@ def add_route(self, uri_template, method_map, resource, name):
 
         self.assertEqual(len(check), 2)
         self.assertIn('my-url-name-arg', check)
+
+    def test_custom_router_takes_req_positional_argument(self):
+        def resource(req, resp):
```
kgriffs: Nitpick: This is a specific HTTP method responder for a resource, not the resource itself, so perhaps this (and the function on L106) should be renamed to `resource_responder`, `responder`, or similar?

kgriffs:
I pulled and verified test coverage locally, looks good!

kgriffs:
@sseg this looks great, thanks for making it happen!

sseg:
Yeah, the results showed up on codecov, not sure why they didn't post here.

https://codecov.io/gh/falconry/falcon/pull/1001

PR #1032, opened 2017-04-26
Response header delete

RonRothman:
This PR adds a single method, `delete_header`, to `Response`.

Previously, `Response` had methods to get and set headers, but no way to delete them, once set. We came across a need to delete response headers that had been set upstream, and without a `delete_header` method we've had to resort to directly manipulating `Response._headers`. This PR alleviates that abstraction-violation.

Tox tests all pass, but this is my first Falcon PR so please let me know if I've missed anything and I'll gladly fix.

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -384,6 +384,20 @@ class Response(object):
         # NOTE(kgriffs): normalize name by lowercasing it
         self._headers[name.lower()] = value
 
+    def delete_header(self, name):
+        """Delete a header for this response.
+
+        If the header was not previously set, do nothing.
+
+        Args:
+            name (str): Header name (case-insensitive).  Must be of type
+                ``str`` or ``StringType`` and contain only US-ASCII characters.
+                Under Python 2.x, the ``unicode`` type is also accepted,
+                although such strings are also limited to US-ASCII.
+        """
+        # NOTE(kgriffs): normalize name by lowercasing it
+        self._headers.pop(name.lower(), None)
+
     def append_header(self, name, value):
         """Set or append a header for this response.
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,10 @@ cache:
 
 matrix:
     include:
-        - python: 2.7
-          env: JYTHON=true
+        # TODO(kgriffs): Uncomment when Travis no longer stalls and
+        # errors out on Jython tests.
+        # - python: 2.7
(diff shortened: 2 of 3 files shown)


falcon/response.py
```diff
@@ -384,6 +384,20 @@ def set_header(self, name, value):
...
+        If the header was not previously set, do nothing.
+
+        Args:
+            name (str): Header name (case-insensitive).  Must be of type
+                ``str`` or ``StringType`` and contain only US-ASCII characters.
+                Under Python 2.x, the ``unicode`` type is also accepted,
+                although such strings are also limited to US-ASCII.
+        """
+        # NOTE(kgriffs): normalize name by lowercasing it
+        self._headers.pop(name.lower(), None)
```
kgriffs: Out of curiosity I benchmarked this approach against wrapping a `del` within a `try...except` block and the latter was slightly more performant, but only by a few ns, so it's nothing to worry about.

RonRothman: Thanks--happy to contribute!

Interesting - did you happen to benchmark both the case where the key is present and the case where it's absent? I wouldn't be surprised if `try...except` performed well for the former but poorly for the latter.

RonRothman: Was curious so I tried benchmarking it myself. If my benchmarks are correct, then my results confirm what you found, @kgriffs. Though it's noteworthy that `try...except` performs much slower in the case when the key in question is not present.

Results:
```
pop_present: 0.9311439991
pop_not_present: 0.939202070236
try_present: 0.687428951263
try_not_present: 4.01770091057
```

My benchmarking code:
```
from timeit import timeit

def pop_present():
    d = {'hello': 'world'}
    d.pop('hello', None)


def pop_not_present():
    d = {'hello': 'world'}
    d.pop('not_present', None)


def try_present():
    d = {'hello': 'world'}
    try:
        del d['hello']
    except KeyError:
        pass


def try_not_present():
    d = {'hello': 'world'}
    try:
        del d['not_present']
    except KeyError:
        pass

N = 5000000
print 'pop_present:', timeit('pop_present()', 'from __main__ import pop_present', number=N)
print 'pop_not_present:', timeit('pop_not_present()', 'from __main__ import pop_not_present', number=N)
print 'try_present:', timeit('try_present()', 'from __main__ import try_present', number=N)
print 'try_not_present:', timeit('try_not_present()', 'from __main__ import try_not_present', number=N)
```

PR #1040, opened 2017-05-03
feat(CompiledRouter): Add an intermediate AST step to the compiler

kgriffs:
Rather than compiling the routing tree directly to Python code, first generate an AST and then use it to produce the code. This provides several benefits, including:

* It makes the compilation process easier to reason about.
* It makes it easier to keep track of indentation and whitespace.
* It sets us up for being able to make transformations that we will need to do to support URI template filters, etc. in the future.

### DIFF
--- a/falcon/routing/compiled.py
+++ b/falcon/routing/compiled.py
@@ -18,7 +18,7 @@ import keyword
 import re
 
 
-_FIELD_REGEX = re.compile('{([^}]*)}')
+_FIELD_PATTERN = re.compile('{([^}]*)}')
 _TAB_STR = ' ' * 4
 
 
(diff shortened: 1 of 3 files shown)


kgriffs:
Note: I checked the generated code against what was output previously to ensure they matched. Also, all the tests pass, but we will probably want to have an extended beta/rc cycle for the next release to ensure no bugs slip through.

falcon/routing/compiled.py
```diff
@@ -294,11 +316,12 @@ def __init__(self, raw_segment,
         self.is_var = False
         self.is_complex = False
         self.var_name = None
+        self.var_pattern = None
 
         # NOTE(kgriffs): CompiledRouter.add_route validates field names,
         # so here we can just assume they are OK and use the simple
-        # _FIELD_REGEX to match them.
-        matches = list(_FIELD_REGEX.finditer(raw_segment))
+        # _FIELD_PATTERN to match them.
```
kgriffs: I took the opportunity to normalize the terminology used in naming variables and such. "Pattern" is more commonly used in Falcon as opposed to "regex" or "expression".

PR #1050, opened 2017-05-16
Request and Response media-type handling

jmvrbanac:
This is a first pass at a non-breaking change that'll allow for a customizable media handling system. This approach combines many of the suggestions brought up by the community in #145.  

One the thing that is left out of this PR is handling full content negotiation (i.e. connecting the request's accept header to the response's content-type). Unfortunately, this is a harder problem to solve in a backwards compatible fashion that doesn't affect performance. However, especially as we move towards v2, I think that would be a great opportunity to revist full negotiation. In the meantime, there are several easy workarounds for people needing this functionality.

### DIFF
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -47,6 +47,15 @@ media type that JSON also supports (i.e. dicts, lists, etc).
 .. autofunction:: falcon.media.validators.jsonschema.validate
 
 
+Content-Type Negotiation
+------------------------
+
+Falcon currently only supports partial negotiation out of the box. By default,
+when the ``media`` attribute is used it attempts to de/serialize based on the
+``Content-Type`` header value. The missing link that Falcon doesn't provide
+is the connection between the :any:`falcon.Request` ``Accept`` header provided
+by a user and the :any:`falcon.Response` ``Content-Type`` header.
+
 
 Replacing The Default Handlers
 ------------------------------
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -1,0 +1,110 @@
+.. _media:
+
+Media
+=====
+
+Falcon allows for easy and customizable media handling. By default
+Falcon only enables a single JSON handler. However, other/multiple handlers
+can be configured through the :any:`falcon.RequestOptions` and
+:any:`falcon.ResponseOptions` objects specified on your :any:`falcon.API`.
+
+.. note::
+
+    For performance concerns, Falcon will not attempt to handle request
+    media until the first time it's accessed. Once it has procesed the
+    request, it'll use the cached result for subsequent interactions.
+
+Usage
+-----
+
+If you're creating a JSON API then zero configuration is needed. Just access
+or set the ``media`` attribute as appropriate and let Falcon do the heavy
+lifting for you.
+
+.. code:: python
+
+    import falcon
+
+
+    class EchoResource(object):
+        def on_post(self, req, resp):
+            message = req.media.get('message')
+
+            resp.media = {'message': message}
+            resp.status = falcon.HTTP_200
+
+Validating Media
+----------------
+
+Falcon currently does not validate media for you as requirements and tooling
+vary quite largely between projects; however, here is an example of how you
+might go about implementing a JSON Schema validator using a decorator.
+
+.. code:: python
+
+    import jsonschema
+
+
+    def validate(schema):
+        def decorator(func):
+            def wrapper(self, req, resp, *args, **kwargs):
+                try:
+                    jsonschema.validate(req.media, schema)
+                except jsonschema.ValidationError as e:
+                    raise falcon.HTTPBadRequest(
+                        'Failed data validation',
+                        e.message
+                    )
+
+                return func(self, req, resp, *args, **kwargs)
+            return wrapper
+        return decorator
+
+
+Given that decorator you could use it on the resource as such:
+
+.. code:: python
+
+    # -- snip --
+
+    @validate(my_post_schema)
+    def on_post(self, req, resp):
+    # -- snip --
+
+
+Replacing the default handlers
+------------------------------
+
+When creating your API object you can either add or completely
+replace all of the handlers. For example, lets say you want to write an API
+that sends and receives MessagePack. We can easily do this by telling our
+Falcon API that we want a default media-type of `application/msgpack` and
+then create a new ``Handlers`` object specifying the desired media type and
+a handler that can process that data.
+
+.. code:: python
+
+    import falcon
+    from falcon import media_handlers
+
+
+    handlers = media_handlers.Handlers({
+        'application/msgpack': media_handlers.MessagePack,
+    })
+
+    api = falcon.API(media_type='application/msgpack')
+
+    api.req_options.media_handlers = handlers
+    api.resp_options.media_handlers = handlers
+
+
+Custom Handlers
+---------------
+
+Currently Falcon only supports a handful of media handlers out of the box;
+however, you can easily create your own. All you need is an object that
+contains the following class methods:
+
+ * ``load(cls)``
+ * ``deserialize(cls, raw)``
+ * ``serialize(cls, media)``
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -82,18 +82,18 @@ Replacing The Default Handlers
 When creating your API object you can either add or completely
 replace all of the handlers. For example, lets say you want to write an API
 that sends and receives MessagePack. We can easily do this by telling our
-Falcon API that we want a default media-type of `application/msgpack` and
-then create a new ``Handlers`` object specifying the desired media type and
+Falcon API that we want a default media-type of ``application/msgpack`` and
+then create a new :any:`Handlers` object specifying the desired media type and
 a handler that can process that data.
 
 .. code:: python
 
     import falcon
-    from falcon import media_handlers
+    from falcon import media
 
 
-    handlers = media_handlers.Handlers({
-        'application/msgpack': media_handlers.MessagePack,
+    handlers = media.Handlers({
+        'application/msgpack': media.MessagePackHandler,
     })
 
     api = falcon.API(media_type='application/msgpack')
@@ -106,9 +106,16 @@ Custom Handlers
 ---------------
 
 Currently Falcon only supports a handful of media handlers out of the box;
-however, you can easily create your own. All you need is an object that
-contains the following class methods:
+however, you can easily create your own using the following abstract base
+class:
 
- * ``load(cls)``
- * ``deserialize(cls, raw)``
- * ``serialize(cls, media)``
+.. autoclass:: falcon.media.BaseHandler
+    :members:
+    :member-order: bysource
+
+
+Handlers
+--------
+
+.. autoclass:: falcon.media.Handlers
+    :members:
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -93,7 +93,7 @@ a handler that can process that data.
 
 
     handlers = media.Handlers({
-        'application/msgpack': media.MessagePackHandler,
+        'application/msgpack': media.MessagePackHandler(),
     })
 
     api = falcon.API(media_type='application/msgpack')
@@ -101,13 +101,21 @@ a handler that can process that data.
     api.req_options.media_handlers = handlers
     api.resp_options.media_handlers = handlers
 
+Supported Handler Types
+-----------------------
 
-Custom Handlers
----------------
+.. autoclass:: falcon.media.JSONHandler
+    :members:
 
-Currently Falcon only supports a handful of media handlers out of the box;
-however, you can easily create your own using the following abstract base
-class:
+.. autoclass:: falcon.media.MessagePackHandler
+    :members:
+
+Custom Handler Type
+-------------------
+
+If Falcon doesn't have a internet media type handler that supports your
+use-case. You can easily implement your own using the abstract base class
+provided by Falcon:
 
 .. autoclass:: falcon.media.BaseHandler
     :members:
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -40,40 +40,12 @@ lifting for you.
 Validating Media
 ----------------
 
-Falcon currently does not validate media for you as requirements and tooling
-vary quite largely between projects; however, here is an example of how you
-might go about implementing a JSON Schema validator using a decorator.
+Falcon currently only supports a JSONSchema media type handler; however,
+JSONSchema is very versatile and can be used to validate any deserialized
+media type that JSON also supports (i.e. dicts, lists, etc).
 
-.. code:: python
+.. autofunction:: falcon.media.validators.jsonschema.validate
 
-    import jsonschema
-
-
-    def validate(schema):
-        def decorator(func):
-            def wrapper(self, req, resp, *args, **kwargs):
-                try:
-                    jsonschema.validate(req.media, schema)
-                except jsonschema.ValidationError as e:
-                    raise falcon.HTTPBadRequest(
-                        'Failed data validation',
-                        e.message
-                    )
-
-                return func(self, req, resp, *args, **kwargs)
-            return wrapper
-        return decorator
-
-
-Given that decorator you could use it on the resource as such:
-
-.. code:: python
-
-    # -- snip --
-
-    @validate(my_post_schema)
-    def on_post(self, req, resp):
-    # -- snip --
 
 
 Replacing The Default Handlers
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -40,8 +40,8 @@ lifting for you.
 Validating Media
 ----------------
 
-Falcon currently only supports a JSONSchema media type handler; however,
-JSONSchema is very versatile and can be used to validate any deserialized
+Falcon currently only provides a JSON Schema media validator; however,
+JSON Schema is very versatile and can be used to validate any deserialized
 media type that JSON also supports (i.e. dicts, lists, etc).
 
 .. autofunction:: falcon.media.validators.jsonschema.validate
@@ -56,8 +56,17 @@ when the ``media`` attribute is used it attempts to de/serialize based on the
 is the connection between the :any:`falcon.Request` ``Accept`` header provided
 by a user and the :any:`falcon.Response` ``Content-Type`` header.
 
+If you do need full negotiation, it is very easy to bridge the gap using
+middleware. Here is an example of how this can be done:
 
-Replacing The Default Handlers
+.. code-block:: python
+
+    class NegotiationMiddleware(object):
+        def process_request(self, req, resp):
+            resp.content_type = req.accept
+
+
+Replacing the Default Handlers
 ------------------------------
 
 When creating your API object you can either add or completely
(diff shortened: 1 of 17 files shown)


docs/api/media.rst
```diff
@@ -0,0 +1,110 @@
...
+Falcon allows for easy and customizable media handling. By default
+Falcon only enables a single JSON handler. However, other/multiple handlers
+can be configured through the :any:`falcon.RequestOptions` and
+:any:`falcon.ResponseOptions` objects specified on your :any:`falcon.API`.
+
+.. note::
+
+    For performance concerns, Falcon will not attempt to handle request
+    media until the first time it's accessed. Once it has procesed the
+    request, it'll use the cached result for subsequent interactions.
```
jmvrbanac: Add comment regarding consumption of the request stream.

falcon/request.py
```diff
@@ -249,6 +250,16 @@ class Request(object):
 
                 doc = json.load(req.bounded_stream)
 
+        media (object): Returns a deserialized form of the request stream.
+            When called, it will attempt to deserialize the request stream
+            using the Content-Type header as well as the media-type handlers
+            configured in the request options.
+
+            Note:
```
jmvrbanac: Change this to a warning

falcon/media_handlers.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def update(self, input_dict):
+        for handler in input_dict.values():
+            handler.load()
+
+        UserDict.update(self, input_dict)
+
+    def __setitem__(self, key, item):
+        item.load()
+        self.data[key] = item
```
jmvrbanac: Make sure people can't do an empty string here.

falcon/media_handlers.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            try:
+                resolved = mimeparse.best_match(
+                    supported_media_types,
+                    media_type
+                )
+            except:
+                pass
+
+        if not resolved:
```
jmvrbanac: Add a comment about why this is `not resolved`

falcon/media_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+            return cls.msgpack.unpackb(raw)
+        except ValueError as err:
+            raise errors.HTTPBadRequest(
+                'Invalid MessagePack',
+                'Could not parse MessagePack body - {0}'.format(err)
+            )
+
+    @classmethod
+    def serialize(cls, media):
+        return cls.msgpack.packb(media)
```
ueg1990: since classes Json and MessagePack are following a certain contract, maybe define an abstract class using ABC which Json, MessagePack and future handlers can import

falcon/media_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+        UserDict.update(self, input_dict)
+
+    def __setitem__(self, key, item):
+        if not key:
+            raise ValueError('Media Type cannot be None or empty string')
+
+        item.load()
+        self.data[key] = item
+
+    def find_by_media_type(self, media_type, default):
```
ueg1990: maybe we should define a default value and make default a keyword argument. In the usages below where find_by_media_type is called, we pass a default_value from the same object (self.options) from where we call media_handlers

falcon/media_handlers.py
BenjamenMeyer: Also, do you really need to import the modules inside the `load()`? It'd be better to just import them at the global level outside the class IMHO, which may negate the need for `load()` - though I can certainly see there may be a good use-case for some media types that require configuration, etc.

That said, I do agree with @ueg1990 that there should be a base class of some sort that can be used to enforce an interface.

falcon/request.py
```diff
@@ -1394,16 +1424,26 @@ class RequestOptions(object):
...
+        'default_media_type',
+        'media_handlers',
     )
 
     def __init__(self):
         self.keep_blank_qs_values = False
         self.auto_parse_form_urlencoded = False
         self.auto_parse_qs_csv = True
         self.strip_url_path_trailing_slash = True
+        self.default_media_type = 'application/json'
```
BenjamenMeyer: This should probably be `None`. Falcon should never assume a default other than `application/text` if any at all.

falcon/media_handlers.py
kgriffs: [fix] Set `use_bin_type=True` so that the client can disambiguate between bytes and a string encoded as UTF-8. I verified the msgpack-python source to ensure it does the right thing.

The fact that the bin type is used should be pointed out in the docstring.

falcon/media_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+class MessagePack(object):
+    @classmethod
+    def load(cls):
+        import msgpack
+        cls.msgpack = msgpack
+
+    @classmethod
+    def deserialize(cls, raw):
+        try:
+            return cls.msgpack.unpackb(raw)
```
kgriffs: [fix] Set `encoding='utf-8'` to decode strings back properly. This assumes usage of the bin type (AKA format family) when encoding, which TBH should be the default in all MessagePack implementations, although that is not always the case. 

Ensure tests cover the case of a single object containing both `bytes` and `unicode`.

This should be mentioned in the docstring so that client developers know they should be encoding with the bin type.

falcon/media_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+            )
+
+    @classmethod
+    def serialize(cls, media):
+        return cls.json.dumps(media)
+
+
+class MessagePack(object):
+    @classmethod
+    def load(cls):
```
kgriffs: [fix] Instantiate instances of `msgpack.Packer` and `msgpack.Unpacker` once, up front, rather than calling `packb()` and `unpackb()` which instantiate instances of the aforementioned classes on every call.

falcon/request.py
```diff
@@ -251,6 +252,16 @@ class Request(object):
...
                 doc = json.load(req.bounded_stream)
 
+        media (object): Returns a deserialized form of the request stream.
+            When called, it will attempt to deserialize the request stream
+            using the Content-Type header as well as the media-type handlers
+            configured in the request options.
+
+            Warning:
+                This operation will consume the request stream the first time
+                it's called and cache the results. Follow-up calls, will just
```
kgriffs: [fix] Remove the extraneous comma.

falcon/request.py
```diff
@@ -251,6 +252,16 @@ class Request(object):
...
+                retrieve a cached version of object.
```
kgriffs: [fix] "of object" ---> "of the object"

falcon/request.py
```diff
@@ -1394,16 +1424,26 @@ class RequestOptions(object):
             forward slash. However, this behavior can be problematic in
             certain cases, such as when working with authentication
             schemes that employ URL-based signatures.
+
+        default_media_type (str): The default media-type to use when
+            deserializing a response.
+
+        media_handlers (Handlers): A dict-like object that allows for you to
```
kgriffs: [fix] Include a description of the default handlers that are configured.

falcon/request.py
```diff
@@ -1394,16 +1424,26 @@ class RequestOptions(object):
             forward slash. However, this behavior can be problematic in
             certain cases, such as when working with authentication
             schemes that employ URL-based signatures.
+
+        default_media_type (str): The default media-type to use when
```
kgriffs: [fix] Include a description of the default media type that this is set to.

falcon/request.py
kgriffs: Should this use the `DEFAULT_MEDIA_TYPE` constant?

falcon/response.py
```diff
@@ -86,6 +87,14 @@ class Response(object):
                 ensure Unicode characters are properly encoded in the
                 HTTP response.
 
+        media (object): A serializable object supported by the media handlers
+            configured via response options.
+
```
kgriffs: [suggestion] Link to the dedicated page on media handling in the docs.

falcon/request.py
```diff
@@ -251,6 +252,16 @@ class Request(object):
 
                 doc = json.load(req.bounded_stream)
 
+        media (object): Returns a deserialized form of the request stream.
+            When called, it will attempt to deserialize the request stream
+            using the Content-Type header as well as the media-type handlers
+            configured in the request options.
+
```
kgriffs: [suggestion] Link to the dedicated page on media handling in the docs.

falcon/response.py
```diff
@@ -86,6 +87,14 @@ class Response(object):
                 ensure Unicode characters are properly encoded in the
                 HTTP response.
 
+        media (object): A serializable object supported by the media handlers
+            configured via response options.
```
kgriffs: [suggestion] Link to the relevant response options.

falcon/request.py
```diff
@@ -251,6 +252,16 @@ class Request(object):
 
                 doc = json.load(req.bounded_stream)
 
+        media (object): Returns a deserialized form of the request stream.
+            When called, it will attempt to deserialize the request stream
+            using the Content-Type header as well as the media-type handlers
+            configured in the request options.
```
kgriffs: [suggestion] Link to the relevant request options.

falcon/response.py
```diff
@@ -86,6 +87,14 @@ class Response(object):
...
+
+            Note:
```
kgriffs: [fix] This note is not relevant in this context.

falcon/media_handlers.py
jmvrbanac: Re: Importing modules inside of `load()` instead of globally.
-  The rational for this was to allow a method for Falcon to easily support a various of media handlers within the Falcon package but not add default installed requirements. I suppose we could accomplish the same thing by moving each one into their own module.

Re: Using an ABC interface
- I actually had started with this approach, but I removed them when I realized that Falcon doesn't use ABCs anywhere else in the code base. Perhaps there is a Falcon wide discussion to be had here?

@kgriffs Do you have any opinions on these items?

falcon/request.py
jmvrbanac: @BenjamenMeyer Interestingly enough Falcon has set a default media type for headers for quite some time. It happens when a new instance of API is created. So this default value is never really used as it's usually set by the API class. However, I agree, this should probably be None or as @kgriffs suggested in one of his comments the `DEFAULT_MEDIA_TYPE` value.

kgriffs: I suppose you could set it to `None` since it is never used. This may actually be a guard against some future change mistakenly leaving the default value in place (presumably seeing `None` would signal that something ain't right in that case).

falcon/media/validators/jsonschema.py
```diff
@@ -0,0 +1,47 @@
...
+        schema (dict): A dictionary that follows the JSONSchema specification.
+            See `json-schema.org <http://json-schema.org/>`_ for more
+            information on defining a compatible dictionary.
+
+    Example:
+        .. code:: python
+
+            from falcon.media.validators import jsonschema
+
+            # -- snip --
```
kgriffs: I like the way you use these `# -- snip --` comments. I'm so totally going to steal this. (You know, because I'm a great artist.) 💯

docs/api/media.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Content-Type Negotiation
+------------------------
+
+Falcon currently only supports partial negotiation out of the box. By default,
+when the ``media`` attribute is used it attempts to de/serialize based on the
+``Content-Type`` header value. The missing link that Falcon doesn't provide
+is the connection between the :any:`falcon.Request` ``Accept`` header provided
+by a user and the :any:`falcon.Response` ``Content-Type`` header.
+
```
kgriffs: [suggestion] We should probably say something about why this isn't the end of the world. Let's not end on a negative note.

docs/api/media.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Replacing The Default Handlers
```
kgriffs: [nit] In title case, articles are not capitalized.

(This is a total grammar dork thing to call out, but I couldn't resist. 😈 )

docs/api/media.rst
```diff
@@ -0,0 +1,110 @@
...
+------------------------------
+
+When creating your API object you can either add or completely
+replace all of the handlers. For example, lets say you want to write an API
+that sends and receives MessagePack. We can easily do this by telling our
+Falcon API that we want a default media-type of ``application/msgpack`` and
+then create a new :any:`Handlers` object specifying the desired media type and
+a handler that can process that data.
+
+.. code:: python
```
kgriffs: [fix] I'd love to also see an example of keeping the default JSON handler and just adding an additional one without having to assign `api.req_options.media_handlers` to a new instance of `media.Handlers`.

falcon/media/msgpack.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import absolute_import
+
+from falcon import errors
+from falcon.media import BaseHandler
+
+
+class MessagePackHandler(BaseHandler):
+    """Handler built using the :py:mod:`msgpack` module from python-msgpack
+
+    Requires the ``python-msgpack`` module to be installed.
```
kgriffs: [suggestion] This probably deserves to be highlighted as a `Note`. Also, `python-msgpack` is technically a package, not a module. 😈

docs/api/media.rst
```diff
@@ -0,0 +1,110 @@
...
+.. autoclass:: falcon.media.JSONHandler
+    :members:
+
+.. autoclass:: falcon.media.MessagePackHandler
+    :members:
+
+Custom Handler Type
+-------------------
+
+If Falcon doesn't have a internet media type handler that supports your
```
kgriffs: [nit] "a internet media" ==> "an internet media"

docs/api/media.rst
```diff
@@ -0,0 +1,110 @@
...
+use-case. You can easily implement your own using the abstract base class
```
kgriffs: [fix] "use-case. You..." ==> "use case, you..."

falcon/request.py
```diff
@@ -1394,16 +1427,31 @@ class RequestOptions(object):
...
             certain cases, such as when working with authentication
             schemes that employ URL-based signatures.
+
+        default_media_type (str): The default media-type to use when
+            deserializing a response. This value is normally set to the media
+            type provided when a :class:`falcon.API` is initialized; however,
+            if created independently, this will default to the
+            ``DEFAULT_MEDIA_TYPE`` specified by Falcon.
+
+        media_handlers (Handlers): A dict-like object that allows for you to
```
kgriffs: [fix] "allows for you" ==> "allows you"

falcon/response.py
```diff
@@ -782,10 +807,25 @@ class ResponseOptions(object):
...
             not requiring HTTPS. Note, however, that this setting can
             be overridden via `set_cookie()`'s `secure` kwarg.
+
+        default_media_type (str): The default media-type to use when
+            deserializing a response. This value is normally set to the media
+            type provided when a :class:`falcon.API` is initialized; however,
+            if created independently, this will default to the
+            ``DEFAULT_MEDIA_TYPE`` specified by Falcon.
+
+        media_handlers (Handlers): A dict-like object that allows for you to
```
kgriffs: [fix] "allows for you" ==> "allows you"

falcon/media/base.py
```diff
@@ -0,0 +1,38 @@
...
+
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class BaseHandler(object):
+    """Abstract Base Class for an internet media type handler"""
+
+    @abc.abstractmethod
+    def load(self):
```
kgriffs: [design] How does this compare and contrast with simply using `__init__()`?

falcon/media/base.py
```diff
@@ -0,0 +1,38 @@
...
+        """Loads any required imports and configuration.
+
+        Allows for implementors to specify runtime configuration
```
kgriffs: [fix] "Allows for implementors..." ==> "Allows implementors..."

falcon/media/handlers.py
```diff
@@ -0,0 +1,70 @@
...
+    Attempts to load any imports for handlers as they are added.
+    """
+    def __init__(self, initial=None):
+        handlers = initial or {
+            'application/json': JSONHandler(),
+            'application/json; charset=UTF-8': JSONHandler(),
+        }
+
+        # NOTE(jmvrbanac): Directly calling UserDict as it's not inheritable.
+        # Also, this results in self.update(...) being called.
```
kgriffs: 👍

falcon/media/handlers.py
```diff
@@ -0,0 +1,70 @@
...
+        return resolved
+
+    def find_by_media_type(self, media_type, default):
+        # PERF(jmvrbanac): Check via a quick methods first for performance
+        try:
+            return self.data[media_type]
+        except KeyError:
+            pass
+
+        if media_type == '*/*' or not media_type:
```
kgriffs: [design] I suspect it is pretty common for clients to just say they accept `*/*`. Given that this check requires very little CPU time, what do you think about moving it above the try...except so that we don't have to pay a raised exception penalty in the case that the client doesn't explicitly request on of the supported media types? Since this check is so fast, it shouldn't add any additional overhead in the case that a specific type *is* requested.

falcon/media/msgpack.py
```diff
@@ -0,0 +1,35 @@
...
+            encoding='utf-8',
+            autoreset=True,
+            use_bin_type=True,
+        )
+
+    def deserialize(self, raw):
+        try:
+            # NOTE(jmvrbanac): Using unpackb since we would need to manage
+            # a buffer for Unpacker() which wouldn't gain us much.
+            return self.msgpack.unpackb(raw)
```
kgriffs: [fix] Set encoding='utf-8' to decode strings back properly. Ensure tests cover the case of a single object containing both byte strings and Unicode strings.

falcon/media/msgpack.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import absolute_import
+
+from falcon import errors
+from falcon.media import BaseHandler
+
+
+class MessagePackHandler(BaseHandler):
+    """Handler built using the :py:mod:`msgpack` module from python-msgpack
+
```
kgriffs: [fix] Add a note saying that this handler assumes that clients encode using the bin type to disambiguate between strings and bytes.

falcon/media/json.py
```diff
@@ -0,0 +1,24 @@
...
+        try:
+            return self.json.loads(raw.decode('utf-8'))
+        except ValueError as err:
+            raise errors.HTTPBadRequest(
+                'Invalid JSON',
+                'Could not parse JSON body - {0}'.format(err)
+            )
+
+    def serialize(self, media):
+        return self.json.dumps(media, ensure_ascii=False).encode('utf-8')
```
kgriffs: 👍

tests/test_response_media.py
```diff
@@ -0,0 +1,106 @@
...
+def test_msgpack(media_type):
+    client = create_client({
+        'application/msgpack': media.MessagePackHandler(),
+        'application/x-msgpack': media.MessagePackHandler(),
+    })
+    client.simulate_get('/')
+
+    resp = client.resource.captured_resp
+    resp.content_type = media_type
+    resp.media = {b'something': True}
```
kgriffs: [fix] Test an object that contains both `unicode` and `bytes` types to ensure the MessagePack bin type is in play.

tests/test_request_media.py
```diff
@@ -0,0 +1,119 @@
...
+    ('application/msgpack'),
+    ('application/msgpack; charset=utf-8'),
+    ('application/x-msgpack'),
+])
+def test_msgpack(media_type):
+    client = create_client({
+        'application/msgpack': media.MessagePackHandler(),
+        'application/x-msgpack': media.MessagePackHandler(),
+    })
+    expected_body = b'\x81\xc4\tsomething\xc3'
```
kgriffs: [fix] Test an object that contains both `unicode` (encoded to UTF-8), and `bytes` (encoded to the MessagePack bin type), to ensure the media handler is correctly decoding and distinguishing the two types.

falcon/media/base.py
jmvrbanac: Good point... It made sense when we were using straight classes, but if we're creating instances now, then we can get rid of ``load()``

jmvrbanac:
@kgriffs I think I've addressed the requested changes.

Issue #1007, opened 2017-03-14
Feature request: get_param_as_datetime


macdjord:
The `get_param_as_date()` function is useful, but what I need for my current use case is a date _and time_.

kgriffs:
In your case, how is the date/time string formatted?

macdjord:
'yyyy_mm_dd_hh_mm_ss', I think. I can no longer check; I've changed jobs.

jmvrbanac:
I believe this is resolved with #1053 and should be included in the next release.

PR #1053, opened 2017-05-21
Added a get_param_as_datetime() method to fetch query param as datetime

kgritesh:
Solution to #1007.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1148,6 +1148,51 @@ class Request(object):
 
         raise errors.HTTPMissingParam(name)
 
+    def get_param_as_datetime(self, name, format_string='%Y-%m-%dT%H:%M:%S',
+                              required=False, store=None):
+        """Return the value of a query string parameter as a datetime.
+
+        Args:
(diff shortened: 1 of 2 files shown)


kgritesh:
@kgriffs fixed as per your review. Pls have a look

falcon/request.py
```diff
@@ -1148,26 +1148,26 @@ def get_param_as_list(self, name,
 
         raise errors.HTTPMissingParam(name)
 
-    def get_param_as_date(self, name, format_string='%Y-%m-%d',
-                          required=False, store=None):
-        """Return the value of a query string parameter as a date.
+    def get_param_as_datetime(self, name, format_string='%Y-%m-%dT%H:%M:%S',
```
kgriffs: Upon further reflection, I think it may be better to require the `Z` UTC timezone designator (i.e., `%Y-%m-%dT%H:%M:%SZ`)  in the default format. The above format string implies ISO 8601, in which case the absence of a time zone designator indicates local time, which is ambiguous and therefore something that we should not encourage.

PR #1058, opened 2017-05-23
feat: Make it easier to set content-type

mayurmahajan:
This is a follow up PR addressing the changes in PR: #965


ofek:
@mayurmahajan @jmvrbanac It would be much nicer if content types were an enum.

mayurmahajan:
Could you please provide an example on how enum will be more helpful here?

ofek:
Following the Zen of Python, "Explicit is better than implicit." Those constant names could be anything; the user doesn't know by the name itself. Also, imports look nicer:

```python
from falcon import ContentTypes
```

vs

```python
from falcon import JSON, HTML, JPEG, PNG
```

jmvrbanac:
@ofek Unfortunately, Enums aren't currently supported in 2.6 and 2.7. While we can add a dependency, I'm not sure it worth it at the moment. Currently, our current constants are like `HTTP_XXX` so it would make sense to follow that pattern for the moment. So, to your point, I agree that we should be a little more explicit with the names. As a result, I think we should rename these to `MEDIA_XXXX` and then I've created an issue (#1059) to investigate refactoring our usage of constants.

falcon/response.py
```diff
@@ -713,24 +720,34 @@ def add_link(self, target, rel, title=None, title_star=None,
...
 
-    def _wsgi_headers(self, media_type=None, py2=PY2):
-        """Convert headers into the format expected by WSGI servers.
+    def _set_media_type(self, media_type=None):
+        """wrapper around set_header to set a content-type
 
         Args:
-            media_type: Default media type to use for the Content-Type
-                header if the header was not set explicitly (default ``None``).
+            media_type: media type to use for the Content-Type
```
kgriffs: [nitpick] Capitalize "media", i.e.: "Media type to use..."

falcon/constants/media_types.py
```diff
@@ -0,0 +1,10 @@
+MEDIA_JSON = 'application/json; charset=UTF-8'
+MEDIA_HTML = 'text/html; charset=utf-8'
+MEDIA_JS = 'text/javascript; charset=utf-8'
+MEDIA_XML = 'application/xml'
+MEDIA_TEXT = 'text/plain; charset=utf-8'
+MEDIA_JPEG = 'image/jpeg'
+MEDIA_PNG = 'image/png'
+MEDIA_YAML = 'application/x-yaml'
```
kgriffs: Would you mind adding MessagePack as well?

falcon/response.py
```diff
@@ -713,24 +720,34 @@ def add_link(self, target, rel, title=None, title_star=None,
 
         """)
 
-    def _wsgi_headers(self, media_type=None, py2=PY2):
-        """Convert headers into the format expected by WSGI servers.
+    def _set_media_type(self, media_type=None):
+        """wrapper around set_header to set a content-type
```
kgriffs: [nitpick] Capitalize "wrapper" and end the first line of the doctoring with a period.

falcon/constants/media_types.py
ofek: @kgriffs https://github.com/msgpack/msgpack/issues/194

`application/x-msgpack` or `application/msgpack`?

kgriffs: Makes sense, let's go with `x-msgpack`.

mayurmahajan: Looks like use of x- is strongly discouraged by RFC : https://tools.ietf.org/html/rfc6838

I will use later instead. If users need to use "application/x-msgpack" they can always set it in media_type.

kgriffs: Good catch. See also: https://tools.ietf.org/html/rfc6648

kgriffs:
Thanks!

Issue #933, opened 2016-10-24
Add referer attribute to the request and response objects


heckfi:
The Request and Response objects have various helper attributes that make working with headers easier. It would be nice if Falcon allowed for us to set the `referer` header via a helper attribute like `content_type`.

PR #1063, opened 2017-05-24
added referer attribute to request

mayurmahajan:
This is to fix issue : https://github.com/falconry/falcon/issues/933

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -296,6 +296,9 @@ class Request(object):
             A dict of name/value cookie pairs.
             See also: :ref:`Getting Cookies <getting-cookies>`
 
+        referer (str): Value of Referer header, or ``None`` if
+            the header is missing.
+
         options (dict): Set of global options passed from the API handler.
     """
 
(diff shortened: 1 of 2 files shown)


falcon/request.py
```diff
@@ -296,6 +296,9 @@ class Request(object):
             A dict of name/value cookie pairs.
             See also: :ref:`Getting Cookies <getting-cookies>`
 
+        referer (str): Value of Referer header, or ``None`` if
```
kgriffs: Would you mind moving this up to be next to the other header property descriptions?

tests/test_req_vars.py
```diff
@@ -622,6 +622,7 @@ def test_attribute_headers(self):
         auth = 'HMAC_SHA1 c590afa9bb59191ffab30f223791e82d3fd3e3af'
         agent = 'testing/1.0.1'
         default_agent = 'curl/7.24.0 (x86_64-apple-darwin12.0)'
+        referer = 'google.com'
```
kgriffs: [nitpick] Actual referer values would be full or partial URIs, e.g.:

```
https://example.com/something/something
```

or

```
/partial/uri
```

Not a big deal, since the property doesn't do any validation, but if you feel like changing it I wouldn't complain. :)

mayurmahajan: cool, i will change it.

PR #1066, opened 2017-05-26
feat: Add 5xx error classes

wssbck:
Adds following error classes, together with tests:

- [x] 501 Not Implemented
- [x] 504 Gateway Timeout
- [x] 505 HTTP Version Not Supported


wssbck:
@jmvrbanac @kgriffs Another PR with error classes.

falcon/errors.py
```diff
@@ -1250,6 +1250,59 @@ def __init__(self, title=None, description=None, **kwargs):
...
+                your own HTTP error class, and install an error handler
+                to convert it into an appropriate HTTP response for the
+                client
+
+            Note:
+                Falcon can process a list of ``tuple`` slightly faster
+                than a ``dict``.
+
+        headers (dict): Extra headers to return in the
+            response to the client (default ``None``).
```
wssbck: @jmvrbanac Those extra fields are present also in other classes, plus directly in the docstring of HTTPError. It was a bit confusing to me as well. Should we clean it all up?

jmvrbanac: @wssbck Wow! I don't know how I never noticed that before... Ok. Go ahead and just remove the extra entries that are in the code you're adding. We'll worry about cleaning up the remaining as a part of #1071. Thanks!

wssbck: @jmvrbanac done in all three cases

PR #1073, opened 2017-06-30
chore(travis): Adding gitter notification for Travis

jmvrbanac:


.travis.yml
```diff
@@ -70,6 +70,11 @@ notifications:
     on_failure: always
     use_notice: true
     skip_join: true
+  webhooks:
+    urls:
+      secure: "R4Hr754P2v69tWoy6vKz/4csb6ZHYRM/+8vTVV6ioDWyqsPqyGXAPiXDfZ617C7TLxJiTlbFIJPjlcH64estPg+RZ1NA4D8BrDKS2nKqHJ2Z5Jv6X5Ds6HkEUNnYXsuwsqNvZzhpPRcCKXXRpvmAbkNUSe8ftn4kz2zCOA9MBSY="
+    on_success: change
```
kgriffs: As far as I can tell, gitter doesn't notify the room when a new PR comes in. It may be nice to always notify when a PR passes the gate so we don't miss new PRs.

PR #1077, opened 2017-07-01
refactor: make the CompiledRouter pickleable

munircontractor:
The CompiledRouter is not pickle-able because the default responder uses
function within functions. As a result, the CompiledRouter cannot be
shared between processes using IPC mechanisms, which utilize pickles
internally. Replace the function within functions with partials of
module level functions so that the CompiledRouter can be pickled. Also,
change the traverse method to use the function wrapped by the partial to
preserve existing functionality.

All tests passed for Python 3.6 and 2.7. No breaking changes found.


falcon/cmd/print_routes.py
```diff
@@ -54,8 +54,8 @@ def traverse(roots, parent='', verbose=False):
                     if func.__name__ != 'method_not_allowed':
                         print('-->{0} {1}:{2}'.format(
                             method,
-                            inspect.getsourcefile(func),
-                            inspect.getsourcelines(func)[1]))
+                            inspect.getsourcefile(func.func),
```
kgriffs: [fix] This will break for regular responder objects, since they don't have a `func` attribute. This didn't show up in the tests because `test_cmd_print_api` is rather basic and doesn't actually test an API that had been configured with routes to actual resources. You'll need to add that test and then fix the issue here that it exposes.

falcon/responders.py
```diff
@@ -38,11 +45,15 @@ def create_method_not_allowed(allowed_methods):
             returned in the Allow header.
 
     """
-    def method_not_allowed(req, resp, **kwargs):
-        """Raise 405 HTTPMethodNotAllowed error"""
-        raise HTTPMethodNotAllowed(allowed_methods)
+    partial_method_not_allowed = partial(method_not_allowed, allowed_methods)
```
kgriffs: [observation] Although partials are less performant than closures, in this case I think it's OK since these code paths are less latency-sensitive.

munircontractor:
Thanks so much!

PR #1079, opened 2017-07-04
feat(routing): Add DateTimeConverter

kgriffs:


falcon/routing/converters.py
```diff
@@ -59,6 +65,28 @@ def convert(self, fragment):
...
+class DateTimeConverter(object):
+    """Converts a field value to a datetime.
+
+    Keyword Args:
+        format_string (str): String used to parse the param value
+            into a datetime. Any format recognized by strptime() is
+            supported (default ``'%Y-%m-%dT%H:%M:%SZ'``).
+    """
+
+    __slots__ = ('_format_string', '_strptime')
```
ueg1990: where is the '_strptime' instance variable defined/used?

falcon/routing/converters.py
```diff
@@ -59,6 +65,28 @@ def convert(self, fragment):
...
+    def convert(self, fragment):
+        try:
+            return strptime(fragment, self._format_string)
+        except ValueError:
+            return None
+
+
 BUILTIN = (
     ('int', IntConverter),
+    ('dt', DateTimeConverter),
```
ueg1990: There is no explicit objection to having this converter but what is the purpose of adding this? And where did the need come from? Only asking to understand requirements

Even though you provided format_string as a keyword argument, a user that may need to provide a different format will have to create a customer class (maybe by inheriting from DateTimeConverter) to provide a format of their choice. Is this ok or expected for usage?

falcon/routing/converters.py
kgriffs: 👍  Nice catch! This was something leftover from an experiment that I forgot to remove.

falcon/routing/converters.py
kgriffs: > There is no explicit objection to having this converter but what is the purpose of adding this? And where did the need come from? Only asking to understand requirements

Timestamps are sometimes used in URLs to group resources chronologically. I'm also thinking ahead to when we will add support for query string parameters in the URI templates, in which case this converter could be used in lieu of `req.get_param_as_datetime()`.

> Even though you provided format_string as a keyword argument, a user that may need to provide a different format will have to create a customer class (maybe by inheriting from DateTimeConverter) to provide a format of their choice. Is this ok or expected for usage?

This should become more clear once I add the documentation (in a future PR), but parameters can be passed into the URI template field expressions, obviating the need to subclass. E.g.,:

```py
api.add_route('/{day:dt("%Y-%m-%d")/messages', messages_resource)
```

ueg1990: nice! this usage made it pretty clear...thank you for the clarification :smile:

tests/test_uri_templates.py
```diff
@@ -164,6 +166,43 @@ def test_converter_rejections(client, uri_template):
...
+    result = client.simulate_get(path)
+
+    if dt_expected is None:
+        assert result.status_code == 404
+        assert not resource.called
+    else:
+        assert result.status_code == 200
+        assert resource.called
+        assert resource.captured_kwargs['start_year'] == 1961
+        assert resource.captured_kwargs['timestamp'] == dt_expected
```
jmvrbanac: Not the biggest fan of using conditional asserts as it generally means you have two different tests; however, it's not a blocking concern.

ueg1990: 👍 for moving the None use case into its own test case

I now remember I did something similar at work with a group of related use cases....I should break them out into separate tests...yay for learning via open source! 😄

PR #1082, opened 2017-07-07
feat(routing): Add UUIDConverter

kgriffs:


kgriffs:
@ueg1990 Not a bad idea. If nothing else, it will provide a self-documenting template for writing custom converters.

ueg1990:
It shouldn't really affect performance right?

ueg1990:
Since this issue is merged, can I open another issue and send a pull request?

kgriffs:
@ueg1990 I don't think you need to create a new issue in this case.

ueg1990:
ok, I will send a pull request :smile:  (i get to keep my streak going of atleast one contritbution to Falcon per year :stuck_out_tongue: )

PR #1083, opened 2017-07-11
feat(Request): Add support for several proxy "forwarded" headers

kgriffs:
Add support for some "forwarded" headers, included several new
attributes and a reworking of some of the existing code to better
facilitate code reuse and performance.

Also clean up a couple nits in the docstrings for the sake of
consistency.


falcon/media/base.py
```diff
@@ -7,7 +7,7 @@
 class BaseHandler(object):
     """Abstract Base Class for an internet media type handler"""
 
-    @abc.abstractmethod
+    @abc.abstractmethod  # pragma: no cover
```
kgriffs: Coverage was giving me grief on this line and the one below for no good reason that I could tell.

falcon/request.py
```diff
@@ -1495,3 +1697,36 @@ def __init__(self):
...
+    """
+
+    # NOTE(kgriffs): Use "client" since "for" is a keyword, and
+    # "scheme" instead of "proto" to be consistent with the
+    # falcon.Request interface.
+    __slots__ = ('by', 'client', 'host', 'scheme')
+
+    def __init__(self):
+        self.by = None
+        self.client = None
```
kgriffs: If you can think of a better name for this attr, please let me know. Per the inline comment, using "for" results in a SyntaxError on reference, and "forwarded_for" seemed a bit unwieldy.

jmvrbanac: Maybe target or destination?

kgriffs: TBH, I think target or destination more accurately describe the "by" parameter. According to RFC 7239:

> The "for" parameter is used to disclose information about the client that initiated the request and subsequent proxies in a chain of proxies.

Whereas:

>  The "by" parameter is used to disclose the interface where the request came in to the proxy server.

What do you think about the following variations?

```py
forwarded = req.forwarded[0]

forwarded.source
forwarded.dest

forwarded.source
forwarded.dest

forwarded.client
forwarded.target

forwarded.client
forwarded.dest

forwarded.client
forwarded.proxy

forwarded.client
forwarded.by
```

jmvrbanac: I think "destination" seems appropriate and generic enough
```
forwarded.source
forwarded.destination
--- or ---
forwarded.dest
```

kgriffs: Sounds good, I'll make the change. Thanks!

kgriffs:
OK, I've updated this now with the new names. Let me know how it looks and if I missed anything!

PR #1094, opened 2017-08-04
fix(constants): Clean up charset params in MEDIA_* and add MEDIA_GIF

kgriffs:


falcon/constants.py
```diff
@@ -10,14 +10,45 @@
     'TRACE',
 )
 
-MEDIA_JSON = 'application/json; charset=UTF-8'
+# NOTE(kgriffs): According to RFC 7159, most JSON parsers assume
+# UTF-8 and so it is the recommended default charset going forward,
+# and indeed, other charsets should not be specified to ensure
+# maximum interoperability. Therefore, we omit the charset parameter
+# below.
+MEDIA_JSON = 'application/json'
```
jmvrbanac: I'm wondering if this is going to have any adverse affects to people's test suites. This value gets propagated as the default content type right? Thoughts?

kgriffs: Hmm, good point. Let me revert it with a note to remove the charset param in 2.0

PR #1096, opened 2017-08-04
perf: Minor optimizations to make up for perf regression in 1.3

kgriffs:


requirements/bench
```diff
@@ -1,4 +1,7 @@
+django
 flask
 pecan
 bottle
-cherrypy
\ No newline at end of file
+
+guppy
```
jmvrbanac: Travis is having problems with guppy :(

kgriffs: 👍

PR #1104, opened 2017-08-23
<fix>(request): Fix unexpected CookieError during cookie parsing

yasek:
Fix unexpected CookieError by splitting Cookie header into parts and parsing them independently.


yasek:
@kgriffs Also, is it possible to include this fix into current release?

falcon/request.py
```diff
@@ -896,7 +896,13 @@ def cookies(self):
             # NOTE(tbug): We might want to look into parsing
             # cookies ourselves. The SimpleCookie is doing a
             # lot if stuff only required to SEND cookies.
-            parser = SimpleCookie(self.get_header('Cookie'))
+            cookie_header = self.get_header('Cookie', default='')
```
kgriffs: test: Could you add a test for the case when the `Cookie` header is missing?

falcon/request.py
```diff
@@ -896,7 +896,13 @@ def cookies(self):
...
+            parser = SimpleCookie()
+            for cookie_part in cookie_header.split('; '):
+                try:
+                    parser.load(cookie_part)
```
kgriffs: fix later: Calling `load()` multiple times will result in a performance degradation. I'll create an issue to address this in 1.4.

falcon/request.py
yasek: Fixed

PR #1108, opened 2017-09-08
update TestClient methods to use simulate_request

timc13:
This allows us to customize any requests by just extending
TestClient and overriding `simulate_request`

Here's an example use case:
I want to inject a custom header to all my requests.  To do that currently, 
I have to pass in a headers keyword argument to every `simulate_*` method call.

With this change, I can do this:
```
class CustomClient(TestClient):
    def simulate_request(self, *args, **kwargs):
        original_headers = kwargs.get('headers', {})
        kwargs['headers'] = original_headers.update({'CustomHeader': 'CustomValue'}
        return super().simulate_request(self, *args, **kwargs)
```

### DIFF
--- a/falcon/testing/client.py
+++ b/falcon/testing/client.py
@@ -592,6 +592,18 @@ class TestClient(object):
         client.simulate_get('/messages')
         client.simulate_head('/messages')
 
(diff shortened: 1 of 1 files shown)


timc13:
done!

Issue #1052, opened 2017-05-20
Update the FAQ to talk about consuming a JSON query string value


kgriffs:
See also: https://github.com/falconry/falcon/issues/1020

santeyio:
Hey @kgriffs, I'm a fan of the framework and would love to help out, thought I might be able to get my toes wet by doing some documentation -- would it be helpful if I sent you a PR for this?

kgriffs:
Hi @santeyio, thanks for contributing! This looks like a great addition to the FAQ. 👍

kgriffs:
Fixed by #1111

PR #1111, opened 2017-09-14
docs(faq): add a section on consuming JSON query string values

santeyio:
See #1052


santeyio:
what added section looks like when built locally: http://bit.ly/2xAyD8U

Issue #667, opened 2015-12-01
Split create_http_method_map?


yohanboniface:
The helper `create_http_method_map` does three things:
1. extract responders from a resource
2. add a default OPTIONS responder if not set
3. patch method map to add 405 responders for non supported verbs

This prevent from using the 2 and 3 when we don't need 1, for example when we take control over routing.

I suggest to split this helper in two, `create_http_method_map` in one hand, that would only do 1, and then call `complete_method_map` (naming suggestions welcome ;) ), that would take care of 2 and 3.

I can work on a PR for that.

Thoughts?

wbaelen:
This sounds like a very good direction.

kgriffs:
Sounds reasonable. The second method might be called `set_default_responders` or similar.

kgriffs:
Note: If we can't do this in a non-breaking way, it will have to wait for Falcon 2.0

santeyio:
@kgriffs I can send a PR for this one -- it looks like no ones currently working on this issue, right?

santeyio:
@kgriffs would it be better if this `set_default_responders` method to take a `method_map` as an argument or a `responder` object?

kgriffs:
Thanks @santeyio, I'll take a look!

kgriffs:
[Implemented]

PR #1115, opened 2017-09-18
refactor(routing): Refactor create_http_method_map into two methods (#667)

santeyio:
Pull out setting default responders from `create_http_method_map` into a separate method `set_default_reponders`. I'm not sure this is how the issue (#667) should be implemented -- mainly I'm not sure if we want `set_default_responders` to take a responder object or a method_map dict... I implemented it as a method_map since that involved the least changing of code. Let me know if I should implement it differently.


falcon/routing/util.py
```diff
@@ -109,6 +109,20 @@ def create_http_method_map(resource):
             if callable(responder):
                 method_map[method] = responder
 
+    return method_map
+
+
+def set_default_responders(method_map):
```
kgriffs: 👍 This looks good as-is.

falcon/routing/util.py
```diff
@@ -92,7 +92,7 @@ def create_http_method_map(resource):
             ``on_post(self, req, resp)``.
 
     Returns:
-        dict: A mapping of HTTP methods to responders.
+        dict: A mapping of HTTP methods to explicitly defined resource responders.
```
kgriffs: I think we had better not change the functionality of `create_http_method_map()` since that would be a breaking change and this issue is scheduled for a minor point release. Instead, what do you think about introducing a new function (say, `map_http_methods()`)? It could be used in lieu of `create_http_method_map()` at `api.py:348` and we could add a note to the docstring for `create_http_method_map()` saying that it's deprecated and will be removed in a future release (use the new methods instead).

santeyio: Right, that makes a lot of sense. Added the new method and reverted `create_http_method_map` back to its original form.

santeyio:
Not sure if something needs to be done about the coverage -- the `create_http_method_map` is now not run through the tests because the only place it's run is in api.py, and now it's been replaced by the `map_http_methods`

falcon/api.py
```diff
@@ -345,7 +345,8 @@ def add_route(self, uri_template, resource, *args, **kwargs):
         if '//' in uri_template:
             raise ValueError("uri_template may not contain '//'")
 
-        method_map = routing.create_http_method_map(resource)
+        method_map = routing.map_http_methods(resource)
+        method_map = routing.set_default_responders(method_map)
```
kgriffs: Assigning `method_map` to the return value makes it unclear to the reader whether `set_default_responders()` mutates the passed-in `method_map` in place. In languages that do not allow side-effects like this, there is no question, but Python is a different matter. What do you think about not having the method return anything and removing the assignment?

```py
routing.set_default_responders(method_map)
```

Alternatively, this method could be renamed to `get_default_responders()` and it would not mutate the passed-in argument, but return a dict containing the extra responders that could then be merged with `method_map`. This would avoid the side-effect but I'm not certain it is worth it.

santeyio:
@kgriffs, made the first change suggested... if you think it's worth changing `set_default_responders` to `get_default_repsponders` I'm down to submit a new PR for that. By the way, thanks so much for taking the time to review my code and give helpful feedback -- I'm sure you could have cleaned it up quickly yourself but I appreciate the opportunity to work on this stuff!

PR #1125, opened 2017-10-13
Integrate Static File Routes

nZac:
There needs to be some test written for this, however, with October growing long and my time being short I find it prudent to propose something.

The one change I wanted to make but don't have the time to, instead of using the prefixing, using a regular expression to match a path after the prefix. My guess is it could be done, but my expertise in the area of regex is limited at best.

LMK what else I am missing, likely docs?

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -25,6 +25,7 @@ from falcon.request import Request, RequestOptions
 import falcon.responders
 from falcon.response import Response, ResponseOptions
 import falcon.status_codes as status
+from falcon.statics import StaticSink
 from falcon.util.misc import get_argnames
 
 
@@ -349,6 +350,9 @@ class API(object):
         routing.set_default_responders(method_map)
         self._router.add_route(uri_template, method_map, resource, *args,
                                **kwargs)
+
+    def add_static_route(self, prefix, folder):
+        self.add_sink(StaticSink(prefix, folder), prefix)
 
     def add_sink(self, sink, prefix=r'/'):
         """Registers a sink method for the API.
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -24,8 +24,8 @@ from falcon.http_status import HTTPStatus
 from falcon.request import Request, RequestOptions
 import falcon.responders
 from falcon.response import Response, ResponseOptions
+from falcon.statics import Statics
 import falcon.status_codes as status
-from falcon.statics import StaticSink
 from falcon.util.misc import get_argnames
 
 
@@ -141,7 +141,8 @@ class API(object):
     __slots__ = ('_request_type', '_response_type',
                  '_error_handlers', '_media_type', '_router', '_sinks',
                  '_serialize_error', 'req_options', 'resp_options',
-                 '_middleware', '_independent_middleware', '_router_search')
+                 '_middleware', '_independent_middleware', '_router_search',
+                 '_static_routes')
 
     def __init__(self, media_type=DEFAULT_MEDIA_TYPE,
                  request_type=Request, response_type=Response,
@@ -149,6 +150,7 @@ class API(object):
                  independent_middleware=False):
         self._sinks = []
         self._media_type = media_type
+        self._static_routes = []
 
         # set middleware
         self._middleware = helpers.prepare_middleware(
@@ -352,7 +354,7 @@ class API(object):
                                **kwargs)
 
     def add_static_route(self, prefix, folder):
-        self.add_sink(StaticSink(prefix, folder), prefix)
+        self._static_routes.add(prefix, Statics(prefix, folder))
 
     def add_sink(self, sink, prefix=r'/'):
         """Registers a sink method for the API.
--- a/falcon/statics.py
+++ b/falcon/statics.py
@@ -1,0 +1,43 @@
+import copy
+import mimetypes
+import os.path
+
+import falcon
+
+
+class StaticSink:
+
+    _additional_types = {
+        '.css': 'text/css',
+        '.js': 'application/javascript',
+        '.html': 'text/html',
+        '.svg': 'image/svg+xml',
+
+        '.png': 'image/png',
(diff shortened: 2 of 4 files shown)


falcon/statics.py
```diff
@@ -0,0 +1,43 @@
...
+        suffix = os.path.splitext(path)[1]
+
+        try:
+            resp.stream = open(path, 'rb')
+        except IOError:
+            raise falcon.HTTPNotFound()
+
+        resp.content_type = self.mime_types.get(suffix, 'application/octet-stream')
+
+        raise falcon.HTTPStatus('200 OK')
```
kgriffs: No need to raise. The framework already defaults to 200 OK (and simply setting `resp.status` is preferred in any case).

falcon/statics.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, prefix, parent_folder):
+        self.prefix = prefix
+        self.parent_folder = os.path.abspath(parent_folder)
+        self.mime_types = copy.copy(mimetypes.encodings_map)
+
+        self.mime_types.update(self._additional_types)
+
+    def __call__(self, req, resp):
+        stripped = req.path.replace(self.prefix, '')
+        cleaned = stripped.replace('../', '').lstrip('/')
```
kgriffs: It may be better to instead run `stripped` through `os.path.normpath()` and then explicitly raise `falcon.HTTPNotFound()` if the result starts with `'..'` or `/` (rather than checking first if the file exists with those characters stripped). This has the added benefits of allowing harmless redundant separators and level references, replacing forward slashes with backslashes on Windows, and ensuring that malformed requests aren't inadvertently mapped to existing files.

falcon/statics.py
```diff
@@ -0,0 +1,43 @@
...
+        '.jpeg': 'image/jpeg',
+        '.jp2': 'image/jpeg',
+        '.gif': 'image/gif',
+        '.tiff': 'image/tiff',
+    }
+
+    def __init__(self, prefix, parent_folder):
+        self.prefix = prefix
+        self.parent_folder = os.path.abspath(parent_folder)
+        self.mime_types = copy.copy(mimetypes.encodings_map)
```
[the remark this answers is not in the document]
kgriffs: Also, I think it would be helpful to factor this out to initialize a new option, `falcon.ResponseOptions.static_media_types`, and use that here. Then, people can add additional types to that dict as needed.

nZac:
@kgriffs I updated as requested and add a test. The tests in this lib have a variety of different methods so I did what felt comfortable for me, if it isn't to standard let me know.

Also, I am not sure where to plug this in now that I converted to using a list of static routes on the `API` instance. My guess would be that `_get_responder` would be the best place.

Docs are still a question mark as well.

falcon/api.py
```diff
@@ -350,6 +353,9 @@ def add_route(self, uri_template, resource, *args, **kwargs):
         self._router.add_route(uri_template, method_map, resource, *args,
                                **kwargs)
 
+    def add_static_route(self, prefix, folder):
+        self._static_routes.add(prefix, Statics(prefix, folder))
```
kgriffs: I think it could be helpful to always insert each new prefix at the head of the list. This would ensure a predictable behavior, which we can (and should) document and include in our tests. With this behavior in place, we can note in the docs that duplicate prefixes will override previous ones, and that a more specific prefix may be added _after_ a more general one (e.g., when "/foo/bar" is added after "/foo", a resource such as "/foo/bar/thing.js" would be mapped to the "/foo/bar" route, while "/foo/tmp/thingtoo.js" would be mapped to the "/foo" route.

```py
self._static_routes.insert(0, Statics(prefix, folder))

Comment (kgriffs) [reply falcon/api.py]: We should also take the opportunity to verify that the prefix begins with a forward slash, and normalize the string so that it always ends with one as well (so that we match on the complete path segment; matching on the complete segment should also be verified in the unit or functional tests).

File: tests/test_statics.py [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+    def _get_stream(self, path):
+        return path
+
+
+def test_serve_static_file():
+    static = MockStatics('/static', '/var/www/statics')
+
+    req = Request(testing.create_environ(
+        host='test.com',
+        path='/static/../css//test.css',
```
kgriffs: We'll want to try lots of crazy paths to ensure a malicious client can't break out. You can use ```pytest.mark.parametrize``` to make tests like this a little easier.

tests/test_statics.py
```diff
@@ -0,0 +1,29 @@
+import os
+
+from falcon.request import Request
+from falcon.response import Response
+from falcon.statics import Statics
+import falcon.testing as testing
+
+
```
kgriffs: This test module looks like a good start. Once you have the unit tests fleshed out, it would be great if you could also add a few functional tests ala `test_sinks.py` just to make sure everything is working end-to-end. Along the way, it may make sense to allow overriding the statics type (similar to ```API._request_type```). This would make testing easier while also providing a way for developers to customize the behavior.

falcon/statics.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def _get_stream(self, path):
+        with open(path, 'rb') as f:
+            return f
+
+    def __call__(self, req, resp):
+        stripped = req.path[len(self.prefix):]
+        normalized = os.path.normpath(stripped).lstrip('/')
+
+        if normalized.startswith('../'):
```
kgriffs: Also ensure it does not start with `'/'`. Assuming `self.prefix` ends in a forward slash, there should be no reason why the normalized path should start with one. And if it does (and there should be a test case for this), you end up jumping out of `self.parent_folder`:

```bash
In [8]: os.path.join('/usr/share/nginx/www', '/etc')
Out[8]: '/etc'

```

kgriffs:
I went ahead and finished the work on this so we could include it in 1.4. Please review #1157

PR #1130, opened 2017-10-30
feat(API): Support additional HTTP verbs from RFC 3253

rhemz:
This PR extends the list of supported HTTP verbs to encompass oft-used extension methods e.g. REPORT.  This has no impact on existing method support or routing, simply allows falcon users to implement said additional handlers if desired.  It is worth noting that many browsers do not support these additional methods yet, but they are recognized by the IANA and supported by virtually all HTTP client libraries/toolkits. 

HTTP Method listing (IANA): https://www.iana.org/assignments/http-methods/http-methods.xhtml

RFC 3253: https://tools.ietf.org/html/rfc3253

### DIFF
--- a/falcon/constants.py
+++ b/falcon/constants.py
@@ -1,3 +1,4 @@
+# RFC 7231, 5789 methods
 HTTP_METHODS = (
     'CONNECT',
     'DELETE',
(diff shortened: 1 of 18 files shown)


rhemz:
The failing Travis CI build looks to be unrelated to this change.

falcon/constants.py
```diff
@@ -10,6 +11,19 @@
     'TRACE',
 )
 
+# RFC 3253 methods
+WEBDAV_METHODS = (
```
kgriffs: Just to be on the safe side, I'd like to see some tests for one or two of these methods. The tests should check that a resource can successfully implement responders for these, and also that a request for one of these methods triggers the default `405 HTTPMethodNotAllowed` responder when the method isn't implemented for a specific resource.

rhemz: I think i've addressed this now with updates to `test_http_method_routing`.

`on_report() was implemented in `ThingsResource`, with a test to exercise the route, and a few of the test_method_not_allowed methods were updated to also iterate through the webdav methods.

Issue #1116, opened 2017-09-21
Incorrect handling of commas inside a Forwarded header


vfaronov:
To implement [`Request.forwarded`](http://falcon.readthedocs.io/en/stable/api/request_and_response.html#Request.forwarded), Falcon splits the incoming Forwarded header on commas. Because of this, it cannot correctly handle forwarded-elements that contain a comma inside a quoted-string.

To reproduce, run the following `fwd_demo.py`:

```
import falcon

class ForwardedResource(object):
    def on_get(self, req, resp):
        resp.body = req.forwarded[-1].src

app = falcon.API()
app.add_route('/fwd', ForwardedResource())
```

under Gunicorn:

```
$ gunicorn fwd_demo:app
```

and send requests like this:

```
curl -s localhost:8000/fwd -H 'Forwarded: for=1.2.3.4;some="extra,info"'
```

The expected response is `1.2.3.4`, but the actual response is empty.

A more correct solution would be to parse the Forwarded header carefully with regard to quoted-strings. However, this is complicated by the fact that the Forwarded header may be partly controlled by an external attacker, who might submit a request with:

```
Forwarded: for=_trusted_addr;by="
```

Unless a reverse proxy eagerly validates and fixes/strips such a header, appending another forwarded-element might give:

```
Forwarded: for=_trusted_addr;by=", for=_real_addr
```

In this case, `req.forwarded[-1].src` ought to be `_real_addr`, not `_trusted_addr`.

See [aiohttp’s implementation of Forwarded](https://github.com/aio-libs/aiohttp/blob/1609acd5802364fa3bd11adfb34a4b2b0d07fc42/aiohttp/web_request.py#L188-L246) for an example of how to deal with this correctly (I hope).

“`Request.forwarded` should not be trusted anyway” would be a poor solution, I think. As long as a reverse proxy is correctly configured to append a header, that header may be and often is trusted by the application for important purposes such as rate limiting.

kgriffs:
Hi @vfaronov, good catch. Let's schedule this issue for the next release.

PR #1131, opened 2017-10-30
fix(Request): Properly handle quoted strings in forwarded header

Just-Sieb:
I have implemented a fix for issue #1116. 

In the case of the malformed forwarded header I decided to just say the whole forwarded is malformed and set it to None.

### DIFF
--- a/falcon/forwarded.py
+++ b/falcon/forwarded.py
@@ -1,0 +1,184 @@
+# Based on code from the aiohttp project, Copyright 2013-2017 by Nikolay Kim and
+# Andrew Svetlov, with modifications for the Falcon project by Kurt Griffiths.
+#
+# See also:
+#
+#   https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_request.py
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+import string
+
+from falcon.util.uri import unquote_string
+
+
+# '-' at the end to prevent interpretation as range in a char class
+_TCHAR = string.digits + string.ascii_letters + r"!#$%&'*+.^_`|~-"
+
+_TOKEN = r'[{tchar}]+'.format(tchar=_TCHAR)
+
+# qdtext includes 0x5C to escape 0x5D ('\]')
+# qdtext excludes obs-text (because obsoleted, and encoding not specified)
+_QDTEXT = r'[{0}]'.format(
+    r''.join(chr(c) for c in (0x09, 0x20, 0x21) + tuple(range(0x23, 0x7F))))
+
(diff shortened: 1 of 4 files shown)


vfaronov:
The [quoted-string](https://tools.ietf.org/html/rfc7230#section-3.2.6) syntax used by RFC 7239 permits escaped quotes:

    Forwarded: for=1.2.3.4;some="other \"info"

Your implementation does not deal with these correctly.

> In the case of the malformed forwarded header I decided to just say the whole forwarded is malformed and set it to None.

May I suggest reflecting this in the [docs](http://falcon.readthedocs.io/en/stable/api/request_and_response.html#Request.forwarded).

kgriffs:
> In the case of the malformed forwarded header I decided to just say the whole forwarded is malformed and set it to None.

That's probably OK, although we should advise that the app should be suspicious of malicious intent in this case and if it decides to respond to the request, it should use safe defaults for any information it uses in the header. 

> Your implementation does not deal with these correctly.

At first glance, I'm not certain that aiohttp deals with escaped quotes either (but I could be wrong). I think we need to handle this in any case. See also https://tools.ietf.org/html/rfc7230#section-3.2.6 

    The backslash octet ("\") can be used as a single-octet quoting
    mechanism within quoted-string and comment constructs.  Recipients
    that process the value of a quoted-string MUST handle a quoted-pair
    as if it were replaced by the octet following the backslash.

     quoted-pair    = "\" ( HTAB / SP / VCHAR / obs-text )

Finally, I'd like to explore the performance of the different approaches (current, this patch, aiohttp's).  If it is faster to do the comma-splitting as currently implemented, we can simply check for the presence of any double-quote's in the string and only use the slower algorithm if any are present.

Thoughts?

vfaronov:
> At first glance, I'm not certain that aiohttp deals with escaped quotes either (but I could be wrong).

It does: see, for example, [`test_single_forwarded_header_custom_param`](https://github.com/aio-libs/aiohttp/blob/dbb8d9a63f6ffba0ae7e4592158045e54c26433a/tests/test_web_request.py#L277).

> we can simply check for the presence of any double-quote's in the string and only use the slower algorithm if any are present

Sure, that’s an option.

kgriffs:
Ah, yes, I misread the regex. It does indeed include VCHARS in the character set.

@Just-Sieb Do you have some bandwidth to get this PR over the finish line this week?

- [ ] Resolve `quoted-pair`'s per [RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.2.6). Add tests to cover this case.
- [ ] Benchmark the relative performance of the current implementation vs. the one in this PR vs. the one used by aiohttp. This could be as simple as pasting some test functions into IPython and using `%timeit`. Be sure to check Python 2.7, Python 3.5 and 3.6 (via, e.g., pyenv).
- [ ] Based on the benchmark results above, optimize this PR's implementation. Use alternative strategies if necessary, based on whether the header value contains any double quotes.

kgriffs:
@Just-Sieb No worries if you don't have the time. If you don't mind, I may just go ahead and finish up the work on this so we can get 1.4 out the door.

kgriffs:
OK, I went ahead and used the aiohttp algorithm with a few modifications. It is faster than Just-Sieb's original solution, and more robust in dealing with malformed values and quoted-pair's. However, the regex solution isn't quite as fast as the current implementation in master, but it is close enough (within a few microseconds) that I didn't think it would be worth maintaining two code paths.

kgriffs:
Hmm, it seems codecov is on the fritz again?

kgriffs:
@jmvrbanac If this looks good to you, feel free to merge without waiting on codecov.

PR #1133, opened 2017-10-31
docs(api/errors): fix syntax error in example code

GriffGeorge:
Use double quotes for strings, since there is a single apostrophe
inside.


GriffGeorge:
Looks like a bare except clause in an unrelated test is causing the pep8 CI check to fail. Added the fix for that to make it pass.

Issue #1107, opened 2017-08-29
Optimize request cookie parsing


kgriffs:
In 1.3, the request cookie parser was changed to parse individual cookies so that malformed ones could be skipped, rather than giving up on parsing the entire header. However, this is significantly slower than parsing the entire header value in one shot, esp. since the SimpleCookie parser is rather inefficient.

We should be able to do better. We can start with the SimpleCookie code from the latest CPython 3.7 code base, optimize it, and make it compatible with Python 2.

See also:

[request.py#L894](https://github.com/falconry/falcon/blob/master/falcon/request.py#L894)

santeyio:
Assuming no one else is working on this, I can take a shot at this

kgriffs:
@santeyio that would be great, thanks!

kgriffs:
@santeyio P.S. If you like, you might want to first register for Hacktoberfest so this will count towards that. 🎃

kgriffs:
The bulk of this work has been moved to a standalone project:

https://github.com/pyrates/biscuits/

We'll watch this project and add support for it in Falcon once it is mature.

PR #1134, opened 2017-11-06
feat(util): Add internal cookie parsing support

santeyio:
Right now I'm keeping to what the python standard library cookie module considers valid cookies. Removes all of the additional features from the standard library module that we don't need or care about for doing simple cookie parsing. 

I wasn't sure where the best place to put the cookie parser and helper methods -- I ended up putting them in with the utils, but let me know if there's somewhere else that I should put them that makes more sense. 

Fixes (#1107)

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -34,19 +34,14 @@ from wsgiref.validate import InputWrapper
 
 import mimeparse
 import six
-from six.moves import http_cookies
 
 from falcon import DEFAULT_MEDIA_TYPE
 from falcon import errors
 from falcon import request_helpers as helpers
 from falcon import util
 from falcon.media import Handlers
+from falcon.util.cookies import parse_cookies
 from falcon.util.uri import parse_host, parse_query_string, unquote_string
-
-# NOTE(tbug): In some cases, http_cookies is not a module
-# but a dict-like structure. This fixes that issue.
-# See issue https://github.com/falconry/falcon/issues/556
-SimpleCookie = http_cookies.SimpleCookie
 
 DEFAULT_ERROR_LOG_FORMAT = (u'{0:%Y-%m-%d %H:%M:%S} [FALCON] [ERROR]'
                             u' {1} {2}{3} => ')
@@ -893,19 +888,8 @@ class Request(object):
     @property
     def cookies(self):
         if self._cookies is None:
-            # NOTE(tbug): We might want to look into parsing
-            # cookies ourselves. The SimpleCookie is doing a
-            # lot if stuff only required to SEND cookies.
             cookie_header = self.get_header('Cookie', default='')
-            parser = SimpleCookie()
-            for cookie_part in cookie_header.split('; '):
-                try:
-                    parser.load(cookie_part)
-                except http_cookies.CookieError:
-                    pass
-            cookies = {}
-            for morsel in parser.values():
-                cookies[morsel.key] = morsel.value
+            cookies = parse_cookies(cookie_header)
 
             self._cookies = cookies
 
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -34,6 +34,7 @@ from wsgiref.validate import InputWrapper
 
 import mimeparse
 import six
+from six.moves import http_cookies
 
 from falcon import DEFAULT_MEDIA_TYPE
 from falcon import errors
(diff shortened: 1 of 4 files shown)


santeyio:
I still need to add a few tests to cover some edge cases that the cookie parsing deals with. Will get to that in the next few days.

santeyio:
I think the only thing that's failing in travis is the stuff fixed in PR #1136

davidbgk:
Hello there, we (@pyrates) are working on a similar feature for [Roll](https://github.com/pyrates/roll) and are still hesitating on how to implement it. There are 3 main challengers as of now:

* using `http.cookies` from the standard lib with [known limitations](http://cookies.readthedocs.io/en/latest/#features)
* using [Cookies](http://cookies.readthedocs.io/en/latest/) but it's not really active to say the least with somehow [critical issues](https://gitlab.com/sashahart/cookies/issues/4) stagnant
* using custom implementations [like did Sanic](https://github.com/channelcat/sanic/blob/master/sanic/cookies.py) and that PR to strip off whatever is possible and speed it up

I'd love to get your feedback on these reflections and why you finally chose the latter. Maybe we can join our efforts and create synergies across micro-frameworks to maintain pico parts loosely coupled :)

falcon/request.py
```diff
@@ -893,19 +894,8 @@ def params(self):
     @property
     def cookies(self):
         if self._cookies is None:
-            # NOTE(tbug): We might want to look into parsing
-            # cookies ourselves. The SimpleCookie is doing a
-            # lot if stuff only required to SEND cookies.
             cookie_header = self.get_header('Cookie', default='')
-            parser = SimpleCookie()
```
davidbgk: I guess you can remove import of `http_cookies`/`SimpleCookie`

falcon/request.py
```diff
@@ -893,19 +894,8 @@ def params(self):
...
-            for cookie_part in cookie_header.split('; '):
-                try:
-                    parser.load(cookie_part)
-                except http_cookies.CookieError:
-                    pass
-            cookies = {}
-            for morsel in parser.values():
-                cookies[morsel.key] = morsel.value
+            cookies = parse_cookies(cookie_header)
```
davidbgk: Avoid parsing if cookie_header equals default (`''`)?

falcon/request.py
santeyio: right, I did do that in an earlier commit but it seems to have broken the tox 'hug' env build, wanted to ask @kgriffs about that before removing the `SimpleCookie` import?

santeyio:
@davidbgk I think the main motivation for implementing our own cookie parsing is twofold. 

1) Performance reasons. `SimpleCookie` has a bunch of stuff built in around sending cookies, and we don't need that, so removing extraneous code. 

2) We want to be able to parse cookie headers that may have a malformed cookie without throwing out the rest of the cookies (`SimpleCookie` tosses the whole header if there's an illegal cookie). 

That's my understanding of the reasoning, not sure if @kgriffs would add / modify anything in the above.

yohanboniface:
> I think the main motivation for implementing our own cookie parsing is twofold.

Those are indeed legit reasons, but have you considered implementing this in a third party module?

Seems we are in a situation where every Python web framework has its own cookies parsing, but at the end all have the exact same need.
Seems to me that by joining the effort we may have a better implementation (and maintained in the long term).
Now that may be with a new dedicated package, or that may be using and improving an existing one (such as [Cookies](http://cookies.readthedocs.io/en/latest/)).

santeyio:
Well, I certainly wouldn't be opposed to setting this up as a standalone library and putting it up on pip. I didn't realize how confusing the state of cookie parsing was in python land until I started researching for this milestone. I don't know falcon's attitude towards using third party packages though -- @kgriffs what do you think about this?

kgriffs:
@davidbgk @santeyio It seems we may be working toward similar goals. Perhaps it would make sense after all to join forces on a standalone package that both projects (and potentially others) could consume. I agree with @yohanboniface that it is silly for the community to duplicate its efforts in this space. As long as the implementation is performant (should include performance tests) and commits to maintaining an exhaustive test suite, I wouldn't have any problem using it in Falcon. It could potentially be an optional "light-up" dependency similar to how ``ujson`` is treated in the framework today.

davidbgk:
@kgriffs thanks for your answer, that's indeed our intention! We initiated https://github.com/pyrates/biscuits/ since then if you want to join the party :)

kgriffs:
Excellent, thanks everyone. We'll watch https://github.com/pyrates/biscuits/ and add support for it in Falcon when it's ready.

PR #1140, opened 2017-11-27
fix(hooks): Allow passing extra args positionally.

kgriffs:
The framework always passes extra args as keyword arguments. However, when an app calls a responder directly, it might use positional arguments instead, so we need to handle that case. This might happen, for example, when overriding a resource and calling a responder via super() in order to introduce a new version of the resource side-by-side.


kgriffs:
Thanks @jmvrbanac !

PR #1150, opened 2017-12-01
test: Cover testing.simulate_* free functions

kgriffs:
In doing some test coverage testing locally, I discovered that somehow we neglected to cover these functions when TestClient was last modified. With this change, my local coverage results now show 100%.


kgriffs:
@jmvrbanac Codecov seems to be rather unreliable lately... we may have to remove the gate for it and test manually on our local machines before merging each PR.

PR #1157, opened 2017-12-07
feat(routing): Implement static routing

### LINKED ISSUE #219: Route to static file
On a PaaS you may not be able to server static files from the web server directly, so let's make it easy to do this in Falcon.

Depends on #181

kgriffs:
Fixes #219


falcon/routing/static.py
```diff
@@ -0,0 +1,100 @@
...
+        self._downloadable = downloadable
+
+    def match(self, path):
+        """Check whether the given path matches this route."""
+        return path.startswith(self._prefix)
+
+    def __call__(self, req, resp):
+        """Resource responder for this route."""
+
+        # import pdb; pdb.set_trace()
```
nZac: You might want to remove this

falcon/api.py
```diff
@@ -350,6 +352,53 @@ def add_route(self, uri_template, resource, *args, **kwargs):
...
+            directory (str): The source directory from which to serve files.
+            downloadable (bool): Set to ``True`` to include a
+                Content-Disposition header in the response. The "filename"
+                directive is simply set to the name of the requested file.
+
+        """
+
+        self._static_routes.insert(
+            0,
+            routing.StaticRoute(prefix, directory, downloadable=downloadable)
```
nZac: I like the addition of a `downloadable` flag, that is a handy feature.

falcon/routing/static.py
kgriffs: Oops, nice catch!

PR #1190, opened 2017-12-22
Shortcut parameter `json` to set JSON body directly in simulate_request

vytas7:


falcon/testing/client.py
```diff
@@ -18,7 +18,16 @@
 WSGI callable, without having to stand up a WSGI server.
 """
 
-import json  # Don't use util.json so we can sanity-test compatibility with ujson output
+try:
+    from ujson import dumps as json_dumps
+except ImportError:
+    from json import dumps as json_dumps
+
+try:
```
vytas7: Note that these separate ``try`` ... ``except``s for ``json_dumps`` and ``json_load`` blocks look somewhat ugly, but I was not hitting the second line in coverage checks otherwise.

Maybe they could be listen on comma? Style suggestions/preferences welcome since I am not that familiar with Falcon codebase.

falcon/testing/client.py
```diff
@@ -318,6 +323,11 @@ def simulate_request(app, method='GET', path='/', query_string=None,
                 prefix=False,
             )
 
+        if json is not None:
+            body = util.json.dumps(json)
```
kgriffs: Should probably set ``ensure_ascii=False`` here.

falcon/testing/client.py
```diff
@@ -282,6 +283,10 @@ def simulate_request(app, method='GET', path='/', query_string=None,
                 Accepts both byte strings and Unicode strings
                 (default: ``None``). If a Unicode string is provided,
                 it will be encoded as UTF-8 in the request.
+            json(JSON serializable): A JSON document to send as the body
```
kgriffs: It may be helpful to replace "send" with "serialize" to reinforce to the reader that they are expected to pass a dict, list, etc. rather than the already-serialized JSON string.

falcon/testing/client.py
```diff
@@ -18,13 +18,13 @@
...
 import platform
 import re
 import wsgiref.validate
 
 from six.moves import http_cookies
 
+from falcon import util
 from falcon.testing import helpers
 from falcon.testing.srmock import StartResponseMock
 from falcon.util import CaseInsensitiveDict, http_date_to_dt, to_query_str
```
kgriffs: I think it would be more consistent with the existing code to import ``json`` here and use it directly, rather than ``from falcon import util``.

tests/test_utils.py
```diff
@@ -529,6 +529,29 @@ def test_query_string_in_path(self):
         with pytest.raises(ValueError):
             client.simulate_get(path='/thing?x=1')
 
+    def test_simulate_json_body(self):
```
kgriffs: Please use ``pytest.mark.parametrize`` to test a few different types for ``document`` (e.g., dict, list, int, string), just to ensure we don't ever have a regression related to this.

falcon/testing/client.py
vytas7: Is that to better test the framework support for request payload encoding?

And in this case I guess one also needs to ensure UTF-8 in the output, and make sure request ``Content-Type`` specifies ``charset=utf-8``?

falcon/testing/client.py
vytas7: The problem is that the parameter is also called ``json`` which would shadow or be shadowed by the import name.

I could alias that to, for instance, ``util_json``, or do like testing/request.py``, I have tried something in that vein (see my first commit in the PR), but it got sort of ugly to  to support both ``json`` and ``ujson``.

How about using just ``json``:

```
@@ -18,13 +18,14 @@ This package includes utilities for simulating HTTP requests against a
 WSGI callable, without having to stand up a WSGI server.
 """
 
+from json import dumps as json_dumps, loads as json_loads
 import platform
 import re
 import wsgiref.validate
```

falcon/testing/client.py
```diff
@@ -18,16 +18,17 @@
...
 import re
 import wsgiref.validate
 
 from six.moves import http_cookies
 
+from falcon.constants import MEDIA_JSON
 from falcon.testing import helpers
 from falcon.testing.srmock import StartResponseMock
 from falcon.util import CaseInsensitiveDict, http_date_to_dt, to_query_str
+from falcon.util import json as util_json
```
vytas7: Have gone this way since the pattern is sort of consistent with other parts of the project (cf ``falcon/response.py:37``: ``from falcon.util.uri import encode as uri_encode``, 
``falcon/response.py:38``: ``from falcon.util.uri import encode_value as uri_encode_value``).

But of course shout if this has to be changed :smiley:

kgriffs:
[Manually tested coverage]

```
TOTAL                                    2680      0    773      0   100%
```

PR #1191, opened 2018-01-04
fix(media): Fix failing response media serialization on py27+ujson

vytas7:


falcon/media/json.py
```diff
@@ -18,4 +18,7 @@ def deserialize(self, raw):
             )
 
     def serialize(self, media):
-        return json.dumps(media, ensure_ascii=False).encode('utf-8')
+        result = json.dumps(media, ensure_ascii=False)
+        if isinstance(result, bytes):
```
kgriffs: Since Python 3 never returns ``bytes``, we might as well add a Python version check here to short-circuit the ``isinstance()`` and save a few cycles.

kgriffs:
Thanks again for your help!

PR #1195, opened 2018-01-05
doc(testing.Result): Clarify type returned by json may not be a dict

kgriffs:


falcon/testing/client.py
```diff
@@ -72,9 +72,9 @@ class Result(object):
             under Python 2.6 and 2.7, and of type ``str`` otherwise.
             If the content type does not specify an encoding, UTF-8 is
             assumed.
-        json (dict): Deserialized JSON body. Will be ``None`` if the body has
-            no content to deserialize. Otherwise, raises an error if the
-            response is not valid JSON.
+        json (JSON serializable): Deserialized JSON body. Will be ``None`` if
```
vytas7: Sorry about forgetting these in my pull request.

I think the same applies to ``falcon/testing/resource.py``: ``85``:
```
        json (dict): Default JSON document to use in responses.
        <...>
```

PR #1202, opened 2018-01-11
JSON custom default and object hook

### LINKED ISSUE #1192: Extend falcon.media.JSONHandler to support specifying a ``default()`` type handler
Per [the docs](https://docs.python.org/3.6/library/json.html#basic-usage):

> If specified, *default* should be a function that gets called for objects that can’t otherwise be serialized. It should return a JSON encodable version of the object or raise a TypeError. If not specified, TypeError is raised.

Also consider providing a way to specify *object_hook* when deserializing to support round-tripping.

In either case, include a note in the docs that specifying one or both of these options will significantly reduce processing speed for JSON documents.

jhildreth:
fixes #1192 

Here's my first attempt at 1192. I made my best guess as to where to document these changes...

A couple questions:

- In the updated documentation, should we link to the Python json docs as in the initial issue, or is what is here sufficient?
- Is there somewhere else the SimpleTestObject class should live, given that it's duplicated in both test_request_media.py and test_response_media.py?

### DIFF
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -126,6 +126,9 @@ objects.
 Note that specifying either or both of these functions will significantly
 increase the time it takes to process JSON documents.
 
+Also note that neither of these are able to be specified if ujson is installed,
+as it does not currently support these parameters.
+
 Supported Handler Types
 -----------------------
 
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -112,6 +112,20 @@ MessagePack, this can be easily done in the following manner:
     api.resp_options.media_handlers.update(extra_handlers)
 
 
+JSON Custom Serialization and Deserialization of Objects
+--------------------------------------------------------
+
+When instantiating a JSONHandler, a function may be passed as the ``default``
+parameter which provides custom serialization of objects which are not
+otherwise JSON serializable.
+
+Similarly, when instantiating a JSONHandler, a function may be passed as the
+``object_hook`` parameter, which provides custom deserialization of JSON into
+objects.
+
+Note that specifying either or both of these functions will significantly
+increase the time it takes to process JSON documents.
+
 Supported Handler Types
 -----------------------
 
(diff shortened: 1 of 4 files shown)


jhildreth:
Ah, looks like ujson does not support the object_hook parameter on ujson.loads or the default parameter on ujson.dumps.
They have a closed [issue](https://github.com/esnme/ultrajson/issues/95) that points out the lack of support for object_hook, but no resolution. There is an open [issue](https://github.com/esnme/ultrajson/issues/124) that also refers to the lack of support for default.
I'm not sure how it would be best to address this.

jhildreth:
So, I've thought about a few different ways to handle the fact that ujson doesn't support these kwargs, and locally I've implemented what I feel like is the best idea I've come up with. I'm checking once if ujson is installed when the JSONHandler is instantiated, using pkgutil.find_loader. I didn't expect that this would introduce a lot of performance overhead. I'm then conditionally setting the contents of a couple of dictionaries in JSONHandler, which can be unpacked in the call to json.dumps in serialize and the call to json.loads in deserialize (so that the kwargs default/object_hook are not passed if ujson is installed).

I'm also raising an exception if either argument is specified when JSONHandler is instantiated if ujson is installed (so that there would be an early indication of a problem, rather than the parameter(s) being silently ignored by JSONHandler).

I've also modified the tests I wrote to expect these exceptions when ujson is installed - having the tests set up like this accomplishes what I'm going for, but feels just a little funny.

I plan to think about this a little bit more before pushing up these changes. I'm certainly open to other suggestions!

jhildreth:
I'm happy to work on the test coverage a bit, but I think I'll wait for a little feedback on if I'm even headed in the direction we want to go, first. :-)

kgriffs:
Hi @jhildreth, thanks for your patience. I think that if we implement https://github.com/falconry/falcon/issues/1206 it will make your work a lot easier. We can stipulate that if a third-party json library is configured, it must support the hooks in question in order to use this feature.

kgriffs:
Thanks for your work on this! It really helped us compare and contrast the different ways we could approach this problem. That being said, after discussing this with the maintainers team we ultimately decided to go with #1404. If you have a minute to take a look and provide your feedback, that would be great.

jhildreth:
Sure thing! I glanced at #1404, but honestly it's been a little while since I looked at this code, so I'm not as familiar with the code as I could be. It looks like a lot of good thought has gone into it!

Issue #584, opened 2015-08-04
Alternate routes to the same resource


kgriffs:
A frequent question from new community members is how to use a single resource class to handle requests for a single item as well as for multiple items. The answer up until now is that the framework doesn't really support this model. The idiomatic way to handle this is as [explained in the FAQ](https://falcon.readthedocs.org/en/stable/community/faq.html#how-do-i-implement-both-posting-and-geting-items-for-the-same-resource).

There have also been attempts to solve this via decorators, as in PR #389. However, it is my belief that this approach requires more typing (is less DRY), and generally makes it more difficult to reason about the entire routing namespace, since the routing definitions are decentralized, being sprinkled about multiple methods in multiple classes and potentially multiple modules. This runs contrary to one of the basic design decisions of the Falcon framework in the way routing is defined vs. other frameworks that prefer object-based or decorator-based routing. 

That being said, I understand there is some value in grouping related responders in a single class, and so we should find a way to do so. Below is my proposal for solving the issue without using decorators, while also being straightforward to support in the framework and somewhat easy for the web developer to reason about:

``` python
class IssueResource(object):
    def on_get(self, req, resp, project_id, issue_id):
        # Set response to a representation of the issue

    def on_put(self, req, resp, project_id, issue_id):
        # Update an existing issue

    def on_get_collection(self, req, resp, project_id):
        # Return a list of issues for the project

    def on_post_collection(self, req, resp, project_id):
        # Create a new issue and return its URL

    def on_get_shortlink(self, req, resp, shortlink):
        # Convert the shortlink to project_id and issue_id,
        # then call self.on_get(req, resp, project_id, issue_id)


issue_resource = IssueResource()
api.add_route('/projects/{project_id}/issues/{issue_id}', issue_resource)
api.add_route('/projects/{project_id}/issues', issue_resource, alias='collection')
api.add_route('/s/{shortlink}', issue_resource, alias='short')

```

swizzard:
What would the `alias` kwarg do?

kgriffs:
So, I'm not entirely sure if the "alias" is the best term, but the idea is that for this route, look for responders that end in the specified (arbitrary) alias, and ignore others. So 'collection' would only recognize on_get_collection and on_post_collection.

swizzard:
Could `alias` (whatever it ends up being called) get passed as an arg/kwarg to `on_get`? Then you'd at least get something potentially more helpful than `AttributeError` when you mess up.

kgriffs:
[...]

PR #1223, opened 2018-02-24
feat(routing): Add support for adding multiple routes for the same resource

santeyio:
Implementation of #584


santeyio:
Made requested changes -- thanks so much for the input @kgriffs, really appreciate it!

falcon/api.py
```diff
@@ -347,7 +352,14 @@ def add_route(self, uri_template, resource, *args, **kwargs):
         if '//' in uri_template:
             raise ValueError("uri_template may not contain '//'")
 
-        method_map = routing.map_http_methods(resource)
+        # NOTE(santeyio): This is a not very nice way to catch the suffix
+        # keyword. In python 3 it can be specified explicitly in the function
+        # definition, e.g.
+        # `add_route(self, uri_template, resource, *args, suffix=None, **kwargs)`
```
vytas7: That is why I was proposing a separate method for it, something like

``add_alternate_route(self, uri_template, suffix, *args, **kwargs)``

This would also eliminate the risk that someone already had a ``suffix`` inside ``kwargs`` for some other purpose. On the other hand, 2.0 is about being able to have breaking changes, and having the same ``add_route`` method probably offers simplicity and brevity over 2 methods.

kgriffs: Yeah, I was kind of thinking it may be a bit more elegant to just add the kwarg. As you say, in the off chance it breaks someone's custom router it is a major release after all. I don't want to just go breaking things left and right, but this seems like a low-risk change.

falcon/routing/util.py
```diff
@@ -183,3 +195,8 @@ def set_default_responders(method_map):
     for method in COMBINED_METHODS:
         if method not in allowed_methods:
             method_map[method] = na_responder
+
+
+class AltMethodNotFoundError(Exception):
```
kgriffs: Sorry to be pedantic, but let's rename this to coincide with the change from "alt" to "suffix".

tests/test_uri_templates.py
```diff
@@ -80,6 +81,40 @@ def on_get(self, req, resp, file_id, ext):
         self.called = True
 
 
+class ResourceWithAltRoutes(object):
```
kgriffs: `alt` ➡️  `suffix`

tests/test_uri_templates.py
```diff
@@ -399,3 +434,41 @@ def test_same_level_complex_var(client, reverse):
     assert details_resource.called
     assert details_resource.file_id == file_id_2
     assert details_resource.ext == ext
+
+
+def test_adding_alternate_routes(client):
```
kgriffs: `alt` ➡️  `suffix`

tests/test_uri_templates.py
```diff
@@ -399,3 +434,41 @@ def test_same_level_complex_var(client, reverse):
...
+    resource_with_alt_routes = ResourceWithAltRoutes()
```
kgriffs: `alt` ➡️  `suffix`

tests/test_uri_templates.py
```diff
@@ -399,3 +434,41 @@ def test_same_level_complex_var(client, reverse):
...
+    # PUT
+    client.simulate_put('/collections/foo345/items/foo567')
+    assert resource_with_alt_routes.collection_id == 'foo345'
+    assert resource_with_alt_routes.item_id == 'foo567'
+    assert resource_with_alt_routes.put_called
+    client.simulate_put('/collections/foo321/items')
+    assert resource_with_alt_routes.collection_id == 'foo321'
+
+
+def test_custom_error_on_alternate_route_not_found(client):
```
kgriffs: `alt` ➡️  `suffix`

tests/test_uri_templates.py
```diff
@@ -399,3 +434,41 @@ def test_same_level_complex_var(client, reverse):
...
+    resource_with_alt_routes = ResourceWithAltRoutes()
```
kgriffs: `alt` ➡️  `suffix`

tests/test_uri_templates.py
```diff
@@ -399,3 +434,41 @@ def test_same_level_complex_var(client, reverse):
...
+    try:
+        client.app.add_route(
+            '/collections/{collection_id}/items', resource_with_alt_routes, suffix='bad-alt')
+    except (AltMethodNotFoundError):
```
vytas7: Is there any particular reason surrounding this with parenthesis?

falcon/routing/util.py
```diff
@@ -140,6 +140,11 @@ def map_http_methods(resource):
             should define ``on_get(self, req, resp)`` and
             ``on_post(self, req, resp)``.
 
+        suffix: A string used to look for alternate methods to the base HTTP methods.
```
vytas7: Should we specify type like in the docstring earlier in this PR: ``suffix (str): ...``?

falcon/routing/util.py
```diff
@@ -140,6 +140,11 @@ def map_http_methods(resource):
...
+            If suffix is specified Falcon will look for resource methods ending in suffix,
+            e.g. if suffix is 'foo', then a "GET" request will be mapped to
```
vytas7: ``"GET"`` and ``"POST"`` are quoted as ``GET`` and ``'GET'`` in this same docstring, maybe it would be worth making the consistent?

falcon/routing/util.py
santeyio: well the reason I didn't is because I was just trying to match the previous argument in the docstring, the one `resource` above it doesn't have a type with out.

falcon/routing/util.py
santeyio: yeah, that makes sense, I made the consistent across the docstring with my latest commit

tests/test_uri_templates.py
santeyio: No reason. Glancing at the general style just now I see that if it's just one exception then in falcon there isn't parentheses, so I removed those in my last commit.

santeyio:
Alrighty, thanks again for your input @kgriffs and @vytas7 ! Much appreciated. Made the changes suggested, let me know if you catch anything else.

Mumrau:
Hey everyone, how did it ended up being implemented? With the `alt` keyword or as another method?

bkcsfi:
are there plans to get this feature into a release?  The tutorial references issue #584, the tag  on this issue is 'merged', but it doesn't look like it's been merged into a deployed version. 

thanks

Mumrau:
@bkcsfi it has been merged onto the master branch of the repository.

Nonetheless, it won't be added before falcon 2.0 release, and won't be merged into the falcon 1.4 actually available through pip as it obviously changes the api.

bkcsfi:
thanks for the update, I'll follow the old-style shown in the docs for now..

PR #1226, opened 2018-03-03
feat(routing): Add support for static route default_filename (#1224)

### LINKED ISSUE #1224: Add default file name in the StaticRoute
Consider adding the ability to specify a default filename for a static route. 

I'm not sure if it would be best to send the file or raise an `HTTPMovedPermanently` (or a `HTTPPermanentRedirect`). 
Probably for performance reasons the redirect would be best (but in this case maybe `default_filename` should become `default_location`)

CaselIT:
Adds an optional `default_filename` argument to the static route to sent that file if none is matched.

Closes #1224


falcon/routing/static.py
```diff
@@ -49,6 +55,10 @@ def __init__(self, prefix, directory, downloadable=False):
         self._prefix = prefix
         self._directory = directory
         self._downloadable = downloadable
+        if default_filename is not None:
+            self._default_filename = os.path.join(directory, default_filename)
```
kgriffs: Should it be possible for subdirectories to provide their own versions of this file? If so, the path will have to be joined on the fly, rather than up front. 

For example, suppose the prefix is `/foo/bar` and a request comes in for `/foo/bar/baz/`. You might want that to map to `/foo/bar/baz/index.html`, while a request for `/foo/bar/` would map to `/foo/bar/index.html`.

falcon/api.py
```diff
@@ -391,12 +391,15 @@ def add_static_route(self, prefix, directory, downloadable=False):
             downloadable (bool): Set to ``True`` to include a
                 Content-Disposition header in the response. The "filename"
                 directive is simply set to the name of the requested file.
+            default_filename (str): Default filename used when the requested file
```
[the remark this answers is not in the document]
CaselIT: It's an error on my part because I thought that join would merge two absolute path, but it does not. 
I meant to only allow files inside the given directory.

Do you think it's best to allow files that are outside of the given directory?

falcon/routing/static.py
```diff
@@ -83,6 +93,9 @@ def __call__(self, req, resp):
         if '..' in file_path or not file_path.startswith(self._directory):
             raise falcon.HTTPNotFound()  # pragma: nocover
 
+        if self._default_filename is not None and not os.path.isfile(file_path):
```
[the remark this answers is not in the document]
CaselIT: I've replayed in #1224

falcon/routing/static.py
CaselIT: The use case I wanted to cover was SPA, where both `/foo/bar` and `/foo/bar/baz/` would need to map to `/foo/bar/index.html` (supposing the SPA home is `/foo/bar`)

falcon/routing/static.py
```diff
@@ -49,10 +55,15 @@ def __init__(self, prefix, directory, downloadable=False):
...
         self._downloadable = downloadable
+        if default_filename is not None:
+            self._default_filename = os.path.join(directory, default_filename)
+        else:
+            self._default_filename = None
 
     def match(self, path):
         """Check whether the given path matches this route."""
-        return path.startswith(self._prefix)
+        prefix = self._prefix if self._default_filename is None else self._prefix[:-1]
```
kgriffs: I don't think this quite addresses the issue. Consider the case in which `self._default_filename is not None` and:

```py
self._prefix = '/foo/bar/'
path = '/foo/barley'
```
In this case, `prefix = '/foo/bar'`, which would result in `match()` returning `True`.

(P.S. Please add a test for this scenario.)

falcon/routing/static.py
```diff
@@ -83,6 +95,9 @@ def __call__(self, req, resp):
         if '..' in file_path or not file_path.startswith(self._directory):
             raise falcon.HTTPNotFound()  # pragma: nocover
 
+        if self._default_filename is not None and not os.path.isfile(file_path):
```
kgriffs: PERF: `open()` already checks whether the file exists. Therefore, you can avoid an extra file system call by attempting to open the fallback file in the `except IOError` block, rather than calling `os.path.isfile()` in advance (filesystem calls are expensive relative to a few extra CPU cycles spent on exception handling).

CaselIT:
I should've addressed the comments.

falcon/routing/static.py
```diff
@@ -61,7 +71,8 @@ def __call__(self, req, resp):
 
         # NOTE(kgriffs): Check surrounding whitespace and strip trailing
         # periods, which are illegal on windows
-        if (not without_prefix or
+        # NOTE(CaselIT): An empty filename is allowed when fallback_filename is provided
```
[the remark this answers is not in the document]
CaselIT: Let's say we mount a static route on `/static` from directory `/tmp/static/` with a fallback filename. If we do not allow an empty filename in the request, and the user asks for`/static`, the route would match, but we would send a 404 instead of the fallback filename.

By allowing the empty file, `file_path` would be set to `/tmp/static/.`, `open` fails since it `file_path` is a directory, so it will send the fallback file in the `except` block

falcon/api.py
```diff
@@ -364,6 +364,8 @@ def add_static_route(self, prefix, directory, downloadable=False):
             Serving files directly from the web server,
             rather than through the Python app, will always be more efficient,
             and therefore should be preferred in production deployments.
+            For security reasons, the prefix forder and the fallback_filename (if provided)
+            should be read only for the account running the application.
```
jmvrbanac: spelling nit: I believe `forder` should be `order`

falcon/routing/static.py
```diff
@@ -49,10 +56,13 @@ def __init__(self, prefix, directory, downloadable=False):
         self._prefix = prefix
         self._directory = directory
         self._downloadable = downloadable
+        self._fallback_filename = None if fallback_filename is None else fallback_path
```
jmvrbanac: could this be paired down to `self._fallback_filename = fallback_filename or fallback_path`?

falcon/api.py
CaselIT: There is an error, instead of prefix I should have used directory

falcon/routing/static.py
CaselIT: No, because we want to assign `fallback_path` when `fallback_filename` is specified by the user

falcon/routing/static.py
```diff
@@ -33,13 +36,17 @@ class StaticRoute(object):
...
         if not prefix.startswith('/'):
             raise ValueError("prefix must start with '/'")
 
         if not os.path.isabs(directory):
             raise ValueError('directory must be an absolute path')
 
+        fallback_path = os.path.join(directory, fallback_filename or '')
+        if fallback_filename is not None and not os.path.isfile(fallback_path):
+            raise ValueError('fallback_filename does not exists')
+
```
kgriffs: Suggestion: It may simplify things a bit and remove the need for L59 altogether if `self._fallback_filename` were assigned directly above, e.g.:

```py
if fallback_filename is None:
    self._fallback_filename = None
else:
    self._fallback_filename = os.path.join(directory, fallback_filename)
    if not os.path.isfile(self._fallback_filename):
        raise ValueError('fallback_filename does not exists')
```

CaselIT: I wanted to keep the same style: parameter check then assignment to instance variable.
I agree that it's cleaner this way

I'll change it

CaselIT:
Thanks for merging it 🥂

PR #1231, opened 2018-03-11
Enable jsonschema.validate() to validate responses.

ZDBioHazard:
With this change, you can pass another schema dict to jsonschema.validate() as 'schema_response' to validate the response data against it.

A falcon.HTTPInternalServerError is raised on schema violations.

This is useful to make sure a bug in your code doesn't somehow accidentally cause your resource to return an out-of-spec response to users.

Let me know if you want me to change the name of the first parameter back. I think it's more specific as 'schema_request', but I understand it's possible some users may have explicitly keyworded it in their apps even though it was the only parameter.

I'm also open to splitting it into it's own 'validate_response' decorator.

### DIFF
--- a/falcon/media/validators/jsonschema.py
+++ b/falcon/media/validators/jsonschema.py
@@ -8,7 +8,7 @@ except ImportError:
     pass
 
 
-def validate(schema):
+def validate(schema_request=None, schema_response=None):
     """Decorator for validating ``req.media`` using JSON Schema.
 
     This decorator provides standard JSON Schema validation via the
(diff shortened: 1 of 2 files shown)


tests/test_validators.py
```diff
@@ -21,31 +21,47 @@
...
+def test_jsonschema_response_validation_success():
+    data = GoodData()
+    assert Resource().response_validated(GoodData(), data) is data
 
-    res = SampleResource()
-    with pytest.raises(falcon.HTTPBadRequest) as err:
-        res.on_get(req, None)
 
-    assert err.value.description == '\'message\' is a required property'
+@skip_py26
```
kgriffs: Once you rebase on master, you can remove these `@skip_py26` decorators, as we have removed 2.6 support.

falcon/media/validators/jsonschema.py
```diff
@@ -8,7 +8,7 @@
     pass
 
 
-def validate(schema):
+def validate(schema_request=None, schema_response=None):
```
kgriffs: Nitpick:  Would you mind swapping the word order in these arguments to follow the _Modifier-Subject_ style, i.e.: `request_schema` and `response_schema` or `req_schema` and `resp_schema`?

kgriffs:
I rebased and added a few additional patches on top, resubmitted as #1246

PR #1233, opened 2018-03-16
Multipart form data upload

branelmoro:
I have added new code to parse POST, PUT and DELETE application/x-www-form-urlencoded and multipart/form-data.
I have used cgi, os, html and sys modules to implement this feature.
Text fields will be available in req.form_data dictionary and upload file buffer stream will be available in req.files dictionary.

Below is sample code to show how to use POST, PUT and DELETE application/x-www-form-urlencoded and multipart/form-data:

```
import falcon

class Resource(object):

    def on_post(self, req, resp):

        # req.form_data will return dictionary of text field names and their values
        print(req.form_data)

        # req.form_data will return dictionary of file field names and
        # their buffer class FileStream objects as values
        print(req.files)

        # support we are uploading a image.jpg in `pancard` file field then
        # req.files["pancard"] will be FileStream buffer object

        # We can use set_max_upload_size method to set maximum allowed
        # file size let say 1Mb = 1*1024*1024 bytes for this file

        req.files["pancard"].set_max_upload_size(1*1024*1024)

        # We can use uploadto method to upload file on required path (Note: absolute filepath is required)
        # This method returns boolean - `True` on successful upload
        # and if upload is unsuccessful then it returns `False` and sets error on failure.
        path = "/tmp/" + req.files["pancard"].name

        response = req.files["pancard"].uploadto("/tmp/" + path)

        print(response)

        # Once file is uploaded sucessfully, we can check it's size
        print(req.files["pancard"].size)

        # If file is not uploaded sucessfully, we can check it's error
        print(req.files["pancard"].error)

        resp.body = "Done file upload"

        resp.status = falcon.HTTP_200

# falcon.API instances are callable WSGI apps
app = falcon.API()

things = Resource()

# things will handle post requests to the '/post_path' URL path
app.add_route('/post_path', things)
```

Do let me know if you have any doubts.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -12,6 +12,11 @@
 
 """Request class."""
 
+import cgi
+import os
+from sys import exc_info, getsizeof
+import tempfile
+
 from datetime import datetime
 
 try:
@@ -26,11 +31,6 @@ except AttributeError:
     import io
     NativeStream = io.BufferedReader
 
-import cgi
-import tempfile
-import os
-from sys import exc_info, getsizeof
-
 from uuid import UUID  # NOQA: I202
 from wsgiref.validate import InputWrapper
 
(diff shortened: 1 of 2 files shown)


kenwoodjw:
when publish this branch to new version?

Issue #953, opened 2016-11-23
Support multipart/form-data


kgriffs:
Depends on: #418 

* Add support to `Request.form` for loading non-file control name/value pairs from a `multipart/form-data` request.
* Add a new property, `files` to `Request`. This should provide an interface for reading one or more files submitted via an HTML form post. We should look at the standard library, werkzeug, and bottle for inspiration re the implementation. There may also be some standalone packages that may be helpful (search PyPI).

lwcolton:
@kgriffs To be clear are you asking for a new multipart form parser implementation, and not to re-use cgi.FieldStorage?  https://github.com/python/cpython/blob/2.7/Lib/cgi.py#L352

kgriffs:
@lwcolton I'm not necessarily opposed to using cgi.FieldStorage but we should see if there are any advantages to forking code from werkzeug, for example.

kgriffs:
See also: #1236

kgriffs:
Note: This will be part of native HTML form post handling work that will have received a sponsorship to fund. @vytas7 will likely be overseeing this work stream.

vytas7:
Resolved in https://github.com/falconry/falcon/pull/1590

I'll create a follow-up issue for ASGI+multipart.

PR #1236, opened 2018-03-28
Multipart form data, file upload and form-urlencoded feature added

branelmoro:
I have added new code to parse POST, PUT and DELETE application/x-www-form-urlencoded and multipart/form-data.
I have used cgi, os, html and sys modules to implement this feature.
Text fields will be available in req.form_data dictionary and upload file buffer stream will be available in req.files dictionary.

Below is sample code to show how to use POST, PUT and DELETE application/x-www-form-urlencoded and multipart/form-data:

```py
import falcon

class Resource(object):

    def on_post(self, req, resp):

        # req.form_data will return dictionary of text field names and their values
        print(req.form_data)

        # req.form_data will return dictionary of file field names and
        # their buffer class FileStream objects as values
        print(req.files)

        # support we are uploading a image.jpg in `pancard` file field then
        # req.files["pancard"] will be FileStream buffer object

        # We can use set_max_upload_size method to set maximum allowed
        # file size let say 1Mb = 1*1024*1024 bytes for this file

        req.files["pancard"].set_max_upload_size(1*1024*1024)

        # We can use uploadto method to upload file on required path (Note: absolute filepath is required)
        # This method returns boolean - `True` on successful upload
        # and if upload is unsuccessful then it returns `False` and sets error on failure.
        path = "/tmp/" + req.files["pancard"].name

        response = req.files["pancard"].uploadto("/tmp/" + path)

        print(response)

        # Once file is uploaded sucessfully, we can check it's size
        print(req.files["pancard"].size)

        # If file is not uploaded sucessfully, we can check it's error
        print(req.files["pancard"].error)

        resp.body = "Done file upload"

        resp.status = falcon.HTTP_200

# falcon.API instances are callable WSGI apps
app = falcon.API()

things = Resource()

# things will handle post requests to the '/post_path' URL path
app.add_route('/post_path', things)
```

Do let me know if you have any doubts.


falcon/request.py
```diff
@@ -1617,6 +1664,85 @@ def _get_wrapped_wsgi_input(self):
...
+                    self._files[fieldname] = list_files
+            else:
+                # NOTE(kgriffs): fieldname has single values
+                value = self._getFieldValue(filedata)
+                if isinstance(value, str):
+                    self._form_data[fieldname] = value
+                elif value:
+                    self._files[fieldname] = value
+
+        for i in list(self._form_data):
```
kgriffs: This loop would probably be more efficient if it were to iterate over `self._form_data.items()`, instead.

falcon/request.py
```diff
@@ -1617,6 +1664,85 @@ def _get_wrapped_wsgi_input(self):
...
+            if isinstance(self._form_data[i], list) and len(self._form_data[i]) == 1:
+                self._form_data[i] = self._form_data[i][0]
+
+        for i in list(self._files):
```
kgriffs: (See above.)

falcon/request.py
```diff
@@ -1646,6 +1772,136 @@ def _parse_form_urlencoded(self):
...
+            key "code" - 1 is set if _max_size is exceeded
+            key "code" - 2 is set if on some exceptional error
+            key "error" contains addition error details
+        size : Size of uploaded file. Actual size is available once upload is successful
+    """
+
+    def __init__(self, file_buffer, max_size, first_byte):
+        self._buffer = file_buffer
+        self._temp_file = None
+        self._max_size = max_size + 1024
```
DmitriiTrofimov: What is this magic?

DmitriiTrofimov:
I think that is wrong way. More pretty handling `multipart/form-data` (and such as `application/x-www-form-urlencoded` and `application/x-url-encoded`) in media handlers, because it's all are media contents (see [OpenApi 3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md) for examples).  For supported handling more various media types - #1312 

I like Django implementation of multipart parser because it's more simplest and clean than werkzeug implementation, but for REST api uses need also provide headers of all parts of multipart content because it's required for automatic decoding their contents (decoding json encoded data, xml etc)

falcon/request.py
```diff
@@ -1617,6 +1664,85 @@ def _get_wrapped_wsgi_input(self):
...
         return helpers.BoundedStream(self.env['wsgi.input'], content_length)
 
+    def _getFieldValue(self, field):
+        """Returns input value for imput text fields and FileStream object
+        for input file and None if no file is selected for upload
+
+        Args:
+            field (cgi.FieldStorage or cgi.MiniFieldStorage).
+        """
+        if field.file and 'filename' in field.disposition_options:
```
kgriffs: Is it necessary to check for the presence of 'filename'?

falcon/request.py
```diff
@@ -1617,6 +1664,85 @@ def _get_wrapped_wsgi_input(self):
...
+                'application/x-www-form-urlencoded' in self.content_type
+            )
+        ):
+            # NOTE(kgriffs): Store input params in _form_data and files in _files
+            self._files = {}
+            self._form_data = {}
+        else:
+            return
+
+        field_storage = cgi.FieldStorage(fp=self.stream, environ=self.env)
```
kgriffs: cgi.FieldStorage will blow up if the env doesn't include QUERY_STRING (see also: https://gist.github.com/kgriffs/fe371bfbb2d63211889b#file-gistfile1-py-L9)

kgriffs:
@branelmoro what do you think about @DmitriiTrofimov's suggestion re adding a media handler to do this? The object returned by `Request.media` would presumably include methods to access file data and params. #1311 would provide access to the request stream so that it could be passed in to `cgi.FieldStorage()`.

kgriffs:
Pushing this to 2.1

falcon/request.py
```diff
@@ -1617,6 +1664,85 @@ def _get_wrapped_wsgi_input(self):
 
         return helpers.BoundedStream(self.env['wsgi.input'], content_length)
 
+    def _getFieldValue(self, field):
```
mikeylight: Also, I suggest rename method name to `_get_field_value`

vytas7:
@branelmoro have you had a chance to review @kgriffs and @DmitriiTrofimov suggestions regarding file storage handling, as well as generalizing this as pluggable media handlers?

We are going to close this PR in favour of the clean start building on the said approaches, but many thanks for helping out!
Should we appropriate anything of the work above in the final version of HTML form handling, care will be taken to attribute the work to you by ``Co-authored-by`` commits.

timc13:
is there a ticket to track this "clean start" ?  I am still looking forward to using multipart form data with falcon.

vytas7:
@timc13 yes, https://github.com/falconry/falcon/issues/953 should be the issue.

I will hopefully start on it soon, just working help with the upcoming Falcon 2.0 release first.

vytas7:
@timc13 FWIW the multipart functionality has now hit the Falcon `master` branch. We are planning to release the first 3.0 alpha soon.

See also:
* https://falcon.readthedocs.io/en/latest/api/multipart.html
* https://falcon.readthedocs.io/en/latest/user/faq.html#multipart-cloud-upload

Issue #1250, opened 2018-05-07
Move routing.map_http_methods into an instance or class method?


xgfone:
I like `Falcon`, but the policy to map http methods. So I want to customize it, and I must override the method `add_route`, which is `WASTEFUL` and `EXPENSIVE`.

I think, `method_map = routing.map_http_methods(resource, suffix=suffix)` should be placed in an instance or class method, such as
```python
class API(object):
    def map_http_methods(self, resource, *args, **kwargs)
        return routing.map_http_methods(resource, suffix=kwargs.pop('suffix', None))

    def add_route(uri_template, resource, *args, **kwargs):
        # ...
        method_map = self.map_http_methods(resource, *args, **kwargs)
        routing.set_default_responders(method_map)
        self._router.add_route(uri_template, method_map, resource, *args, **kwargs)
```

Now, I only need to override the method `map_http_methods`, don't change or repeat the process in `add_route`. For example,
```python
class MyAPI(API):
    def map_http_methods(self, resource, *args, **kwargs):
        return {
            kwargs["method"]: getattr(resource, kwargs["action"])
        }

        # OR

        return {method: getattr(resource, action) for method, action in kwargs["map"].items()}
```

I know someone does not like this method mapping, but you should not forbid others to like it.

xgfone:
@kgriffs I have submitted a RP #1252 , please take a look. Thanks.

This is my overriding example,
```python
class MyAPI(falcon.API):
    
    def map_http_methods(self, resource, args, kwargs):
        _map = kwargs.pop("map", None)
        if _map:
            return {k.upper(): getattr(resource, v) for k, v in _map.items()}
        return super(MyAPI, self).map_http_methods(resource, args, kwargs)
```

It supports not only the original HTTP method mapping but also the new self-defined. For example,
```python
resource = Resource()
app = MyAPI()
app.add_route("/path/to/1", resource)  # The original
app.add_route("/path/to/2", resource, map={"GET": "my_get_method", "POST": "my_post_method"})
```

@kgriffs I'm thinking whether we maybe use the following method instead of that in PR.
```python
def map_http_methods(self, resource, args, kwargs):
    _map = kwargs.pop("map", None)
    if _map:
        return {k.upper(): getattr(resource, v) for k, v in _map.items()}
        
    suffix = kwargs.pop('suffix', None)
    return routing.map_http_methods(resource, suffix=suffix)
```

kgriffs:
[...]

PR #1252, opened 2018-05-08
feat(api): allow to override the default method mapping

xgfone:
Move `routing.map_http_methods()` to an instance method, so it will allow the user to override it to implement the self-defined method mapping.

------

Implement #1250

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -301,6 +301,26 @@ class API(object):
     def router_options(self):
         return self._router.options
 
+    def map_http_methods(self, resource, **kwargs):
+        """Map HTTP methods (e.g., GET, POST) to methods of a resource object.
+
+        You maybe override it to implement yourself method mapping.
+
+        Args:
+            resource (instance): Object which represents a REST resource.
+                The default maps the HTTP method ``GET`` to ``on_get()``,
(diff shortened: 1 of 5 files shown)


falcon/api.py
```diff
@@ -359,14 +391,11 @@ class can use suffixed responders to distinguish requests
...
-        # but python 2 won't accept args like this.
-        suffix = kwargs.pop('suffix', None)
+        # NOTE(xgfone): we should allow the method map_http_methods to fix them,
+        # in order to prevent some arguments to be passed to the router.
+        args = list(args)
+        method_map = self.map_http_methods(resource, args, kwargs)
 
-        method_map = routing.map_http_methods(resource, suffix=suffix)
         routing.set_default_responders(method_map)
         self._router.add_route(uri_template, method_map, resource, *args,
```
kgriffs: What do you think about making the new, override-able method a member of the router rather than `falcon.API`? That may give us a more cohesive design. All we would need to do is move the calls to get the mapping and to set the default responders into `add_route()`.

xgfone: Yes, you are right. I'm doing it, but it will break down the API `CompiledRouter.add_route()`, because the argument `method_map` is unnecessary.

kgriffs:
Rebased and merged as part of #1329

Issue #1276, opened 2018-05-12
Change the default of the independent_middleware kwarg in API.__init__ to True


kgriffs:
`independent_middleware=True` tends to lead to behavior that is easier to reason about and less surprising.

KingAkeem:
May I work on this PR?

PR #1277, opened 2018-05-13
Set mid default to API.__init__ to True for independent_middleware kwarg

KingAkeem:
#1276


KingAkeem:
@kgriffs I fixed the failing tests :)

PR #1285, opened 2018-05-14
chore(request): Add a new method, get_param_as_float()

ozzzik:
test: Add relevant tests.

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -228,7 +228,7 @@ class API(object):
                 # HTTP exception signalling the problem, e.g. a 404.
                 responder, params, resource, req.uri_template = self._get_responder(req)
             except Exception as ex:
-                if not self._handle_exception(ex, req, resp, params):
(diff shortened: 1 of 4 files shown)


falcon/request.py
```diff
@@ -1232,6 +1232,75 @@ def get_param_as_int(self, name,
...
+                           required=False, min=None, max=None, store=None):
+        """Return the value of a query string parameter as an float.
+
+        Args:
+            name (str): Parameter name, case-sensitive (e.g., 'limit').
+
+        Keyword Args:
+            required (bool): Set to ``True`` to raise
+                ``HTTPBadRequest`` instead of returning ``None`` when the
+                parameter is not found or is not an integer (default
```
jmvrbanac: nit: replace `integer` with `float`

ozzzik: @jmvrbanac, Fixed.

Issue #763, opened 2016-04-26
Normalize order of params for custom error handlers and custom error serializers


kgriffs:
The current order of params is inconsistent between the two types of callbacks:

`def my_error_handler(ex, req, resp, params)`

vs.

`def my_serializer(req, resp, exception)`

steven-upside:
I'll do this

PR #1286, opened 2018-05-14
refactor(error_handler): Reorder Custom ErrorHandler params

steven-upside:
This is a breaking Change.
Closes: https://github.com/falconry/falcon/issues/763

### DIFF
--- a/falcon/api.py
+++ b/falcon/api.py
@@ -228,7 +228,7 @@ class API(object):
                 # HTTP exception signalling the problem, e.g. a 404.
                 responder, params, resource, req.uri_template = self._get_responder(req)
             except Exception as ex:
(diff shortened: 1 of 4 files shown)


steven-upside:
@kgriffs It would be possible to support old parameter order and new parameter order simultaneously by calling using the keyword style:

```
        for err_type, err_handler in self._error_handlers:
            if isinstance(ex, err_type):
                try:
                    err_handler(req=req, resp=resp, ex=ex, params=params)
                <...>
```
however error handlers would then have to use those exact names for their parameters. Users who don't currently do that would still experience a breaking change, so I'm not sure if its better or worse compatibility-wise, and it feels very different from other methods where order matters to have exactly one function which cares about names instead.

I think forced re-ordering is the preferable breaking change

Issue #1266, opened 2018-05-12
feat: Add a content_length property to Response to set the Content-Length header


kgriffs:
Useful for responding to HEAD requests when you aren't actually providing the response body.

pshello:
Taking a look at this

PR #1288, opened 2018-05-14
Add a content_length property to Response Issue #1266

pshello:
The new content_length property should be an adequate replacement for Response.stream_len, which could be deprecated. Using content_length would be preferable, as it more directly references the HTTP header that both stand in for. I have updated the documentation that relates to Response.stream_len to refer to content_length where I have found it.

Breaking Changes: Changes method signature from `stream_len` to `content_length`.

Closes: #1266


docs/user/tutorial.rst
```diff
@@ -1203,9 +1203,9 @@ URI parameters in a moment.
 Inside the ``on_get()`` responder,
 we set the Content-Type header based on the filename extension, and then
 stream out the image directly from an open file handle. Note the use of
-``resp.stream_len``. Whenever using ``resp.stream`` instead of ``resp.body`` or
-``resp.data``, you typically also specify the expected length of the stream so
-that the web client knows how much data to read from the response.
+``resp.content_length``. Whenever using ``resp.stream`` instead of ``resp.body`` or
+``resp.data``, you typically also specify the expected length of the stream using the
+content length header, so that the web client knows how much data to read from the response.
```
kgriffs: Nitpick: Use the actual header name here, i.e.: **Content-Length**.

falcon/response.py
```diff
@@ -185,23 +188,28 @@ def media(self, obj):
...
         return '<%s: %s>' % (self.__class__.__name__, self.status)
 
-    def set_stream(self, stream, stream_len):
-        """Convenience method for setting both `stream` and `stream_len`.
+    def set_stream(self, stream, content_length):
+        """Convenience method for setting both `stream` and `content_length`.
+
+        Args:
+            stream (str): stream
+            content_length (int): set the stream length, if known
```
kgriffs: Nitpick: I think this could be improved a bit, e.g.: "Length of the stream to use for the Content-Length header in the response."

falcon/response.py
```diff
@@ -185,23 +188,28 @@ def media(self, obj):
     def __repr__(self):
         return '<%s: %s>' % (self.__class__.__name__, self.status)
 
-    def set_stream(self, stream, stream_len):
-        """Convenience method for setting both `stream` and `stream_len`.
+    def set_stream(self, stream, content_length):
+        """Convenience method for setting both `stream` and `content_length`.
+
+        Args:
+            stream (str): stream
```
kgriffs: Change to: "A readable file-like object."

falcon/response.py
```diff
@@ -185,23 +188,28 @@ def media(self, obj):
     def __repr__(self):
         return '<%s: %s>' % (self.__class__.__name__, self.status)
 
-    def set_stream(self, stream, stream_len):
-        """Convenience method for setting both `stream` and `stream_len`.
+    def set_stream(self, stream, content_length):
+        """Convenience method for setting both `stream` and `content_length`.
+
+        Args:
```
kgriffs: Would you mind moving this Args section to the bottom of the docstring, below the prose sections?

tests/test_headers.py
```diff
@@ -202,6 +202,20 @@ def test_content_length(self, client):
         content_length = str(len(SAMPLE_BODY))
         assert result.headers['Content-Length'] == content_length
 
+    def test_declare_content_length(self, client):
+        resource = testing.SimpleTestResource(headers={'Content-Length': 42})
```
kgriffs: This actually just sets the header directly rather than going through the `content_length` property. To test that, you will need to create a new resource class that sets `req.content_length` from within its `on_get()` responder, e.g.:

```py
class SomeTestResource(object):
    def on_get(self, req, resp):
        resp.content_length = 42
```

falcon/response.py
```diff
@@ -106,9 +106,12 @@ class Response(object):
             blocks as byte strings. Falcon will use *wsgi.file_wrapper*, if
             provided by the WSGI server, in order to efficiently serve
             file-like objects.
-        stream_len (int): Expected length of `stream`. If `stream` is set,
+
+        stream_len (int): Deprecated alias for :py:attr:`content_length`
```
kgriffs: Let's go ahead and move the text following the deprecation notice (i.e., starting with "If stream is..." up into a `Note:` section under the docstring for the `content_length` property.

So, this would become simply (note the trailing period):

```py
stream_len (int): Deprecated alias for :py:attr:`content_length`.
```

PR #1295, opened 2018-05-15
feat(Response) Serialize response when data is referenced, not when media is set

blemasle:
This pull request allows falcon to only proceeds with the serialization when `Response.data` is first accessed (consistent with how `Request.media` works).

This offers several benefits
 - `Response.media` can be affected several times without any performance impact
 - `Response.content_type` does not need to be set **before** `Response.media`

Failing tests in this PR relates to slight behaviour changes. 2 tests are failing with these changes because they assumes that `HTTPUnsupportedMediaType` will be raised on `Response.media` affectation. The last test, that checks `Response.json` I cannot really figure what needs to be changed.

I intentionally did not want to fix these tests as these changes can be seen as a behavior change  and would, I guess, needs to be thoroughly evaluated.

I might be completely wrong here, but I cannot see a reason why not to do this in that order. Affecting `Response.media` should not implies serializing it (ie the "media" is **not** the actual response)

Regards

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -133,7 +133,6 @@ class Response(object):
 
     __slots__ = (
         'body',
-        'data',
         '_headers',
         '_cookies',
         'status',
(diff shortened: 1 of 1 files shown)


kgriffs:
Hi @blemasle, if the goal is to avoid re-serialization of the resource representation when the media property is set multiple times, it may be cleaner to simply move the serialization to the point in `falcon.API` where the response is being hydrated. In that case, the media property could be refactored to simply hold the object to be serialized for later.

@jmvrbanac thoughts?

blemasle:
Great news. Thanks !

PR #1299, opened 2018-05-15
Update from upstream

steven-upside:


steven-upside:
Sorry, had this backwards

Issue #1243, opened 2018-04-08
Misleading doc string in JSON media handler


sciatro:
Looks like the [JSON media handler's doc string](https://github.com/falconry/falcon/blob/3ba0412acee1eaddf36b9a35d9859fd43798e4a9/falcon/media/json.py#L11) (``"""Handler built using Python's :py:mod:`json` module."""``) is incorrect since ``falcon.util.json`` [could be ujson](https://github.com/falconry/falcon/blob/3ba0412acee1eaddf36b9a35d9859fd43798e4a9/falcon/util/__init__.py#L24).

Looks like it became incorrect in 24d51a4808e72d14108d61c2a089158cb6757435.

kgriffs:
Good catch, we'll get that fixed!

Nateyo:
This looks completed and already in master. Just needs CHANGES.rst I guess?

vytas7:
It seems that this has been rectified in the linked #1303 .

It might need a yet another update after https://github.com/falconry/falcon/issues/1206 , but lets handle it in that issues depending on the decisions made.

Issue #1206, opened 2018-01-16
Third party json support


hynek:
Congratulations on shipping 1.4!  I’ve noticed you’ve added direct support for ujson which I find unfortunate for at least three reasons:

1. It’s mostly abandoned: latest release was in 2016, latest commit almost a year ago.  Maintainers aren’t very responsive on the bug tracker.
1. It doesn’t allow for customising serialisation (`default` and `object_hook`) – and this is my biggest point.
1. It has/had some standards issues (I’ve lost the link to info on that and I’m too lazy to look for it – but I think the other reasons are enough :))

Personally I’ve come to use [python-rapidjson](https://pypi.org/project/python-rapidjson/) which is comparable in terms of performance, under active development, and supports `default` and `object_hook`.  Its only possible downside is that it only supports Python 3 – but hey, 2020 is looming. :)

***

I’ve looked how falcon implemented ujson and it seems like it’s just a protected imports.

Now, I could ask you to add another `import/except ImportError` block, but maybe this whole thing could be solved differently and more elegantly.  Maybe by allowing to parametrize dumps/loads functions when they are used?  I don’t know your code well enough but I suspect there could be a way to make everyone happy without adding support to everyone’s pet JSON library. :)

kgriffs:
Thanks for the feedback and I agree. It would be great to have a more flexible approach and give everyone the ability to choose from any of a number of alternative JSON libraries. The danger of picking one, as you point out, is that it may stop being maintained, or may not support certain features one might need, etc. 

Let's explore dropping ujson in 2.0 and looking for a way to make alternative loads/dumps functions configurable. I don't think it would be too terribly difficult to make it happen.

hynek:
oh and totally related to #1192 – I just forgot to link it :)

kgriffs:
@mattgiles I really like mujson and would consider using it if you can get it to Production/Stable status. It would be especially great if it could detect PyPy (by checking for the presence of ``sys.pypy_version_info``, for example) and modifying what it loads accordingly.

steven-upside:
@kgriffs What specifically do you think is required to get `mujson` to Production/Stable status? 

As indicated in the issue, `mujson` is very useful for me outside of falcon, and since I'm planning to add PyPy detection anyways--which you also expressed interest in--I figured it would make sense to see if I could take it all the way to `1.0`.

If you have a good idea of what's required to make it usable within falcon, I'll create a Pull Request resolving the issue I just opened on the `mujson` project.
[...]

PR #1303, opened 2018-05-15
docs(handler): Clarify that handler may use `ujson`

steven-upside:
#1243


steven-upside:
This may be a change that gets obsoleted by the resolution to https://github.com/falconry/falcon/issues/1206.

falcon/media/json.py
```diff
@@ -8,7 +8,7 @@
 
 
 class JSONHandler(BaseHandler):
-    """Handler built using Python's :py:mod:`json` module."""
+    """Handler built using Python's :py:mod: `ujson` if available, :py:mod: `json` otherwise."""
```
jmvrbanac: This would probably be better phrased 
```
Handler built using Python's :py:mod: `json` by default, but will use :py:mod: `ujson` if available.
```

Issue #1177, opened 2017-12-19
Remove the deprecated routing.create_http_method_map method


kgriffs:

vytas7:
Closed by @MichaelYusko in #1310 .

PR #1310, opened 2018-05-19
Removed the deprecated create_http_method_map method #1177

MichaelYusko:
- [x] Remove from `routing.create_http_method_map`
- [x] Remove from `routing.util.create_http_method_map`
- [x] Remove from the docs


MichaelYusko:
@kgriffs Hello, I'm not sure that, I was needed delete a `create_http_method_map` from the `routing.util`, if we need keep the method in the file, please, let me know and I'll amend one more commit in order to fix it.

MichaelYusko:
@kgriffs @vytas7 Awesome :tada:

PR #1311, opened 2018-05-20
Providing of real content-type to media handlers.

DmitriiTrofimov:
#1312 
Providing of real content-type to media handlers for deserializing and serializing media by corresponding to real content-type or specific parameters (such as "boundary" for multipart/form-data). Also provided content-length for special cases.

For avoiding reading big data to memory to media handlers now providing bounded stream instead raw data. Also consumption of stream proceed by small chunks.

### DIFF
--- a/falcon/request_helpers.py
+++ b/falcon/request_helpers.py
@@ -157,6 +157,25 @@ class BoundedStream(io.IOBase):
 
         raise IOError('Stream is not writeable')
 
+    def exhaust(self, chunk_size=64 * 1024):
+        """Exhaust the stream.
+
+        This consumes all the data left until the limit is reached.
+
+        Args:
+            chunk_size (int): The size for a chunk (default: 64 KB).
+                It will read the chunk until the stream is exhausted.
+        """
+        while True:
+            chunk = self.read(chunk_size)
+            if not chunk:
+                break
+
+    @property
+    def is_exhausted(self):
+        """If the stream is exhausted this attribute is ``True``."""
+        return self._bytes_remaining <= 0
+
 
 # NOTE(kgriffs): Alias for backwards-compat
 Body = BoundedStream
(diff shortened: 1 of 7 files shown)


falcon/request_helpers.py
```diff
@@ -157,6 +157,24 @@ def write(self, data):
 
         raise IOError('Stream is not writeable')
 
+    def exhaust(self, chunk_size=64 * 1024):
+        """Exhaust the stream. This consumes all the data left until the
```
kgriffs: Nitpick/Style: The first line should not wrap (simply put the second sentence on its own line, i.e.:

```py
"""Exhaust the stream. 

Consume all of the remaining data, up to the content size limit.

...

"""
```

falcon/request_helpers.py
```diff
@@ -157,6 +157,24 @@ def write(self, data):
...
+        limit is reached.
+
+        Args:
+            chunk_size (int): The size for a chunk (default: 64 KB).
+                It will read the chunk until the stream is exhausted.
+        """
+        while 1:
```
kgriffs: Nitpick: `while True` may be considered a bit more explicit/pythonic.

falcon/request_helpers.py
DmitriiTrofimov: Done

falcon/request_helpers.py
DmitriiTrofimov: Done

DmitriiTrofimov:
@kgriffs All is done. Thanks for the review!

vytas7:
Thank you once again for this @DmitriiTrofimov :+1: :100: 
In retrospect, this was a small, but a very important milestone towards more flexible media handling.

I could only proudly add that both `multipart/form-data` and `application/x-www-form-urlencoded` media types are handled out of the box in Falcon 3.0.0a1:
* [`multipart/form-data`](https://falcon.readthedocs.io/en/3.0.0a1/api/multipart.html)
* [`application/x-www-form-urlencoded`](https://falcon.readthedocs.io/en/3.0.0a1/user/faq.html#how-can-i-access-posted-form-params)

Issue #1313, opened 2018-05-22
fallback_filename in api.add_static_route() always served as application/octet-stream


snobu:
When `fallback_filename` is defined,

```python
api.add_static_route('/', '/absolute/path', downloadable=False,
        fallback_filename='index.html')
```

`suffix` always comes up empty which results in the fallback file being sent over the wire as `application/octet-stream` (browser prompts for file download instead of rendering the HTML).

https://github.com/falconry/falcon/blob/master/falcon/routing/static.py#L109

The caveman-simple fix would be:

<s>

```python
        # suffix = os.path.splitext(file_path)[1] # comes up empty
        suffix = os.path.splitext(self._fallback_filename)[1]
        resp.content_type = resp.options.static_media_types.get(
            suffix,
            'application/octet-stream'
        )
```
I guess this completely skips the normalization logic so a more elegant fix should be in place.

</s>

^ Scratch that, the solution needs a bit more research.

I could take a short at a PR once you validate these findings.
I'm on Python 3.5.2, Linux with the current master HEAD as my falcon module.

vytas7:
Fixed in https://github.com/falconry/falcon/pull/1348 (see also https://github.com/falconry/falcon/pull/1314, https://github.com/falconry/falcon/pull/1325 & https://github.com/falconry/falcon/pull/1341).

PR #1314, opened 2018-05-22
Fix content-type for fallback static file

snobu:
Fixes https://github.com/falconry/falcon/issues/1313

### DIFF
--- a/falcon/routing/static.py
+++ b/falcon/routing/static.py
@@ -84,7 +84,6 @@ class StaticRoute(object):
             raise falcon.HTTPNotFound()
 
         normalized = os.path.normpath(without_prefix)
-
         if normalized.startswith('../') or normalized.startswith('/'):
             raise falcon.HTTPNotFound()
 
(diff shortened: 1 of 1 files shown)


falcon/routing/static.py
```diff
@@ -96,16 +95,21 @@ def __call__(self, req, resp):
...
         except IOError:
             if self._fallback_filename is None:
                 raise falcon.HTTPNotFound()
             try:
                 resp.stream = io.open(self._fallback_filename, 'rb')
+                serve_fallback_file = True
             except IOError:
                 raise falcon.HTTPNotFound()
 
+        if (serve_fallback_file):
```
kgriffs: Nitpick: It would probably be more pythonic to remove the parens here.

vytas7:
Would it be possible to add some sort of regression test, for instance, a new unit test that would have been failing with the old code?

vytas7:
The new PR reflecting the joint effort is out: https://github.com/falconry/falcon/pull/1348

Thanks to everyone involved for the genuine effort in fixing this bug! I have merely put the pieces together and added some unit tests & documentation notes.

kgriffs:
[superseded by  #1348]

Issue #842, opened 2016-07-07
Change the auto_parse_qs_csv Request option default to False


kgriffs:
It seems like more people use form-style lists than csv-style, so let's switching the default.

PR #1318, opened 2018-05-24
Set auto_parse_qs_csv to False, instead True #842

mikeylight:


MichaelYusko:
@nZac No problems, I'll looking for new tickets!

vytas7:
We have a couple of tests setting *auto_parse_qs_csv* to ``False``. I was thinking to myself if we could morph one of them, or add a new test, to check that the behaviour is corresponding to the new default even without setting it explicitly?

tests/test_query_params.py
```diff
@@ -136,6 +136,7 @@ def test_simple(self, simulate_request, client, resource):
     def test_percent_encoded(self, simulate_request, client, resource):
         query_string = 'id=23,42&q=%e8%b1%86+%e7%93%a3'
         client.app.add_route('/', resource)
+        client.app.req_options.auto_parse_qs_csv = True
```
vytas7: We have a couple of tests below setting *auto_parse_qs_csv* to False. I was thinking to myself if we could morph one of them, or add a new test, to check that the behaviour is corresponding to the new default even without setting it explicitly?

MichaelYusko: @vytas7 Yeah, agree! I'll add one more test without setting it explicitly, also I think that I'll push a commit in 1-2~ weeks, it works for you?

vytas7: Sure that works, your effort is much appreciated!

MichaelYusko:
@vytas7 added one more test.

falcon/request.py
```diff
@@ -1773,9 +1773,10 @@ class RequestOptions(object):
                 encoded according to the standard W3C algorithm (see
                 also http://goo.gl/6rlcux).
 
-        auto_parse_qs_csv: Set to ``False`` to treat commas in a query
-            string value as literal characters, rather than as a comma-
-            separated list (default ``True``). When this option is
+        auto_parse_qs_csv: Set to ``True`` to split on any
+            non-percent-encoded commas.
```
vytas7: Sorry for a bit of a late finding, but the sentence seems to be somehow broken here with a full stop, and continued in lower-case in a new line afterwards.

vytas7: I am probably too picky... but I would still like to polish this docstring even further :slightly_smiling_face: 

Right now it reads:
> Set to ``True`` to split on any non-percent-encoded commas. As a comma separated list, rather than a string value as literal characters (default ``False``).

The sentence got a bit fragmented, maybe we could join it into one sentence or rewrite in some other way?

vytas7: @MichaelYusko How about:

> Set to ``True`` to split a query string value on any non-percent-encoded commas.
> This option is disabled by default (``False``), and commas in the value will be treated as literal characters, rather than as a comma- separated list. The default behavior allows encoding lists as multiple occurrences of the same parameter, as well as cases when values may be encoded in alternative formats in which the comma character is significant.

@nZac suggestions are very welcome should you have any idea how to word this nicer :slightly_smiling_face:

MichaelYusko: @vytas7 Yeah, I like it! and maybe we will wait an answer from @nZac , maybe he will have some suggestions.

nZac: > Pass in `True` to split a query string value on any non-percent-encoded commas, this option is disabled by default (`False`). 

^ This looks good, keep it as the first line, below are two alternate phrasings. If we can give it a little white space, it might help readability, but I am not sure how to format that so renders correctly.

> When `False` query strings containing commas are untouched, this enables encoding lists as multiple occurrences of the same parameter and cases where the comma character is significant such as alternate encodings (i.e. `/?t=1,2,3&t=4` becomes `['1,2,3', '4']`).
> 
> When `True`, the query string value is split on the commas and returned as a list (i.e. `/?t=1,2,3&t=4` becomes `['1', '2', '3', '4']`). 


-- or --

> Consider this query string `/?t=1,2,3&t=4` and the impact on `req.params`. When `auto_parse_qs_csv` is `True` it is represented as `{ 't': ['1', '2', '3', '4'] }`, when `False` it will be `{ 't': ['1,2,3', '4'] }`. 
> 
> While auto-parsing can be a helpful shortcut, the default of `False` works better with alternative encodings and representing multiple parameters with the same key.

-- or --

You can use the previous one 😄.

MichaelYusko: @vytas7 @nZac I prefer this one, because documentation with some examples is always helpful.

```
When False query strings containing commas are untouched, this enables encoding lists as multiple occurrences of the same parameter and cases where the comma character is significant such as alternate encodings (i.e. /?t=1,2,3&t=4 becomes ['1,2,3', '4']).
```

falcon/request.py
```diff
@@ -1773,14 +1773,15 @@ class RequestOptions(object):
...
-            enabled, the value will be split on any non-percent-encoded
-            commas. Disable this option when encoding lists as multiple
-            occurrences of the same parameter, and when values may be
-            encoded in alternative formats in which the comma character
-            is significant.
+        auto_parse_qs_csv: Pass in ``True`` to split
+            a query string value on any non-percent-encoded commas,
+            this option is disabled by default (``False``).
+            When ``False`` query strings containing commas are untouched,
+            his enables encoding lists as multiple of the same parameter
```
vytas7: *his*?

falcon/request.py
```diff
@@ -1773,14 +1773,15 @@ class RequestOptions(object):
...
+            and cases where the comma character is significant
+            such as alternate encodings (i.e. ``/?t=1,2,3&t=4 becomes ['1,2,3', '4']``).
```
vytas7: *becomes* inbetween should not lie within backticks, see the examples by @nZac

falcon/request.py
```diff
@@ -1773,14 +1773,15 @@ class RequestOptions(object):
...
+            When ``True``, the query string value is split on the commas
+            and returned as a list (i.e. ``/?t=1,2,3&t=4 becomes ['1', '2', '3', '4']``).
```
vytas7: Same as above: *becomes* inbetween should not lie within backticks, see the examples by @nZac

MichaelYusko: @vytas7 Sorry, my bed. It's already fixed.

Issue #1321, opened 2018-06-05
feat: Add a "default" kwarg to Response.get_header()


kgriffs:
This provided parity with Request.get_header()

TheMushrr00m:
Hi, do you need help with this? 
i want to help to improve this great framework :)

TheMushrr00m:
#1322 #

PR #1322, opened 2018-06-12
feat(response): Add a "default" kwarg to Response.get_header() #1321

TheMushrr00m:


tests/test_headers.py
```diff
@@ -472,7 +472,9 @@ def test_response_set_and_get_header(self, client):
             assert result.headers['X-Auth-Token'] == 'toomanysecrets'
 
             assert resource.resp.location is None
-            assert resource.resp.get_header('not-real') is None
+            assert resource.resp.get_header('X-Header-Not-Set') is None
+            assert resource.resp.get_header('X-Header-Not-Set', 'Yes') == 'Yes'
+            assert resource.resp.get_header('X-Header-Not-Set', default='') == ''
```
vytas7: Could we also add a test case where the default is provided, but the header *is* set?

kgriffs: I went ahead and took care of this.

falcon/response.py
```diff
@@ -413,7 +413,7 @@ def unset_cookie(self, name):
         # thus removing it from future request objects.
         self._cookies[name]['expires'] = -1
 
-    def get_header(self, name):
+    def get_header(self, name, default=None):
```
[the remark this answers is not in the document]
kgriffs: I went ahead and took care of this.

PR #1325, opened 2018-06-18
Fix fallback filename content type inference

bendemaree:
When using a fallback filename (#1226), the path used when setting the response's `Content-Type` is an empty string, so all responses are returned as `application/octet-stream`. This changes the content type update to refer to the the fallback filename, if given, or the path.

### DIFF
--- a/falcon/routing/static.py
+++ b/falcon/routing/static.py
@@ -106,7 +106,11 @@ class StaticRoute(object):
             except IOError:
                 raise falcon.HTTPNotFound()
 
-        suffix = os.path.splitext(file_path)[1]
+        path_with_extension = file_path
+        if self._fallback_filename is not None:
+            path_with_extension = self._fallback_filename
+
+        suffix = os.path.splitext(path_with_extension)[1]
         resp.content_type = resp.options.static_media_types.get(
             suffix,
             'application/octet-stream'
--- a/tests/test_static.py
+++ b/tests/test_static.py
@@ -281,6 +281,9 @@ def test_e2e_fallback_filename(client, monkeypatch, strip_slash, path, fallback,
             assert response.status == falcon.HTTP_200
             assert response.text == directory + expected
 
+            if path.endswith('.html'):
+                assert response.headers['Content-Type'] == 'text/html'
+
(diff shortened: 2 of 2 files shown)


falcon/routing/static.py
```diff
@@ -106,7 +106,11 @@ def __call__(self, req, resp):
             except IOError:
                 raise falcon.HTTPNotFound()
 
-        suffix = os.path.splitext(file_path)[1]
+        path_with_extention = file_path
+        if self._fallback_filename is not None:
+            path_with_extention = self._fallback_filename
+
+        suffix = os.path.splitext(path_with_extention)[1]
```
jmvrbanac: Small spelling nit: `extention` should be `extension`

bendemaree:
@jmvrbanac serves me right for working on this too late in the evening. Thanks for catching that; updated!

falcon/routing/static.py
```diff
@@ -106,7 +106,11 @@ def __call__(self, req, resp):
             except IOError:
                 raise falcon.HTTPNotFound()
 
-        suffix = os.path.splitext(file_path)[1]
+        path_with_extension = file_path
```
nZac: This solution makes sense, however a little further down, `downloadable_as` pulls the file name from the `file_path`, wouldn't we want to update that as well?

For example, if someone requests a CSS file and we fallback to an HTML file, the content type will be `text/html` however the extension will be `.css` When in a browser that isn't a big deal I wouldn't think, but once it hits a filesystem, the OS will infer it incorrectly.

When we set the `resp.stream` in the `try...except` above, we might want to set the `file_path` there as well. That would also eliminate the `if`.

bendemaree:
@kgriffs fine with me! Can I help in some way? Should I close out this PR?

nZac:
Thanks @bendemaree for your world on this, it is very appreciated!

vytas7:
The new PR reflecting the joint effort is out: https://github.com/falconry/falcon/pull/1348

Thanks to everyone involved for the genuine effort in fixing this bug! I have merely put the pieces together and added some unit tests & documentation notes.

kgriffs:
[superseded by  #1348]

Issue #1135, opened 2017-11-08
Support http.HTTPStatus


floqqi:
It would be great to support [`http.HTTPStatus`](https://docs.python.org/3/library/http.html#http.HTTPStatus) from the standard library (3.5+).

Could possibly look like:

```python
from http import HTTPStatus

class MyResource:
    def on_get(self, req, res):
        res.status = HTTPStatus.NOT_FOUND  # instead of falcon.HTTP_404
```

kgriffs:
Not a bad idea. Since 1.4 is on the home stretch, let's look at this for 2.0. For the sake of backwards-compatibility, we will probably want to allow both the enum and the raw string.

vytas7:
As discussed on Gitter, I can take a stab at this.

Basic design:
* Python 3.5+ ``http.HTTPStatus`` or its subclass in Falcon will be THE way to set status
* Python 3.5+ ``http.HTTPStatus`` will be backported to be used in older Python versions
* Falcon 1.x status constants ``falcon.HTTP_*`` will be aliased to the new enum values to maintain backward compatibility for the most common usecases

Note that, strictly speaking, it will still be a breaking change in type compatibility, so setting status to raw strings not obtained from ``falcon.HTTP_*`` constants will yield a type error.

vytas7:
@kgriffs on Gitter, we agreed that a lookup table to speed things up is a reasonable approach to proceed.

You also indirectly implied it could be a status property:

> Kurt Griffiths
> Agreed, it should be good enough. One area where Falcon is opinionated is encouraging apps to follow the RFCs. Inventing app-specific codes sort of violates the uniform interface.
> In a pinch, I suppose someone could subclass Request (probably *Response* // Vytas) and override the status prop

Another alternative I had in my mind was having the lookup table on the API side.
The drawback is that should one desire to use some completely nonstandard type (yes, we are going to maintain 7XX, teapots and coffee machines), it will be even harder and dirtier to hook that in. Unless we add an API for that.

Thoughts?

In the attached ``test.py``, the lookup table on the API (no use of properties) seems to outdo even the current approach (at least on ``py27``) due to the absence ``if six.PY2`` clause.

xgfone:
A good idea!

When using falcon, I like `int` as the status. But falcon does not support it, so I wrap it.
```python
import falcon

STATUS_CODES = {}
for v in vars(falcon.status_codes).values():
    if isinstance(v, str):
        try:
[...]

PR #1334, opened 2018-07-02
 feat(API): Support http.HTTPStatus (a proof-of-concept prototype) (#1135)

vytas7:
This is not a fully polished version, but rather a prototype. Old documentation, tutorials and Python docstrings are still largely intact. It would be awesome to first discuss the proposed changeset with @kgriffs and @jmvrbanac, and distill an acceptable technical solution first; I would be happy to contribute to the documentation afterwards.

The main design ideas laid out below are roughly in-line with the plan found in #1135 :
* ``Response.status`` shall be set to ``http.HTTPStatus``, or its backport to older Python versions found in ``falcon.util.status_enum``
* Integer status codes implicitly work too (both a convenience and logical inheritance wise: ``http.HTTPStatus`` subclasses an ``int`` via ``IntEnum``)
* In the same ``falcon.util.status_enum``, a base HTTP status enum class is defined similarly to ``http.HTTPStatus``; that is used to support other non-standard status codes previously included in Falcon [1]
* Falcon usual status code constants are aliased to the new enum values [2], so a fair share of users could hopefully continue using most of their code as-is

Somewhat controversial ramifications:
* It may be cumbersome to define arbitrary codes; with the proposed changeset, one would need to resort to manually adding the string status to ``falcon.status_codes.COMBINED_STATUS_CODE`` lookup table. This could be made easier by either implementing support for legacy raw status, or probably by moving some part of this machinery to ``Response`` class, and documenting how and what to override to cram arbitrary status in; most probably both alternatives would come at a performance cost.

Other issues:
* Some of https://github.com/timothycrosley/hug tests are failing because of the same reason many Falcon tests were failing (assuming that ``falcon.HTTP_XXX`` are raw string statuses and using those values to check response status). @timothycrosley might now more, but from what I have seen in *hug* source code, it uses ``falcon.HTTP_XXX`` constants in a clean fashion, so it is just a problem with a couple of unit tests, not Hug API functionality.

1. As advertised!
> What other framework has integrated support for 786 TRY IT NOW?

2. On Python 3:
```python
>>> import falcon
>>> import http
>>> falcon.HTTP_404
<HTTPStatus.NOT_FOUND: 404>
>>> falcon.HTTP_404.phrase
'Not Found'
>>> int(falcon.HTTP_404)
404
>>> falcon.HTTP_FORBIDDEN
<HTTPStatus.FORBIDDEN: 403>
>>> falcon.HTTP_FORBIDDEN is http.HTTPStatus(403)
True
```

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -34,6 +34,7 @@ from falcon.response_helpers import (
     header_property,
     is_ascii_encodable,
 )
+from falcon.status_codes import HTTP_OK, COMBINED_STATUS_CODES
 from falcon.util import dt_to_http, TimezoneGMT
 from falcon.util.uri import encode as uri_encode
 from falcon.util.uri import encode_value as uri_encode_value
@@ -151,7 +152,7 @@ class Response(object):
     context_type = dict
 
     def __init__(self, options=None):
-        self.status = '200 OK'
+        self.status = HTTP_OK
         self._headers = {}
 
         self.options = options if options else ResponseOptions()
@@ -802,6 +803,10 @@ class Response(object):
             type.
 
         """)
+
+    # @property
+    # def status_line(self):
+    #     return COMBINED_STATUS_CODES[self.status]
 
     def _set_media_type(self, media_type=None):
         """Wrapper around set_header to set a content-type.
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -34,7 +34,7 @@ from falcon.response_helpers import (
     header_property,
     is_ascii_encodable,
 )
-from falcon.status_codes import HTTP_OK, COMBINED_STATUS_CODES
+from falcon.status_codes import HTTP_OK
 from falcon.util import dt_to_http, TimezoneGMT
 from falcon.util.uri import encode as uri_encode
 from falcon.util.uri import encode_value as uri_encode_value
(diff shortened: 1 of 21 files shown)


falcon/response.py
```diff
@@ -803,6 +804,10 @@ def add_link(self, target, rel, title=None, title_star=None,
 
         """)
 
+    # @property
```
vytas7: An open question: could something like this be of use to API users?

falcon/util/misc.py
```diff
@@ -320,34 +322,50 @@ def get_argnames(func):
...
     try:
         return getattr(status_codes, 'HTTP_' + str(code))
     except AttributeError:
-        # not found
-        return str(code) + ' ' + default_reason
+        raise ValueError('get_http_status failed: %s is not a '
+                         'supported status code' % status_code)
+
+
+def get_http_status_line(status):
```
vytas7: Suggestions for a better name are very welcome as this is a bit of a misnomer.

Technically HTTP status line includes HTTP protocol, so a full status would look like ``HTTP/1.1 404 Not Found``.

vytas7:
Killing this for now, I will try to port optimizations and design changes in small increments instead.

Issue #1340, opened 2018-07-12
Falcon's TestClient breaks when request headers contain unicode values in Python 2.7


vishal-bala:
This is caused by internal validation checks in `wsgiref`, which assert that stored environment variables are of `str` type ([source code](https://github.com/python/cpython/blob/2.7/Lib/wsgiref/validate.py#L318-L320)).

When trying to write code that is both Python 2 and 3 compatible, it's common to use
```python
from __future import unicode_literals
```
However, this causes the TestClient to start breaking once any headers are passed with a simulated request.

## Reproduction

Reproduced below using Python 2.7.15 and Falcon v1.4.1.

#### Without unicode strings
```python
>>> import falcon
>>> from falcon.testing import TestClient
>>> app = falcon.API()
>>> client = TestClient(app)
>>> client.simulate_get('/')
<falcon.testing.client.Result object at 0x1055bfcd0>
>>> client.simulate_get('/', headers={'Authorization': 'token blah'})
<falcon.testing.client.Result object at 0x105708250>
```

#### With unicode strings
```python
>>> from __future__ import unicode_literals
>>> import falcon
>>> from falcon.testing import TestClient
>>> app = falcon.API()
>>> client = TestClient(app)
>>> client.simulate_get('/')
<falcon.testing.client.Result object at 0x10af21cd0>
>>> client.simulate_get('/', headers={'Authorization': 'token blah'})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/vbala/.pyenv/versions/atlas2/lib/python2.7/site-packages/falcon/testing/client.py", line 639, in simulate_get
    return self.simulate_request('GET', path, **kwargs)
  File "/Users/vbala/.pyenv/versions/atlas2/lib/python2.7/site-packages/falcon/testing/client.py", line 702, in simulate_request
    return simulate_request(self.app, *args, **kwargs)
  File "/Users/vbala/.pyenv/versions/atlas2/lib/python2.7/site-packages/falcon/testing/client.py", line 341, in simulate_request
    iterable = validator(env, srmock)
  File "/Users/vbala/.pyenv/versions/2.7.15/lib/python2.7/wsgiref/validate.py", line 148, in lint_app
    check_environ(environ)
  File "/Users/vbala/.pyenv/versions/2.7.15/lib/python2.7/wsgiref/validate.py", line 320, in check_environ
    % (key, type(environ[key]), environ[key]))
[...]

PR #1342, opened 2018-07-16
Convert TestClient Header Values to `str` in Py2

nZac:
Fixes GH-1340

### DIFF
--- a/falcon/testing/client.py
+++ b/falcon/testing/client.py
@@ -21,6 +21,7 @@ WSGI callable, without having to stand up a WSGI server.
 import warnings
 import wsgiref.validate
 
+import six
 from six.moves import http_cookies
 
 from falcon.constants import MEDIA_JSON
@@ -684,5 +685,9 @@ class TestClient(object):
             merged_headers.update(additional_headers)
 
             kwargs['headers'] = merged_headers
+
+        if six.PY2 and 'headers' in kwargs:
+            for key, value in six.iteritems(kwargs['headers']):
+                kwargs['headers'][key] = str(value) if isinstance(value, unicode) else value
 
         return simulate_request(self.app, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


falcon/testing/client.py
```diff
@@ -685,4 +686,8 @@ def simulate_request(self, *args, **kwargs):
 
             kwargs['headers'] = merged_headers
 
+        if six.PY2 and 'headers' in kwargs:
+            for key, value in six.iteritems(kwargs['headers']):
+                kwargs['headers'][key] = str(value) if type(value) == six.text_type else value
```
vytas7: It is more idiomatic to use ``isinstance(...)`` than comparing types this way.

Furthermore you are targeting ``PY2`` specifically, so you could just say ``unicode`` directly. The latter remark is a matter of taste, not a requirement though. And performance is unimportant in testing client.

tests/test_py2_wsgi.py
```diff
@@ -0,0 +1,29 @@
...
+
+import pytest
+import six
+
+
+def test_test_client_works_in_py2():
+    app = falcon.API()
+    client = TestClient(app)
+    client.simulate_get('/', headers={
+        'Authorization': 'token blah'
```
vytas7: Another matter of taste thing: instead of ``unicode_literals``, you could just parametrize the test and test for both ``'token_blah'`` and ``u'token_blah'``, which would bring some redundancy on PY3, but something we could probably live with.

tests/test_py2_wsgi.py
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Test unicode_literals in python2 invalidating headers which come as Unicode values"""
+from __future__ import unicode_literals
+
+import falcon
+from falcon.testing import TestClient
+
+import pytest
+import six
```
vytas7: Unused imports?

nZac:
@kgriffs in the test file of this PR is that the correct file preamble to use?

kgriffs:
@nZac For new files the author controls the copyright, so you can use your own name or your organization.

Issue #1269, opened 2018-05-12
Add a headers property to Response


kgriffs:
Return a list of all headers already set for the Response. Don't return the internal data structure directly to avoid breaking the abstraction and enabling hard-to-debug problems in apps.

alysivji:
Looking into this issue.

PR #1343, opened 2018-08-04
Add headers property to Response class

alysivji:
Closes https://github.com/falconry/falcon/issues/1269

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,7 @@ Changes to Supported Platforms
 New & Improved
 --------------
 
-- Added new ``falcon.Reponse.headers`` property to ``Response`` class.
+- Added a new ``headers`` property to the ``Response`` class.
 
 Fixed
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,7 @@ Changes to Supported Platforms
 New & Improved
 --------------
 
-- Added new ``falcon.Reponse.headers`` property to ``Response`` class
+- Added new ``falcon.Reponse.headers`` property to ``Response`` class.
 
 Fixed
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,8 @@ Changes to Supported Platforms
 New & Improved
 --------------
 
+- Added new property ``falcon.Reponse.headers`` property to ``Response`` class
+
 Fixed
 -----
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,7 @@ Changes to Supported Platforms
 New & Improved
 --------------
 
-- Added new property ``falcon.Reponse.headers`` property to ``Response`` class
+- Added new ``falcon.Reponse.headers`` property to ``Response`` class
 
 Fixed
 -----
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -133,7 +133,9 @@ class Response(object):
 
         options (dict): Set of global options passed from the API handler.
 
-        headers (dict): Read-only list of all headers set for Response.
+        headers (dict): Copy of all headers set for the response,
+            sans cookies. Note that a new copy is created and returned each
+            time this property is referenced.
     """
 
     __slots__ = (
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -133,7 +133,7 @@ class Response(object):
 
         options (dict): Set of global options passed from the API handler.
 
-        headers (list): Read-only list of all headers set for Response
+        headers (list): Read-only list of all headers set for Response.
     """
 
     __slots__ = (
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -132,6 +132,8 @@ class Response(object):
                 the Response instance itself (self).
 
         options (dict): Set of global options passed from the API handler.
+
+        headers (list): Read-only list of all headers set for Response
     """
 
     __slots__ = (
@@ -201,6 +203,10 @@ class Response(object):
     @data.setter
     def data(self, value):
         self._data = value
+
+    @property
+    def headers(self):
+        return [header for header, value in self._headers.items()]
 
     @property
     def media(self):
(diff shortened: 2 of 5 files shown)


falcon/response.py
```diff
@@ -202,6 +204,10 @@ def data(self):
     def data(self, value):
         self._data = value
 
+    @property
+    def headers(self):
+        return [header for header, value in self._headers.items()]
```
vytas7: Do we really need to iterate *items* just to create a list?
Also note that `items()` does a full copy on PY2, so we would be allocation twice completely needlessly.

Could we potentially simplify as:

```python
    @property
    def headers(self):
        return list(self._headers)
```

(Or *tuple(...)*, I have not give much thought about that).

alysivji: Didn't realize items did a full copy. Will change. Great catch!

vytas7: I have briefly gone through with what we have in other Falcon *Request* / *Response* properties and ``list`` seems more in line with the existing stuff, so lets go with it.

``tuple(dict argument)`` is, on the other hand, about 15-20% faster. But considering this is a property for information purposes / user convenience rather than a crucial piece used in forming every response, performance is not that important to optimize in this case. Unless @kgriffs has any insights on this.

falcon/response.py
```diff
@@ -132,6 +132,8 @@ class Response(object):
                 the Response instance itself (self).
 
         options (dict): Set of global options passed from the API handler.
+
+        headers (list): Read-only list of all headers set for Response
```
vytas7: Documentation sentences for other members of the class end with a full stop (period), let us add one here too.

CHANGES.rst
```diff
@@ -13,6 +13,8 @@ Changes to Supported Platforms
 New & Improved
 --------------
 
+- Added new ``falcon.Reponse.headers`` property to ``Response`` class
```
vytas7: A full stop (period) missing here too.

falcon/response.py
```diff
@@ -202,6 +204,10 @@ def data(self):
     def data(self, value):
         self._data = value
 
+    @property
+    def headers(self):
+        return list(self._headers)
```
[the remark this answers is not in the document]
vytas7: (Sorry for not bringing this up earlier -- this came to my head the last second before merging :grimacing: ).

A side note: the importance of performance of this property, as I pointed before, is close to negligible, but I have benchmarked anyway :slightly_smiling_face: If we agree on making a ``dict`` copy, just ``self._headers.copy()`` seems to be the faster alternative, as opposed to dict(self._headers) (a close second) and other methods.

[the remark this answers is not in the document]
alysivji: Made headers into a `dict`. Thanks @vytas7 for benchmarking performance! Made my job a bit easier.

falcon/response.py
```diff
@@ -202,6 +204,10 @@ def data(self):
     def data(self, value):
         self._data = value
 
+    @property
+    def headers(self):
+        return dict(self._headers)
```
vytas7: Most probably I was unclear in my comment. But I meant I had found ``.copy()`` to be the fastest.

falcon/response.py
```diff
@@ -132,6 +132,8 @@ class Response(object):
                 the Response instance itself (self).
 
         options (dict): Set of global options passed from the API handler.
+
+        headers (list): Read-only list of all headers set for Response.
```
vytas7: Let us update the docstrings to match the changes in design: _headers (~~list~~ dict):..._ etc

falcon/response.py
alysivji: Just misread. Apologies. This is fixed.

CHANGES.rst
```diff
@@ -13,6 +13,8 @@ Changes to Supported Platforms
 New & Improved
 --------------
 
+- Added new ``falcon.Reponse.headers`` property to ``Response`` class.
```
kgriffs: Would you mind adding a line to `docs/changes/2.0.0.rst` as well?

falcon/response.py
```diff
@@ -132,6 +132,8 @@ class Response(object):
                 the Response instance itself (self).
 
         options (dict): Set of global options passed from the API handler.
+
+        headers (dict): Read-only list of all headers set for Response.
```
kgriffs: Nitpick: We probably should say "dict" or "collection" or something, instead of "list".

kgriffs: Would you mind adding a note about the header names being lowercased for performance reasons? People will need to know that we don't use a case-insensitive dict (again, for performance reasons).

tests/test_response.py
```diff
@@ -11,3 +13,31 @@ def test_response_set_content_type_set():
 def test_response_set_content_type_not_set():
     resp = falcon.Response()
     assert 'content-type' not in resp._headers
+
+
+def test_response_get_headers():
```
kgriffs: Would you mind adding some tests that verify that appending a camel-cased header results in a lowercased header, since that is part of the contract? We should also test that attempting to look up a value using the same non-lowercased form used to set the header raises an instance of `KeyError`.

tests/test_response.py
```diff
@@ -11,3 +13,31 @@ def test_response_set_content_type_set():
...
+    headers = resp.headers
+    assert headers['x-things1'] == 'thing-1'
+    assert headers['x-things2'] == 'thing-2'
+    assert headers['x-things3'] == 'thing-3'
+
+
+def test_response_attempt_to_set_read_only_headers():
+    resp = falcon.Response()
+
+    resp.append_header('x-things1', 'thing-1')
```
kgriffs: For completeness, could you append multiple values to the same header as well, and check that it comes out as a comma-separated string value?

kgriffs:
I added a few more commits per my most recent comments.

alysivji:
Apologies. Was going to get to this over the weekend. Thanks for taking care of it!

kgriffs:
@alysivji No worries, thanks for all your work on this!

PR #1344, opened 2018-08-07
using raw instead of encoding='utf-8' and removed deprecated encoding…

navyad:
… for packer


vytas7:
Looking good now :+1:

PR #1346, opened 2018-08-17
feat(json): refactor third party JSON support and extend falcon.media.JSONHandler

mattgiles:
Possible solution for #1192 and #1206.

I think there are two obvious open questions.

1. JSON functions are consumed from `falcon.util` in a few different places in the application. We have only parameterized the `media.JSONHandler` to address #1192. Does it make any sense to try to expose parameters such that they would also be used consistently in `Request` and `HTTPError` objects?

2. Right now, `mujson` declares explicit default rankings as to which JSON libs are preferred for different functions across different Python versions. The behavior defined in this PR is more or less a drop-in replacement for old `ujson` behavior, and it enables identical *implicit* behavior for users that install other libs like [yajl](https://github.com/rtyler/py-yajl) or [rapidjson](https://github.com/python-rapidjson/python-rapidjson). Which addresses #1206. However, `mujson` does not contemplate JSON libs which may arrive in the future (e.g. [hyperjson](https://github.com/mre/hyperjson)). The parameters for `media.JSONHandler` do include `dumps` and `loads` functions, so a user can explicitly override default rankings. But, depending on the answer to `1.` above it may make sense to give users the option to just specify `json_lib="hyperjson"` in a single place.

All of which is to say that this is a heavy PR insofar as it introduces a new top level requirement, but it is light insofar as it avoids a more opinionated refactoring of the configuration interface exposed to Falcon users -- which may or may not be desirable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -202,11 +202,11 @@ Installing it is as simple as:
 
     $ pip install falcon
 
-If `ujson <https://pypi.python.org/pypi/ujson>`__ is available, Falcon
-will use it to speed up media (de)serialization, error serialization,
-and query string parsing. Note that ``ujson`` can actually be slower
-on PyPy than the standard ``json`` module due to ctypes overhead, and
-so we recommend only using ``ujson`` with CPython deployments:
+Falcon uses `mujson <https://pypi.python.org/pypi/mujson>`__ to select the most
+performant JSON functions available at import time. For instance, if `ujson
+<https://pypi.python.org/pypi/ujson>`__ is available, Falcon will use it to
+speed up media (de)serialization, error serialization, and query string
+parsing.
 
 .. code:: bash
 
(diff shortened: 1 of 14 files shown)


vytas7:
@mattgiles let us keep it open for a bit longer I would say, it is not impossible that we would offer recipes involving *mujson*, or even ask you if you are OK with us vendoring it inside Falcon, and so on.

We are still very much in the middle of prototyping and brainstorming custom JSON handling for Falcon 2.0: https://github.com/falconry/falcon/pull/1404
The work is being overseen by @nZac .

kgriffs:
Thanks for your work on this! It really helped us compare and contrast the different ways we could approach this problem. That being said, after discussing this with the maintainers team we ultimately decided to go with #1404. If you have a minute to take a look and provide your feedback, that would be great.

Thanks again!

kgriffs:
[superseded by #1404]

Issue #1351, opened 2018-09-12
Resource level hook doesn't apply to suffixed methods


mumrau:
Hello,

I believe the title of the issue is pretty clear, a @falcon.before hook wouldn't apply to a suffixed method (on_get_suffix) handler while it would perfectly apply to the classic legacy handlers (on_get, on_post,...).
Consider that piece of code:
```python
import json
from wsgiref import simple_server
import falcon

def hook(req, resp, resource, params):
    print("Inside the hook for resource: ", type(resource))
    resp.body = json.dumps("HOOKED")

@falcon.before(hook)
class HookedResource:
    def on_get(self, req, resp):
        resp.status = falcon.HTTP_200
    def on_get_suffix(self, req, resp):
        resp.status = falcon.HTTP_202

class HookedMethod:
    @falcon.before(hook)
    def on_get_suffix(self, req, resp):
        resp.status = falcon.HTTP_202

api = falcon.API()

hkr = HookedResource()
hkm = HookedMethod()

api.add_route("/hooked_resource", hkr)
api.add_route("/hooked_resource_suffix", hkr, suffix="suffix")
api.add_route("/hooked_method_suffix", hkm, suffix="suffix")

if __name__ == '__main__':
    httpd = simple_server.make_server('127.0.0.1', 8000, api)
    httpd.serve_forever()
```

The hooked_resource route (resource level hook) would have a functional hook, but the hooked_resource_suffix won't.
Nevertheless, the hooked_method_suffix (method level hook) will behave correctly, which gives a lot of hope in terms of implementation effort for a fix.

I hope I have been clear enough.

Falcon still rocks!

EDIT:
I believe this code is responsible in https://github.com/falconry/falcon/blob/master/falcon/hooks.py :
```python
    def _before(responder_or_resource):
        if isinstance(responder_or_resource, six.class_types):
            resource = responder_or_resource

            for method in COMBINED_METHODS:
                responder_name = 'on_' + method.lower()
```
It only iterates through COMBINED_METHODS which relates to the default responders which are constants.
Maybe it would be worth to iterate through every functions of the class and select starting with `on_HTTP_VERB_` as well.
On handlers registrations, we could also register in the class an attribute like "additional_responders" or something like that. I peeked at the source code but can't really take any decision cause I'm not fully aware of the guidelines in terms of implementation.

vytas7:
Thank you for the detailed report! It seems this has been indeed overlooked when adding suffixed resources, well spotted.

[...]

PR #1352, opened 2018-09-13
Fix for resource-level before hooks not applying to suffixed responders

mumrau:
This PR is intended for this issue https://github.com/falconry/falcon/issues/1351 and was also discussed on gitter briefly (@vytas7 )
Basically the hooks were not applying to the responders added with a suffix,  a standalone fairly clear "test" can be found below.

This looks like a decent fix and makes sense to me, I hope it does to you as well as I don't clearly see all the eventual consequences of this change.
Couldn't come up with a test, cause I couldn't understand deeply enough to come up with one of my own, plus I didn't wanted to "waste" time considering the fact that my approach might be terrible.
The tests on the repository are passing, and my little standalone one as well.

I applied it to the @before hook, I can totally apply the same change to the @after hook for consistency of course.

I'm looking forward to your feedback and I am totally open for discussion, on gitter or here.

```python
import json
from wsgiref import simple_server
import falcon

def hook(req, resp, resource, params):
    print("Inside the hook for resource: ", type(resource))
    resp.body = json.dumps("HOOKED")

@falcon.before(hook)
class HookedResource:
    def on_get(self, req, resp):
        resp.status = falcon.HTTP_200
    def on_get_suffix(self, req, resp):
        resp.status = falcon.HTTP_202

class HookedMethod:
    @falcon.before(hook)
    def on_get_suffix(self, req, resp):
        resp.status = falcon.HTTP_202

api = falcon.API()
hkr = HookedResource()
hkm = HookedMethod()
api.add_route("/hooked_resource", hkr)
api.add_route("/hooked_resource_suffix", hkr, suffix="suffix")
api.add_route("/hooked_method_suffix", hkm, suffix="suffix")

if __name__ == '__main__':
    httpd = simple_server.make_server('127.0.0.1', 8000, api)
    httpd.serve_forever()
```

### DIFF
--- a/falcon/hooks.py
+++ b/falcon/hooks.py
@@ -13,13 +13,27 @@
 # limitations under the License.
 
 """Hook decorators."""
-import inspect
+
 from functools import wraps
+import inspect
 
 import six
 
 from falcon import COMBINED_METHODS
 from falcon.util.misc import get_argnames
+
+
+def predicate_method_function(member):
+    """Predicate to use for inspection of a resource in order to extract the responders.
+
+    Args:
+        member (member): A member tuple as returned per the inspect.getmembers function
+        where the first element of the tuple is the name and the second element the member itself.
+
+        Note:
+            Should be only used as a predicate of inspect.getmembers(to_inspect, predicate=this function.
+    """
+    return inspect.ismethod(member) or inspect.isfunction(member)
 
 
 def before(action, *args, **kwargs):
@@ -57,13 +71,13 @@ def before(action, *args, **kwargs):
         if isinstance(responder_or_resource, six.class_types):
             resource = responder_or_resource
 
-            members = inspect.getmembers(resource, predicate=inspect.isfunction)
+            members = inspect.getmembers(resource, predicate=predicate_method_function)
 
             responders = []
 
             for member in members:
                 for method in COMBINED_METHODS:
-                    if member[0].startswith("on_" + method.lower()):
+                    if member[0].startswith('on_' + method.lower()):
                         responders.append(member)
 
             for responder_name, responder in responders:
--- a/falcon/hooks.py
+++ b/falcon/hooks.py
@@ -31,7 +31,8 @@ def predicate_method_function(member):
         where the first element of the tuple is the name and the second element the member itself.
 
         Note:
-            Should be only used as a predicate of inspect.getmembers(to_inspect, predicate=this function.
+            Should only be used as a predicate of
+            inspect.getmembers(to_inspect, predicate=this function).
     """
     return inspect.ismethod(member) or inspect.isfunction(member)
 
--- a/falcon/hooks.py
+++ b/falcon/hooks.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 """Hook decorators."""
-
+import inspect
 from functools import wraps
 
 import six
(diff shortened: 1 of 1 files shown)


falcon/hooks.py
```diff
@@ -57,28 +72,27 @@ def _before(responder_or_resource):
...
-                        # variable that is shared between iterations of the
-                        # for loop, above.
-                        def let(responder=responder):
-                            do_before_all = _wrap_with_before(responder, action, args, kwargs)
+            responders = []
 
-                            setattr(resource, responder_name, do_before_all)
+            for member in members:
+                for method in COMBINED_METHODS:
+                    if member[0].startswith('on_' + method.lower()):
```
vytas7: Probably ``'on_' + method.lower() + '_'`` to make it even more specific, if one wants to go this way.

vytas7:
Hi again @Mumrau !
As you write, we have briefly discussed this on Gitter, but thanks once again for your contribution to debugging this issue!

As for the solution itself, we'll have to discuss it further with other maintainers and contributors.

One has a few different paths of solution here:
* As you propose, apply hooks to anything that resembles a (potentially suffixed) responder
* Require ``suffix`` parameter from the hook itself when decorating a whole class
* Somehow hook-ify all methods that resemble a responder, but record the parsed suffix in the closure or some attribute, and then trigger these on only upon adding an instance via ``add_route`` provided the suffix matches; but maybe too complex/too much dark sorcery needed?

Mumrau:
I'm sorry I couldn't find time to test it thoroughly (even tho it's live on my production servers, which doesn't make it tested but  _tried_).

- The first solution looks good to me (which is kinda why I took that path haha), we just need to make sure we're clear on that because it is quite strong in terms of constraints: if something resembles a responder but is not, it is involved as well. That doesn't seem counter-intuitive to me as you don't want to mess with the framework in terms of naming and conventions, but it could mess people's installations.
- I'm not sure about the `suffix` parameter on the hook, I think I'd rather have a more globally controlled way of applying hooks, but this is not up to me to decide that.
- Third solution would work, but it seems complex to me, and the gain from doing that seems quite marginal.

(Thanks! Seems good to me, I haven't signed-off the commits yet, I'll do so if necessary.)

kgriffs:
Some variation of (C) would be easiest for the app developer (because it would not require specifying suffixes explicitly), while also being correct, meaning it would avoid hooking methods that aren't actually meant to be suffixed responders. This solution is complicated by the fact that multiple routes could be added with multiple suffixes, so those would have to filtered dynamically (with a minor performance penalty) or statically (faster, but even more complicated).

(B) Would also result in correct behavior, provided that you can pass a list of suffixes (rather than just one). However, it does seem a bit tedious and requires keeping the hook decorators in sync with the add_route() calls.

(A) Is the simplest to implement as well as to employ in apps. However, this approach makes it possible to accidentally hook non-responders. That being said, it is probably rare for all of these things to be true at once: (1) A public (i.e., not prefixed with `_`) method is exposed on a resource class that is not a responder, (2) Said public method follows the on_* naming convention but is not a responder, (3) a hook is applied at the class level.

So, in summary we have:

(A) Simplest to implement, maintain and use. May hook non-responder methods.
(B) Harder to use, somewhat simple to implement, but not DRY. Never hooks non-responder methods.
(C) Hardest to implement and maintain, yet easy to use. DRY. Never hooks non-responder methods.

Personally I would lean toward (A) and then switch to (C) later if we get reports that it is hooking the wrong methods in a fair number of apps. That being said, this behavior should be clearly described in the hook docs to avoid surprises.

falcon/hooks.py
vytas7: Ah, true.
What I trying to suggest though was that we should match ``on_post`` and ``on_post_suffixed``, but not ``on_postman``.

kgriffs:
@vytas7 @nZac Thoughts? I'd like to get to a consensus and land this patch ASAP. Thanks!

vytas7:
> Personally I would lean toward (A) and then switch to (C) later if we get reports that it is hooking the wrong methods in a fair number of apps. That being said, this behavior should be clearly described in the hook docs to avoid surprises.

I agree with your view on this.
(Tbh none of the options is particularly pretty, that's why I laid them out so we could choose)

vytas7:
Hi again @mumrau ! 

Would you have some time to add some unit tests? It seems we are leaning towards your approach (A), and most of the unit tests would test the behaviour so they would be essentially independent of the chosen approach, maybe save for (B), but I doubt we would pursue that way (too tedious from the API user perspective).

mumrau:
Hello!
Sorry, I'm being quite busy with out of office stuff.
It really depends on the deadline, I'll be back to business within a week, and I probably can ensure that the fixes could be applied by the end of January.
Let me know if these would work for you as a deadline.
I'll of course make sure it also works on the @after hooks, I might ask questions (on gitter most likely) towards the testing approach.

vytas7:
No worries @mumrau , thanks for the extremely quick reply. There is no strict deadline, but ideally we would like have this fixed rather quicker than sooner, and push out 2.0.

I will grab this myself then for now, but let me know as soon as you are back so we can coordinate.

And thanks for your great effort already put into this so far 👍

falcon/hooks.py
```diff
@@ -15,13 +15,28 @@
...
 from functools import wraps
+import inspect
 
 import six
 
 from falcon import COMBINED_METHODS
 from falcon.util.misc import get_argnames
 
 
+def predicate_method_function(member):
```
vytas7: I am thinking to revert this approach by @mumrau in the follow-up PR. While I personally like it being very clean, nice and explicit, but I think just listing all the members, and checking those that quack a *callable*, are probably good enough to decorate.

Consider the following example:

```python
import inspect

import falcon


def predicate_method_function(member):
    return inspect.ismethod(member) or inspect.isfunction(member)


class HelloSayer(object):

    def __call__(self, req, resp):
        resp.media = {'Hello': 'World'}


class Resource(object):

    on_get = HelloSayer()


api = falcon.API()
api.add_route('/hello', Resource())

print(inspect.getmembers(Resource, predicate_method_function))
```

It is undeniably a bit contrived, but OTOH we cannot know how exactly people are building their APIs.

It does say Hello, but ``inspect.getmembers`` returns an empty list with the exemplified predicate function.

mumrau: Makes sense. Totally didn't think of that approach, stayed focused on the "on_http(_optional_follow_up_subhandler)".
I agree with the callables inspections then.

vytas7:
Superseded by https://github.com/falconry/falcon/pull/1402

PR #1360, opened 2018-10-02
update user faq, correct example use of auto_parse_qs_csv

bkcsfi:
change user faq so that example of api.auto_parse_qs_csv is changed to correct use as api.req_options.auto_parse_qs_csv

### DIFF
--- a/docs/user/faq.rst
+++ b/docs/user/faq.rst
@@ -564,7 +564,7 @@ handles commas in a query string by setting the
 
 .. code:: python
 
-    api.auto_parse_qs_csv = False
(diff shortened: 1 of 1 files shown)


nZac:
Thanks @bkcsfi for submitting this, we are thrilled that you would take time to submit a patch.

#1354 beat you to the task on this one. If you would like to continue to contribute, we have a bunch of [open documentation issues](https://github.com/falconry/falcon/issues?q=is%3Aopen+is%3Aissue+label%3Adocumentation)  that need fixing.

I am going to close this for now, reply back if you have something additional to add or if I missed something.

PR #1373, opened 2018-10-24
Decode request path to a unicode under python 2.7

### LINKED ISSUE #1031: Decode non-ASCII UTF-8 characters in the request path before routing (Python 2.6/7 only)
HTTP/1.1 allows for non-ASCII characters to be included in URLs via UTF-8 percent-encoding. WSGI servers decode the URL to bytes, leaving a UTF-8 bytestring under Python 2.6/7 that must be decoded to a `unicode` string for proper routing, etc. when non-ASCII characters are present. This is a breaking change.

This means that req.path would now always be a `unicode` type, and we'll need to note this in the docstring.

kandziu:
Closes #1031

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -183,6 +183,10 @@ class Request(object):
             Note:
                 `req.path` may be set to a new value by a `process_request()`
                 middleware method in order to influence routing.
+
+            Note:
+                `req.path` always decode to a unicode.
+
         query_string (str): Query string portion of the request URI, without
             the preceding '?' character.
         uri_template (str): The template for the route that was matched for
@@ -426,15 +430,15 @@ class Request(object):
         # empty string, so normalize it in that case.
         path = env['PATH_INFO'] or '/'
 
-        if six.PY2:
-            # NOTE(kandziu): WSGI servers decode the URL to bytes, leaving
-            # a "UTF-8 bytestring". Under Python 2.6/7 that must be decoded
-            # to a unicode.
-            path = path.decode('utf-8', 'replace')
-        else:
+        if six.PY3:
             # PEP 3333 specifies that PATH_INFO variable are always
             # "bytes tunneled as latin-1" and must be encoded back
             path = path.encode('latin1').decode('utf-8', 'replace')
+        else:
+            # NOTE(kandziu): WSGI servers decode the URL to bytes, leaving
+            # a "UTF-8 bytestring". Under Python 2.7 that must be decoded
+            # to a unicode.
+            path = path.decode('utf-8', 'replace')
 
         if (self.options.strip_url_path_trailing_slash and
                 len(path) != 1 and path.endswith('/')):
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -426,7 +426,12 @@ class Request(object):
         # empty string, so normalize it in that case.
         path = env['PATH_INFO'] or '/'
 
-        if six.PY3:
+        if six.PY2:
+            # NOTE(kandziu): WSGI servers decode the URL to bytes, leaving
+            # a "UTF-8 bytestring". Under Python 2.6/7 that must be decoded
+            # to a unicode.
+            path = path.decode('utf-8', 'replace')
+        else:
             # PEP 3333 specifies that PATH_INFO variable are always
             # "bytes tunneled as latin-1" and must be encoded back
             path = path.encode('latin1').decode('utf-8', 'replace')
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -26,7 +26,6 @@ in the `falcon` module, and so must be explicitly imported::
 
 import six
 
-
 # NOTE(kgriffs): See also RFC 3986
 _UNRESERVED = ('ABCDEFGHIJKLMNOPQRSTUVWXYZ'
                'abcdefghijklmnopqrstuvwxyz'
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -27,9 +27,6 @@ in the `falcon` module, and so must be explicitly imported::
 import six
 
 
-# NOTE(kandziu): Invalid unicode character
-FFFD = u'\ufffd'
-
 # NOTE(kgriffs): See also RFC 3986
 _UNRESERVED = ('ABCDEFGHIJKLMNOPQRSTUVWXYZ'
                'abcdefghijklmnopqrstuvwxyz'
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -26,6 +26,10 @@ in the `falcon` module, and so must be explicitly imported::
 
 import six
 
+
+# NOTE(kandziu): Invalid unicode character
+FFFD = u'\ufffd'
+
 # NOTE(kgriffs): See also RFC 3986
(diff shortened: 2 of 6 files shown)


vytas7:
@kandziu thanks for the effort! And no, the pep8 problem was caused by the new Flake8 release which had introduced breaking changes into how the ``--ignore`` lists are treated wrt default ignores, as well as the new checks which we were not in compliance with. This has been addressed, and I have merged the master into your branch. Let us keep the fingers crossed Travis passes this time :slightly_smiling_face: 

P.S. I have taken the liberty to edit out Python 2.6 from your title as Falcon no longer supports it.

kandziu:
Friendly ping! @vytas7 can you please review this PR?

falcon/request.py
```diff
@@ -426,7 +426,12 @@ def __init__(self, env, options=None):
         # empty string, so normalize it in that case.
         path = env['PATH_INFO'] or '/'
 
-        if six.PY3:
+        if six.PY2:
```
vytas7: Could we flip this conditional statement to check for ``PY3`` first, since
* Python 3 is supposedly getting more popular than Python 2 (the latter would be end of life in 2020)
* Python 3 is less performant

falcon/util/uri.py
```diff
@@ -26,6 +26,10 @@
 
 import six
 
+
+# NOTE(kandziu): Invalid unicode character
+FFFD = u'\ufffd'
```
vytas7: Any reason for singling out ``FFFD``?
There are more non-characters in Unicode: https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Non-characters

What is more, do these pose any practical problem for URI handling? We could make sure we do not allow them inside routes though.

tests/test_request_attrs.py
```diff
@@ -143,6 +143,23 @@ def test_nonlatin_path(self, test_path):
...
         assert req.path == falcon.uri.decode(test_path)
 
+    @pytest.mark.skipif(not six.PY2, reason='Test only applies to Python 2')
+    @pytest.mark.parametrize('test_path', [
+        u'/hello_\u043f\u0440\u0438\u0432\u0435\u0442',
+        u'/world_\u043c\u0438\u0440',
+        u'/test_\u0442\u0435\u0441\u0442',
+    ])
+    def test_utf_8_path(self, test_path):
+        # NOTE(kandziu): Test that under python 2.6/7 a path with
```
vytas7: You could change this to mention ``2.7`` only since we dropped support for ``2.6``.

vytas7:
It is none of your fault, but I get a feeling this issue is not as trivial as it looks at a first glance.

Consider a simple end point to serve burgers:

```python
import falcon


class Hamburger(object):

    def on_get(self, req, resp):
        resp.body = 'Have a burger :)'
        resp.content_type = falcon.MEDIA_TEXT


api = falcon.API()
api.add_route(u'/\U0001f354', Hamburger())
```

This does not work on Python 2, complaining about non-ASCII routes. Which could be perceived as somewhat inconsistent with *Request.path* always being ``unicode``.

It does work on Python 3 though:

```
http http://localhost:8000/🍔
HTTP/1.1 200 OK
Connection: close
Date: Sat, 22 Dec 2018 19:01:54 GMT
Server: gunicorn/19.9.0
content-length: 16
content-type: text/plain; charset=utf-8

Have a burger :)
```

I will raise this on Gitter as well https://gitter.im/falconry/dev

falcon/util/uri.py
kandziu: By default unicode encoder replaces invalid characters on `U+FFFD` [refs](https://docs.python.org/2/howto/unicode.html#the-unicode-type).  How we should process a such `/static/.\ufffdssh/authorized_keys` uri path?

kandziu:
@vytas7 non-ASCII routes issue is not related to this PR but could be easy fixed if we include `unicode_literals` in the `routing/compiled.py`. All test will be success and my result
```
$ python --version
Python 2.7.15
$ curl -v -XGET "http://127.0.0.1:8000/🍔"
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> GET /🍔 HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: gunicorn/19.9.0
< Date: Sun, 23 Dec 2018 14:33:41 GMT
< Connection: close
< content-length: 16
< content-type: text/plain; charset=utf-8
<
* Closing connection 0
Have a burger :)
```
Should I fix it in this PR?

falcon/util/uri.py
vytas7: I am no expert on this tbh. But have you seen [Corrigendum #9: Clarification About Noncharacters](https://www.unicode.org/versions/corrigendum9.html) though? They are not invalid _per se_, just private.

I have to apologize though I haven't notice in the beginning these you apply it only to static [file] routes. It makes sense in this case I guess.

If the path is now strictly guaranteed to be ``unicode`` though, could we just compile this ``U+FFFD`` into ``StaticRoute._DISALLOWED_CHARS_PATTERN``?

vytas7:
I would like to have more opinions on this. I have asked other maintainers, and on Falcon developer community chat (you are welcome to hang there too :slightly_smiling_face: )

I can be argued that it is related, and it can be argued that it is not. Another thing to consider is that we are currently moving to release 2.0 which would be the last major release supporting Python 2, and we are considering to drop Python 2 support for Falcon 3.0 and beyond: https://github.com/falconry/falcon/issues/1395 . In that light, one probably wouldn't want to invest too much into Python 2 branches.

tests/test_request_attrs.py
vytas7: Could we expand these tests for your fix?

I haven't checked in detail myself, but could one craft a contrived test case where Unicode character expansion in UTF-8 includes the byte ``/`` as part of multiple-byte representation, which cause routing to a wrong end point or something like that?

falcon/util/uri.py
kandziu: Yes we can compile it into `StaticRoute._DISALLOWED_CHARS_PATTERN`

kgriffs:
It does seem like it would be better to be consistent if we are going to go ahead and fix this issue. I'm curious to see how this may affect performance. I'll run a few quick tests to see.

vytas7:
Having invested some time in the review & discussions as well, I feel uneasy about this too.
But I would vote for **not pursuing** this any further for Python 2:

* Great potential for unknown regressions since we have not really explored/tested all the consequences in depth
* More work ahead, like a couple of additional places in ``compiled.py`` as mentioned by @kgriffs 
* The upcoming Falcon 2.0 is the last planned major Falcon release to support Python 2.7, we will not enjoy the fruits of this labour for too long

Furthermore, it seems the great work by @kandziu is not completely wasted. The handling of invalid characters wrt static routes should apply to Python 3, too.

kgriffs:
> The handling of invalid characters wrt static routes should apply to Python 3, too.

Indeed, that was a good catch.

kgriffs:
[superseded by #1415]

Issue #1259, opened 2018-05-11
docs: Add a warning to the Request.path docstring re url encoding


kgriffs:
If the original path requested by the client was url-encoded, it will be decoded before being returned by `Request`'s `path` attribute. Therefore, if this attribute is to be used by the app for any upstream requests (say, when implementing a proxy), any non-url-safe characters in the path must be url-encoded back, before making the request. Not all HTTP client libraries handle this automatically.

danilito19:
Hello! Was looking to contribute to this issue, but I believe it's already been addressed [here](https://github.com/falconry/falcon/blob/master/falcon/request.py#L178). Should this issue be closed?

vytas7:
Indeed, this has already been tackled by @brooksryba in https://github.com/falconry/falcon/pull/1374 !

PR #1374, opened 2018-10-28
Add a warning to the Request.path docstring re url encoding #1259

brooksryba:


falcon/request.py
```diff
@@ -182,7 +182,12 @@ class Request(object):
 
             Note:
                 `req.path` may be set to a new value by a `process_request()`
-                middleware method in order to influence routing.
+                middleware method in order to influence routing. If the
+                original path requested was url-encoded, it will be decoded
+                before being returned this attribute. If this attribute is to
```
vytas7: > If the original path requested was url-encoded, it will be decoded before being returned :question: this attribute.

It feels like some preposition is missing here in-between, like *it will be decoded before being returned **as/in/by** this attribute*.

vytas7:
Hi @brooksryba !
Apart from that tiny nitpick I had, your documentation change is a very welcome addition to the project, would you mind if we reopened this pull request?

Alternatively we can go and fix this ourselves in a new branch if you are OK with it. You would be credited for the work in the joint commit using ``Co-authored-by``.

brooksryba:
@vytas7 Please do re-open! I accidentally closed this while clearing out my PR listing. I don't have much time to make the requested changes for a few weeks

falcon/request.py
kgriffs: I went ahead and addressed this.

PR #1378, opened 2018-11-02
refactor(msgpack): deprecated encoding options

### LINKED ISSUE #1337: Use raw=False instead of encoding='utf-8' for MessagePackHandler
See also "Deprecating encoding option" and "string and binary type" in the msgpack README:

https://pypi.org/project/msgpack/

sbensoussan:
Change MessagePack encoding options to address two pending deprecation warnings:
- `msgpack.Packer`: encoding is deprecated, use `raw=False` instead.
- `msgpack.unpackb`: encoding is deprecated, `use_bin_type=True`.
See https://github.com/msgpack/msgpack-python#deprecating-encoding-option.

Fixes #1337


kgriffs:
Now that #1344 has been merged, I think we can land this one for the tutorial updates.

PR #1388, opened 2018-12-10
doc(RequestOptions): Update docstring for auto_parse_qs_csv

kgriffs:


vytas7:
Should we update ``CHANGES.rst`` while we are at it?

kgriffs:
I suppose that could be considered a related patch, since it is part of the docs.

Issue #1215, opened 2018-01-24
testing.simulate_request allows not to modify the env


ghost:
The function "falcon.testing.client.simulate_request" of the testing module currently accepts multiple optional arguments that are written into the env, which is ultimately used in the simulated request. Other env options are fixed.

It would be great if this function would accept an environment created by testing.create_environ. 

This way one could simulate e.g. ...

- ... requests from different hosts 
- ... requests with SSO

vytas7:
That is a very valid issue, +1 from me as well.

An possible ugly workaround, at least I have gone that way, would be to monkey-patch environment creation:

```
    def override_environ(update):
        def wrapper(*args, **kwargs):
            env = original_create_environ(*args, **kwargs)
            env.update(update)
            return env

        # Save the original function since it will be temporarily patched
        original_create_environ = falcon.testing.helpers.create_environ
        return wrapper
```

Then ``mock.patch`` ``falcon.testing.helpers.create_environ`` with this one.

kgriffs:
This sounds like a reasonable feature to add. I will schedule it for 2.x. If anyone is interested in implementing this for 2.0 or 2.1, feel free to submit a PR and we'll take a look. Thanks!

vytas7:
This is done now, leaving the issue open as a checklist item to also mention this one in ``CHANGES.rst``.

kgriffs:
Usually I just review the commit log to make sure everything is enumerated in CHANGES.rst in the process of finalizing a new release, so it should be safe to close this.

vytas7:
Aha, thanks, closing then!

PR #1397, opened 2018-12-22
feat(testing): allow overriding host, remote_addr, and arbitrary environ variables

vytas7:
Attempt to address https://github.com/falconry/falcon/issues/1215

### DIFF
--- a/falcon/testing/client.py
+++ b/falcon/testing/client.py
@@ -232,7 +232,8 @@ class Cookie(object):
 def simulate_request(app, method='GET', path='/', query_string=None,
                      headers=None, body=None, json=None, file_wrapper=None,
                      wsgierrors=None, params=None, params_csv=True,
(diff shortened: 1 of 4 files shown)


falcon/testing/client.py
```diff
@@ -275,8 +276,14 @@ def simulate_request(app, method='GET', path='/', query_string=None,
                 environ (default: ``None``). This can be used to test
                 high-performance file transmission when `resp.stream` is
                 set to a file-like object.
+            host(str): A string to use for the hostname part of the fully
```
vytas7: I these additions look OK, I will update copy-paste them to specialized ``simulate_*`` methods.

kgriffs: LGTM 👍

vytas7: Done with copy-paste now.

PR #1399, opened 2018-12-27
fix(cython): properly test cythonized Falcon & make print_routes work with cythonized responders

### LINKED ISSUE #1400: Tox (including Travis) not testing cythonized variants
As described in ``README.rst``, Falcon can be cythonized for ~20% performance gain (or actually even more). Installing Falcon from *sdist* into an environment with Cython does the trick:

```python
>>> import falcon
>>> falcon.api  # As we can see, falcon.api is coming from the dynamically-linked (cythonized) library api.so
<module 'falcon.api' from '/home/vytas/.virtualenvs/fresh/local/lib/python2.7/site-packages/falcon/api.so'>
```

However, this does not hold under Tox ``py27_cython`` and ``py36_cython`` environments, including runs in Travis, as the properly cythonized Falcon is shadowed by the local source directory. This could potentially be worked around by changing dir in Tox, but apparently pytest is even more stubborn as it is correctly determining the root dir of tests, and changing to it.

See also discussions here:
* https://github.com/tox-dev/tox/issues/54
* https://github.com/tox-dev/tox/issues/514

The last comment on the latter also explains the possible patterns to work this around: https://github.com/tox-dev/tox/issues/514#issuecomment-327779367 (links to the useful https://docs.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules ).

vytas7:
Closes #1400 

(see the issue for problem background and discussion).

Note that the solution is somewhat counter-intuitive and strange: removing `test/__init__.py`. This prevents `pytest` from treating `tests` as a module which in turns makes `pytest` change directory to its parent, at the same time shadowing cythonized Falcon with the local source.

Could this removal be a problem?

### DIFF
--- a/falcon/cmd/print_routes.py
+++ b/falcon/cmd/print_routes.py
@@ -58,13 +58,19 @@ def traverse(roots, parent='', verbose=False):
                         else:
                             real_func = func
 
-                        source_file = inspect.getsourcefile(real_func)
+                        try:
+                            source_file = inspect.getsourcefile(real_func)
+                            source_lines = inspect.getsourcelines(real_func)
+                            source_info = '{}:{}'.format(source_file,
+                                                         source_lines[1])
+                        except TypeError:
+                            # NOTE(vytas): If Falcon is cythonized, all default
+                            # responders coming from cythonized modules will
+                            # appear as built-in functions, and raise a
+                            # TypeError when trying to locate the source file.
+                            source_info = '[unknown file]'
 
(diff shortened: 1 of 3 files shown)


kgriffs:
Unfortunately I don't recall why the `__init__.py` was originally added, but that was certainly before we switched to pytest and anyway, I can't think of any problems removing it might cause.

kgriffs:
(nice sleuthing, btw!) 💯

PR #1404, opened 2019-01-02
feat(media) Initial Implementation of custom json handling

nZac:
This PR represents the proof-of-concept for custom JSON handling (#1206) and additional kwargs for `loads` and `dumps` (#1192). 

TODO: 
- [x] Proof-of-concept
- [x] Tests
- [x] Complete documention
- [x] Changelog / Contributor updates

### DIFF
--- a/falcon/media/json.py
+++ b/falcon/media/json.py
@@ -10,16 +10,20 @@ from falcon.util import json
 class JSONHandler(BaseHandler):
     """JSON media handler.
 
-    This handler uses Python's :py:mod:`json` by default, but will
-    use :py:mod:`ujson` if available.
+    This handler uses Python's standard :py:mod:`json` library by default, but
+    can be easily configured to use any of a number of third-party JSON
+    libraries, depending on your needs. For example, you can often
+    realize a significant performance boost under CPython by using an
+    alternative library. Good options in this respect include `orjson`,
+    `python-rapidjson`, and `mujson`.
 
-    Keyword Arguments:
-        dumps (func): the function to use when serializing JSON responses.
-        loads (func): the function to use when deserializing JSON requests.
+    Note:
+        If you are deploying to PyPy, we recommend sticking with the standard
+        library's JSON implementation, since it will be faster in most cases
+        as compared to a third-party library.
 
-    You can override the JSON library used by changing the ``dumps`` and
-    ``loads`` functions. Some good options are orjson, python-rapidjson,
-    and mujson.::
+    Overriding the default JSON implementation is simply a matter of specifying
+    the desired ``dumps`` and ``loads`` functions::
 
         import falcon
         from falcon import media
@@ -42,11 +46,9 @@ class JSONHandler(BaseHandler):
     If you override the ``dumps`` function, you will need to explicitly set
     ``ensure_ascii`` to ``False`` in order to enable the serialization of
     Unicode characters to UTF-8. This is easily done by using
-    ``functools.partial`` to curry the keyword arguments. This gives you
-    complete control over the behavior of the ``dumps`` and ``loads``
-    functions.
-
-    ::
+    ``functools.partial`` to apply the desired keyword argument. In fact, you
+    can use this same technique to customize any option supported by the
+    ``dumps`` and ``loads`` functions::
 
         from functools import partial
         import ujson
@@ -57,6 +59,10 @@ class JSONHandler(BaseHandler):
                 ensure_ascii=False, escape_forward_slashes=True
             ),
         )
+
+    Keyword Arguments:
+        dumps (func): Function to use when serializing JSON responses.
+        loads (func): Function to use when deserializing JSON requests.
     """
 
     def __init__(self, dumps=None, loads=None):
--- a/falcon/media/json.py
+++ b/falcon/media/json.py
@@ -51,12 +51,14 @@ class JSONHandler(BaseHandler):
     ``dumps`` and ``loads`` functions::
 
         from functools import partial
-        import ujson
+
+        from falcon import media
+        import rapidjson
 
         json_handler = media.JSONHandler(
             dumps=partial(
-                ujson.dumps,
-                ensure_ascii=False, escape_forward_slashes=True
+                rapidjson.dumps,
+                ensure_ascii=False, sort_keys=True
             ),
         )
 
--- a/falcon/media/json.py
+++ b/falcon/media/json.py
@@ -17,14 +17,18 @@ class JSONHandler(BaseHandler):
 
         loads (func): the function to use when deserializing JSON requests
 
-        dumps_kwargs (dict): keyword arguments passed to ``dumps`` when serializing to JSON.
-            ``ensure_ascii`` is set to ``false`` by default. To override this setting, pass ``True``
-            explicitly.
+        dumps_kwargs (dict): keyword arguments passed to ``dumps`` when
+            serializing to JSON.  ``ensure_ascii`` is set to ``false`` by
+            default.  To override this setting, pass ``{ensure_ascii=True}``
+            explicitly. You can completely disable extra parameters by passing
+            ``False``.
 
-        loads_kwargs (func): keyword arguments to pass to ``loads`` when deserializing from JSON.
+        loads_kwargs (func): keyword arguments to pass to ``loads`` when
+            deserializing from JSON.  You can completely disable extra
+            parameters by passing ``False``.
 
-
-    You can override the JSON library used by changing the ``dumps`` and ``loads`` functions::
+    You can override the JSON library used by changing the ``dumps`` and
+    ``loads`` functions::
 
         import falcon
         from falcon import media
@@ -35,6 +39,7 @@ class JSONHandler(BaseHandler):
             dumps=rapidjson.dumps,
             loads=rapidjson.loads,
             dumps_kwargs=dict(ensure_ascii=False)
+            loads_kwargs=False,
         )
         extra_handlers = {
             'application/json': json_handler,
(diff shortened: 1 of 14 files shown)


falcon/media/json.py
```diff
@@ -12,19 +12,61 @@ class JSONHandler(BaseHandler):
...
+
+
+        json_handler = media.JSONHandler(
+            dumps=rapidjson.dumps,
+            loads=rapidjson.loads,
+            dumps_kwargs=dict(ensure_ascii=False)
+        )
+        extra_handlers = {
+            'application/json': json_handler,
+            'application/json; charset=UTF-8': json_handler,
```
vytas7: Probably outside of scope of this issue, but do we need to single out this ``application/json; charset=UTF-8``? 
And I think I've seen an issue or a code comment somewhere to phase this explicit charset out, wouldn't ``application/json`` capture the latter as well?

falcon/media/json.py
```diff
@@ -12,19 +12,61 @@ class JSONHandler(BaseHandler):
...
+        self.dumps_kwargs.update(dumps_kwargs or dict())
+
+        self.loads = loads or json.loads
+        self.loads_kwargs = dict()
+        self.loads_kwargs.update(loads_kwargs or dict())
+
     def deserialize(self, stream, content_type, content_length):
         try:
-            return json.loads(stream.read().decode('utf-8'))
+            return self.loads(stream.read().decode('utf-8'), **self.loads_kwargs)
```
vytas7: Just pondering about this ``decode('utf-8')``... Do all 3rd party JSON libraries require a *unicode* here? I could think some hypothetical ultrafast decoder may shave off some nanosecond here by operating directly off *bytes*, but I haven't done any investigation.

vytas7:
I think the overall structure looks simple and clean :+1: 

Assorted open questions for consideration:

* Would we vendor any implementation (beyond standard lib ``json``) of our own?
* Which environments would we set up in Tox/Travis?
* Would we document any recommendations to squeeze the best performance out of CPython? PyPy?

falcon/media/json.py
```diff
@@ -12,19 +12,61 @@ class JSONHandler(BaseHandler):
...
+
+        api = falcon.API()
+        api.req_options.media_handlers.update(extra_handlers)
+        api.resp_options.media_handlers.update(extra_handlers)
     """
 
+    def __init__(self, dumps=None, dumps_kwargs=None, loads=None, loads_kwargs=None):
+        self.dumps = dumps or json.dumps
+        self.dumps_kwargs = dict(ensure_ascii=False)
+        self.dumps_kwargs.update(dumps_kwargs or dict())
```
ijl: Could this set the self.dumps/loads_kwargs arguments to None if not explicitly set, and call self.dumps/loads without **kwargs if kwargs is None? This would support orjson, which doesn't take kwargs.

falcon/media/json.py
ijl: json, ujson, and rapidjson don't require unicode, but edge cases on how they decode bytes to unicode are inconsistent, so doing it here is probably safer.

vytas7: Aha, fair enough.

@ijl Thanks for your insights, btw! Are you the author of ``orjson``?

falcon/media/json.py
vytas7: In a pinch, one could wrap ``orjson.dumps`` in a kwargs-discarding closure, and pass that as the *dumps* parameter, but I agree it would be cleaner to have a way to set empty *kwargs*.

falcon/media/json.py
ijl: Yes.

falcon/media/json.py
```diff
@@ -12,19 +12,61 @@ class JSONHandler(BaseHandler):
...
+            return self.loads(stream.read().decode('utf-8'), **self.loads_kwargs)
         except ValueError as err:
             raise errors.HTTPBadRequest(
                 'Invalid JSON',
                 'Could not parse JSON body - {0}'.format(err)
             )
 
     def serialize(self, media, content_type):
-        result = json.dumps(media, ensure_ascii=False)
+        result = self.dumps(media, **self.dumps_kwargs)
```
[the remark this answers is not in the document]
nZac: What would you suggest we do with a `UnicodeEncodeError`? That seems like a valid exception to raise at this spot in the request cycle.

nZac:
@vytas7,

- I think we can test against whatever additional libraries people would like us to if they want to submit a PR for it. I think stdlib (and maybe `ujson` since we supported it previously) makes sense for testing. My vote is just stdlib.
- Whatever we decide to include in the test suite, I can incorporate into CI.
- This seems like something that changes regularly and I wouldn't want to offer outdated advice. Regardless, mentioning that PyPy is really fast by itself might be good and that there are a bunch of options if you are in CPython and just link out to them? Otherwise, bench marking seems to be the next best option.

falcon/media/json.py
```diff
@@ -12,19 +12,88 @@ class JSONHandler(BaseHandler):
...
 
+    dumps_kwargs = False
+    loads_kwargs = False
+
+    def __init__(self, dumps=None, dumps_kwargs=None, loads=None, loads_kwargs=None):
+        self.dumps = dumps or json.dumps
+
+        # NOTE(nZac): Turn off kwargs support by manually passing `False` to
+        # either `_kwargs` argument.
+        if dumps_kwargs is not False:
```
vytas7: This works, but to be honest I would prefer another solution than giving ``False`` a special meaning this way, provided a cleaner solution exists, of course.

As an idea, how about having empty *kwargs* as defaults, and instead configuring the desired ones inside ``class Handlers(UserDict)`` constructor (``falcon/media/handlers.py``)?

tests/test_media_handlers.py
```diff
@@ -1,3 +1,6 @@
+import unittest.mock as mock
```
vytas7: Note: ``mock`` does not exist in the Python 2 world this way, needs to be pip-installed.

falcon/media/json.py
```diff
@@ -12,19 +12,88 @@ class JSONHandler(BaseHandler):
...
+        loads_kwargs (func): keyword arguments to pass to ``loads`` when
+            deserializing from JSON.  You can completely disable extra
+            parameters by passing ``False``.
+
+    You can override the JSON library used by changing the ``dumps`` and
+    ``loads`` functions::
+
+        import falcon
+        from falcon import media
+        import rapidjson
```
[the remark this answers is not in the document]
vytas7: IMHO we could consider testing with ``orson`` as well. One, its author @ijl came here to share insights on the PR, and two, ``orson`` seems to be one the fastest kids on block indeed. Not to mention some Falcon maintainers have closet sympathy for Rust :smiling_imp:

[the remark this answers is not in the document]
kgriffs: 👍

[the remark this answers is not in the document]
kgriffs: For reference: https://github.com/ijl/orjson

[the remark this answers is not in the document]
kgriffs: Actually, after thinking about this some more and looking at mujson's implementation more closely, I think we may want to just list it as something people might want to look at (and add it to our tests), and just stick with python-rapidjson in this example. mujson's approach may not be as straightforward and predictable as some people may like. orjson is super fast, but unfortunately not fully compliant as a drop-in replacement to stdlib json.

So, to summarize, we demonstrate python-rapidjson in this example, and then list these in the docs (as well as include them in a few smoke tests):

* orjson (fastest, but doesn't support some of the features provided by stdlib)
* python-rapidjson
* mujson (does not currently include orjson in the mix)

Plus we will add ujson to our smoke tests since we previously supported it, although it is not well-maintained these days:

* ujson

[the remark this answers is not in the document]
nZac: Done! Thanks for the feedback

falcon/media/json.py
```diff
@@ -2,30 +2,79 @@
...
+        }
+
+        api = falcon.API()
+        api.req_options.media_handlers.update(extra_handlers)
+        api.resp_options.media_handlers.update(extra_handlers)
+
+
+    By default, ``ensure_ascii`` is passed to the ``json.dumps`` function.
+    If you override the dumps function you might want to include that as a
+    default parameter. A simple way is by using ``functools.partial`` to curry
```
vytas7: I think I might have misused the term *to curry* usage in my earlier comment, sorry about that: https://en.wikipedia.org/wiki/Currying#Contrast_with_partial_function_application

falcon/media/json.py
```diff
@@ -11,21 +12,70 @@ class JSONHandler(BaseHandler):
...
+
+        api = falcon.API()
+        api.req_options.media_handlers.update(extra_handlers)
+        api.resp_options.media_handlers.update(extra_handlers)
+
+    By default, ``ensure_ascii`` is passed to the ``json.dumps`` function.
+    If you override the ``dumps`` function, you will need to explicitly set
+    ``ensure_ascii`` to ``False`` in order to enable the serialization of
+    Unicode characters to UTF-8. This is easily done by using
+    ``functools.partial`` to curry the keyword arguments. This gives you
```
vytas7: As said in a comment before, I misled you with this *curry*. Apparently, it has a slightly different meaning than partial application.

tests/test_media_handlers.py
```diff
@@ -1,13 +1,80 @@
...
+     b'{"key": "value"}',
+     {'key': 'VALUE'}),
+
+    (mujson.loads, b'{"test": "value"}', {'test': 'value'}),
+    (ujson.loads, b'{"test": "value"}', {'test': 'value'}),
+]
+
+YEN = b'\xc2\xa5'
+
+if compat.PY3:
```
vytas7: If I were to write this from scratch, I would probably define ``skip`` based on the Python version, but it is mostly a matter of taste, not a requirement from me.

falcon/media/json.py
```diff
@@ -11,21 +12,70 @@ class JSONHandler(BaseHandler):
 
     This handler uses Python's :py:mod:`json` by default, but will
```
vytas7: This comment about ``ujson`` no longer holds as I understand?

kgriffs: [fixed]

tests/test_media_handlers.py
kgriffs: [fixed]

kgriffs:
@nZac  @vytas7 I took the liberty of making a few additional tweaks to help get this over the finish line.

vytas7:
Thanks for the combined effort @kgriffs @nZac ! The current version passes my review.

vytas7:
There was unfortunately one lap left until the finish line, so I had to join the relay team :grimacing: 
Hopefully sprinting over the line for real now!

Issue #1120, opened 2017-10-02
Consider changing context type to a bare class


kgriffs:
Consider changing from using a dict to a bare class and setting attributes on it, also removing the ability to use a custom class (since allowing for attribute-based access was the primary motivation for allowing the use of a custom context class in the first place).

I performed a few benchmarks under Python 2.7.13 and python 3.6.2 and discovered that using a bare class is actually faster in most cases. It was slightly slower (on the order of nanoseconds) for setting a value.

kgriffs:
On second thought, we may retain the ability to use a custom class so that app users can ease migration by setting it back to a dict.

vytas7:
@steven-upside have you had a chance to look into this issue?

I am thinking to hack something myself if this is still desired for 2.0.

And I would strongly advocate retaining the custom class ability since I know that some users (read: me :slightly_smiling_face: ) have already adopted this functionality. It would be a no-op them.

vytas7:
Closing as the referenced PR has been merged.

PR #1408, opened 2019-01-06
feat(request, response): change context type to a subclass of dict

vytas7:
ref https://github.com/falconry/falcon/issues/1120

Work in progress:

- [x] Make ``Request`` / ``Response`` use ``context_type`` subclassing dict
- [x] Update unit tests
- [x] Update FAQ
- [x] Update benchmark code
- [x] Update examples in ``README.rst`` and elsewhere
- [x] ~~Resolve Hug tests (Hug is relying on dict context)~~ :arrow_left: not any longer an issue with dict subclass approach

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,17 @@ Breaking Changes
   ``False``.
 - ``RequestOptions.auto_parse_qs_csv`` now defaults to ``False`` instead of
   ``True``.
+- ``Request.context_type`` was changed from dict to a subclass of dict.
+- ``Response.context_type`` was changed from dict to a subclass of dict.
+
+Changes to Supported Platforms
+------------------------------
+
+New & Improved
+--------------
+
+- Added a new ``headers`` property to the ``Response`` class.
+- Removed ``six`` as a dependency.
 - ``Request.context_type`` now defaults to a bare class allowing to set
   attributes on the request context object::
 
@@ -46,15 +57,6 @@ Breaking Changes
   deprecated, and may be removed in a future release. It is also noteworthy
   that object attributes and dict items are not automagically linked in any
   special way, and setting one does not affect the other.
-
-Changes to Supported Platforms
-------------------------------
-
-New & Improved
---------------
-
-- Added a new ``headers`` property to the ``Response`` class.
-- Removed ``six`` as a dependency.
 
 Fixed
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,10 @@ Breaking Changes
   ``False``.
 - ``RequestOptions.auto_parse_qs_csv`` now defaults to ``False`` instead of
   ``True``.
+- ``Request.context_type`` now defaults to a subclass of dict allowing to
+  set attributes on the request context object.
+- ``Response.context_type`` now defaults to a subclass of dict allowing to
+  set attributes on the response context object.
 
 Changes to Supported Platforms
 ------------------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,10 +16,36 @@ Breaking Changes
   ``False``.
 - ``RequestOptions.auto_parse_qs_csv`` now defaults to ``False`` instead of
   ``True``.
-- ``Request.context_type`` now defaults to a subclass of dict allowing to
-  set attributes on the request context object.
-- ``Response.context_type`` now defaults to a subclass of dict allowing to
-  set attributes on the response context object.
+- ``Request.context_type`` now defaults to a bare class allowing to set
+  attributes on the request context object::
+
+    # Before
+    req.context['role'] = 'trial'
+    req.context['user'] = 'guest'
+
+    # Falcon 2.0
+    req.context.role = 'trial'
+    req.context.user = 'guest'
+
+  To ease the migration path, the previous behavior is supported by subclassing
+  dict, however, as of Falcon 2.0, the dict context interface is considered
+  deprecated, and may be removed in a future release. It is also noteworthy
+  that object attributes and dict items are not automagically linked in any
+  special way, and setting one does not affect the other.
+- ``Response.context_type`` now defaults to a bare class allowing
+  to set attributes on the response context object::
+
+    # Before
+    resp.context['cache_strategy'] = 'lru'
+
+    # Falcon 2.0
+    resp.context.cache_strategy = 'lru'
+
+  To ease the migration path, the previous behavior is supported by subclassing
+  dict, however, as of Falcon 2.0, the dict context interface is considered
+  deprecated, and may be removed in a future release. It is also noteworthy
+  that object attributes and dict items are not automagically linked in any
+  special way, and setting one does not affect the other.
 
 Changes to Supported Platforms
 ------------------------------
--- a/README.rst
+++ b/README.rst
@@ -532,7 +532,7 @@ bodies.
                                             'A valid JSON document is required.')
 
             try:
-                req.context['doc'] = json.loads(body.decode('utf-8'))
+                req.context.doc = json.loads(body.decode('utf-8'))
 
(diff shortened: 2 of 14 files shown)


docs/user/faq.rst
```diff
@@ -584,6 +585,43 @@ In the meantime, the workaround is to percent-encode the forward slash. If you
...
+
+   # Before Falcon 2.0
+   req.context['cache_backend'] = MyUltraFastCache.connect()
+
+   # Falcon 2.0
+   req.context.cache_backend = MyUltraFastCache.connect()
+
+Since the default context type derives from dict, the existing code will work
+unmodified with Falcon 2.0. Nevertheless, it is recommended to change the usage
+as recommended above since the ability to use context as dictionary may be
```
vytas7: _recommended_ to change... as _recommended_ above -- I will reword

falcon/request.py
```diff
@@ -75,15 +75,17 @@ class Request(object):
     Attributes:
         env (dict): Reference to the WSGI environ ``dict`` passed in from the
             server. (See also PEP-3333.)
-        context (dict): Dictionary to hold any data about the request which is
-            specific to your app (e.g. session object). Falcon itself will
-            not interact with this attribute after it has been initialized.
+        context (object): Empty object to hold any data (in its attributes)
```
kgriffs: I think it would be helpful to inline an example here to illustrate using attributes in lieu of the (dict) mapping interface.

falcon/response.py
```diff
@@ -108,15 +108,17 @@ class Response(object):
 
         stream_len (int): Deprecated alias for :py:attr:`content_length`.
 
-        context (dict): Dictionary to hold any data about the response which is
-            specific to your app. Falcon itself will not interact with this
-            attribute after it has been initialized.
+        context (object): Empty object to hold any data (in its attributes)
```
kgriffs: I think it would be helpful to inline an example here to illustrate using attributes in lieu of the (dict) mapping interface.

CHANGES.rst
```diff
@@ -16,6 +16,10 @@ Breaking Changes
   ``False``.
 - ``RequestOptions.auto_parse_qs_csv`` now defaults to ``False`` instead of
   ``True``.
+- ``Request.context_type`` now defaults to a subclass of dict allowing to
+  set attributes on the request context object.
```
kgriffs: We should include an additional note about the fact that the mapping and attribute protocols are not magically linked, and so you can not mix-and-match (i.e., we are not trying to be like JavaScript here).

docs/user/faq.rst
```diff
@@ -416,8 +416,9 @@ See also the `WSGI middleware example <https://www.python.org/dev/peps/pep-3333/
 How can I pass data from a hook to a responder, and between hooks?
 ------------------------------------------------------------------
 You can inject extra responder kwargs from a hook by adding them
-to the *params* dict passed into the hook. You can also add custom data to
-the ``req.context`` dict, as a way of passing contextual information around.
+to the *params* dict passed into the hook. You can also add custom attributes
+to the ``req.context`` object, as a way of passing contextual information
+around.
```
kgriffs: Let's include an example here as well.

docs/user/faq.rst
```diff
@@ -584,6 +585,43 @@ In the meantime, the workaround is to percent-encode the forward slash. If you
...
+   # Before Falcon 2.0
+   req.context['cache_backend'] = MyUltraFastCache.connect()
+
+   # Falcon 2.0
+   req.context.cache_backend = MyUltraFastCache.connect()
+
+Since the default context type derives from dict, the existing code will work
+unmodified with Falcon 2.0. Nevertheless, it is recommended to change the usage
+as recommended above since the ability to use context as dictionary may be
+removed in the future major releases.
```
kgriffs: Let's include a warning here as well about mixing-and-matching mapping vs. attributes.

docs/user/faq.rst
```diff
@@ -584,6 +585,43 @@ In the meantime, the workaround is to percent-encode the forward slash. If you
 don’t control the clients and can't enforce this, you can implement a Falcon
 middleware component to rewrite the path before it is routed.
 
+How do I adapt my code to default context type changes in Falcon 2.0?
+---------------------------------------------------------------------
+
+The default request/response context type has been changed from dictionary to a
```
kgriffs: dictionary ➡️ ``dict``

docs/user/faq.rst
```diff
@@ -584,6 +585,43 @@ In the meantime, the workaround is to percent-encode the forward slash. If you
...
+class subclassing dict in Falcon 2.0. Instead of setting dictionary items, one
```
kgriffs: Suggested edit:

"one shall set attributes on the object now:" ➡️ "you can now simply set attributes on the object:"

docs/user/faq.rst
kgriffs: "in the future major releases" ➡️ "in a future release"

CHANGES.rst
kgriffs: Also, we should note that while the context type currently subclasses ``dict`` in order to ease the migration path, the type may be changed to a bare object in a future release (i.e., use of the mapping interface is deprecated as of Falcon 2.0).

README.rst
```diff
@@ -542,10 +542,10 @@ bodies.
                                        'UTF-8.')
 
         def process_response(self, req, resp, resource):
-            if 'result' not in resp.context:
```
kgriffs: 🤔 It may make sense to do a followup PR to change the attribute name to ``doc`` to improve the symmetry between the request and response examples.

vytas7: Right, I will revise these in a followup PR:
* ``README.rst``, ``quickstart.rst`` and ``test_example.py`` all contain slightly different versions of the same file
* The example is not tested (bar ``flake8`` on ``test_example.py``)

falcon/request.py
vytas7: Sure, will address. Just wondering if you meant just exemplifying the "new" way, or also contrasting to the dict interface as in FAQ and changelogs?

falcon/response.py
vytas7: Same question as: https://github.com/falconry/falcon/pull/1408/files#r251250026

falcon/request.py
kgriffs: I think we could add a `Note:` section that says "New in 2.0" or something, explaining that this changed and going forward people should use the attribute-based interface. The regular example would just show the "new" way.

README.rst
vytas7: I have created a followup issue: https://github.com/falconry/falcon/issues/1423

falcon/request.py
```diff
@@ -75,15 +75,27 @@ class Request(object):
...
+            about the request which is specific to your app (e.g. session
+            object). Falcon itself will not interact with this attribute after
+            it has been initialized.
+
+            Note:
+                **New in 2.0:** the default `context_type` (see below) was
+                changed from dict to a bare class, and the preferred way to
+                pass request-specific data is setting attributes on the
```
kgriffs: Nitpick: "is setting attributes" ➡️ "is now to set attributes directly"

falcon/response.py
```diff
@@ -108,15 +108,26 @@ class Response(object):
...
+            about the response which is specific to your app (e.g. session
+            object). Falcon itself will not interact with this attribute after
+            it has been initialized.
+
+            Note:
+                **New in 2.0:** the default `context_type` (see below) was
+                changed from dict to a bare class, and the preferred way to
+                pass response-specific data is setting attributes on the
```
kgriffs: (See nitpick above.)

vytas7:
@nZac I think I found a semi-elegant way to do that, but I think that would bog down users with warnings being spammed upon every request provided they use dict context. So I think it is best to go forward with this now and give users some breathing space to adapt to the new changes, and start emitting these warnings in a later release.

PR #1409, opened 2019-01-10
refactor: Import abstract base classes from collections.abc under Python 3

aparkerlue:
Update the import path for [abstract base classes](https://docs.python.org/3/library/collections.abc.html), which changed from `collections` to `collections.abc` in Python 3.3.

This resolves warnings like this:

```
.../site-packages/falcon/util/structures.py:35: DeprecationWarning:
  Using or importing the ABCs from 'collections' instead of from
  'collections.abc' is deprecated, and in 3.8 it will stop working
    class CaseInsensitiveDict(collections.MutableMapping):  # pragma: no cover
```

### DIFF
--- a/falcon/util/structures.py
+++ b/falcon/util/structures.py
@@ -27,17 +27,17 @@ for convenience::
 
 """
 
-import collections
+from collections.abc import Mapping, MutableMapping
 
 
 # TODO(kgriffs): If we ever diverge from what is upstream in Requests,
 # then we will need write tests and remove the "no cover" pragma.
-class CaseInsensitiveDict(collections.MutableMapping):  # pragma: no cover
+class CaseInsensitiveDict(MutableMapping):  # pragma: no cover
     """
     A case-insensitive ``dict``-like object.
 
     Implements all methods and operations of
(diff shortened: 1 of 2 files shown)


falcon/util/structures.py
```diff
@@ -27,17 +27,17 @@
 
 """
 
-import collections
+from collections.abc import Mapping, MutableMapping
```
vytas7: Hi @aparkerlue !
And thanks for your effort to rectify this. However, it seems we will need a more sophisticated approach here since we are still supporting Python 2.7 and PyPy2.

On Python 2:
```python
>>> from collections.abc import Mapping, MutableMapping
ImportError: No module named abc
```

See also the CI reports.

Could we consider a some sort of conditional import based on Python version, or ``try`` ... ``except ImportError`` approach?

kgriffs: I went ahead and addressed this by refactoring the change to use the new ``compat`` module. Let me know what you think.

vytas7:
Looking good now! :+1:

kgriffs:
Thanks for catching this!

Issue #1331, opened 2018-06-30
perf: Revisit micro-optimizations


kgriffs:
Newer versions of 2.7.x and 3.x exhibit different performance characteristics vs. 2.6 and 3.4. Revisit micro-optimizations such as the following and adjust where it makes sense:

> #NOTE(kgriffs): Reference via self since that is faster than module global...

vytas7:
Since the example above originates from my HTTPStatus experiments, I will split some of the improvements off in a separate PR.

vytas7:
Resolved in the referenced PR.

PR #1416, opened 2019-01-19
perf(API): revisit request-response cycle micro optimizations

vytas7:
ref https://github.com/falconry/falcon/issues/1331

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -873,7 +873,9 @@ class Response(object):
         """
 
         headers = self._headers
-        self._set_media_type(media_type)
+        # PERF(vytas): uglier inline version of Response._set_media_type
+        if media_type is not None and 'content-type' not in self._headers:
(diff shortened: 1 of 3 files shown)


falcon/response.py
```diff
@@ -873,7 +873,9 @@ def _wsgi_headers(self, media_type=None, py2=PY2):
         """
 
         headers = self._headers
-        self._set_media_type(media_type)
+        # PERF(vytas): uglier inline version of Response._set_media_type
+        if media_type is not None and 'content-type' not in self._headers:
+            self.set_header('content-type', media_type)
```
kgriffs: Setting self._headers directly would offer an additional performance boost. I think ``_set_media_type()`` was the only place that called ``self.set_header().``

vytas7: Thanks, addressed!

PR #1417, opened 2019-01-22
refactor(Response): Remove stream_len property

kgriffs:
This patch "removes" the deprecated stream_len property, fully replacing it's use by the content-length property/header. Along the way, the way API uses content-length (in lieu of stream_len) was optimized.


kgriffs:
Note: I have temporarily disabled the hug gate. See also: https://github.com/timothycrosley/hug/pull/721

PR #1418, opened 2019-01-25
allow passing iterable of exceptions to add_error_handler

carlodri:
This PR extends the `add_error_handler` to accept a generic iterable of exceptions.

Todo:

- [x] update docs
- [x] update tests


falcon/api.py
```diff
@@ -696,7 +696,7 @@ def _handle_exception(self, req, resp, ex, params):
         """
 
         for err_type, err_handler in self._error_handlers:
-            if isinstance(ex, err_type):
+            if isinstance(ex, tuple(err_type)):
```
kgriffs: I think it would be better to just normalize to a tuple up front when the error handler is added.

vytas7: An idea how to implement that part, one way could be something like
```python
# ...
if not issubclass(exc_type, BaseException):  # not an exception type, assume iterable
    exc_type = tuple(exc_type)
```

carlodri: thanks for the suggestions guys! Sorry I started the work but then dropped it due to lack of time....I will soon get back on it!

kgriffs:
Hi @carlodri, do you think you'll be able to get this over the finish line over the next day or two? If not, we'll need to push this to the next release after 2.0.

carlodri:
@kgriffs sorry but I'm quite busy with my job this week... Maybe it's better if we push it forward. If I find the time in the next few days, I will let you know here ASAP.

carlodri:
Now that I think of it I'm actually puzzled as to _why_ this even works with exception tuples as it is, because 
```python
isinstance(MyException, (MyBaseException, AnotherException))
```
is `False`... What am I missing here?

tests/test_error_handlers.py
```diff
@@ -114,3 +114,23 @@ def test_error_order_subclass_masked(self, client):
...
+
+    def test_handler_exception_list(self, client):
+        exception_list = [Exception, CustomException]
+        client.app.add_error_handler(exception_list, capture_error)
+
+        result = client.simulate_get()
+        assert result.status_code == 723
+
+        result = client.simulate_delete()
+        assert result.status_code == 723
```
vytas7: No new line at the end of file.

vytas7:
@carlodri Could you show a code example where it happens?
Might it be that you happened to compare a *type* against these exceptions classes?
You would want to use `issubclass` in that case, such as the example in my previous comment, see also, for instance:
```python
>>> class MyException(KeyboardInterrupt):
...     """Press that Ctrl+C!"""
... 
>>> issubclass(MyException, (ValueError, BaseException))
True
>>> isinstance(MyException, (ValueError, BaseException))
False
>>> # But:
... 
>>> isinstance(MyException(), (ValueError, BaseException))
True
```

falcon/api.py
```diff
@@ -736,7 +736,7 @@ def _handle_exception(self, req, resp, ex, params):
...
```
vytas7: I am probably stating the obvious, but I think we would also need to continue supporting the current case of a single exception class, not just iterables as your prototype would imply.

carlodri: sure, I will make sure of that, thanks for the reminder!

carlodri:
@vytas7 thanks for the example above, in the rush I was indeed mixing up _instances_ and _classes_... I need to think for a moment of which is the best solution. I also like @kgriffs's idea above of normalizing to tuple upstream in `add_error_handler`.

carlodri:
@vytas7 @kgriffs this is ready for review, let me know your thoughts!

vytas7:
This is really shaping up, thanks!
(I would just like to understand the design choice behind the current EAFP implementation, but I am not opposed to it).

falcon/api.py
```diff
@@ -556,7 +558,10 @@ def handle(req, resp, ex, params):
 
         # Insert at the head of the list in case we get duplicate
         # adds (will cause the most recently added one to win).
-        self._error_handlers.insert(0, (exception, handler))
```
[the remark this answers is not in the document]
carlodri: Thanks for your comments @vytas7. Well, the idea of using EAFP stemmed from the fact that I cannot say `issubclass((ValueError, TypeError), BaseException)` (arg 1 must be a class), so in some way I have to first check if what the user has passed to the function is an iterable or not. AFAIK, a simple, pythonic way to do that is EAFP. 

I agree with you that it would be better to have some stricter checks on the nature of the `exception` argument. One way to go could be first simply normalize to a tuple with EAFP, and then check `all(issubclass(exc, BaseException) for exc in exception_tuple)`: if yes, I would add them to the list of handlers, otherwise, raise a descriptive `TypeError`. What do you think?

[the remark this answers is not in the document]
vytas7: > arg 1 must be a class
Ouch, I should have tested that before suggesting... :man_facepalming: 

But maybe we could still build upon that idea?
```python
if isinstance(exception, type) and issubclass(exception, BaseException):
    self._error_handlers.insert(0, (exception, handler))
else:
    self._error_handlers.insert(0, (tuple(exception), handler))
```

I think I would prefer this form. But I realize it is much about stylistic preferences, and not necessarily a huge difference in logic.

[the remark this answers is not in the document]
carlodri: so, the point here is that we want to allow `exception` to be a generic iterable of exceptions, or a single exception, that can be made into a tuple which can then be used by `_handle_error()` to check against the currently occurring exception. 

We cannot use `isinstance` or `issubclass` to test for "cast-ability" to a tuple, so EAFP would be the way to go for this. The we can also check that `exception` (or each element of `exception`) is indeed a subclass of `BaseException`. Give me 10min and I'll try to show you the code..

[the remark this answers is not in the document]
carlodri: the alternative to simplify things is to allow only single exceptions or tuples, and document this. I actually started this PR because in one case I found it useful to accumulate/remove some exceptions into a list and pass it directly to the `add_error_handler`. I'm available to both options though, I have no strong opinions about it.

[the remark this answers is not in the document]
kgriffs: In this case I think EAFP is justified, given that either way the only way to reliably test for an iterable is to attempt it and see if it raises an exception. If passing a single exception were the most common use case, and this logic were on a hot code path, checking the type up front would be ideal. But under the circumstances that's a moot point, so I'd be cool with either of the proposed approaches.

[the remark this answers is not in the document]
vytas7: I'm not a fan of EAFP, but I see here it allows to continue to part two in an elegant fashion (my previous comment precedes the latest commit), so I'm fine with it.

I just a have a small stylistic nitpick, and would like to add one more test for the `TypeError` case, and I'm fine to approve this :+1:

falcon/api.py
```diff
@@ -556,7 +558,15 @@ def handle(req, resp, ex, params):
...
+        try:
+            exception_tuple = tuple(exception)
+        except TypeError:
+            exception_tuple = tuple([exception, ])
+
+        if all(issubclass(exc, BaseException) for exc in exception_tuple):
```
vytas7: I like the way we are checking all members of the given iterable here in advance :+1:  Could we add a test for the mixed case, for instance, when not all members of a tuple are exception classes?

falcon/api.py
```diff
@@ -556,7 +558,15 @@ def handle(req, resp, ex, params):
 
         # Insert at the head of the list in case we get duplicate
         # adds (will cause the most recently added one to win).
-        self._error_handlers.insert(0, (exception, handler))
+        try:
+            exception_tuple = tuple(exception)
+        except TypeError:
+            exception_tuple = tuple([exception, ])
```
vytas7: Style suggestion: maybe just `(exception,)` ?

falcon/api.py
```diff
@@ -556,7 +558,15 @@ def handle(req, resp, ex, params):
...
+
+        if all(issubclass(exc, BaseException) for exc in exception_tuple):
+            self._error_handlers.insert(0, (exception_tuple, handler))
```
vytas7: I realized there is a minor performance issue lurking here: for single exception classes (not tuples), we may want to insert just `exception`, not a tuple with one element.

The penalty is very minor (~10 ns), OTOH we would hit it every time when handling an error.

vytas7:
Let's do that indeed! 
Great effort btw @carlodri !

falcon/api.py
carlodri: solved in 73b7d64

falcon/api.py
carlodri: well, a mixed case would be a pathological situation where we should raise an error, right? Or am I missing something? IIUC `exception` _must_ be a subclass of `BaseException`, nothing else would be meaningful here (at least from the little I know about falcon!). Please do correct me if I'm wrong...

carlodri:
@vytas7 I also fixed the last issue you raised concerning the unwanted overhead in the case of as single-element tuple. Let me know if it's ok!

falcon/api.py
vytas7: Yes, exactly, we should raise an error exactly as your code does.
I.e. we want to write a test or two here to test even error handling / malformed input. You could use, for instance, [`pytest.raises`](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) to assert that the correct type of error was raised.

falcon/api.py
```diff
@@ -554,9 +556,23 @@ def handle(req, resp, ex, params):
                                      'method named "handle" that is a '
                                      'member of the given exception class.')
 
-        # Insert at the head of the list in case we get duplicate
-        # adds (will cause the most recently added one to win).
-        self._error_handlers.insert(0, (exception, handler))
+        try:
+            exception_tuple = tuple(exception)
+        except TypeError:
+            exception_tuple = tuple((exception, ))
```
vytas7: Sorry for being unclear here.

My suggestion was getting rid of this outer `tuple` altogether, i.e. `(exception,)` is already a `tuple`.
Let me illustrate with an example:
```python
>>> exception = ZeroDivisionError
>>> type((exception,))
<type 'tuple'>
>>> type(tuple((exception,)))
<type 'tuple'>
>>> (exception,) == tuple((exception,))
True
>>> (exception,) is tuple((exception,))
False
```

falcon/api.py
carlodri: oh sure, got it. I will add those breifly.

falcon/api.py
carlodri: yeah sure, sorry for not seeing that....! 🤦‍♂️

carlodri:
thanks to you for making `falcon` available to the community! Happy to contribute with a small improvement.

PR #1419, opened 2019-01-25
refactor(falcon.request): Change etag headers to return a list of ETag

### LINKED ISSUE #1211: Change Request.if_none_match to return a list of ETag
The list should contain each ETag with quotes stripped. In the case of an asterisk, the result will be ``['*']``. 

TODO: How to handle the `W/` prefix? I believe it can simply be ignored, since the recipient is supposed to always use the weak comparison algorithm regardless of this prefix.

kandziu:
Closes #1211

`W/` prefix was simply ignored as proposed in the issue by @kgriffs.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -364,6 +364,11 @@ class Request(object):
             This property is determined by the value of ``HTTP_IF_MATCH``
             in the WSGI environment dict.
 
+            Note:
+                This property includes strong and weakness entity-tags and that
+                means that two entity-tags are equivalent if both are not weak
+                and their opaque-tags match character-by-character.
+
         if_none_match (list): A list containing all the etags in the
             If-None-Match header.
 
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -368,7 +368,7 @@ class Request(object):
                 This property includes strong and weak entity-tags. Per
                 `RFC 7239`_, two entity-tags are equivalent if both are not
                 weak and their opaque-tags match character-by-character.
-            
+
             (See also: RFC 7239, Section 3.1)
 
         if_none_match (list): Value of the If-None-Match header, as a parsed
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -358,22 +358,33 @@ class Request(object):
             accessed.)
         range_unit (str): Unit of the range parsed from the value of the
             Range header, or ``None`` if the header is missing
-        if_match (list): A list containing all the etags in the If-Match
-            header.
+        if_match (list): Value of the If-Match header, as a parsed list of
+            :class:`falcon.ETag` objects or ``None`` if the header is missing.
 
             This property is determined by the value of ``HTTP_IF_MATCH``
             in the WSGI environment dict.
 
             Note:
-                This property includes strong and weakness entity-tags and that
-                means that two entity-tags are equivalent if both are not weak
-                and their opaque-tags match character-by-character.
+                This property includes strong and weak entity-tags. Per
+                `RFC 7239`_, two entity-tags are equivalent if both are not
+                weak and their opaque-tags match character-by-character.
+            
+            (See also: RFC 7239, Section 3.1)
 
-        if_none_match (list): A list containing all the etags in the
-            If-None-Match header.
+        if_none_match (list): Value of the If-None-Match header, as a parsed
+            list of :class:`falcon.ETag` objects or ``None`` if the header is
+            missing.
 
             This property is determined by the value of ``HTTP_IF_NONE_MATCH``
             in the WSGI environment dict.
+
+            Note:
+                This property includes strong and weak entity-tags. Per
+                `RFC 7239`_, two entity-tags are equivalent if their
+                opaque-tags match character-by-character, regardless of
+                either or both being tagged as weak.
+
+            (See also: RFC 7239, Section 3.2)
 
         if_modified_since (datetime): Value of the If-Modified-Since header,
             or ``None`` if the header is missing.
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -358,10 +358,10 @@ class Request(object):
             accessed.)
         range_unit (str): Unit of the range parsed from the value of the
             Range header, or ``None`` if the header is missing
-        if_match (str): Value of the If-Match header, or ``None`` if the
-            header is missing.
-        if_none_match (str): Value of the If-None-Match header, or ``None``
-            if the header is missing.
+        if_match (list): A list containing all the etags in the If-Match
+            header.
+        if_none_match (list): A list containing all the etags in the
+            If-None-Match header.
         if_modified_since (datetime): Value of the If-Modified-Since header,
             or ``None`` if the header is missing.
         if_unmodified_since (datetime): Value of the If-Unmodified-Since
@@ -529,8 +529,6 @@ class Request(object):
 
     expect = helpers.header_property('HTTP_EXPECT')
 
-    if_match = helpers.header_property('HTTP_IF_MATCH')
-    if_none_match = helpers.header_property('HTTP_IF_NONE_MATCH')
     if_range = helpers.header_property('HTTP_IF_RANGE')
 
     referer = helpers.header_property('HTTP_REFERER')
@@ -619,6 +617,14 @@ class Request(object):
     @property
     def date(self):
         return self.get_header_as_datetime('Date')
+
+    @property
+    def if_match(self):
+        return helpers.parse_etags(self.env.get('HTTP_IF_MATCH'))
+
+    @property
+    def if_none_match(self):
+        return helpers.parse_etags(self.env.get('HTTP_IF_NONE_MATCH'))
 
     @property
     def if_modified_since(self):
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -360,8 +360,16 @@ class Request(object):
             Range header, or ``None`` if the header is missing
         if_match (list): A list containing all the etags in the If-Match
             header.
+
+            This property is determined by the value of ``HTTP_IF_MATCH``
+            in the WSGI environment dict.
+
         if_none_match (list): A list containing all the etags in the
             If-None-Match header.
+
+            This property is determined by the value of ``HTTP_IF_NONE_MATCH``
+            in the WSGI environment dict.
+
         if_modified_since (datetime): Value of the If-Modified-Since header,
             or ``None`` if the header is missing.
         if_unmodified_since (datetime): Value of the If-Unmodified-Since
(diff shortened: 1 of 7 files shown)


falcon/request_helpers.py
```diff
@@ -38,6 +42,29 @@ def fget(self):
     return property(fget)
 
 
+def parse_etags(etag_str):
+    """
+    Parse a string of ETags given in an If-Match or If-None-Match header as
```
vytas7: I am next to hopeless at properly using articles, but I believe we should use **the** here, if we know which header it is.
So :arrow_right: **the** ``If-Match``... etc

falcon/request_helpers.py
```diff
@@ -38,6 +42,29 @@ def fget(self):
...
+    defined by RFC 7232.
+
+    Args:
+        etag_str (str): A string of ASCII characters placed between double quotes
```
vytas7: Suggested rewording:

> A string of ASCII characters placed between double quotes and may be prefixed by ``W/`` to indicate that the weak comparison algorithm should be used.

:arrow_down: 

An ASCII header value to parse ETags from. ETag values within may be prefixed by ``W/`` to indicate that the weak comparison function should be used.

tests/test_request_attrs.py
```diff
@@ -785,6 +783,23 @@ def test_app_missing(self):
...
         assert req.app == ''
 
+    @pytest.mark.parametrize('etag,expected', [
+        ('W/"67ab43"', ['67ab43']),
+        ('w/"67ab43"', ['67ab43']),
+        ('W/"67ab43", "54ed21"', ['67ab43', '54ed21']),
+        ('  W/"67ab43" , "54ed21"', ['67ab43', '54ed21']),
+        ('W/"67ab43"  ,  "54ed21"', ['67ab43', '54ed21']),
+        ('W/"67ab43" ,"54ed21"', ['67ab43', '54ed21']),
+        ('W/"67ab43", unquoted, "54ed21"', ['67ab43', '54ed21']),
```
vytas7: Looking at bug trackers of various open source projects, or at any type of projects with public bug trackers, it is very common to find issues mentioning that they are not properly quoting ``ETag``s.

On one hand, RFCs are probably the only area where Falcon is opinionated -- that we should follow the established standards, but OTOH rejecting unquoted tags might be perceived as harsh. Could we maybe consider allowing a single unquoted ``ETag`` value in the header?

Regardless how we go forward with the unquoted case, we should IMHO document the behaviour clearly, and potentially reconsider how we handle parse errors. Right now, we would silently swallow any unrecognized tags, and thus mask the corresponding parse errors. Masking errors this way might lead to hard-to-debug scenarios.

vytas7:
Thanks for working on this, and thus making the 2.0 release closer and closer!

In addition to the question wrt parse error handling posted on the review, I am somewhat unsure about ``"W/"`` interpretation. I posted that on the issue itself: (WiP insert a link)

falcon/request_helpers.py
```diff
@@ -38,6 +42,29 @@ def fget(self):
...
+            and may be prefixed by ``W/`` to indicate that the weak comparison
+            algorithm should be used.
```
vytas7: It is good to reuse code for both `If-Match` and `If-None-Match`, however, my impression is that the comparison function handling is different according to the RFC [[1]](https://tools.ietf.org/html/rfc7232):

> `If-Match`
> <...> An origin server MUST use the **strong** comparison function when
   comparing entity-tags for If-Match <...>

> `If-None-Match`
> <...> A recipient MUST use the **weak** comparison function when comparing
   entity-tags for If-None-Match <...>

In your proposed changeset, we would present the stripped `If-Match` values to the API user as if they were all suitable for strong comparison, but it may not be the case.

In addition, see also the other comment regarding parsing, we should be crystal clear what we do with weak `ETag`s in this case, since just silently discarding them might also contribute to confusion (not as much as masking parse errors though).

1. https://tools.ietf.org/html/rfc7232

tests/test_request_attrs.py
kgriffs: It isn't terribly out of the ordinary for updated RFCs to acknowledge dubious behavior that has proliferated and so must be allowed for the time being, while encouraging a stricter adherence to the preferred syntax going forward. With that in mind, I think we should probably be a little more forgiving in this specific case. I like @vytas7 suggestion:

> Could we maybe consider allowing a single unquoted ETag value in the header?

falcon/request_helpers.py
```diff
@@ -38,6 +42,29 @@ def fget(self):
...
+    Returns:
+        A list of unquoted ETags or ``['*']`` if all ETags should be matched.
+    """
+    if not etag_str:
+        return []
+
+    if etag_str.strip() == '*':
+        return ['*']
+
+    matches = re.findall(_ETAG_PATTERN, etag_str.strip())
```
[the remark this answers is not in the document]
kgriffs: PERF: It may improve performance to short-circuit and avoid using a regex for the common case of a single etag. Something like:

1. If the string does not contain a comma, go to step 2.
2. If the string starts with `W/"` assume it also ends in `"` and do a quick string slice to produce the result. Otherwise, go to step 3.
3. If the string starts with `"`, assume it also ends in `"` and return the string slice. Otherwise go to step 4.
4. The string is unquoted. But since we are nice, we will just return the string as-is rather than returning an empty list. 

To be strictly correct, we could also verify that strings that start with quotes do indeed also end with them, but IMHO if the etag is malformed in that way, there is something seriously broken downstream and it shouldn't be Falcon's job to work around it.

kgriffs:
After some discussion in the dev channel, I think I'm favoring the approach suggested by @vytas7 to subclass `str`, e.g.:

```py
class ETag(str):
    def to_header(self):
        if self.weak:
            return 'W/"%s"' % self
        return '"%s"' % self

def make_etag():
    result = ETag('hello')
    result.weak = False
    return result
```

kgriffs:
Also, I'd love to try to get this finished up over the next few days. If that is not practical, we'll have to postpone to 3.0 so we can get 2.0 out the door.

vytas7:
I am actually starting to like my own suggestion more than when it popped in my head :slightly_smiling_face:  At least for `If-None-Match` (the original issue was only mentioning this header FWIW), IMHO that would give us a solid solution there.

However, I have some misgivings about `If-Match`, where, as I quoted above, the RFC mandates use of the **strong** comparison function. Even if we return a list of strings, objects, or tuples that hold the `weak` value, the user would still have to do the filtering, making this helper property slightly less helpful. But maybe that is not a big deal, and we could get away with clear documentation and good recipes.

kandziu:
@vytas7, my opinion that it's not the ETag parsing issue and should be resolved in a comparison. It will be closer to the rfc: `Strong comparison: two entity-tags are equivalent if both are not weak and their opaque-tags match character-by-character.` In the [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match#Directives) we also can find that the `If-Match` header can accepts tags prefixed by `W/`.

I agree with you that it will be great if we will documented it.

falcon/request.py
```diff
@@ -358,10 +358,23 @@ class Request(object):
...
-        if_none_match (str): Value of the If-None-Match header, or ``None``
-            if the header is missing.
+        if_match (list): A list containing all the etags in the If-Match
+            header.
+
+            This property is determined by the value of ``HTTP_IF_MATCH``
+            in the WSGI environment dict.
+
+            Note:
+                This property includes strong and weakness entity-tags and that
```
vytas7: Just ~~weakness~~ :arrow_right: weak IMHO

falcon/request.py
```diff
@@ -358,10 +358,23 @@ class Request(object):
             accessed.)
         range_unit (str): Unit of the range parsed from the value of the
             Range header, or ``None`` if the header is missing
-        if_match (str): Value of the If-Match header, or ``None`` if the
-            header is missing.
-        if_none_match (str): Value of the If-None-Match header, or ``None``
-            if the header is missing.
+        if_match (list): A list containing all the etags in the If-Match
+            header.
```
vytas7: Regardless of the chosen design, we need to talk more about the entities within the list, i.e. strings with the _weak_ attribute, or w/e we choose.

falcon/request.py
vytas7: If this sentence is rephrased from the relevant RFC, could we also consider linking to that RFC?

Furthermore, I would start a new sentence instead of _and that_, and would cite the RFC there.

falcon/request.py
```diff
@@ -358,10 +358,23 @@ class Request(object):
...
+
+            This property is determined by the value of ``HTTP_IF_MATCH``
+            in the WSGI environment dict.
+
+            Note:
+                This property includes strong and weakness entity-tags and that
+                means that two entity-tags are equivalent if both are not weak
+                and their opaque-tags match character-by-character.
```
vytas7: As discussed, we would also need to include code recipes, for instance, how to check if the provided strong entity-tags match, and so on.

falcon/request_helpers.py
```diff
@@ -38,6 +44,63 @@ def fget(self):
     return property(fget)
 
 
+def make_etag(value, is_weak=False):
```
vytas7: This function should be private IMHO, or we if we need to make it public, we need to document its parameters.

Furthermore, it is used just twice, so it could be inlined instead.

falcon/util/structures.py
```diff
@@ -105,3 +105,19 @@ def copy(self):
...
+
+
+class ETag(str):
+    """An entity-tag ``str``-like object with the weakness indicator.
+
+    Attributes:
+        is_weak (bool): The weakness indicator.
+
+    """
+    is_weak = False
```
vytas7: I don't we need to use the discussed class variable shadowing trick here, if we make sure we always set it. Unless I missed important motivation to do it this way?

vytas7:
Thanks, this is really starting to shape up :+1: 
I just think we need to improve documentation a little bit, like being more specific and verbose when explaining the returned data types, as well as illustrating the concepts with code examples, how, for instance, the returned list could look like, or how to perform the most common chores, like checking if none match, or filtering just strong tags.

falcon/request_helpers.py
kandziu: This function also imported in other source file and probably it would be a good builder for our public class `ETag`.

falcon/util/structures.py
kandziu: It's a public class with public method `to_header` which require this attribute. Since we don't override the class constructor (it worse for performance) and don't ensure that it always will be created through a builder function we must define it. I mean that some users can used it for creating entity tags and set them as a header. Otherwise we should make this class private.

PR #1425, opened 2019-02-04
Testing framework preserves + in URL paths

aparkerlue:
When testing using a URL that contains a "+" in the path component (as distinct from the query component), Falcon replaces "+" with a space—this does not appear to be appropriate behavior.

Consider the call

```python
import falcon.testing

client = falcon.testing.TestClient(app)
client.simulate_get("/a+b")
```

for some `app`. This code initiates the sequence of method invocations:

1. `falcon.testing.client.TestClient.simulate_get()`
2. `falcon.testing.client.TestClient.simulate_request()`
3. `falcon.testing.client.simulate_request()`
4. `falcon.testing.helpers.create_environ()`
5. `falcon.util.uri.decode()`

The last method replaces "+" with a space so that the path `/a+b` becomes `/a b`.

As I understand it, plus signs should be unquoted in query strings but not in the path component of a URI. See [urlencode - When to encode space to plus (+) or %20? - Stack Overflow](https://stackoverflow.com/questions/2678551/when-to-encode-space-to-plus-or-20) for a discussion.

This pull request modifies the `decode()` method by adding the parameter `unquote_plus` so that the method signature becomes `decode(encoded_uri, unquote_plus=True)`. This solution is just a suggestion; I welcome alternatives.

### DIFF
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -192,7 +192,7 @@ if compat.PY2:  # NOQA: C901 - Work around a bug in flake8 McCabe scoring
 
         # PERF(kgriffs): Don't take the time to instantiate a new
         # string unless we have to.
-        if unquote_plus and '+' in decoded_uri:
+        if '+' in decoded_uri and unquote_plus:
             decoded_uri = decoded_uri.replace('+', ' ')
 
         # Short-circuit if we can
@@ -254,7 +254,7 @@ else:
 
         # PERF(kgriffs): Don't take the time to instantiate a new
         # string unless we have to.
-        if unquote_plus and '+' in decoded_uri:
+        if '+' in decoded_uri and unquote_plus:
             decoded_uri = decoded_uri.replace('+', ' ')
 
         # Short-circuit if we can
--- a/falcon/util/uri.py
+++ b/falcon/util/uri.py
@@ -171,7 +171,7 @@ if compat.PY2:  # NOQA: C901 - Work around a bug in flake8 McCabe scoring
                         for a in _HEX_DIGITS
                         for b in _HEX_DIGITS)
 
-    def decode(encoded_uri):
+    def decode(encoded_uri, unquote_plus=True):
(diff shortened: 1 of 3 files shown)


falcon/util/uri.py
```diff
@@ -192,7 +192,7 @@ def decode(encoded_uri):
 
         # PERF(kgriffs): Don't take the time to instantiate a new
         # string unless we have to.
-        if '+' in decoded_uri:
+        if unquote_plus and '+' in decoded_uri:
```
vytas7: Since there is a `PERF` note above, we might want to try to optimize this as much as possible.

As `unquote_plus=False` path is used mostly in testing, we might want to put `'+' in decoded_uri` first to short-circuit `AND`.

tests/test_utils.py
```diff
@@ -201,6 +201,20 @@ def test_uri_decode(self):
             'http://example.com?x=ab%2Bcd%3D42%2C9'
         ) == 'http://example.com?x=ab+cd=42,9'
 
+    def test_uri_decode_unquote_plus(self):
```
nZac: It would be nice to see pytest parametrize here instead of individual `assert`, but that is a nit-pick beyond nitpicks!

vytas7: @nZac indeed :+1:  I was considering that route, but I wanted to include tests for both specifying the kwarg parameter, and omitting it, which made parametrization somewhat more cumbersome.

But maybe I could fully parametrize `*args` and `**kwargs`? It is easy to change, let me know if you think it is worth pursuing.

nZac: That's fine. I am good without it.

vytas7:
@aparkerlue I have to also thank you for this fix, well spotted!

aparkerlue:
Thanks for tying up this pr and merging @vytas7! I meant to address your comment @kgriffs about adding tests but got overtaken in the past couple weeks.

Issue #1431, opened 2019-02-08
jsonschema.validate decorator removes docstring, encounter while using Sphinx


paneru-rajan:
The `jsonschema.validate` decorator does not have `@wraps(func)` which removes the docstring when applied to the methods  like `on_get`, `on_post` etc.. 
This is detected while creating docs with **Sphinx** on methods with the decorator `@jsonschema.validate`.

I have modified the decorator to this:
```
def decorator(func):
        @wraps(func)
        def wrapper(self, req, resp, *args, **kwargs):
            try:
                jsonschema.validate(req.media, schema, format_checker=jsonschema.FormatChecker())
```
Just added `@wraps(func)` inside the function, and it start working...

References:
[https://github.com/sphinx-doc/sphinx/issues/3783](https://github.com/sphinx-doc/sphinx/issues/3783)
[https://docs.python.org/3/library/functools.html#functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps)

vytas7:
Hi @paneru-rajan !
And thank you for the report.

Since you have already solved the issue yourself, would you mind creating a pull request with your change?

vytas7:
Fixed by https://github.com/falconry/falcon/pull/1432

PR #1432, opened 2019-02-08
fix(sphinx): add wraps in jsonschema.validate decorator

paneru-rajan:
the usage of `@jsonschema.validate` on methods loses the doc-string
    when using sphinx to create documents.
`@wraps` will preserve the doc-string

issue #1431


vytas7:
Thanks @paneru-rajan .
Normally we try to add tests for the bugs fixed, but this is bordering trivial, so it looks good as-is to me!

PR #1434, opened 2019-02-09
docs(style): block highlighting

### LINKED ISSUE #1323: doc: Highlighted source washes out the syntax highlighting
Example:

http://falcon.readthedocs.io/en/stable/_modules/falcon/redirects.html#HTTPPermanentRedirect

Nateyo:
Closes #1323 

I checked several different pages, and for anything that doesn't overflow it looks good. Since I changed the color just slightly so `self` was more readable here's a screenshot.

new_style

We have some overflow in the current docs, just based on our source line length and what the current docs css can handle. May want to create a new ticket for fixing overflow.

See:
overflow

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -55,7 +55,9 @@ code.xref.docutils.literal {
 }
 
 div.viewcode-block:target {
-    background: inherit;
+    background:#dadada;
+    border-radius: 5px;
+    padding: 5px;
 }
 
 a:hover, div.sphinxsidebar a:hover, a.reference:hover, a.reference.internal:hover code {
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -55,7 +55,8 @@ code.xref.docutils.literal {
 }
 
 div.viewcode-block:target {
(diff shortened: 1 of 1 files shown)


docs/_static/custom.css
```diff
@@ -55,7 +55,9 @@ code.xref.docutils.literal {
 }
 
 div.viewcode-block:target {
-    background: inherit;
+    background:#dadada;
```
vytas7: Could we add a space before `#dadada` for consistency with the rest of the stylesheet?

vytas7: Other elements defined in this stylesheet seem to be specific in form of `background-color` as opposed to `background`, should we follow the same?

Nateyo: Good point. Actually `background` inherits many different background attributes, so I added the inherit back in (although it's not really inheriting anything, but could in the future) for background and set background-color explicitly.

Nateyo: Fixed. VSCode plops in a color picker to the left so I didn't see the missed space hah

PR #1436, opened 2019-02-11
feat(middleware): Implement request short-circuiting via Response.complete

### LINKED ISSUE #1296: feat: Add a way for Middleware process_request() to indicate whether routing should proceed or the response should just be rendered immediately
We can either set an attribute on `resp`, or simply support returning a truthy value to indicate whether or not the routing should be short-circuited. Other ideas?

Should we also do this for process_resource()?

kgriffs:
Closes #1296

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,6 +59,9 @@ New & Improved
 
 - Added a new ``headers`` property to the ``Response`` class.
 - Removed the ``six`` and ``python-mimeparse`` dependencies.
+- Added a new ``complete`` property to the ``Response`` class. This can be used
+  to short-circuit request processing when the response has been pre-constructed.
+- Removed ``six`` as a dependency.
 - ``Request.context_type`` now defaults to a bare class allowing to set
   attributes on the request context object::
 
--- a/docs/api/middleware.rst
+++ b/docs/api/middleware.rst
@@ -117,13 +117,43 @@ like this::
         mob2.process_response
     mob1.process_response
 
+Short-circuiting
+----------------
+
+A *process_request* middleware method may short-circuit further request
+processing by setting :attr:`~.Response.complete` to ``True``, e.g.::
+
+      ``resp.complete = True``
+
+After the method returns, setting this flag will cause the framework to skip
+any remaining *process_request* and *process_resource* methods, as well as
+the responder method that the request would have been routed to. However, any
+*process_response* middleware methods will still be called.
+
+In a similar manner, setting :attr:`~.Response.complete` to ``True`` from
+within a *process_resource* method will short-circuit further request processing
+at that point.
+
+This feature affords use cases in which the response may be pre-constructed,
+such as in the case of caching.
+
+Exception Handling
+------------------
+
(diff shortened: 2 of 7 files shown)


docs/api/middleware.rst
```diff
@@ -112,13 +112,43 @@ like this::
...
+exception, it will be processed according to the exception type. If
 the type matches a registered error handler, that handler will
 be invoked and then the framework will begin to unwind the
 stack, skipping any lower layers. The error handler may itself
-raise an instance of HTTPError, in which case the framework
-will use the latter exception to update the *resp* object.
+raise an instance of :class:`~.HTTPError` or :class:`~.HTTPStatus`, in
+which case the framework will use the latter exception to update the
+*resp* object.
+
```
kgriffs: While I was updating the docs for this PR, I made a couple of additional updates that aren't directly related, but I decided to include them in the same PR in the interest of getting things merged more quickly for 2.0.

CHANGES.rst
```diff
@@ -32,6 +32,8 @@ New & Improved
 --------------
 
 - Added a new ``headers`` property to the ``Response`` class.
+- Added a new ``complete`` property to the ``Response`` class. This can be used
```
vytas7: Change not mirrored to `2.0.0` changelog yet.

falcon/api.py
```diff
@@ -220,21 +220,24 @@ def __call__(self, env, start_response):  # noqa: C901
                 if self._independent_middleware:
                     for process_request in mw_req_stack:
                         process_request(req, resp)
+                        if resp.complete:
+                            break
                 else:
                     for process_request, process_response in mw_req_stack:
-                        if process_request:
+                        if process_request and not resp.complete:
```
vytas7: Unless I missed something I'm getting an impression that this combination has no test coverage:
* `independent_middleware=False`
* skipping `process_request` due to `resp.complete`

falcon/api.py
```diff
@@ -244,12 +247,16 @@ def __call__(self, env, start_response):  # noqa: C901
                     # route, a default responder is returned and the
                     # resource is None. In that case, we skip the
                     # resource middleware methods.
-                    if resource is not None:
+                    if resource is not None and not resp.complete:
```
vytas7: Same here, I could remove `and not resp.complete` and still get everything to pass.

falcon/api.py
kgriffs: Yikes, good catch! Covered now.

falcon/api.py
kgriffs: In this case, when `resp.complete == True`, `resource` will always be `None`, which is why the check is superfluous. I went ahead and removed it and updated the comment.

tox.ini
```diff
@@ -88,7 +88,7 @@ deps = {[with-debug-tools]deps}
        jsonschema
 
 [testenv:py3_debug]
-basepython = python3.6
+basepython = python3.7
```
vytas7: Good catch, I must have missed these :+1: 

While we are at it, could we also change the new `check_vendored` environment which was developed on a parallel track?

Issue #1405, opened 2019-01-04
Official Support for Python 3.7 


kgriffs:
- [x] README
- [x] Docker images
- [x] Docs
- [x] Tox
- [x] Travis
- [x] setup.py

soultech67:
@vytas7 is this something I can help with? I have a vested interest in 3.7.2 support...

vytas7:
Sure, you are very welcome to help!

See also the checklist by @kgriffs above, that may be helpful in which changes are needed.

vytas7:
@soultech67 have you had a chance to look at this?
I think this issue is important to get out for Falcon 2.0, so do not hesitate to ask for help, or you are constrained by time, could we join forces on this one?

vytas7:
@soultech67 ping!
I am sorry if I am interrupting your work on this due to the imminence of Falcon 2.0. I'll try to quickly sketch something working provided there are no hidden rocks with 3.7.

vytas7:
Done in https://github.com/falconry/falcon/pull/1437

PR #1437, opened 2019-02-11
chore: add official support for CPython 3.7

vytas7:
Addresses https://github.com/falconry/falcon/issues/1405

For the reference: I am leaving Hug tests at CPython 3.6, it seems that Hug's pinned build requirements fail to install under 3.7.
Filed as: https://github.com/timothycrosley/hug/issues/725

### DIFF
--- a/docker/bench_py27_cython.Dockerfile
+++ b/docker/bench_py27_cython.Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Falcon Framework Maintainers
 RUN pip install cython
 
 # We don't currently benchmark JSON deserialization, but in the future we might
-RUN pip install ujson
+# RUN pip install ujson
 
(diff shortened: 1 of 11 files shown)


docker/bench_py27_cython.Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Falcon Framework Maintainers
 RUN pip install cython
 
 # We don't currently benchmark JSON deserialization, but in the future we might
-RUN pip install ujson
+# RUN pip install ujson
```
kgriffs: You may want to change this to orjson to mirror docker/bench_py37_cython.Dockerfile

vytas7: No, I may not I think. `orjson` does not support Python 2.7.

PR #1439, opened 2019-02-12
chore: Vendor python-mimeparse

### LINKED ISSUE #1420: Consider vendoring python-mimeparse
As brought up on https://github.com/falconry/falcon/pull/1383 and discussions on Gitter, we may consider vendoring ``python-mimeparse`` dependency.

Normally, vendoring Python dependencies is just an unnecessary burden:
* Laborious process to keep integrating upstream fixes
* *Pip* is good at handling dependencies
* Redundant duplication of libraries if the application is also using them

However, in this particular case of ``python-mimeparse`` it might be a compelling choice:
* As we are no longer relying on ``six``, this would make Falcon completely dependency-free, which, while not being much of a benefit in the *virtualenv* + *Pip* workflow, might ease distributing Falcon in other ways, such as packaging for Linux distributions (see for instance: https://github.com/dbtsai/python-mimeparse/issues/20 :arrow_left: note by @kgriffs )
* ``python-mimeparse`` is a mature package with a concise codebase (cone 190-line file), the pace of recent developments is rather moderate (at the time of writing, Jan 2019, the latest stable release on PyPi is from 2016)
* We would cythonize it together with the rest of Falcon, see also https://github.com/falconry/falcon/issues/1183

If we decide to do it, we need to keep an eye on upstream developments. We could introduce a Tox or Travis check that we have not fallen behind its stable PyPi version.

kgriffs:
Closes #1420

### DIFF
--- a/falcon/vendor/mimeparse/__init__.py
+++ b/falcon/vendor/mimeparse/__init__.py
@@ -1,0 +1,10 @@
+"""
+This module wraps code from the MIT-licensed python-mimeparse project. The
+original project source code may be obtained from GitHub by visiting the
+following URL:
+
+    https://github.com/dbtsai/python-mimeparse
+
+"""
+
+from .mimeparse import *  # NOQA
--- a/tools/check-vendored.sh
+++ b/tools/check-vendored.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+_EXPECTED_VERSION_MIMEPARSE="1.6.0"
+
+_VERSION_OUTPUT=$(pip install python-mimeparse== 2>&1)
+if [[ $_VERSION_OUTPUT == *", $_EXPECTED_VERSION_MIMEPARSE)"* ]]; then
+	echo "Latest version of python-mimeparse has not changed ($_EXPECTED_VERSION_MIMEPARSE)"
+    exit 0
+fi
+
+echo "Latest version of python-mimeparse is newer than expected."
+exit 1
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,7 @@
 [run]
 branch = True
 source = falcon
-omit = falcon/tests*,falcon/cmd*,falcon/bench*
+omit = falcon/tests*,falcon/cmd*,falcon/bench*,falcon/vendor/*
 
 parallel = True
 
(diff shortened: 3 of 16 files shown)


tools/check-vendored.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+_EXPECTED_VERSION_MIMEPARSE="1.6.0"
```
kgriffs: In the interest of KISS, the version number is hard-coded here instead of going to the trouble up reading it from `mimeparse.py`. We can always change this later.

falcon/vendor/mimeparse/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
+This module wraps code from the MIT-licensed python-mimeparse project. The
+original project source code may be obtained from GitHub by visiting the
+following URL:
+
+    https://github.com/dbtsai/python-mimeparse
+
+"""
+
+from .mimeparse import *  # NOQA
```
vytas7: No new line at the end of file (in the Unix world, it is a punishable offense :smiling_imp: )

tools/check-vendored.sh
```diff
@@ -0,0 +1,12 @@
...
+
+_VERSION_OUTPUT=$(pip install python-mimeparse== 2>&1)
+if [[ $_VERSION_OUTPUT == *", $_EXPECTED_VERSION_MIMEPARSE)"* ]]; then
+	echo "Latest version of python-mimeparse has not changed ($_EXPECTED_VERSION_MIMEPARSE)"
+    exit 0
+fi
+
+echo "Latest version of python-mimeparse is newer than expected."
+exit 1
```
vytas7: No new line at the end of file.

kgriffs: fixed

falcon/vendor/mimeparse/__init__.py
kgriffs: Sheesh. Usually I add an extra line without thinking. Fixed.

kgriffs:
> it seemed that the vendored `mimeparse` would not cythonize

Nice catch, fixed.

falcon/vendor/mimeparse/__init__.py
vytas7: Editors normally add that, at least for source files. And `flake8` would complain, but `NOQA` on the last line felled us here. At least Github rendered it in a very visual way, making my job easy :slightly_smiling_face:

PR #1441, opened 2019-02-12
doc(middleware): Clarify how process_request can be used to re-route requests

kgriffs:


vytas7:
LGTM, thanks!
(I remember I took a note of this being raised on our *user* channel, just have never got back to it... :grimacing: )

PR #1442, opened 2019-02-12
cleanup: Remove backwards-compat shims

kgriffs:


tests/test_middleware.py
```diff
@@ -619,6 +619,7 @@ def handler(req, resp, ex, params):
             'ExecutedLastMiddleware.process_response',
             'ExecutedFirstMiddleware.process_response'
         ]
+        print(context['executed_methods'])
```
vytas7: This can be handy for debugging, but probably better added when need arises ☺

Issue #1423, opened 2019-02-02
Revise & clean up code examples and tutorials


vytas7:
Go through existing code samples and code tutorials:
* Check code quality of other inlined code samples, could flake8-rst be used?

kgriffs:
@vytas7  I think it would make sense to split this into two issues; one for critical stuff that is also quick enough to do for 2.0, the other for the remainder that we can assign to the 3.0 milestone.

vytas7:
Rescheduling all items except the most pressing one (hopelessly obsolete docs front page example) to 3.0, we won't be able to spare the bandwidth for 2.0.

A "quick fix" for the said example is proposed in https://github.com/falconry/falcon/pull/1438

Ciemaar:
I'm going to take a look at this after lunch

vytas7:
We have some open PRs that would improve the situation wrt our `look` example, but we still have work to do there:
* Streamline tests, consider implementing that integration test process wrapper
* Revise `requirements` (some of them may be out of date, see https://github.com/falconry/falcon/issues/1605)
* Add tests to CI

vytas7:
The most pressing points were worked on during PyCon US 2019 Sprints, as well as parts of other issues.
However, we won't be able to focus on perfecting this until the final 3.0 release is out. Rescheduling to 3.1.

PR #1446, opened 2019-02-13
doc(errors): Fix wrong indent

kandziu:


vytas7:
Nice catch indeed, thanks!
For the reference, the topic is touched in https://github.com/falconry/falcon/issues/1423 , we may want to use `flake8-rst` or a similar tool to validate the integrity of documentation examples.

PR #1447, opened 2019-02-14
feat(Request): Improve request cookie handling

### LINKED ISSUE #1046: Multiple cookies with same name in request returns only one
This looks like it could be relatively simple fix, but it would involve changing the cookie API a bit.

If my browser has two cookies set using the same name - let's call this `_csrf` for example - they are both sent to the server. This is the string returned by `req.get_header("Cookie")`:

```
_csrf=2HBpry7fn6OBCIdS2SnFI-jCdQcq3bt0m03gWgCtjRA; _csrf=Oq_o_Lo_dhhDXqZHQ95NhBp8Nvzntx1FUG-2ryAi6nA
```

However, only one of these is made available in `req.cookies`.

It would be preferable to have a method of getting both of these cookies somehow.

### LINKED ISSUE #1107: Optimize request cookie parsing

kgriffs:
This patch provides support for getting all values for a given cookie (when
the same cookie is included multiple times in the request), and also optimizes
request cookie parsing to make it an order of magnitude faster.

BREAKING CHANGE: Request.cookies now gives precedence to the first value
  encountered in the Cookie header for a given cookie name, rather than the
  last.

BREAKING CHANGE: Request cookie parsing no longer uses the standard library
	for most of the parsing logic. This may lead to subtley different results
	for archaic cookie header formats, since the new implementation is based on
	the latest RFC 6265.

Closes #1107
Closes #1046

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -27,11 +27,6 @@ from falcon.util import json
 from falcon.util.uri import parse_host, parse_query_string
 from falcon.vendor import mimeparse
 
-# NOTE(tbug): In some cases, compat.http_cookies is not a module
-# but a dict-like structure. This fixes that issue.
-# See issue https://github.com/falconry/falcon/issues/556
-SimpleCookie = compat.http_cookies.SimpleCookie
-
 DEFAULT_ERROR_LOG_FORMAT = (u'{0:%Y-%m-%d %H:%M:%S} [FALCON] [ERROR]'
                             u' {1} {2}{3} => ')
 
@@ -256,8 +251,12 @@ class Request(object):
         client_accepts_xml (bool): ``True`` if the Accept header indicates that
             the client is willing to receive XML, otherwise ``False``.
         cookies (dict):
-            A dict of name/value cookie pairs. (See also:
-            :ref:`Getting Cookies <getting-cookies>`)
+            A dict of name/value cookie pairs. The returned object should be
+            treated as read-only to avoid unintended side-effects.
+            If a cookie appears more than once in the request, only the first
+            value encountered will be made available here.
+
+            See also: :meth:`~get_cookie_values`
         content_type (str): Value of the Content-Type header, or ``None`` if
             the header is missing.
         content_length (int): Value of the Content-Length header converted
@@ -394,7 +393,6 @@ class Request(object):
         '_cached_prefix',
         '_cached_relative_uri',
         '_cached_uri',
-        '_cookies',
         '_params',
         '_wsgierrors',
         'content_type',
@@ -408,6 +406,9 @@ class Request(object):
         'uri_template',
         '_media',
     )
+
+    _cookies = None
+    _cookies_collapsed = None
 
     # Child classes may override this
     context_type = type('RequestContext', (dict,), {})
@@ -455,8 +456,6 @@ class Request(object):
 
             else:
                 self._params = {}
-
-        self._cookies = None
 
         self._cached_access_route = None
         self._cached_forwarded = None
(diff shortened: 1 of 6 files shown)


falcon/request_helpers.py
```diff
@@ -15,6 +15,74 @@
 """Utilities for the Request class."""
 
 import io
+import re
+
+from falcon.util.compat import http_cookies
+
+# https://tools.ietf.org/html/rfc6265#section-4.1.1
+_COOKIE_NAME_RESERVED_CHARS = re.compile('[\x00-\x1F\x7F-\xFF()<>@,;:\\\\"/[\\]?={} \x09]')
```
kgriffs: Someone should sanity-check these chars against the RFC.

falcon/request.py
```diff
@@ -1077,6 +1069,37 @@ def get_header_as_datetime(self, header, required=False, obs_date=False):
                    'Section 7.1.1.1')
             raise errors.HTTPInvalidHeader(msg, header)
 
+    def get_cookie(self, name):
```
vytas7: I have some misgivings about the function name `get_cookie` and its signature.
In Falcon, one can draw parallels between cookie functions and headers functions, they often set or retrieve a **name** :arrow_right: **value** pair, see for instance, [`get_header`](https://falcon.readthedocs.io/en/stable/api/request_and_response.html#falcon.Request.get_header), `set_header`, `set_cookie`, and so on.

However, `get_cookie` would now return a `list`, making it somewhat of an outlier. Could we reconsider alternative names implying a list (such as I remember `get_cookie_values` has been proposed before)?

falcon/request_helpers.py
vytas7: We could potentially simplify the expression somewhat by using `r'raw strings'`. It might make it easier to read those quad backslashes.

vytas7: One potential pitfall could be unicode string handling on Python 3. Unless it is already guaranteed by the WSGI mechanics that we would not handle any case like that?

```python
>>> cookie_and_burger ='cookie&🍔' 
>>> _COOKIE_NAME_RESERVED_CHARS = re.compile('[\x00-\x1F\x7F-\xFF()<>@,;:\\\\"/[\\]?={} \x09]')
>>> _COOKIE_NAME_RESERVED_CHARS.search(cookie_and_burger)
```

That happily passes on Python 3, but the standard says only US_ASCII.

vytas7: I have otherwise validated the expression to conform re *separators*.

falcon/request_helpers.py
```diff
@@ -15,6 +15,73 @@
...
+
+        # NOTE(kgriffs): Skip malformed cookie-pair
+        if not name:
+            continue
+
+        # NOTE(kgriffs): Skip cookies with invalid names
+        if _COOKIE_NAME_RESERVED_CHARS.search(name):
+            continue
+
+        # NOTE(kgriffs): To maxmize compatibility, we mimic the support in the
```
vytas7: _maxmize_ (sic) spelling :arrow_right: _maximize_

falcon/request.py
kgriffs: Good point. `get_cookie_values()` might work. Let me see if I can think of any other possible names (suggestions welcome).

falcon/request_helpers.py
kgriffs: Fortunately we don't have to worry about code points in header strings outside the range 0x0000 - 0x00FF per PEP 3333. I'll add an inline comment for future travelers.

kgriffs: Perhaps I'm missing something, but I don't believe there is a way to specify hex values in raw string literals?

falcon/request_helpers.py
kgriffs: 👍

falcon/request.py
vytas7: I don't have any extremely good suggestions unfortunately. I remember I was suggesting *get_cookies*, but that would actually be even more confusing as it would imply all at once. *get_cookie_as_list* would technically fill the bill well, but it sounds somewhat awkward as we don't have other *get_cookie_as_*.

falcon/request_helpers.py
vytas7: > Perhaps I'm missing something, but I don't believe there is a way to specify hex values in raw string literals?

Oops... that wasn't the brightest suggestion on my part :man_facepalming:

vytas7:
LGTM now!

Issue #1051, opened 2017-05-20
Implement a Quickstart CLI


jmvrbanac:
Setting up a new Falcon API takes longer than it probably should. Also, quite a few newer contributors ask questions about how to start up and structure a new Falcon API project. While we have some documentation around this, it would be great if we make this much easier. This would be sort of like `sphinx-quickstart` only for Falcon.

kgriffs:
Just a thought, but perhaps we could roll this into a "falcon-cli" that could start a development server, generate a QuickStart project template, and maybe even pull down templates from GitHub repos or something.

kgriffs:
Just got a request for this feature via the Gitter channel, copying here for the record:

> do we have a sample "manage.py" to run one-off CLI commands in falcon, similar to django or flask ?

phoenix911:
I'm working on something similar. [falacon_setup](https://github.com/phoenix911/falacon_setup) . M planning to add cli later.

kgriffs:
Based on discussions with other members of the maintainer's team during PyCon I think we have come to a rough consensus that our first iteration on this will simply output a simple scaffolding with inline comments showing where people can plug in their favorite libraries for config, logging, DB, etc.

kgriffs:
P.S. - We should also provide a basic layout suggestion for a project in the docs.

nZac:
@phoenix911 that is cool, thanks for sharing. I noticed you organized things around Falcon concepts (hooks, controllers, middlewares) instead of app/business concepts (user, account, etc) which then contain hooks/controllers/middlewares. 

Just curious if you had thoughts structuring based on app concepts?

PR #1455, opened 2019-02-21
An API Builder object that allows for custom methods on routes

cbefus:
This work wraps the current API object with a builder class.  This affords a few benefits over using the raw API class.

1. We are separating the concerns of building an `API` and using and `API`.  The current `API` class has both setters and doers.  Ideally we would avoid such hybrids.  This builder does not affect the `API` class (for backwards compatibility) but we could at some point deprecate the setters on the `API` object.
2. To better support a builder pattern, the setters in the new `APIBuilder` class all `return self` which means it can be used in a chained way. For example:
```
 app = APIBuilder() \
     .add_get_route('/', foo_resource.get_foos) \
     .add_put_route('/update_bar', update_bar) \
     .build()
```
3. Functions in resources no longer need to be named after HTTP methods and prefixed with the `on_` prefix.  This means that a resource can now have functions named `updated_username`, or `create_user` and have those methods be agnostic of the HTTP method that is directed to them.
4. The builder frees up the tight-REST resource constraint of Falcon, where a resource must be a class with at most one function per HTTP method.
  4.1 A route can now just be a top level python function which takes a `request` and a `response`.
  4.2 A resource can have multiple methods which are using the same HTTP method as long as they are
      using a different uri template. 

For this change I have maintained 100% test coverage by:
1. Adding new tests in `tests/test_api_builder.py` to cover any code paths of the builder.
2. Adding a doubled test in every other test where we instantiate an `API` but using the `APIBuilder` instead.  This should shore up any confidence issues in this new `APIBuilder` providing the same underlying `API` as would have been constructed by calling the `API` directly.

### DIFF
--- a/falcon/__init__.py
+++ b/falcon/__init__.py
@@ -28,6 +28,7 @@ the framework's classes, functions, and variables::
 from falcon.version import __version__  # NOQA
 from falcon.constants import *  # NOQA
 from falcon.api import API  # NOQA
+from falcon.api_builder import APIBuilder  # NOQA
 from falcon.status_codes import *  # NOQA
 from falcon.errors import *  # NOQA
 from falcon.redirects import *  # NOQA
--- a/falcon/api_builder.py
+++ b/falcon/api_builder.py
@@ -1,0 +1,219 @@
+
+from falcon import api_helpers as helpers, API, DEFAULT_MEDIA_TYPE, routing
+from falcon.request import Request
+from falcon.response import Response
+
+import sys
+import inspect
+
+
+class APIBuilder:
+
+    def __init__(self):
+        self._media_type = DEFAULT_MEDIA_TYPE
+        self._call_response_middleware_on_request_middleware_exception = True
+        self._middlewares = []
+        self._request_type = Request
+        self._response_type = Response
+        self._router = routing.DefaultRouter()
+        self._error_serializer = helpers.default_serialize_error
+        self._routes = {}
+        self._error_handling_functions = []
+        self._sinks = []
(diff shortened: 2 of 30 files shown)


nZac:
Wow, this is really neat, thanks for submitting it! Sorry for it taking so long to get you feedback, we have been working hard on 2.0.

A change like this is rather large and would be hard to integrate directly, so we can't take this _as is_.  However, there are some neat patterns (like the `APIBuilder`) in this PR that might be well suited for a separate package and we would be interested in seeing that grow in the community apart from the base framework and talk about inclusion when it has stablized.

We have heard from the community that it would be useful to have simple functions for resources. I have opened an issue for this (#1532).

We have also been discussing how to encourage people to structure falcon projects. You can track that in #1051, I think that loosely dovetails into your API builder.

> where a resource must be a class with at most one function per HTTP method

You can now use the `suffix` kwarg on `Api.add_route` to tell Falcon that there is more than one responder function on this resource.

```python
class MyResource:
    def on_get(self, req, resp, obj_id):
        resp.media = { 'label': 'MySingleObject'}

    def on_get_collection(self, req, resp):
        resp.media = [{ 'label': 'MyObject1'}, {'label': 'MyObject2'}]

api = falcon.Api()
my_resource = MyResource()
api.add_route("/obj", my_resource, suffix='collection')
api.add_route("/obj/{obj_id}", my_resource)
```

I am going to close this PR for now since we can't merge this as is, it is just to big of a change. Catch one of the @falconry/maintainers on the Gittr channel if you have more questions.

PR #1459, opened 2019-02-24
feat(etag): Improve etag performance and docs, plus usability of the ETag class

kgriffs:


falcon/util/structures.py
```diff
@@ -108,22 +108,114 @@ def __repr__(self):
...
+            this one.
+
+        Returns:
+            bool: ``True`` if the two entity-tags match, otherwise ``False``.
+
+        """
+
+        return self == other and not (self.is_weak or other.is_weak)
+
+    def dumps(self):
```
kgriffs: I wasn't sure what to name these methods, so I just defaulted to something that hopefully seems pythonic. Suggestions welcome if these names don't make sense.

falcon/util/structures.py
```diff
@@ -108,22 +108,114 @@ def __repr__(self):
...
-            return 'W/"%s"' % self
-        return '"%s"' % self
+            # PERF(kgriffs): Simple concatenation like this is slightly faster
+            #   than %s string formatting.
+            return 'W/"' + self + '"'
+
+        return '"' + self + '"'
+
+    @staticmethod
+    def loads(etag_str):
```
kgriffs: I decided to add this as part of the public interface because it can be useful when constructing ETag objects to perform strong comparisons with something that came in on the request. Note that this lives at a different level of abstraction than the make_etags() method that was originally part of the public interface.

tests/test_request_attrs.py
```diff
@@ -5,14 +5,31 @@
...
-from falcon.request_helpers import make_etag
 import falcon.testing as testing
 import falcon.uri
 from falcon.util import compat
+from falcon.util.structures import ETag
 
 _PROTOCOLS = ['HTTP/1.0', 'HTTP/1.1']
 
 
+def _make_etag(value, is_weak=False):
```
kgriffs: Moved this in here because it was still useful for constructing test data, if nothing else.

falcon/request_helpers.py
```diff
@@ -113,69 +118,59 @@ def fget(self):
...
     Args:
         etag_str (str): An ASCII header value to parse ETags from. ETag values
             within may be prefixed by ``W/`` to indicate that the weak comparison
             function should be used.
 
     Returns:
-        A list of unquoted ETags or ``['*']`` if all ETags should be matched.
+        list: A list of unquoted ETags or ``['*']`` if all ETags should be
+        matched. If the string to be parse is ``None``, empty, or contains only
+        whitespace, ``None`` will be returned instead.
```
kgriffs: I added some additional cases to when ``None`` would be returned because it seemed to me the way a caller would deal with them would essential boil down to the same use case of "there wasn't an etag provided".

falcon/request_helpers.py
```diff
@@ -113,69 +118,59 @@ def fget(self):
...
+
+    # PERF(kgriffs): Parsing out the weak string like this turns out to be more
+    #   performant than grabbing the entire entity-tag and passing it to
+    #   ETag.loads(). This is also faster than parsing etag_str manually via
+    #   str.find() and slicing.
+    for weak, value in _ENTITY_TAG_PATTERN.findall(etag_str):
+        t = ETag(value)
+        t.is_weak = bool(weak)
+        etags.append(t)
+
```
kgriffs: `findall()` is generally faster when there are only a handful of results expected. Plus, memory usage should not be a big deal since the set of strings generated is not going to be very large, and each string is not very long.  Anyway, this along with the other changes made here resulted in a decent speedup.

falcon/request.py
```diff
@@ -355,32 +355,34 @@ class Request(object):
...
             Range header, or ``None`` if the header is missing
         if_match (list): Value of the If-Match header, as a parsed list of
-            :class:`falcon.ETag` objects or ``None`` if the header is missing.
+            :class:`falcon.ETag` objects or ``None`` if the header is missing
+            or its value is blank.
 
-            This property is determined by the value of ``HTTP_IF_MATCH``
-            in the WSGI environment dict.
+            The header is parsed on-demand, and so if you need to reuse the
+            result it is best to assign it to a variable rather than reference
```
vytas7: I am quote hopeless at English punctuation, so this is just a flag for you to check if you intentionally meant it this way -- I would put a comma after _result_.

falcon/util/structures.py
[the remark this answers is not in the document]
kgriffs: TBH I wasn't thinking about subclassing, but it wouldn't hurt to support it. I'll make the change. 👍

falcon/request.py
kgriffs: Due to the clauses being joined by *and*, placing a comma there would make it a little choppier than I would like. However, the sentence in its current form is admittedly rather long. It may be best to simply split it into two sentences.

kgriffs:
Updated per feedback.

falcon/request.py
```diff
@@ -619,11 +623,11 @@ def date(self):
 
     @property
     def if_match(self):
-        return helpers.parse_etags(self.env.get('HTTP_IF_MATCH'))
```
nZac: Nit-pick: Any reason we can't just store these in a private variable on first access instead of parsing every time?

kgriffs: Well, I was debating whether to add the extra overhead of another variable slot and having to initialize it. I guess we could use the class variable shadowing trick?

vytas7: I think we could just try releasing as-is now, and revise in the future if needed.
I wouldn't take it for granted that this property is accessed many times, I could imagine one would just check this value only once to render the `Not Modified` or similar.

kgriffs: Fixed

falcon/request.py
```diff
@@ -619,11 +613,30 @@ def date(self):
...
+        #   pattern for us.
+        # PERF(kgriffs): It probably isn't worth it to set
+        #   self._cached_if_match to a special type/object to distinguish
+        #   between the variable being unset and the header not being
+        #   present in the request. The reason is that if the app
+        #   gets a None back on the first reference to property, it
+        #   probably isn't going to access the property again (TBD).
+        if self._cached_if_match is None:
+            header_value = self.env.get('HTTP_IF_MATCH')
+            if header_value:
```
vytas7: This `if` check seems to be a bit superfluous, since `_parse_etags` can understand `None`, and explicitly so by design:
```python
if etag_str is None:
        return None
```

Or is it by design in order not to add an attribute needlessly?
However, in that case the snippet above is not really needed in `_parse_etags`.

falcon/request.py
```diff
@@ -619,11 +613,30 @@ def date(self):
...
+                self._cached_if_match = helpers._parse_etags(header_value)
+
+        return self._cached_if_match
 
     @property
     def if_none_match(self):
-        return helpers.parse_etags(self.env.get('HTTP_IF_NONE_MATCH'))
+        if self._cached_if_none_match is None:
+            header_value = self.env.get('HTTP_IF_NONE_MATCH')
+            if header_value:
```
vytas7: Same as discussed for `If-Match`.

falcon/request.py
kgriffs: This is a bit of cruft left over from the refactor. The if statement is an optimization to avoid both a function call and setting an attribute. All things considered, I think it probably makes sense to just remove the statement inside `_parse_etags()` that you quoted above.

kgriffs: Updated. Let me know what you think.

vytas7: Looking good now IMHO!

PR #1463, opened 2019-02-28
refactor(Response): Restore the stream_len alias

kgriffs:
Restore the stream_len alias since it was only deprecated as of 2.0, not 1.4, and we should allow at least one release cycle before removing deprecated parts of the interface.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,7 @@ matrix:
           env: TOXENV=py37_smoke_cython
         - python: 2.7
           env: TOXENV=docs
-        # TODO(kgriffs): Re-enable once we have mitigated the stream_len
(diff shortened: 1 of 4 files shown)


vytas7:
I have just remembered Hug gate in our Travis setup, was this breaking change the only thing blocking it, or did we have more?

kgriffs:
Let's try re-enabling the hug gate.

vytas7:
Ah, I guess we'll have to forget the Hug gate for now, it seems Hug is building heavily upon older signatures of hooks, middlewares, and error handlers which where removed in https://github.com/falconry/falcon/commit/6bb2ac770ff12ef2419e30c122a23356d3251cca

PR #1468, opened 2019-03-03
docs: attempt to fix Codecov badge on PyPi

vytas7:


vytas7:
Hmm... now it does not render in GitHub file preview... :grimacing: 

I was trying to model after this project: https://github.com/oasis-open/cti-stix-elevator
They seem to get the badge both on GitHub page, and PyPi, but probably there's more to it...

PR #1472, opened 2019-03-04
doc(middleware): Add example and fix markup typo

kgriffs:


docs/api/middleware.rst
```diff
@@ -71,6 +71,11 @@ Falcon's middleware interface is defined as follows:
     component modifies ``req.path`` in its *process_request* method,
     the framework will use the modified value to route the request.
 
+    For example::
+
+        # Route requests based on the host header
+        req.path = '/' + req.host + '/' + req.path
```
vytas7: `Request.path` normally already starts with a `'/'`, so the `'/' + req.path` part of the expression would yield a double slash, which may confuse users.

kgriffs: Ah shoot, wasn't thinking straight. I blame the flu. 😉  

Fixed.

vytas7: Looks good now :+1:

PR #1473, opened 2019-03-05
docs(setup): omit raw HTML based patron list in long description

vytas7:


vytas7:
This is probably the ugliest silver taping I've done in a while. Suggestions to make this cleaner are very welcome!

setup.py
```diff
@@ -74,13 +74,27 @@ def list_modules(dirname):
...
     in_raw = False
 
     description_lines = []
 
     # NOTE(kgriffs): PyPI does not support the raw directive
     for readme_line in io.open('README.rst', 'r', encoding='utf-8'):
-        if readme_line.startswith('.. raw::'):
+
+        # NOTE(vytas): The patron list largely builds upon raw sections
+        if 'big thank you to our patrons' in readme_line.strip().lower():
```
kgriffs: One thing that might make parsing a bit simpler is to add some cross-reference bookmarks to `README.rst`. They would not influence the normal rendering on GitHub, but you could check for them here as a way of knowing what sections to leave out, e.g.:

```rst
.. _begin_patrons:

blah blah

.. _end_patrons:
```

vytas7: I have now hopefully improved the data and code separation in that vein, even though it is still somewhat awkward... but maybe it is in the nature of this substitution itself.

kgriffs: Thanks, I think this is sufficient for now.

PR #1474, opened 2019-03-06
fix: omit the charset parameter from JSON media type

grktsh:


vytas7:
Thanks for this PR @grktsh ,
we have indeed forgotten to include this change in the first alpha release, even though this change was clearly slated to be made for 2.0, good catch! Let's see if we can still carry this change into the 2.0 beta.

I think I have found one more case which needs to be ironed out wrt explicitly specifying charset, looking good otherwise!

kgriffs:
According to my reading of RFC 7303 and https://www.w3.org/TR/xml/ (section 4.3.3) it is safe to leave off the charset for `application/xml` when the encoding is utf-8. I can take care of this in a separate PR. We'll also need to update the changelog for both of these changes.

falcon/api_helpers.py
```diff
@@ -126,12 +124,11 @@ def default_serialize_error(req, resp, exception):
...
-            representation = exception.to_json()
+            resp.body = exception.to_json()
+            resp.content_type = preferred
         else:
-            representation = exception.to_xml()
-
-        resp.body = representation
-        resp.content_type = preferred + '; charset=UTF-8'
+            resp.body = exception.to_xml()
+            resp.content_type = preferred + '; charset=UTF-8'
```
kgriffs: As I commented earlier, I think we can actually remove the charset parameter in this case as well. But we can just merge this PR as-is and then I'll rebase #1476 to suit.

kgriffs:
@grktsh sorry about the mixup!

grktsh:
@kgriffs Sorry!

vytas7:
LGTM now, as my comments were addressed in https://github.com/falconry/falcon/pull/1476 :+1:

PR #1480, opened 2019-03-12
Ensure response status and headers are assigned after `start_response` was called

### LINKED ISSUE #1479: Test client fails when WSGI application is a generator
Here is the test that proves it:

```


def another_dummy_wsgi_app(environ, start_response):
    start_response('200 OK', [('Content-Type', 'text/plain')])

    yield 'It works!'


def test_falcon_testing_client():
    client = TestClient(another_dummy_wsgi_app)

    client.simulate_get('/nevermind')
```

Here is [the link](/tribals/falcon/blob/feature%2Fgenerator-app/tests/test_testing.py) for convenience.

tribals:
Closes #1479

### DIFF
--- a/falcon/testing/srmock.py
+++ b/falcon/testing/srmock.py
@@ -21,7 +21,17 @@ used, along with a mock environ dict, to simulate a WSGI request.
 from falcon import util
 
 
-class StartResponseMock(object):
+class ResponseMeta(object):
+    @property
+    def status(self):
+        raise NotImplementedError
+
+    @property
+    def headers(self):
+        raise NotImplementedError
+
+
+class StartResponseMock(ResponseMeta):
     """Mock object representing a WSGI `start_response` callable.
 
     Attributes:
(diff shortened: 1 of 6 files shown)


requirements/tests
```diff
@@ -12,3 +12,6 @@ ujson
 # Python 3 Only Handlers
 python-rapidjson; python_version >= '3'
 orjson; python_version >= '3' and python_version != '3.4' and platform_python_implementation != 'PyPy'
+
+# Python 2 Only
+mock; python_version < '3.3'
```
vytas7: This would of course go away with Python 3, but maybe we could do without this `mock` dependency at all?
Other tests, such as `tests/test_static.py`, we are employing [`pytest` monkey-patching facilities](https://docs.pytest.org/en/latest/monkeypatch.html), maybe that could be good enough?

falcon/testing/srmock.py
```diff
@@ -21,7 +21,17 @@
 from falcon import util
 
 
-class StartResponseMock(object):
+class ResponseMeta(object):
+    @property
+    def status(self):
```
vytas7: It seems refactoring these `status` properties breaks some existing unit tests, @tribals could you please take a look at Travis output?

requirements/tests
tribals: It is not a production dependency, it is for test only, so personally I don't see any issues. Of course, final decision is yours.

I don't know how exactly you are using pytest monkey patching because the tests are written using "old style" `unittest` tests. And old `unittest`-style tests doesn't support (pytest's amazing fixture injection)[https://docs.pytest.org/en/latest/fixture.html#fixtures-a-prime-example-of-dependency-injection].

Also, about `monkeypatch` vs `mock` - the later is much easier to use, IMHO. But, it is just IMHO, though.

falcon/testing/srmock.py
tribals: I ran full test suite in both `py27` and `py36` envs before submitting pull request. I fixed all (as I thought) tests broken by this change.

I will take a look at Travis, of course.

requirements/tests
vytas7: I agree with you that `mock` is easier/more modern, I prefer it too, but my comment was more about staying consistent with the existing tests, which are largely `pytest`-based.

Btw here we are actually using just `mock.sentinel`, and that only once if I understand it correctly. That could probably be done other way without dragging in `mock` just for that sole purpose?

vytas7:
Superseded by https://github.com/falconry/falcon/pull/1488

PR #1481, opened 2019-03-15
fix(API): Revert breaking change vs. 1.4 regarding forcing the Content-Length header

kgriffs:


kgriffs:
This patch reverts the 2.0 behavior to that of 1.4. I found this issue while testing a project that was relying on the Content-Length being overridden. It resulted in a lot of weird behavior by HAProxy / NGINX that is difficult to diagnose.

tests/test_headers.py
```diff
@@ -207,14 +207,24 @@ def on_get(self, req, resp):
...
-        resp.body = self._body
+
+        if self._body:
+            resp.body = self._body
+
+        if self._data:
+            resp.data = self._data
+
+    def on_head(self, req, resp):
+        # NOTE(kgriffs): Use stream_len for now to cover the deprecated alias
```
vytas7: Intent of this note is somewhat unclear (suspected copy-paste from L216?)

vytas7:
What is even worse, reverse-proxies, front-end servers or HTTP clients might hang for a long time, or time-out waiting for "full" content if you set the `Content-Length` to a greater value than the actual content length being sent.

tests/test_headers.py
kgriffs: oops, yeah. Copy-paste #fail - Let me fix that real quick

kgriffs: fixed

PR #1486, opened 2019-03-27
feat: improve dict-compatibility for the default Request and Response context type

vytas7:


docs/user/faq.rst
```diff
@@ -617,19 +617,17 @@ now simply set attributes on the object:
...
 
-Since the default context type derives from dict, the existing code will work
-unmodified with Falcon 2.0. Nevertheless, it is recommended to change usage
-as outlined above since the ability to use context as dictionary may be
-removed in a future release.
+The new default context type emulates the dict interface in a way that context
+attributes are linked to dict items, i.e. setting an object attribute also sets
+the corresponding dict item, and vice versa. As a result, the existing code
+will largely work unmodified with Falcon 2.0. Nevertheless, it is recommended
+to change usage as outlined above since the ability to use context as dict may
```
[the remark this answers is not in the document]
kgriffs: I would suggest emphasizing the interface (mapping) vs. the type (dict) to avoid any confusion. With that in mind, you might change this:

> since the ability to use context as dict may be removed in a future release.

To this:

> since the dict-like mapping interface may be removed from the context type in a future release.

docs/user/faq.rst
```diff
@@ -617,19 +617,17 @@ now simply set attributes on the object:
...
-   Context attributes are not linked to dict items in any special way,
-   i.e. setting an object attribute does not set the corresponding dict item,
-   nor vice versa.
-
-   Furthermore, if you need to mix-and-match both approaches under migration,
-   beware that setting attributes such as *items* or *values* would obviously
-   shadow the corresponding dict functions.
+   If you need to mix-and-match both approaches under migration, beware that
+   setting attributes such as *items* or *values* would obviously shadow the
+   corresponding dict functions.
```
kgriffs: dict ➡️ mapping

falcon/util/structures.py
```diff
@@ -107,6 +107,126 @@ def __repr__(self):
         return '%s(%r)' % (self.__class__.__name__, dict(self.items()))
 
 
+class Context(object):
```
kgriffs: It may be helpful to add an inline `# NOTE` comment to explain why this does not inherit from `MutableMapping`. I assume it is (1) to avoid breaking any app code later on that may rely on checking to see if `Context` subclasses `MutableMapping` if we remove the mapping interface, and (2) to avoid an unnecessary inheritance chain, since for the sake of efficiency all of the default implementations provided by the ABC have to be overridden.

CHANGES.rst
```diff
@@ -67,8 +67,10 @@ Breaking Changes
 - The deprecated ``stream_len`` property was removed from the ``Response``
   class. Please use ``Response.set_stream()`` or ``Response.content_length``
   instead.
-- ``Request.context_type`` was changed from dict to a subclass of dict.
-- ``Response.context_type`` was changed from dict to a subclass of dict.
+- ``Request.context_type`` was changed from dict to a bare class implementing
+  dict interface.
```
kgriffs: "implementing dict interface." ➡️ "implementing the mapping interface."

CHANGES.rst
```diff
@@ -67,8 +67,10 @@ Breaking Changes
...
+- ``Response.context_type`` was changed from dict to a bare class implementing
+  dict interface.
```
kgriffs: See above.

docs/user/faq.rst
```diff
@@ -617,19 +617,17 @@ now simply set attributes on the object:
    # Falcon 2.0
    req.context.cache_backend = MyUltraFastCache.connect()
 
-Since the default context type derives from dict, the existing code will work
-unmodified with Falcon 2.0. Nevertheless, it is recommended to change usage
-as outlined above since the ability to use context as dictionary may be
-removed in a future release.
+The new default context type emulates the dict interface in a way that context
```
kgriffs: It's probably OK to talk about the "dict interface" here instead of the "mapping interface", but you may still want to reword it for the sake of consistency. E.g.:

> The new default context type implements a dict-like mapping interface in a way...

falcon/util/structures.py
```diff
@@ -107,6 +107,126 @@ def __repr__(self):
...
+    def __hash__(self):
+        return hash(self.__dict__)
+
+    def __iter__(self):
+        return self.__dict__.__iter__()
+
+    def __len__(self):
+        return self.__dict__.__len__()
+
+    def __ne__(self, other):
```
kgriffs: Nitpick: It may improve readability to move `__ne__()` to sit below `__eq__()`.

falcon/util/structures.py
```diff
@@ -107,6 +107,126 @@ def __repr__(self):
...
+        if isinstance(other, type(self)):
+            return self.__dict__.__ne__(other.__dict__)
+        return self.__dict__.__ne__(other)
+
+    def __repr__(self):
+        return '{}({})'.format(type(self).__name__, self.__dict__.__repr__())
+
+    def __setitem__(self, key, value):
```
kgriffs: Nitpick: Group related methods in the implementation (e.g., colocate `__getitem__`,  `__setitem__`,  `__delitem__`).

falcon/util/structures.py
```diff
@@ -107,6 +107,126 @@ def __repr__(self):
...
+
+    def viewitems(self):
+        return self.__dict__.viewitems()
+
+    def viewkeys(self):
+        return self.__dict__.viewkeys()
+
+    def viewvalues(self):
+        return self.__dict__.viewvalues()
+
```
kgriffs: Looks like all the mapping interface methods are here. 👍

falcon/util/structures.py
kgriffs: It may also be helpful to add a `# PERF` comment explaining the overhead introduced by this implementation vs. using a `dict` directly. From a quick test on my MBP, it looks like the extra level of indirection adds about 200 ns, so probably not enough for anyone to notice.

falcon/util/structures.py
```diff
@@ -107,6 +107,126 @@ def __repr__(self):
...
+    Note:
+        Python 2 specific ``dict`` methods are exposed regardless of the
+        Python language version, however, as they are delegated to the
+        underlying ``__dict__``, a similar error would be raised as if
+        attempting to use these methods for a usual Python 3 dict.
+    """
+
+    def __contains__(self, key):
+        return self.__dict__.__contains__(key)
```
kgriffs: Using `self.__dict__` is quite elegant. 💯

falcon/util/structures.py
vytas7: Sure, can do. I started off in the alphabetic order, and it stuck that way.

falcon/util/structures.py
vytas7: Almost :slightly_smiling_face: I shied away from some purely syntactical ones, like `__format__`, since it doesn't seem to affect much.

falcon/util/structures.py
vytas7: Tbh the concern regarding `MutableMapping` is a good question, I remember I was looking at it briefly, and made a decision not to inherit from it, but haven't documented why. I'll recheck if there is any performance penalty associated to `MutableMapping`. If not anything else, yes (1), I didn't want to cement the fact that it is implementing the mapping interface as I wanted this to be viewed more as a bare class with convenience methods to ease migration.

falcon/util/structures.py
```diff
@@ -107,6 +107,126 @@ def __repr__(self):
...
+    >>> 'cache_strategy' in context
+    True
+
+    Note:
+        Python 2 specific ``dict`` methods are exposed regardless of the
+        Python language version, however, as they are delegated to the
+        underlying ``__dict__``, a similar error would be raised as if
+        attempting to use these methods for a usual Python 3 dict.
+    """
+
```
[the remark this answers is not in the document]
vytas7: Indeed, that was the main reason I took this route. I believe instantiating `Context` is still close to instantiating a bare class (and faster than the previous subclassing `dict` approach).

docs/user/faq.rst
```diff
@@ -617,19 +617,17 @@ now simply set attributes on the object:
...
+attributes are linked to dict items, i.e. setting an object attribute also sets
+the corresponding dict item, and vice versa. As a result, the existing code
```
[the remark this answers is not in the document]
vytas7: Absolutely, I happened to sneak this mistake into master in the previous PR.
I.e. it is "the existing code" in the existing code :smile:

tests/test_utils.py
```diff
@@ -741,3 +741,151 @@ def __call__(self, a, b):
...
+        assert copy == ctx
+        assert copy == {'foo': 'bar', 'details': None, 1: 'one', 2: 'two'}
+        copy.pop('foo')
+        assert copy != ctx
+
+        assert set(key for key in ctx) == {'foo', 'details', 1, 2}
+
+        assert ctx.get('foo') == 'bar'
+        assert ctx.get('bar') is None
+        assert not ctx.get('bar', False)
```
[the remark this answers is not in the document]
vytas7: It would... I actually have a vague memory of dismissing this as good enough when developing :blush:

Improved now.

falcon/util/structures.py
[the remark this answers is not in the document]
kgriffs: 👍

PR #1487, opened 2019-03-28
refactor(API): add a compatibility shim to support the Falcon<=1.4.1 error handler signature

vytas7:


falcon/api.py
```diff
@@ -562,6 +568,12 @@ def handle(req, resp, ex, params):
                                      'method named "handle" that is a '
                                      'member of the given exception class.')
 
+        # TODO(vytas): Remove this shimming in a future Falcon version.
+        arg_names = tuple(misc.get_argnames(handler))
+        if (arg_names[0:1] in (('error',), ('ex',), ('exception',)) or
```
kgriffs: Let's also add checks for 'err' and 'e'.

PR #1488, opened 2019-04-01
fix(testing): allow testing generator-based WSGI apps

### LINKED ISSUE #1479: Test client fails when WSGI application is a generator

vytas7:
Closes #1479

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,8 +67,10 @@ Breaking Changes
 - The deprecated ``stream_len`` property was removed from the ``Response``
   class. Please use ``Response.set_stream()`` or ``Response.content_length``
   instead.
-- ``Request.context_type`` was changed from dict to a subclass of dict.
-- ``Response.context_type`` was changed from dict to a subclass of dict.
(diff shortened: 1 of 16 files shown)


vytas7:
My attempt at fixing https://github.com/falconry/falcon/issues/1479, a competing PR to https://github.com/falconry/falcon/pull/1480.

I ripped off a fair part of `test_testing.py` from @tribals , and the co-authorship is attributed using `Co-authored-by`.

My approach is somewhat clunky, but the advantage is that there are (almost) no breaking changes.
Should we choose this approach I can write changelog items, too.

PR #1493, opened 2019-04-11
chore: Remove Python 2 support

### LINKED ISSUE #1395: Drop support for Python 2

### LINKED ISSUE #1465: Remove Python 2 jobs from tox, travis (and migrate pep8, to Python 3.7)
To Do

- [x] Coverage
- [x] Tox
- [x] Travis
- [x] Migrate pep8 tox job
- [x] Migrate pep8-examples tox job

kgriffs:
BREAKING CHANGE: This patch will cause apps to break when running under Python 2. Apps must upgrade to Python 3.5+ to use Falcon 3.0+.

Closes #1465
Closes #1395


CHANGES.rst
```diff
@@ -1,3 +1,27 @@
...
+=====
+
+Summary
+-------
+
+Changes to Supported Platforms
+------------------------------
+
+- Starting with Falcon 3.0, Python 2 is no longer supported. Therefore,
+  support for CPython 2.7 and PyPy2.7 was removed in this release.
```
vytas7: Maybe we could add a recommendation to use Falcon 2.x in case Python 2.7 support is desired?

falcon/util/uri.py
```diff
@@ -345,8 +261,7 @@ def parse_query_string(query_string, keep_blank=False, csv=True):
     is_encoded = '+' in query_string or '%' in query_string
 
     # PERF(kgriffs): This was found to be faster than using a regex, for
-    # both short and long query strings. Tested on both CPython 2.7 and 3.4,
-    # and on PyPy 2.3.
+    # both short and long query strings. Tested on CPython 3.4.
```
vytas7: Maybe we should retest on CPython 3.5+, otherwise we would making claims based on a Python version that is no longer supported by this code.

vytas7:
If we are doing everything in one sweep, we could remove extraneous (in the Python 3 sense) `# -*- coding: utf-8 -*-` in `falcon/api.py`, `falcon/bench/bench.py` and other files inside `bench`.

falcon/util/uri.py
kgriffs: Agreed. However, I think that belongs in a separate patch if a regex ends up being faster on 3.5+. Do you want to benchmark that?

kgriffs:
> * `CONTRIBUTING.md` is still advising to run `py27` benchmarks

Done

> * benchmarks locking Django to `<2.0` to support Python 2

Done

> * `falcon/response.py:952` mentions `dict.setdefault (tested on py27)`

I did some testing and have an alternative implementation ready, but I'd like to land that as a separate PR.

> * Docker files need to be cleaned up to remove Python 2 ones

Done

kgriffs:
> If we are doing everything in one sweep, we could remove extraneous (in the Python 3 sense) `# -*- coding: utf-8 -*-` in `falcon/api.py`, `falcon/bench/bench.py` and other files inside `bench`, while we are at it.

Done

falcon/util/uri.py
vytas7: Fine to do this in a separate PR. :+1:

kgriffs:
I just noticed that tools/build.sh also needs to be updated, but I'll get that in a separate patch.

PR #1496, opened 2019-04-19
perf(Response): Optimize _set_media_type()

kgriffs:


kgriffs:
I suppose we could remove it for Falcon 3.0. Other than dropping Python 2 support, hopefully the list of breaking changes is going to be much smaller anyway vs. 2.0.

kgriffs:
P.S. We should move and keep the perf comment though for posterity.

PR #1502, opened 2019-04-25
Fix for falcon unable to identify JSON

VariableDeclared:
In certain cases clients will send ('application/json; charset=UTF-8',)
causing mimeparse to except, and falcon to wrongly reject the request.

This attempts to fix this by stripping invalid characters, needs review.

### DIFF
--- a/falcon/media/handlers.py
+++ b/falcon/media/handlers.py
@@ -19,14 +19,13 @@ class Handlers(UserDict):
 
     def _resolve_media_type(self, media_type, all_media_types):
         resolved = None
-        prim_sanitised_media = media_type.strip(")'(")
 
         try:
             # NOTE(jmvrbanac): Mimeparse will return an empty string if it can
             # parse the media type, but cannot find a suitable type.
             resolved = mimeparse.best_match(
                 all_media_types,
-                prim_sanitised_media
+                media_type
             )
         except ValueError:
             pass
@@ -42,13 +41,13 @@ class Handlers(UserDict):
             return self.data[media_type]
         except KeyError:
             pass
-
+        sanitised_media = media_type.strip(")',(")
         # PERF(jmvrbanac): Fallback to the slower method
-        resolved = self._resolve_media_type(media_type, self.data.keys())
+        resolved = self._resolve_media_type(sanitised_media, self.data.keys())
 
         if not resolved:
             raise errors.HTTPUnsupportedMediaType(
-                '{0} is an unsupported media type.'.format(media_type)
+                '{0} is an unsupported media type.'.format(sanitised_media)
             )
 
         return self.data[resolved]
--- a/falcon/media/handlers.py
+++ b/falcon/media/handlers.py
@@ -20,12 +20,14 @@ class Handlers(UserDict):
     def _resolve_media_type(self, media_type, all_media_types):
         resolved = None
 
+        prim_sanitised_media = media_type.strip(")'(")
+
         try:
             # NOTE(jmvrbanac): Mimeparse will return an empty string if it can
             # parse the media type, but cannot find a suitable type.
             resolved = mimeparse.best_match(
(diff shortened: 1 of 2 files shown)


VariableDeclared:
Do not commit yet, not working, testing fix now.

kgriffs:
Could you provide a little more background on the type of client(s) that are submitting these non-standard media type strings? I'm just trying to get a feel for how common the use case is.

VariableDeclared:
Hi @kgriffs!

Sure, I bumped into this issue using cURL, not sure of the exact reason that this appeared.

I will have to check if I bumped into this using other clients, I think I may have bumped into it first on Chrome, but cannot remember specifics.

Cheers

nZac:
@VariableDeclared if you could help us to reproduce this we would be very thankful. 

The relevant RFC, I think, is [RFC 6838#4.2](https://tools.ietf.org/html/rfc6838#section-4.2) and [RFC 2045#5.1](https://tools.ietf.org/html/rfc2045#section-5.1). As you can see in 2045, the only invalid token from your example is `'`. `(),` are all valid portions of `tspecials`, so I do not think stripping them is what we want to do.

If we can reproduce this in a common HTTP client, then we will want to get that addressed there and for Falcons sake, `HTTPUnsupportedMediaType` is the correct action when sending incorrect media types.

VariableDeclared:
@nZac, Sure, thank you for making me aware of this. 

I will get back to you tomorrow when I am back to uni work.

Have a lovely day!

VariableDeclared:
OK So, went back and investigated, tested on the version of Falcon I had on my project (v2.0.0).

Running:
```
curl -XPOST -v -i http://localhost:8085/reporting/location -d '{
    "polygon": [
    {
        "lat": -34.19811476786663,
        "lng": 150.9571103515625
    },
    {
        "lat": -34.052600992487285,
        "lng": 150.7758359375
    },
    {
        "lat": -33.706003988404404,
        "lng": 151.3095022330342
    },
    {
        "lat": -33.77452274386239,
        "lng": 151.4688039908467
    },
    {
        "lat": -34.23639671005719,
        "lng": 150.96502005419552
    }
    ]
}'
```
Results in the error shown in the terminal screenshot below illustrate the issue.
image

So I'm not sure whether this is the correct response or not, it happened running with the `-H "Content-Type: application/json"` and without. It would seem to me to be a bug of sorts.

Cheers.

nZac:
@VariableDeclared, I am still not sure this is Falcon... here is a bare minimum example:

The Falcon App:
```python
import falcon

class Resc:
    def on_post(self, req, resp):
        resp.media = req.media

app = application = falcon.API()
app.add_route('/', Resc())
```

Start a simple server:
```bash
$ pip freeze | grep falcon
falcon==2.0.0
$ gunicorn myapp:app
[2019-05-20 00:10:48 -0500] [29615] [INFO] Starting gunicorn 19.9.0
[2019-05-20 00:10:48 -0500] [29615] [INFO] Listening at: http://127.0.0.1:8000 (29615)
[2019-05-20 00:10:48 -0500] [29615] [INFO] Using worker: sync
[2019-05-20 00:10:48 -0500] [29656] [INFO] Booting worker with pid: 29656
```

Send Request
```sh
# Curl Version
$ curl -V
curl 7.54.0 (x86_64-apple-darwin18.0) libcurl/7.54.0 LibreSSL/2.6.5 zlib/1.2.11 nghttp2/1.24.1

# Send the request 
$ curl -XPOST -H "Content-Type: application/json" localhost:8000 -d '{}'
{}%
```

Does this work for you? If it does, then something else in your app is causing the bad content type.

VariableDeclared:
Closing pending further research as I don't have the time right now. :smile:

VariableDeclared:
Closing pending further research as I don't have the time right now. smile

PR #1512, opened 2019-05-06
added support to RFCs 2518 & 4918 WEBDAV METHODS

### LINKED ISSUE #1426: WebDAV methods not supported
Falcon defines supported HTTP methods in `falcon/constants.py`: supported are "usual" `HTTP_METHODS` and, in addition to that, `WEBDAV_METHODS`. However, only WebDAV versioning extension methods from RFC 3253 are supported, but not the "ordinary" WebDAV ones (i.e. from RFCs 2518 & 4918) like `COPY`, `LOCK`, `MKCOL`, `MOVE` etc.

Supporting only an extension, but not the core upon which that extension builds looks somewhat inconsistent.

screamingskulls:
Fixes #1426

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -389,6 +389,8 @@ Fixed
   compliance with RFC 7232.
 - Fixed ``TypeError`` being raised when using Falcon's testing framework
   to simulate a request to a generator-based WSGI app.
+- Falcon now supports all WebDAV methods (RFC 2518 and RFC 4918), such as
+  COPY, LOCK, MKCOL, MOVE, PROPFIND, PROPPATCH and UNLOCK.
 
 Contributors to this Release
 ----------------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -433,6 +433,7 @@ Many thanks to all of our talented and stylish contributors to this release!
 - `steven-upside <https://github.com/steven-upside>`_
 - `tribals <https://github.com/tribals>`_
 - `vytas7 <https://github.com/vytas7>`_
+- `screamingskulls <https://github.com/screamingskulls>`_
 
 1.4.1
 =====
--- a/falcon/constants.py
+++ b/falcon/constants.py
@@ -11,18 +11,25 @@ HTTP_METHODS = (
     'TRACE',
 )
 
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -431,6 +433,7 @@ Many thanks to all of our talented and stylish contributors to this release!
 - `steven-upside <https://github.com/steven-upside>`_
 - `tribals <https://github.com/tribals>`_
 - `vytas7 <https://github.com/vytas7>`_
+- `screamingskulls <https://github.com/screamingskulls>`_
```
vytas7: This list was specific to Falcon 2.0, we should create a new one under 3.0.

CHANGES.rst
```diff
@@ -389,6 +389,8 @@ Fixed
   compliance with RFC 7232.
 - Fixed ``TypeError`` being raised when using Falcon's testing framework
   to simulate a request to a generator-based WSGI app.
+- Falcon now supports all WebDAV methods (RFC 2518 and RFC 4918), such as
```
vytas7: * We should move this note to the 3.0.0 section. 2.0 has already been released.
* We should also add this note to `3.0.0` RST file.

falcon/constants.py
```diff
@@ -11,19 +11,26 @@
...
     'REPORT',
     'UNCHECKIN',
+    'UNLOCK'
     'UPDATE',
     'VERSION-CONTROL',
 )
 
 COMBINED_METHODS = HTTP_METHODS + WEBDAV_METHODS
 
-
```
vytas7: Any specific reason for re-styling whitespace here?
(Or just trying to make a compelling case for `black` :wink: )

CHANGES.rst
nZac: Actually, we need a new `docs/_newsfragment/` for this. I would think this should be classified as a bug fix.

vytas7: Now with the Towncrier setup, yes :slightly_smiling_face: 
Agreed wrt bug fix.

vytas7:
Closing this as it has been superseded by https://github.com/falconry/falcon/pull/1588

PR #1522, opened 2019-05-06
feat(issue-1503) Support API users defining and using custom HTTP methods

seanharrison:
* adds support for an environment variable, FALCON_CUSTOM_HTTP_METHODS, which if defined is a comma-delimited string with custom HTTP methods that the app should support.
* add tests, which simulate this using the other method that works, which is to add these method names to the falcon.constants.COMBINED_METHODS.
* add documentation for using either method at the end of docs/api/routing


sharrisoncaktus:
Glad to pitch in! It was great working with you guys, hope to continue it as we go along.

Issue #1514, opened 2019-05-06
Attempt ordering exception handlers by exception class inheritance


vytas7:
At the time of writing, exception handlers are matched in LIFO order, i.e. when searching for an error handler to match a raised exception, and more than one handler matches the exception type, the framework will choose the one that was most recently registered.

In case one has many error handlers for classes potentially inheriting from each other, this may become very error-prone and confusing in case a wrong handler swallows the exception.

Investigate if it is possible to order exceptions statically as they are being added (not at the time of catching).

Prior art: Flask/Werkzeug claim to handle this; investigate how it is done there.

csojinb:
Since this is relevant to the other work I did with error-handling, I can dig into werkzeug and try to figure out what's going on there. Don't promise to finish said research though 😅

csojinb:
[Here](https://github.com/pallets/flask/blob/1dda032b00cb4f7a783e018a7c370ee88f15e797/flask/app.py#L1678-L1682) are the relevant lines in Flask for using the most specific handler available. They have some additional levels of categorization for blueprints, or storing handlers by code instead of exception class, which don't seem relevant for Falcon.

So, basically, the error handlers are stored in a dict of `{<exception class>: <handler callable>}`. Then they iterate through `<raised exception>.__mro__` (which produces a tuple of the class hierarchy from least to most general) until a match is found in the error handler dict.

If y'all decide to do this, it seems like it should be pretty straightforward.

csojinb:
@vytas7 I might be misunderstanding the issue here... what is the problem with having handlers associated with more than one exception? Can't you just store that as `{Exception1: handle_exceptions_1_and_2, Exception2: handle_exceptions_1_and_2}`?

vytas7:
@csojinb No, there is no problem per se, that is what my suggestion #2 is basically aiming at.
You could store expand that to handlers in a way you depict, however, if possible, for performance reasons it would be good to keep exceptions grouped in case they sort equal on the inheritance scale.

kgriffs:
Where did we end up on this issue? Is https://github.com/falconry/falcon/pull/1527 still waiting on this?

vytas7:
@kgriffs this is, strictly speaking, not blocking #1527 .
But it would make the implementation more straightforward and consistent, but not having to care the default bare handler is invoked last.

kgriffs:
OK, I'd love to have this smarter ordering implemented for 3.0 if we can swing it, otherwise push to 3.1.

[...]

Issue #1507, opened 2019-05-05
Add a generic error handler that returns the application's default media type


jmvrbanac:
If a generic Python exception is raised while running in gunicorn. The user will see an HTML Internal Server Error even though the default media type is `application/json`. This is due to the exception getting raised to the WSGI server which is defaulting the output. We should gracefully handle this with the default media type. According to the the WSGI spec applications should `In general, applications should try to trap their own, internal errors, and display a helpful message in the browser.`

Tasks:

1. Add a default handler that can be overridden
2. Output a notice and hint to wsgi.errors / stderr?
3. ~Add Recipe in docs~ Add to documentation about adding error handling.

csojinb:
I am taking a look at this

csojinb:
@jmvrbanac for #3 -- do you mean a recipe for overriding the handler?

jmvrbanac:
@csojinb Sorry, I should've called that "example" or something a bit clearer instead. The idea is that this is something people end up doing more often than we thought, so just having a code example showing how a user would go about doing this.

vytas7:
Resolved in https://github.com/falconry/falcon/pull/1527 [Merged]

PR #1527, opened 2019-05-06
fix(API): return 500 response to server for uncaught exceptions

csojinb:
Per newly-understood WSGI spec, applications should trap internal errors instead of passing them on to the server. Add a default exception handler for the base Python ``Exception``, which is placed first in the error handler list (and is thus evaluated last, per LIFO operation).

The 500 error response sends no specific error information back to the client, but it does log a traceback to ``wsgi.errors``. This logging behavior currently has no tests.

Update error-handling documentation to reflect new default behavior. No "recipe" is included as yet -- need more details on requirements here.

Currently, errors are serialized in JSON or XML, depending on the Accept header in the request. It defaults to JSON if the request does not prefer XML, even if the app's default media type is XML. Changing this behavior appears to be non-trivial, so it is not attempted here, though it was discussed during development.

Addresses https://github.com/falconry/falcon/issues/1507


falcon/api.py
```diff
@@ -186,6 +187,7 @@ def __init__(self, media_type=DEFAULT_MEDIA_TYPE,
         self.resp_options.default_media_type = media_type
 
         # NOTE(kgriffs): Add default error handlers
+        self.add_error_handler(Exception, self._python_error_handler)
```
jmvrbanac: I'm wondering if this should be `Exception` or `BaseException`... @kgriffs thoughts?

vytas7: I would suggest just _Exception_. The base one could make the WSGI app catch stuff like system exit, keyboard interrupt / Ctrl+C etc, and return JSON in those cases. That doesn't sound like safe default behaviour, but maybe we document how to, in case anyone still wants that thing.

csojinb: So, are we okay with this as-is?

falcon/api.py
```diff
@@ -610,6 +613,11 @@ def set_error_serializer(self, serializer):
             property value, and it may choose to override the
             representation logic.
 
+        Note:
```
csojinb: https://github.com/falconry/falcon/blob/f078b2e6f1d54100f097b2fbb560a2b56fe7d922/falcon/api.py#L616-L619

This is not directly related to the changes in this PR, but I didn't see that this specific interaction was documented, so it seemed worth including.

csojinb:
So, the hug tests are failing on this. Do I need to make a PR to there before this can be merged?

csojinb:
Okay, so a brief scan of the test output suggests that a number of hug tests are depending on the behavior where falcon passes out I caught errors. 

I think that I could make a PR to hug that would override the now-default (in this PR) behavior to catch those exceptions for the purposes of their tests. (Hopefully I can do that with minimal disruption.) However, that does create the annoying situation where the default HTTPError and HTTPStatus handlers are also overridden, which is just kinda gross.

So, I’m thinking maybe it would make sense to resolve https://github.com/falconry/falcon/issues/1514 first and then come back to this PR? I looked into how werkzeug does it, and it’s super-simple, so I think I could change it pretty quickly. Catching all of the references to LIFO order in the docs seems like the hardest part :joy:. That issue is still in “Needs Decision” status, but I think it would be a significant usability win (assuming there are no objections to storing the error handlers in a dict).

falcon/api.py
jmvrbanac: Yeah, we're good with `Exception`, my bad.

kgriffs:
Perhaps we should disable the hug test temporarily and open an issue over there? We can re-enable the hug tests in 3.1 once all of Falcon's breaking changes have been addressed. @timothycrosley what would you suggest?

timothycrosley:
@kgriffs I think that sounds good! Once Falcon approaches a release candidate for the next version I'll start working on resolving any breaking issues that arise. Once I feel like hug is in good shape to support the new version I can make a pull request to uncomment the tests.

vytas7:
Thanks for input @timothycrosley !
And :+1: from me for this path, we are probably going to hit more breaking changes on the road to 3.0 anyway.

Bandwidth permitting, I will try to submit a Hug PR if the changes are trivial (as I did for Falcon 2.0), otherwise we are going to need your help @timothycrosley :slightly_smiling_face:

kgriffs:
OK, sounds like a plan. I will disable the hug test for the moment.

kgriffs:
Depends on #1599

docs/user/faq.rst
```diff
@@ -446,9 +446,14 @@ for this exception type. Alternatively, you may be able to configure your web
 server to transform the response for you (e.g., using Nginx's ``error_page``
 directive).
 
-500 errors are typically the result of an unhandled exception making its way
-up to the web server. To handle these errors more gracefully, you can add a
-custom error handler for Python's base :class:`Exception` type.
+By default, Falcon handles all exceptions not inheriting from
```
vytas7: Technically, _all exceptions_ is not 100% correct since we are not installing any handler for `BaseException` (neither should we). But probably we don't need to dig that deep in the FAQ, this will do :+1:

csojinb: Fair point. I have to fix the commit message style anyway, so I can go make this technically correct

vytas7: Oh, sorry for stepping on your toes... I have hopefully addressed Sphinx complaints via the Web UI

csojinb: No worries :-)

vytas7: And btw don't worry about the commit message style. We maintainers get an extra chance to edit that when squash-merging a pull request. (But we seldom get it right ourselves...)

csojinb: Okay, I'm just double-checking that the docs are formatted how I wanted, and then I'll push and we should be good to go

csojinb:
Okay @vytas7, I modified the documentation to be more technically correct re: default error-handling. If it all looks good to you, then I'm happy with it.

Issue #1534, opened 2019-05-07
Middleware process_response function is missing parameter


pbjr23:
The `process_response` function used in the middleware in the [readme file](https://github.com/falconry/falcon/blob/master/README.rst) and on the [quickstart page](https://falcon.readthedocs.io/en/stable/user/quickstart.html) is missing the `req_succeeded` parameter, which was introduced in Falcon 2.0.0.

It's currently written as `def process_response(self, req, resp, resource):`, but should be `def process_response(self, req, resp, resource, req_succeeded):` (see [current middleware documentation](https://falcon.readthedocs.io/en/stable/api/middleware.html)). This is currently causing the test example to not function correctly, returning this exception: `TypeError: process_response() takes 4 positional arguments but 5 were given`

PR #1535, opened 2019-05-07
docs: fix middleware function example

pbjr23:
Fixes https://github.com/falconry/falcon/issues/1534

* Update all of the `process_response` functions to use the correct arguments
* Move second examples to examples directory so it can be tested


xzycn:
👍 I have just encountered the problem.

PR #1526, opened 2019-05-06
docs(readme.rst): clean first things.py example

Ciemaar:
* update comments for things.py to match expected use
* extract things.py from readme.rst to examples/things.py
* create examples/requirements.txt

Partially addresses https://github.com/falconry/falcon/issues/1423

PR #1536, opened 2019-05-07
tests(examples/things.py): tests for first things.py example

Ciemaar:
* tests for example code
* reconfigure flake8 to make this possible

Testing for #1423


Ciemaar:
This PR is based off and will require https://github.com/falconry/falcon/pull/1526 to be practically merged.

tests/test_things_example.py
```diff
@@ -0,0 +1,22 @@
+from unittest.mock import Mock
+
+from examples.things import app, things, ThingsResource
+import falcon
+
+
+def test_app():
+    assert app
+    (resource, method_map, params, uri_template) = app._router.find('/things')
```
jmvrbanac: It would probably be cleaner to do: `resource, *_ = app._router.find('/things')`

tests/test_things_example.py
```diff
@@ -0,0 +1,22 @@
...
+def test_things():
+    assert isinstance(things, ThingsResource)
+
+
+def test_ThingsResource_on_get():
+    test_thing = ThingsResource()
+    mock_response = Mock()
+    test_thing.on_get(None, mock_response)
+    assert mock_response.status == falcon.HTTP_200
+    assert len(mock_response.body) > 10
```
jmvrbanac: We should probably assert the actual body rather than just the length.

vytas7:
@Ciemaar `flake8` was complaining about your docstring-style string, not sure what it was expecting, so I went on to convert it to strings inside parenthesis.

tests/test_things_example.py
```diff
@@ -0,0 +1,25 @@
...
+    assert app
+    resource, *_ = app._router.find('/things')
+    assert resource is things
+
+
+def test_things():
+    assert isinstance(things, ThingsResource)
+
+
+def test_ThingsResource_on_get():
```
nZac: Hi @Ciemaar, thanks for contributing! Tests like these are often neglected and really important for our users.

You might find this a simpler way to write tests for resources:


```python
import falcon.testing as testing


# (Optional) First create a pytest fixture which returns a TestClient which 
# can be used to send requests to the passed application
@pytest.fixture()
def client():


# add the fixture name, in this case "client" as an argument to the test. 
# pytest will see that and inject the fixture you created above.
#
# NOTE: you can create the `client` in the test if you want but a 
# fixture lends itself to better code reuse.
def test_things_resource_response(client):
    # client = testing.TestClient(app)
    # Then use the client to simulate a request to the api.
    resp = client.simulate_get('/things')

    # Then you can assert as you did before.
    assert resp.status_code == 200
    assert resp.text == (
        '\nTwo things awe me most, the starry sky above me and the moral law within me.'
        '\n\n    ~ Immanuel Kant\n\n'
    )
```

tests/test_things_example.py
```diff
@@ -0,0 +1,25 @@
+from unittest.mock import Mock
+
+from examples.things import app, things, ThingsResource
+import falcon
+
+
+def test_app():
```
nZac: If testing like I describe below, I don't think this adds substantial value. We  will test the routing is connected properly by `assert`ing on the result of the simulated request.

Issue #1216, opened 2018-02-02
Processing empty req.media results in 400


alysivji:
The following code snippet results in a 400 when making a POST request with an empty body.

```python
import falcon
import logging

logger = logging.getLogger()


class QuoteResource:
    def on_post(self, req, resp):
        """Handles GET requests"""
        quote = {
            'quote': (
                "I've always been more interested in "
                "the future than in the past."
            ),
            'author': 'Grace Hopper'
        }

        logger.info(req.media)

        resp.media = quote

api = falcon.API()
api.add_route('/quote', QuoteResource())
```

I would expect `req.media` to return an empty dictionary if the body is empty.

vytas7:
An empty body is, strictly speaking, not a valid JSON representation.

I do not think an empty dict is fitting in this scenario. However, one could maybe consider making ``req.media`` reduce to ``None`` in this case, as ``test.Request.json`` will start behaving in 2.0 (see the related discussion in, for instance, https://github.com/falconry/falcon/issues/1205 )

On the other hand, as said above, the correct way to serialize ``None`` in JSON is ``null``, not an empty string:

```
>>> json.loads('')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

An unrelated note: although YAML is *almost* a superset of JSON, it does treat empty payload differently:

```
>>> yaml.load('') is None
True
```

foresmac:
I think there are a few ways to maybe change some behavior and make the framework a little more flexible in dealing with incoming request data. A solution here might use one or more of the following:

1. Don't allow/rely on the handler to raise `BadRequest`
1. set `req.media` to some consistent value (e.g. `None`) when the media handler can't return valid results.
1. Let the resource handler explicit raise Bad Request based on the value of `req.media`.
[...]

PR #1537, opened 2019-05-07
feat(media): add URL-encoded form data handler

vytas7:
~~Just dropping a quick sketch from PyCon sprints, this still a bit rough on edges.~~

We could probably polish documentation & recipes etc, but otherwise this is ready to review as in proposal for inclusion in Falcon 3.0 alpha1.


vytas7:
Just dropping a quick sketch from PyCon sprints, this still a bit rough on edges.

tests/test_media_urlencoded.py
```diff
@@ -0,0 +1,71 @@
...
+    ({'number': [1, 2]}, 'number=1&number=2'),
+])
+def test_urlencoded_form_handler_serialize(data, expected):
+    handler = media.URLEncodedFormHandler()
+    assert handler.serialize(data, falcon.MEDIA_URLENCODED) == expected
+
+
+class MediaMirror:
+
+    @classmethod
```
jmvrbanac: Perhaps I'm overlooking something, but why do we need the `classmethod`?

vytas7: No specific reason tbh, it could well be a `staticmethod` or a normal class method with `self`. I think I was planning to subclass it, but that part never got finished.

I can change to a normal method.

mikeylight: @vytas7 I'd prefer to use `staticmethod` decorator here, it recommended to use for functions when `self` haven't used explicitly.

IMHO.

mikeylight: @vytas7 I'd prefer to use `staticmethod` decorator here, it recommended to use for functions when `self` haven't used explicitly.

IMHO.

vytas7: @mikeylight it's a matter of taste, if you use class variables (which we don't do here), I prefer to define a `@classmethod`. I'll change this one to a normal method with `self` though not to complicate the matters too much.

vytas7:
@nZac it is so shockingly simple because we already had that functionality, albeit in a different form, and because it **is** simple :slightly_smiling_face:
Now, `multipart/form-data` is a completely different beast... and a really ugly one. I'm still wading through RFCs, and polishing something that at least could be demoed interface and functionality wise, although full correctness and improved performance are going to be challenging.

falcon/media/urlencoded.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # NOTE(kgriffs): According to http://goo.gl/6rlcux the
+        # body should be US-ASCII. Enforcing this also helps
+        # catch malicious input.
+        body = body.decode('ascii')
+
+        # NOTE(vytas): Due to the Request.media property mechanics, this path
+        # is normally never hit.
+        if not body:
+            return None
```
ch3pjw: Would it be more type-correct to return `{}`?

ch3pjw:
This looks like it would be super-useful for a project I'm working on right now! Thanks!

falcon/media/urlencoded.py
vytas7: No, honestly, I think I disagree an empty dict is any more correct. JSON media can also be a JSON list, a JSON string or any other JSON-serializable document. Unfortunately I don't have a one-size-fits-all suggestion...
See also the discussion on the same matter here: https://github.com/falconry/falcon/issues/1216

In the code highlighted above, though, I am just cementing the fact how Falcon processes media now. As stated in the comment, this path is normally never hit when using this (or other) handler due to it being short-circuited before the handler even gets a chance to see the request media.

vytas7: I'm actually thinking to go ahead and create an issue for this.
You're in principle right @ch3pjw -- an empty URL-encoded form string should deserialize to an empty dict (while I do *not* think it should be the default behaviour for JSON).

vytas7:
Implements #1580

falcon/media/urlencoded.py
kgriffs: @vytas7 If we are to proceed as I proposed in #1589, you could just go ahead and make the change now to return `{}`.

vytas7: Will do.
Another alternative is just to remove this check altogether and let the `parse_query_string` handle this, considering it is not a very common case for submitted media?

kgriffs: Good point. Letting `parse_query_string()` handle it would require a few more CPU cycles, but given that it should not be a common case, the extra overhead isn't worth bothering about.

vytas7: Uh oh. I'll have to short circuit for now anyway to keep the things moving... :grimacing: 
And file this as a new issue I guess:
```python
>>> from urllib import parse
>>> from falcon.util import uri
>>> parse.parse_qs('', keep_blank_values=True)
{}
>>> uri.parse_query_string('', keep_blank=True)
{'': ''}
```

vytas7:
Thanks @kgriffs , merging then!

I'll create a follow-up PR to promote this to default handlers (if we want that?) and to deprecate the old parameter merge approach as per #1580

kgriffs:
Yes, I think we want this in default handlers. People can elect not to use it by avoiding req.media - just need to make sure we document that it is enabled by default.

vytas7:
(What is more, people can also cull it from default handlers by creating a new set of handlers by their own. So :+1: :arrow_double_up:  )

PR #1540, opened 2019-05-08
fix(custom-http-methods): Making sure we clean up the env var input string

jmvrbanac:
Making sure we're strip and uppercase the custom methods specified from the environmental variable. Also, rewording docs a touch, and expanding the tests.

### DIFF
--- a/falcon/constants.py
+++ b/falcon/constants.py
@@ -26,8 +26,8 @@ WEBDAV_METHODS = [
 # if FALCON_CUSTOM_HTTP_METHODS is defined, treat it as a comma-
 # delimited string of additional supported methods in this env.
 FALCON_CUSTOM_HTTP_METHODS = [
-    method for method in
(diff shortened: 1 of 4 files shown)


falcon/constants.py
```diff
@@ -26,8 +26,8 @@
 # if FALCON_CUSTOM_HTTP_METHODS is defined, treat it as a comma-
 # delimited string of additional supported methods in this env.
 FALCON_CUSTOM_HTTP_METHODS = [
-    method for method in
-    (os.environ.get('FALCON_CUSTOM_HTTP_METHODS') or '').split(',')
+    method.strip().upper()
+    for method in os.environ.get('FALCON_CUSTOM_HTTP_METHODS', '').split(',')
     if method != ''
```
vytas7: This check could be somewhat more robust in the spirit of the above, so like `if method.strip() != ''`

PR #1549, opened 2019-06-01
feat(media): implement multipart form media handler

vytas7:
@kgriffs @jmvrbanac @nZac and others in the community who are following the development of this feature ( @sim0nx ):
this is still rough on edges when it comes to documentation, but the interface and the implementation are hopefully alpha-level ready.

I have also created an extended test/benchmark suite for tuning performance as well as doubling as extra validation for code correctness: https://github.com/vytas7/falcon-multipart-tests

There is still one notable case of many small reads from the part stream which I stopped short of optimizing. The machinery is largely ready, I could just force reads to be aligned on the part stream. But I would be happy to get feedback from the community how the functionality is used in the wild in order to get a better impression whether we need to focus on this specific usage pattern.

**The Gist** illustrating the interface and the demo therein are still valid: https://gist.github.com/vytas7/34c60e5ac3a4bc2f2eb0af2428d77003 (I have actually just slightly updated it).

I have also started implementing **async multipart form parsing for ASGI** (i.e. marrying this to async media support in https://github.com/falconry/falcon/pull/1573).
This class could probably be used verbatim for ASGI from #1573, but better don't even try it at home. In the best case, your file would be buffered in memory in a bytes buffer.


kgriffs:
@vytas7 For community members who are following this work, could you summarize what is left to do for this feature? Could the async support be done in a followup patch to allow us to cut a 3.0 alpha sooner, or would that not be very much extra work?

vytas7:
@kgriffs I'm actively working on it again. I hope to provide an update soon.

vytas7:
Implements #953

falcon/cyutil/streams.pyx
```diff
@@ -0,0 +1,312 @@
...
+        cdef Py_ssize_t read_amount
+
+        amount = int(amount)
+        if amount < 0 or amount > self._chunk_size:
+            amount = self._chunk_size
+
+        if self._buffer_len - self._buffer_pos < amount:
+            read_amount = (self._chunk_size - self._buffer_len +
+                           self._buffer_pos)
+            if read_amount >= self._max_bytes_remaining:
```
mikeyusko: @vytas7 Hi, an amazing PR!

I have one more question, maybe in the future would be better to split `if/if/else` blocks to 3 different private methods of an instance?

e.g
```
if read_amount >= self._max_bytes_remaining:
    self._some_cool_method()
if self._buffer_pos == 0:
   self._some_cool_method2()
else:
   self.some_cool_method3()
```

What do you think?

mikeyusko: @vytas7 Understood, in that case it make sense.

vytas7:
Superseded by a new pull request since I have no idea how to resolve the bogus conflicts above (the Git command line shows no conflicts): https://github.com/falconry/falcon/pull/1590

Issue #1556, opened 2019-06-13
Include "SameSite" keyword parameter option for the set_cookie() method


pushshift:
There is a somewhat new option for cookies to have the SameSite parameter set which helps protect against CSRF attacks (More information here: https://www.netsparker.com/blog/web-security/same-site-cookie-attribute-prevent-cross-site-request-forgery/)

Adding this keyword option for the set_cookie method would be a great addition.

pushshift:
Thanks @kgriffs for the fast action on this! Much appreciated.

5onic:
@kgriffs @vytas7 Is someone working on it? otherwise I can pick up this issue.

vytas7:
@5onic I don't think anyone is working on this. You're welcome to grab it :+1:

vytas7:
Implemented in https://github.com/falconry/falcon/pull/1557

PR #1557, opened 2019-06-27
Adding SameSite keyword

mikeyusko:
This is PR for #1556 

Added `same_site` keyword to `Response.set_cookie` method.


5onic:
@vytas7 this PR is still in the WIP stage.

Also, I have one more question - Should we handle [here](https://github.com/falconry/falcon/pull/1557/commits/9893e15b2906a61ba25d6a42621b3e9994e74417#diff-4a6833e063be95e9d02ddb25e9a8d8d3R428) whether a value is equal to `Lax` or `Strict`? because my God tells me that, we need to do it, and raise an error if a value wasn't equal.

falcon/response.py
```diff
@@ -279,7 +279,7 @@ def set_stream(self, stream, content_length):
         self._headers['content-length'] = str(content_length)
 
     def set_cookie(self, name, value, expires=None, max_age=None,
-                   domain=None, path=None, secure=None, http_only=True):
+                   domain=None, path=None, secure=None, http_only=True, same_site=None):
```
RonRothman: Please document the `same_site` parameter in the docstring. Without that, it's not clear what type it expects.

falcon/response.py
```diff
@@ -425,6 +425,9 @@ def set_cookie(self, name, value, expires=None, max_age=None,
         if http_only:
             self._cookies[name]['httponly'] = http_only
 
+        if same_site:
+            self._cookies[name]['samesite'] = same_site[0]
```
RonRothman: Why did you choose to make `same_site` a tuple instead of just a string?

5onic: Yes, it's my mistake. We can pass it as a string, but the main question is - do we need to raise an error if a string doesn't equal to `Lax` or `Strict`?

falcon/response.py
5onic: Yeah, a PR isn't complete, will do.

falcon/response.py
RonRothman: > do we need to raise an error if a string doesn't equal to Lax or Strict?

No. Note that no other parameters in this method are validated.

mikeyusko:
@RonRothman @vytas7 Sorry for extremely delay with that Pull request, I plan to fix the review comments in 2-3 days. After that it will be ready for code review.

mikeyusko:
Hm, I can't understand why I the PR failed, on my local machine everything were passed.

vytas7:
@mikeyusko Tests seem to fail on all Python environments because of quite mundane reasons.
Is it possible that you have some uncommitted fixture or tweak in your local checkout that helps in passing tests, or some local library on your system? Could you take a look?

mikeyusko:
@vytas7 Ok, I did an investigation, and unfortunately, we will able to add `SameSite` cookie since python3.8 and later.

The reason is - we're using a built-in `http` package, and the cookie was introduced in the python3.8 version.

```
Python 3.7.2 (v3.7.2:9a3ffc0492, Dec 24 2018, 02:44:43)
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from http import cookies
>>> cookies.Morsel._reserved
{'expires': 'expires', 'path': 'Path', 'comment': 'Comment', 'domain': 'Domain', 'max-age': 'Max-Age', 'secure': 'Secure', 'httponly': 'HttpOnly', 'version': 'Version'}
```

```
Python 3.8.0a1 (v3.8.0a1:e75eeb00b5, Feb  3 2019, 12:33:49)
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from http import cookies
>>> cookies.Morsel._reserved
{'expires': 'expires', 'path': 'Path', 'comment': 'Comment', 'domain': 'Domain', 'max-age': 'Max-Age', 'secure': 'Secure', 'httponly': 'HttpOnly', 'version': 'Version', 'samesite': 'SameSite'}
>>>
```

inirudebwoy:
Hi @mikeyusko , 
There is a new option coming for `SameSite` according to this article https://blog.chromium.org/2019/10/developers-get-ready-for-new.html

AFAIK your code will support it but the docstring needs to be corrected.

tests/test_cookies.py
```diff
@@ -30,7 +30,7 @@ def dst(self, dt):
 class CookieResource:
 
     def on_get(self, req, resp):
-        resp.set_cookie('foo', 'bar', domain='example.com', path='/')
+        resp.set_cookie('foo', 'bar', domain='example.com', path='/', same_site=('Lex',))
```
vytas7: Are you explicitly testing an unsupported value (`'Lex'`) here?

The supported values are [`'Lax'` and `'Strict'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie). It is a good idea to add some tests for unsupported values too, but maybe we should add a comment so the reader wouldn't have to guess whether it is a misspelling from the Falcon authors or not.

falcon/response.py
vytas7: As @RonRothman points out, let's add documentation for public methods.
I'm reiterating the obvious here but I do so that these stay as requested changes.

It would also be cool to link to the relevant RFC section as docs for other parameters do.

falcon/response.py
vytas7: I second @RonRothman on the parameter type, a string parameter sounds easier for the API users unless there are use cases where a list might be useful?

I'm admittedly a bit divided on the validation question. @kgriffs  thoughts?
I would say we do actually validate some parameters, for instance, `max_age` is cast to `int` etc. Now that we are on Python 3, defining an `enum` with supported values is another option to consider. OTOH we are not using `enum`s anywhere else in Falcon (at least not yet).
NB if we decide to validate anything, we could also consider supporting the experimental `'None'` value too.

falcon/response.py
```diff
@@ -425,6 +432,9 @@ def set_cookie(self, name, value, expires=None, max_age=None,
         if http_only:
             self._cookies[name]['httponly'] = http_only
 
+        if same_site:
+            self._cookies[name]['samesite'] = same_site
```
vytas7: I'm admittedly a bit divided on the validation question. @kgriffs  thoughts?
I would say we do actually validate some parameters, for instance, `max_age` is cast to `int` etc. Now that we are on Python 3, defining an `enum` with supported values is another option to consider. OTOH we are not using `enum`s anywhere else in Falcon (at least not yet).
NB if we decide to validate anything, we could also consider supporting the experimental `'None'` value too.

mikeyusko: The easiest way to check if a value is correct.
e.g
```
if same_site and same_site in {'Lax', 'Strict', 'None'}
```
otherwise, skip it, that's it;)

RonRothman: `...same_site.lower() in {`lax', 'strict', 'none`}` - the value is case insensitive, right?

mikeyusko: if the value will be case insensitive, we will need to do something like your wrote.

It will allow us, to set value in different ways.

e.g
```
resp.set_cookie('samesite', 'Lax')
# OR
resp.set_cookie('samesite', 'lax')
```

I like it.

kgriffs: I don't think validation will hurt from a performance standpoint and is a simple thing to do. An enum may provide some benefits from a type-checking perspective if/when we add type hints to the framework, but on the other hand using an enum would require few more keystrokes from the developer vs. a simple string, and then we would have to spend a few extra cycles serializing out the enum when setting the cookie attribute.

Regardless, I think we really should do validation because a typo in this case can impact security (considering that the user agent maps unknown values to the "None" behavior).

My preference would be to keep the `str`-based approach (don't use an enum) and do the validation as proposed above, e.g.:

```py
same_site.lower() in {'lax', 'strict', 'none'}
```

tests/test_cookies.py
```diff
@@ -151,6 +157,7 @@ def test_cookie_expires_naive(client):
     assert cookie.value == 'bar'
     assert cookie.domain is None
     assert cookie.expires == datetime(year=2050, month=1, day=1)
+    assert cookie.same_site == 'Strict'
```
kgriffs: Please also add a couple of tests to demonstrate case-insensitivity (i.e., that the framework does not require case-sensitive values be used when setting the attribute).

falcon/response.py
RonRothman: I like the enum idea, since it obviates the need to do any value checking. With respect to time spent serializing the enum to string (which is effectively just a dict lookup under the hood), that would be offset at least somewhat by the avoidance of explicit string value checking.

RonRothman: Also, if we're really trying to squeeze every last cycle out of the code, then we might find that the lookup is faster in a tuple than in a set, with such a small N. Worth measuring?
```
same_site.lower() in ('lax', 'strict', 'none')
```

mikeyusko: I, unfortunately, disagree here, I don't think so that we need `enum` module.

This is only one place where we do this kind of check if I remember correctly.

And the main reason - some people will be wondering why we are using this kind of enum, but this is just a check.

IMO.

vytas7: @RonRothman Lookup is fastest on `frozenset`s and `set`s if you cache them in a global variable. Benchmarked that before.

RonRothman: @vytas7 Ack, (frozen)set it is.

kgriffs:
Hi @mikeyusko, just checking in to see where we you with this? I'd love to get it over the finish line soon if we can.

mikeyusko:
> Hi @mikeyusko, just checking in to see where we you with this? I'd love to get it over the finish line soon if we can.

Hi @kgriffs I'm going to finish it by the end of the week.

vytas7:
Re the absence of [`Morsel.samesite`](https://docs.python.org/3.8/library/http.cookies.html#http.cookies.Morsel) in Python < 3.8.0... I guess we could ship our own `Morsel` class to enable our test client? Particularly considering PyPy is probably a couple of light-years apart from 3.8.0 compatibility...

If we don't provide any alternatives, any API user testing on pre-3.8.0 would run into the very same issue in their own unit tests.

kgriffs:
@mikeyusko I've got some time to work on this. If you don't mind, I can add a couple patches to address remaining comments.

falcon/util/__init__.py
```diff
@@ -20,9 +20,18 @@
...
 from falcon.util.structures import *  # NOQA
 from falcon.util.misc import *  # NOQA
 from falcon.util.time import *  # NOQA
+
+
+# NOTE(kgriffs): Backport support for the new 'SameSite' attribute
+#   for Python versions prior to 3.8. We do it this way because
+#   SimpleCookie does not give us a simple way to specify our own
+#   subclass of Morsel.
+if 'samesite' not in http_cookies.Morsel._reserved:  # pragma: no cover
```
vytas7: I'm thinking whether we should somehow propagate this warning into our docs that we monkey-patch this way.
I don't really see this backport doing any harm, but OTOH it's hard to exhaust all possible things people are doing with the standard lib.

But I'm not insisting on this change for merging this PR. We can do this in a follow-up PR if we decide to.

kgriffs: Not a bad idea to at least call it out in the docs.

PR #1560, opened 2019-07-22
Use full license text in LICENSE

BigBlueHat:
# Summary of Changes

Previous content was the per-file header/preamble (from the appendix of the license). This brings the full legal text of the Apache License 2.0 into the LICENSE file. It also makes lawyers and robots happier.

# Related Issues

This fulfills the intent of #1014 which was about including the license in the distribution.

# Pull Request Checklist

None of these apply to this particular PR--but I checked them anyway per the template's note about doing so. 😁 

Cheers!
🎩


nZac:
I'm going to close this for now. Happy to hear more if you think that is in error.

BigBlueHat:
That's for applying the Apache License to a code file--hence the statement about "The text should be enclosed in the appropriate comment syntax for the file format."

There's a clearer explanation in the [License FAQ](https://www.apache.org/foundation/license-faq.html#Apply-My-Software)--i.e. the `LICENSE` file is meant to hold the full text of the license for the project.

One practical advantage is that this repository would be properly classified by GitHub and other tools as "Apache License 2.0" and not "other" (in the API) as it is now. For instance, see the "View License" link in the bottom right corner of the screenshot below.
falcon-license-bar

That link would read "Apache-2.0" if the `LICENSE` file contained the license text. Here's an example of that link when `LICENSE` has the full text: https://github.com/apache/incubator-annotator

Additionally, the `license` object in the repositories GitHub API output will be correct. Here's the current output of those two repos:

**Falcon** - https://api.github.com/repos/falconry/falcon
```json
{
  "license": {
    "key": "other",
    "name": "Other",
    "spdx_id": "NOASSERTION",
    "url": null,
    "node_id": "MDc6TGljZW5zZTA="
  }
}
```
**Apache Annotator** - https://api.github.com/repos/apache/incubator-annotator
```json
{
  "license": {
    "key": "apache-2.0",
    "name": "Apache License 2.0",
    "spdx_id": "Apache-2.0",
    "url": "https://api.github.com/licenses/apache-2.0",
    "node_id": "MDc6TGljZW5zZTI="
  }
}
```

Hope that clarifies!
🎩

nZac:
@BigBlueHat, thanks for the clarification and that makes sense.

@kgriffs can you approve this or does someone else need to okay messing with the LICENSE file?  Also, I don't think we tag every file, do you?

nZac:
Assuming this is okay with @kgriffs the only thing left for this is to get the build to pass. It is failing because of this line in the docs: https://github.com/falconry/falcon/blame/f9e5ef0d1084ae637eff56e2b30538ad964efecf/docs/user/intro.rst#L85 (sorry for a blame link, that is the only way to like to an RST).

I think it is safe to just use the existing short LICENSE and remove the `.. include` all together.

PR #1566, opened 2019-09-01
chore(Sphinx): explicitly specify autodoc directives for hook functions and redirect exceptions

vytas7:
Do so to stop Sphinx 2.1+ from complaining about duplicate object description of `falcon`.

### DIFF
--- a/docs/api/hooks.rst
+++ b/docs/api/hooks.rst
@@ -90,6 +90,6 @@ requests.
     are intended to hold request and response data specific to your app as it
     passes through the framework.
 
-.. automodule:: falcon
-    :members: before, after
-    :undoc-members:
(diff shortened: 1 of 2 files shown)


vytas7:
NB There is an ugly visual glitch with Sphinx 2.1.0 (not technically related to the problematic of this PR) that needs to be addressed: https://files.gitter.im/falconry/dev/OvLI/image.png

This is caused by CSS changes introduced in https://github.com/sphinx-doc/sphinx/pull/6625 ... We need to either work it around somehow, or to lock Sphinx to a previous version (that would negate the necessity of this PR too FWIW).

kgriffs:
Thanks @nZac ! I can put up a PR for that CSS fix once this is merged.

Issue #1358, opened 2018-10-01
ASGI+HTTP Support


kgriffs:
This will need to work with middleware, hooks, error handlers, media handlers, etc.

https://asgi.readthedocs.io/en/latest/specs/www.html

See also: #1055

kgriffs:
I will take this one.

nmcbride:
Hey there kgriffs,

I was wondering what you thought the timeline on this would be? I have a project where I am going to use web sockets for real-time collaboration. I'm not trying to rush you or anything like that I know this release will include really big changes, but I'm currently writing the core api in falcon and trying to decide if I should use Sanic or something else as a microservice specifically for the web socket stuff just so I have a way ahead.

Thanks for taking the time to respond.

Nate

kgriffs:
Hi, sorry for the delayed response. We decided to focus on shipping 2.0 first (RC1 is currently available on PyPI) so that we could drop Python 2.7 and 3.4 support in 3.0 and do a cleaner ASGI implementation.

3.0 development just opened on the master branch and we anticipate this milestone cycle to be significantly shorter than 2.0 (i.e., 2-3 months instead of 12-14 months), with the primary focus being on ASGI and improved form data parsing. We'll be sprinting on the 3.0 milestone at PyCon US so stay tuned!

You can see what is currently planned here: https://github.com/falconry/falcon/milestone/33

julioasotodv:
+1 for this. If Falcon supported natively Async requests using ASGI, it would become my go-to framework for almost everything web-related.

kgriffs:
Hi, for those interested, here is a gist demonstrated the proposed interface. Feedback is welcome. In the meantime, I'm prototyping the implementation and should have some code to share soon.

https://gist.github.com/kgriffs/4f99da6dde2266201ddddc42784e5aee

Sushi2all:
+1. Same as @julioasotodv.

julioasotodv:
@kgriffs I have just looked at your gist. I really like the style: very clean and “asyncionic” (I believe that word does not exist).

With an implementation like that, I would really never look back to aiohttp 😊

kgriffs:
OK everyone, please take a look at #1573

kgriffs:
Note that falcon.testing has been updated to handle ASGI. The event loop is handled behind the scenes to make it a lot easier for app developers to write their functional test cases.

kgriffs:
[...]

PR #1573, opened 2019-10-05
feat: ASGI support

kgriffs:
This patch adds the much-anticipated async support to Falcon by way of a
new ASGI interface, additional testing helpers, and updated internals.

Note that only the HTTP ASGI interface is implemented. WebSocket support
is planned to follow.

Docs will be fully fleshed-out in a follow-up PR.

Changelog snippets will also be added in a follow-up PR.

In order to reconcile differences between the WSGI and ASGI interfaces,
several breaking changes were made in this patch, as follows:

BREAKING CHANGE: create_environ no longer sets a default user agent header

BREAKING CHANGE: Renamed `protocol` kwarg for create_environ() to
	`http_version` and also the renamed kwarg only takes the version string
	(no longer prefixed with "HTTP/")

BREAKING CHANGE: Renamed `app` kwarg for create_environ() to `root_path`.
	and deprecated, may be removed in a future release.

BREAKING CHANGE: get_http_status() is deprecated, no longer accepts floats

BREAKING CHANGE: BoundedStream.writeable() changed to writable() per the
	standard file-like I/O interface (the old name was a misspelling).

BREAKING CHANGE: api_helpers.prepare_middleware() no longer accepts a single
	object; the value that is passed must be an iterable.

BREAKING CHANGE: Removed outer "finally" block from API and APP; add an
	exception handler for the base Exception type if you need to deal with
	unhandled exceptions.

BREAKING CHANGE: falcon.request.access_route will now include the value of
  the remote_addr property as the last element in the route, if not already present
  in one of the headers that are checked.

BREAKING CHANGE: When the 'REMOTE_ADDR' field is not present in the WSGI
	environ, Falcon will assume '127.0.0.1' for the value, rather than
	simply returning `None` for Request.remote_addr.


kgriffs:
OK folks, this is ready for review!

kgriffs:
Implements #1358

falcon/asgi/structures.py
```diff
@@ -0,0 +1,93 @@
...
+            #   does provide a check to ensure it is valid UTF-8. I'm also
+            #   assuming for the moment that most people will not use this
+            #   attribute, but rather the text and json ones instead. If that
+            #   is true, it makes sense to construct the entire string
+            #   first, then encode it all in one go at the end.
+            block += 'data: ' + self.data.decode() + '\n'
+        elif self.text is not None:
+            block += 'data: ' + self.text + '\n'
+        elif self.json is not None:
+            block += 'data: ' + json_dumps(self.json, ensure_ascii=False) + '\n'
```
vytas7: Composing string this way from many items (`local_var += '...something...'` etc) is optimized on CPython if you keep concatenating towards a local variable, but there is no guarantee it wouldn't perform poorly on PyPy (unless you benchmarked it?)

See also [here](https://pypy.org/performance.html) -- **String concatenation is expensive**, the recommended way is
```python
parts = []
for x in mylist:
    parts.append(foo(x))
s = "".join(parts)
```

falcon/routing/compiled.py
```diff
@@ -227,6 +238,51 @@ def find(self, uri, req=None):
     # Private
     # -----------------------------------------------------------------
 
+    def _check_coroutine_responders(self, method_map):
+        # NOTE(kgriffs): We use an environ variable because it would be
+        #   really convulted to try to pass this down the call stack
```
vytas7: _convulted_ (sic) :arrow_right: convoluted?

falcon/testing/client.py
```diff
@@ -749,3 +1140,49 @@ def simulate_request(self, *args, **kwargs):
...
+
+
+def _is_asgi_app(app):
+    app_args = inspect.getfullargspec(app).args
+    num_app_args = len(app_args)
+
+    # NOTE(kgriffs): Technically someone could name the "self"
+    #   arg something else, but we will make the simplifying
+    #   assumption that this is rare enough to not worry about.
+    if app_args[0] == 'self':
```
vytas7: As a fan of `@classmethod`s, would that be supported OoB, or do we need to add support for `app_args[0] == 'cls'`?
I would vote in favour in that case :slightly_smiling_face:

falcon/asgi/structures.py
kgriffs: IIRC PyPy actually optimizes string concatenation pretty well, at least for small numbers of concatenations. In the past when I've benchmarked this strategy, concatenation tended to be faster for at least up to ~5 concatenations (at the expense of a tiny bit of memory overhead), and it doesn't seem likely that we would go beyond that here. That being said, I will do some specific benchmarking with this method as-is vs. using `join()`.

vytas7: 5 small string concatenations would definitely compare in favour or `+=`.
I was just thinking here we actually have more concatenations, like `block += 'data: ' + self.text + '\n'` is actually 3 `+=` concatenations, and so on.
But I wouldn't be surprised it still outperforms `join()` (at least on CPython).

falcon/util/misc.py
```diff
@@ -305,3 +311,89 @@ def get_http_status(status_code, default_reason='Unknown'):
     except AttributeError:
         # not found
         return str(code) + ' ' + default_reason
+
+
+@functools.lru_cache(maxsize=64)
```
vytas7: I've commented on https://github.com/falconry/falcon/issues/1135#issuecomment-550187832, `functools.lru_cache` had thread-safety issues which where fixed post-3.6 release.
Maybe the issue is less serious here with async being mixed-and-matched with threads less often, and us mandating Python 3.6+, but still thought it might be worth bringing attention to the issue. Maybe we should check which CPython 3.5.x and 3.6.x versions are susceptible to the issue, and make a decision based on the findings.

vytas7:
I started the whole thing (testing in progress), issues found so far:

Serious issues:
None so far!

Cosmetic issues/desired improvements etc:
* Trying to use a WSGI resource with `asgi.App` results in a fatal warning with explanation what went wrong, however, the opposite case happily proceeds until a request is handled, and the results are unpredictable (calling such a "responder" produces a coroutine object, which is basically no-op in the sense of request handling, but then very strange warnings are produced)

falcon/media/base.py
```diff
@@ -16,15 +17,54 @@ def serialize(self, media, content_type):
...
+            stream (object): Asynchronous file-like object to deserialize.
+            content_type (str): Type of request content.
+            content_length (int): Length of request content.
+
+        Returns:
+            object: A deserialized object.
+        """
+
+        data = await stream.read()
+        return self.deserialize(io.BytesIO(data), content_type, content_length)
```
vytas7: Would it make sense to override `content_length` here :arrow_right: `len(data)`?

kgriffs:
I've rebased this on master (non-trivial!) and made a few tweaks. I still need to take care of the following:

- [ ] Address the remainder of @vytas7's comments
- [ ] Add an `async_to_sync()` helper
- [ ] Require the use of awaitables for responders, middleware, and error handlers
- [ ] Make sure that recent changes on the WSGI side are all ported over as necessary
- [ ] Add all the docstrings (with full docs probably waiting for a followup PR)
- [ ] Add towncrier fragments.

tests/asgi/test_asgi.py
```diff
@@ -0,0 +1,217 @@
...
+import pytest
+import requests
+import uvicorn
+
+import falcon
+import falcon.asgi
+import falcon.testing as testing
+
+_SERVER_HOST = '127.0.0.1'
+_SERVER_PORT = 9000 + os.getpid() % 100  # Facilitates parallel test execution
```
vytas7: Note that while OS PIDs are normally assigned sequentially on Unix-like systems, one needs to walk around existing PIDs for other programs, furthermore, allocation is restarted (often to PID 300) at some point to recycle IDs. The approach presented here is probably good enough for practical use, but it be worth adding some disclaimer note?

falcon/util/misc.py
kgriffs: Good catch. I think we will have to do something about this.

kgriffs: I opted to just use a noop for older pythons. The performance decrease probably won't be noticeable for most people, and over time those platforms will be used less and less.

kgriffs:
@vytas7 

> One general question regarding Cython, do we need to make them [Cython and ASGI] mutually exclusive?
> I'm thinking if we couldn't just exclude ASGI files (basically anything involving `async` or `await` keywords) from cythonization? If async stuff is calling "normal" Python functions, those can be well be cythonized, or can't they?

I'm hoping this is the case; I've been looking into it over the past few days. Hopefully I can get it to work!

falcon/asgi/structures.py
```diff
@@ -0,0 +1,93 @@
...
+        self.data = data
+        self.text = text
+        self.json = json
+        self.event = event
+        self.event_id = event_id
+
+        # NOTE(kgriffs) The SSE spec states that retry must be an integer;
+        #   this is a quick way to check (and coerce, if necessary) without
+        #   spending cycles on an isinstance() call.
+        self.retry = int(retry) if retry else None
```
vytas7: Just FWIW, if the value is `int`, `isinstance` seems to outperform allocation of a new `int` (CPython 3.8):
```python
>>> timeit.timeit(lambda: int(A), number=int(1e8))
8.146974843999487
>>> timeit.timeit(lambda: int(A), number=int(1e8))
8.008457058000204
>>> timeit.timeit(lambda: int(A), number=int(1e8))
8.354208095996
>>> timeit.timeit(lambda: isinstance(A, int), number=int(1e8))
6.547206504001224
>>> timeit.timeit(lambda: isinstance(A, int), number=int(1e8))
6.532925082006841
>>> timeit.timeit(lambda: isinstance(A, int), number=int(1e8))
6.530731582999579
>>> A
60
```

falcon/util/misc.py
```diff
@@ -305,3 +336,70 @@ def get_http_status(status_code, default_reason='Unknown'):
     except AttributeError:
         # not found
         return str(code) + ' ' + default_reason
+
+
+@_lru_cache_safe(maxsize=64)
+def http_status_to_code(status):
+    """Normalized an HTTP status to an integer code.
```
vytas7: Probably just imperative could suit better "Normalize an HTTP..."

falcon/asgi/response.py
```diff
@@ -0,0 +1,256 @@
...
+    def media(self):
+        return self._media
+
+    @media.setter
+    def media(self, value):
+        self._media = value
+        self._media_rendered = None
+
+    @property
+    def data(self):
```
vytas7: We'll have to also document the differences between this property and the WSGI `Request.data`, i.e. that accessing this property does not "materialize" `resp.media`, and one should use `render_body` instead. This bit me while trying to implement the async flavour of the "Look" tutorial :slightly_smiling_face:

falcon/asgi/app.py
```diff
@@ -0,0 +1,600 @@
...
+            triggered independently for the individual event loop associated
+            with each process.
+
+        A lifespan handler is any class that implements the interface below.
+        Note that it is only necessary to implement the methods for the
+        events that you wish to handle; Falcon will simply skip over any
+        missing methods in the lifespan handler::
+
+            class ExampleHandler:
+                def process_startup(self, scope, event):
```
vytas7: Needs to be `async def`

falcon/asgi/app.py
```diff
@@ -0,0 +1,600 @@
...
+
+                    Arguments:
+                        scope (dict): The ASGI scope dictionary for the
+                            lifespan protocol. The lifespan scope exists
+                            for the duration of the event loop.
+                        event (dict): The ASGI event dictionary for the
+                            startup event.
+                    \"\"\"
+
+                def process_shutdown(self, scope, event):
```
vytas7: Same as :arrow_up:, `async def`.

tests/asgi/test_asgi.py
kgriffs: I went ahead and just switched to using a simple RNG with a fairly large range. That should be sufficient and be easier to reason about.

falcon/asgi/structures.py
kgriffs: I did a quick check and looks like it is a tiny bit faster on 3.6 as well. This is actually good news; isinstance used to be a lot slower! I'll make the change.

kgriffs:
The gate is all green again. I think I've managed to get Cython working, but let me know if you run into any other Cython issues!

falcon/routing/compiled.py
```diff
@@ -15,12 +15,14 @@
 """Default routing engine."""
 
 from collections import UserDict
+from inspect import iscoroutinefunction
 import keyword
 import re
 import textwrap
 
 from falcon.routing import converters
```
vytas7: We'll probably need to document that converters are always sync since they are supposed to do simple path parsing and validation (as opposed to I/O).

kgriffs:
@vytas7 

> Cosmetic issues/desired improvements etc:
> 
> * Trying to use a WSGI resource with `asgi.App` results in a fatal warning with an explanation what went wrong, however, the opposite case happily proceeds until a request is handled, and the results are unpredictable (calling such a "responder" produces a coroutine object, which is basically no-op in the sense of request handling, but then strange warnings are produced)
> * Could we consider streamlining implementation of custom media handlers somewhat? Right now, `serialize` and `deserialize` are defined as abstract methods, so one needs to take care of them even in case one is only interested in the async variants for an ASGI app

Both of these should be addressed in the feature branch now.

falcon/routing/compiled.py
kgriffs: 👍 I've added this to my list of things to document.

falcon/asgi/app.py
kgriffs: fixed

falcon/asgi/app.py
kgriffs: fixed

falcon/util/misc.py
kgriffs: fixed

falcon/asgi/response.py
kgriffs: Do you think we should change this on the WSGI side to be the same as for ASGI, or just leave as-is with some warnings in the docs?

vytas7: IMHO it is fine to just document this for now (the first alpha) as we will probably have to make slight tweaks based on community feedback anyway.

tests/test_python_version_requirements.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from falcon import ASGI_SUPPORTED
+
+# from _util import CYTHON  # NOQA
```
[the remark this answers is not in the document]
kgriffs: fixed

falcon/testing/client.py
```diff
@@ -784,3 +1176,49 @@ def simulate_request(self, *args, **kwargs) -> Result:
...
+    if app_args[0] in {'cls', 'self'}:
+        num_app_args -= 1
+
+    is_asgi = (num_app_args == 3)
+
+    return is_asgi
+
+
+async def _wait_for_startup(events):
+    while True:
```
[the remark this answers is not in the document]
kgriffs: Strange... I guess I'll have to add a pragma

falcon/media/msgpack.py
```diff
@@ -41,5 +41,21 @@ def deserialize(self, stream, content_type, content_length):
                 'Could not parse MessagePack body - {0}'.format(err)
             )
 
+    async def deserialize_async(self, stream, content_type, content_length):
```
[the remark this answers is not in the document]
kgriffs: done

falcon/asgi/app.py
```diff
@@ -0,0 +1,604 @@
...
+            # NOTE(kgriffs): According to the ASGI spec: "Applications should
+            #   actively reject any protocol that they do not understand with
+            #   an Exception (of any type)."
+            raise UnsupportedScopeError(
+                f'The ASGI "{scope_type}" scope type is not supported.'
+            )
+
+        # PERF(kgriffs): This is slighter faster than using dict.get()
+        # TODO(kgriffs): Use this to determine what features are supported by
+        #   the servver (e.g., the headers key in the WebSocket Accept
```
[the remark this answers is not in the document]
kgriffs: heh, fixed

tests/test_error_handlers.py
```diff
@@ -73,6 +79,36 @@ def test_uncaught_python_error(self, client,
...
+        client = testing.TestClient(app)
+
+        result = client.simulate_get()
+        assert result.text == 'error: Plain Exception'
+
+        result = client.simulate_head()
+        assert result.status_code == 723
+        assert not result.content
+
+    # def test_uncaught_error(self, client):
```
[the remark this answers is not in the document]
kgriffs: I re-enabled them but it is a little wonky because by default we don't let exceptions propagate any more.

falcon/asgi/app.py
```diff
@@ -0,0 +1,604 @@
...
+            elif (
+                req.method == 'HEAD'and
+                resp_status not in _BODILESS_STATUS_CODES and
+                'content-length' not in resp._headers
+            ):
+                # NOTE(kgriffs): We really should be returning a Content-Length
+                #   in this case according to my reading of the RFCs. By
+                #   optionally using len(data) we let a resource simulate HEAD
+                #   by turning around and calling it's own on_get().
+                resp._headers['content-length'] = str(len(data)) if data else '0'
```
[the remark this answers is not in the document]
kgriffs: fixed and tests added

falcon/testing/client.py
```diff
@@ -338,48 +376,157 @@ def simulate_request(app, method='GET', path='/', query_string=None,
...
+    req_event_emitter = helpers.ASGIRequestEventEmitter(
+        (body or b''),
+        disconnect_at,
+        chunk_size=asgi_chunk_size
+    )
+
+    resp_event_collector = helpers.ASGIResponseEventCollector()
+
+    async def conductor():
+        # NOTE(kgriffs): We assume this is a Falcon WSGI app, which supports
```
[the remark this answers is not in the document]
kgriffs: fixed

falcon/asgi/app.py
```diff
@@ -0,0 +1,604 @@
...
+            if hasattr(stream, 'close'):
+                await stream.close()
+
+        await send(_EVT_RESP_EOF)
+        self._schedule_callbacks(resp)
+
+    def add_lifespan_handler(self, handler):
+        """Add a lifespan handler object.
+
+        A lifespan handler performs startup and/or shutdown activities for the
```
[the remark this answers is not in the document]
kgriffs: I suppose for something like a caching component, having everything in the same class makes sense. Perhaps I should just remove the distinction and update the docs accordingly? (There would no longer be a separate `add_lifespan_handler()` method.)

falcon/asgi/request.py
```diff
@@ -0,0 +1,607 @@
...
+                #   safe than sorry).
+                self._asgi_server_cached = tuple(self.scope['server'])
+            except (KeyError, TypeError):
+                # NOTE(kgriffs): Not found, or was None
+                default_port = 80 if self.scheme == 'http' else 443
+                self._asgi_server_cached = ('localhost', default_port)
+
+        return self._asgi_server_cached
+
+    async def _parse_form_urlencoded(self):
```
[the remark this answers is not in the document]
kgriffs: done

falcon/asgi/app.py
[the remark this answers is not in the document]
vytas7: Hmm... yeah, I think I like the idea :+1: (But I may be missing the downsides, if any).
You've even named the methods `process_*` which is essentially consistent with the middleware pattern. The onion diagram for handlers' order looks in principle similar, too.

falcon/media/base.py
vytas7: To explain a bit more how I was reasoning here.
Although we'll be changing the media API to not always mandate the `content_length`, but this might be helpful to legacy handlers that may be expecting the perfect `content_length`.

OTOH multipart form part media handling has the same requirement even on the WSGI side, so maybe there is not point in shimming this way if we cannot consistently guarantee it anyway.

kgriffs:
Heh, I made a similar fix for uvicorn in parallel but you beat me to it!

falcon/media/base.py
kgriffs: Can't hurt. I went ahead and made the change.

vytas7:
Timezones worked in my advantage :slightly_smiling_face:

falcon/asgi/app.py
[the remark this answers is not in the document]
kgriffs: I went ahead and merged the two concepts.

vytas7:
@kgriffs not to spoil the fun you're having, but `daphne` might be blocking on the clogged `stdout` (`subprocess.PIPE`) since we don't drain it.

I get all the tests to pass in Docker by commenting these out:
```diff
--- a/tests/asgi/test_asgi_servers.py
+++ b/tests/asgi/test_asgi_servers.py
@@ -271,8 +271,8 @@ def _daphne_factory(host, port):

             _MODULE_NAME + ':application'
         ),
-        stdout=subprocess.PIPE,
-        stderr=subprocess.STDOUT,
+        # stdout=subprocess.PIPE,
+        # stderr=subprocess.STDOUT,
         cwd=_MODULE_DIR,
     )
```

kgriffs:
@vytas7 OK, I think I've addressed all your feedback so far.

kgriffs:
I'm going to start working on docstrings and towncrier fragments. Everyone please let me know if you see any other implementation/design issues.

tests/test_headers.py
```diff
@@ -6,19 +6,21 @@
...
+def client(asgi):
+    app = create_app(asgi)
     return testing.TestClient(app)
 
 
 @pytest.fixture(scope='function')
-def cors_client():
-    app = falcon.App(cors_enable=True)
+def cors_client(asgi):
+    app = create_app(asgi, cors_enable=True)
```
karmux: I get `CompatibilityError` when using `cors_enable=True` with asgi App.

kgriffs:
@vytas7 re SSE testing that may be worth creating a standalone issue so we don't forget about it?

tests/test_headers.py
kgriffs: Good catch!

kgriffs: fixed

vytas7:
> @vytas7 re SSE testing that may be worth creating a standalone issue so we don't forget about it?

Created https://github.com/falconry/falcon/issues/1634.

kgriffs:
Due to the cython bug re `iscoroutinefunction()`, we may have to add some warnings to the docs about cythonizing ASGI apps, since that would hit the same issues we've been experiencing in the framework itself. 

In some cases we could simply not do the check (when it is just to highlight a developer mistake), but in other cases it is less clear how we would avoid using `iscoroutinefunction()` (e.g., `jsonschema.validate`, `hooks`) without creating explicitly named functions (e.g., `validate_async()`). I suppose we *could* do that, but it puts more burden on the developer to remember to use the right function name.

See also:

* https://github.com/cython/cython/issues/2273
* https://bugs.python.org/issue38225

vytas7:
Yeah, I guess we'll just have to be transparent in the docs and make sure the warnings are easily discoverable.

OTOH we *could* make everything explicit, and always expect coroutines where applicable, i.e.
* Do not perform developer mistake checks
* Do `jsonschema` the `validate_async()` way
* Hooks could be applied with `falcon.asgi.after` / `falcon.asgi.before` instead of `falcon.after` / `falcon.before`

Going this path would make porting clumsier, and easier to make mistakes due to lacking checks though...

FWIW I was reading more on the subject, and although Cython maintainers state on https://github.com/cython/cython/issues/2273 they would accept a PR exposing a faux `_is_coroutine` attribute, I'm not sure if this is a sustainable solution given the whole `asyncio.coroutine` thing is deprecated and slated for removal in Python 3.10.

kgriffs:
> I'm not sure if this is a sustainable solution given the whole asyncio.coroutine thing is deprecated and slated for removal in Python 3.10.

Agreed. It seems that the proposed solution will only work for `asyncio.iscoroutinefunction()` and not for `inspect.iscoroutinefunction()`, with the former possibly going away in Python 3.10. The alternative is setting `co_flags` bits, but apparently that will not work in Python 3.8+ because of the new `inspect._has_code_flag()` function that ignores flags when the function is not a Python function (but a Cython function). So then you would have to check it directly ala hug, but that isn't a general enough solution so I can understand why scoder is reluctant to go there...

...hence the proposal to try and create an attribute-based protocol for it in core python. If a PR does land that makes `asyncio.iscoroutinefunction()` work, I suppose we could use that in the interim while we wait on https://bugs.python.org/issue38225 .

Regardless, I think it does make sense to provide an escape hatch for anyone using cythonized async functions in their app. I could provide alternative functions that just assume the thing is a coroutine. I would then sprinkle the docs with warnings re the issue and the workaround (using the alternative function names). I could even catch any TypeErrors that are raised while attempting to await the result of calling a non-coroutine function, check for cython in the environment, and finally raise an error with a message that talks about the workaround.

If we are able to come to a rough consensus on this, I can tackle it following 3.0.0a1.

tests/asgi/test_asgi_servers.py
```diff
@@ -0,0 +1,185 @@
...
+@pytest.fixture(params=[_uvicorn_factory, _daphne_factory])
+def server_base_url(request):
+    process_factory = request.param
+
+    # NOTE(kgriffs): This facilitates parallel test execution as well as
+    #   mitigating the problem of trying to reuse a port that the system
+    #   hasn't cleaned up yet.
+    # NOTE(kgriffs): Use our own Random instance because we don't want
+    #   pytest messing with the seed.
+    server_port = _random.randint(50000, 60000)
```
vytas7: We'll probably need to make this more robust in the future, because it may randomly clash with other stuff on the system; although the probability is low, it may clash between tests as well.

PR #1574, opened 2019-10-05
Support setting the crossorigin attribute (closes #1410)

### LINKED ISSUE #1410: Update Response:add_link() to support setting the crossorigin attribute

nagaabhinaya:
Fixes #1410 

# Summary of Changes

- Added `crossorigin` attribute to `Response:add_link()` method. 
- It can either be `None` (default) or `anonymous` or `use-credentials`.
- No breaking change introduced.

For more info, see [here](https://www.w3.org/TR/html50/infrastructure.html#cors-settings-attribute)

Is this how you visualize the implementation? Open to feedback on this

# Pull Request Checklist

- [ ] Changes (and possible deprecations) have [towncrier](https://pypi.org/project/towncrier/) news fragments under `docs/_newsfragments/`. (Run `towncrier --draft` to ensure it renders correctly.)


falcon/response.py
```diff
@@ -746,6 +749,10 @@ def add_link(self, target, rel, title=None, title_star=None,
         if anchor is not None:
             value += '; anchor="' + uri_encode(anchor) + '"'
 
+        if crossorigin is not None:
+            if crossorigin in ['anonymous', 'use-credentials']:
```
kgriffs: Since setting crossorigin to a blank value is the same as `anonymous', we can check for that case and avoid performing the extra concatenation, e.g.:

```py
value += '; crossorigin'
```

kgriffs: Rather than failing silently when the value is invalid, I think we ought to raise an instance of ValueError in that case.

falcon/response.py
```diff
@@ -771,6 +776,19 @@ def add_link(self, target, rel, title=None, title_star=None,
         if anchor is not None:
             value += '; anchor="' + uri_encode(anchor) + '"'
 
+        if crossorigin is not None:
+            crossorigin = crossorigin.lower()
```
vytas7: Note that I added support for case-insensitive `crossorigin` values.

The `Link` header itself seems to be quite poorly documented, however the available documentation states that it should be treated exactly in the same way as the corresponding `<link>` HTML5 element [`crossorigin` attribute](https://www.w3.org/TR/html50/document-metadata.html#attr-link-crossorigin). Which in turn is an [enumerated attribute](https://www.w3.org/TR/html50/infrastructure.html#enumerated-attribute), hence [ASCII case-insensitive](https://www.w3.org/TR/html50/infrastructure.html#ascii-case-insensitive).

kgriffs: That is correct, it is treated by the user agent the same as the `<link>` element, although some browser versions support certain features only when expressed as an element vs. a header.

falcon/response.py
```diff
@@ -771,6 +776,19 @@ def add_link(self, target, rel, title=None, title_star=None,
...
+            if crossorigin not in _RESERVED_CROSSORIGIN_VALUES:
+                raise ValueError(
+                    'crossorigin must be set to either '
+                    "'anonymous' or 'use-credentials'")
+            if crossorigin == 'anonymous':
+                value += '; crossorigin'
+            else:  # crossorigin == 'use-credentials'
+                # PERF(vytas): the only remaining value is inlined.
+                # Un-inline in case more values are supported in the future.
+                value += '; crossorigin="use-credentials"'
```
kgriffs: IIRC, the double-quotes are unnecessary, but they don't hurt either.

vytas7: I was actually reading upon this one too :slightly_smiling_face: 

But in the the majority of examples in the wild used quotes, and @nagaabhinaya proposed this way as well, so I decided those two bytes were not worth saving.

PR #1575, opened 2019-10-05
Fix sphinx error - duplicate object description of falcon

nagaabhinaya:
# Summary of Changes

Sphinx run was failing for `TOXENV=docs`. This fixes that.


kgriffs:
Thanks! Looks like this is a duplicate of #1566 (albeit using a different approach). If you don't mind, I think we'll go ahead and merge the other PR since it was submitted first.

PR #1590, opened 2019-10-25
feat(media): implement multipart form media handler

vytas7:
@kgriffs @jmvrbanac @nZac and others in the community who are following the development of this feature:
this is still rough on edges when it comes to documentation, but the interface and the implementation are hopefully alpha-level ready.

I have also created an extended test/benchmark suite for tuning performance as well as doubling as extra validation for code correctness: https://github.com/vytas7/falcon-multipart-tests

There is still one notable case of many small reads from the part stream which I stopped short of optimizing. The machinery is largely ready, I could just force reads to be aligned on the part stream. But I would be happy to get feedback from the community how the functionality is used in the wild in order to get a better impression whether we need to focus on this specific usage pattern.

**The Gist** illustrating the interface and the demo therein are still valid: https://gist.github.com/vytas7/34c60e5ac3a4bc2f2eb0af2428d77003 (I have actually just slightly updated it).

I have also started implementing **async multipart form parsing for ASGI** (i.e. marrying this to async media support in https://github.com/falconry/falcon/pull/1573).
This class could probably be used verbatim for ASGI from #1573, but better don't even try it at home. In the best case, your file would be buffered in memory in a bytes buffer.

_NB: creating a new pull request because my previous branch became too complex for GitHub to handle (the GH PR interface was warning of bogus conflicts)._


vytas7:
Implements #953

falcon/media/multipart.py
```diff
@@ -0,0 +1,387 @@
...
+            self._media = handler.deserialize(
+                self.stream, self.content_type, None)
+
+        return self._media
+
+
+class MultipartForm:
+
+    def __init__(self, stream, boundary, content_length, parse_options):
+        # if not isinstance(stream, BufferedStream):
```
nZac: Is this a coding artifact or here on purpose?

falcon/media/multipart.py
```diff
@@ -0,0 +1,387 @@
...
+        if not hasattr(stream, 'read_until'):
+            if isinstance(stream, request_helpers.BoundedStream):
+                stream = BufferedStream(stream.stream.read, content_length)
+            else:
+                stream = BufferedStream(stream.read, content_length)
+
+        self._stream = stream
+        self._boundary = boundary
+        self._dash_boundary = b'--' + boundary
```
nZac: We might want to add a reference to the [RFC 7578#4.1](https://tools.ietf.org/html/rfc7578#section-4.1) for why this is necessary.

falcon/media/multipart.py
```diff
@@ -0,0 +1,387 @@
...
+        max_headers_size = self._parse_options.max_body_part_headers_size
+        remaining_part_count = self._parse_options.max_body_part_count
+
+        while True:
+            # NOTE(vytas): Either exhaust the unused stream part, or skip
+            #   the prologue.
+            stream.pipe_until(delimiter)
+            stream.read(len(delimiter))
+
+            if not delimiter.startswith(b'\r\n'):
```
nZac: Instead of having b`'\r\n'` all over the place, could you make "constant" `CLRF` at the top of the file and use that? It's three less characters you need to type :)

vytas7: Sure, thanks, can do that :+1: (I would go with the more common abbreviation `CRLF` though).

falcon/media/multipart.py
vytas7: Thanks, will do.

falcon/media/multipart.py
vytas7: It is just a debugging artefact indeed. It reflects another, stricter approach that I had initially, but later switched to just testing if the object quacks `read_until` because that allowed for much easier testing both cythonized and pure-Python variants in the same extended test suite (https://github.com/vytas7/falcon-multipart-tests).

I'll clean up.

vytas7:
> I am still slowly walking through this. Here are some very minor pieces of feedback until I can give this a more thorough look.
> 
> In the meantime, can you give me a synopsis on the addition of `BufferedStream`? Is it all the handle the `boundry` stuff?

@nZac Thanks for looking at this PR! Yeah, `BufferedStream` is to handle the boundary stuff. Once you read a chunk from the WSGI input stream, and find the part boundary, you need to buffer the remaining part to be able to transparently present it as the part (sub-)stream.

The ASGI variant has a bit simpler variant of dealing with the same problem (the ASGI HTTP request input is coming in the form of chunked messages).
We'll probably have to sync approaches with @kgriffs (see `falcon.asgi.stream`). Also, we may want to retire the existing BoundedStream in favour of this BufferedStream (or w/e name we choose for it). See also this issue:

falcon/media/multipart.py
nZac: Whoops! `CRLF` is correct... apparently I can't type straight.

docs/api/multipart.rst
```diff
@@ -0,0 +1,51 @@
...
+through multipart form
+:class:`body parts <falcon.media.multipart.BodyPart>`:
+
+.. code:: python
+
+    for part in req.media:
+        if part.content_type == 'application/json':
+            # TODO: Body part is a JSON document, do something useful with it
+            resp.media = part.media
+        elif part.name = 'datafile':
```
kgriffs: Looks like this needs another `=`

docs/api/multipart.rst
```diff
@@ -0,0 +1,51 @@
...
+                chunk = part.stream.read(8192)
+                if not chunk:
+                    break
+                feed_data(chunk)
+        else:
+            # TODO: Do something else
+
+.. note::
+   In further development versions of Falcon 3.0 series,
+   :class:`falcon.media.MultpartFormHandler` may be promoted to the default
```
kgriffs: Probably a good idea, but I guess we can wait for community feedback.

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+            stream.read(len(delimiter))
+
+            if not delimiter.startswith(CRLF):
+                # NOTE(vytas): RFC 7578, section 4.1.
+                #   As with other multipart types, the parts are delimited with
+                #   a boundary delimiter, constructed using CRLF, "--", and the
+                #   value of the "boundary" parameter.
+                delimiter = CRLF + delimiter
+
+            separator = stream.read_until(CRLF, 2, MultipartParseError)
```
kgriffs: How does this approach of passing `MultipartParseError` compare to raising the same error type each time and letting the caller catch the error and do with it what it will?

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
...
+                self._buffer_len = next_chunk_len
+            else:
+                self._buffer = self._buffer[self._buffer_pos:] + next_chunk
+                self._buffer_len = (self._buffer_len - self._buffer_pos +
+                                    next_chunk_len)
+                self._buffer_pos = 0
+
+        return ret_value
+
+    def _read_until(self, delimiter, amount, missing_delimiter_error=None):
```
kgriffs: "size" may be a more common name to use vs. "amount"

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def __init__(self, description=None, headers=None, **kwargs):
+        super().__init__('Malformed multipart/form-data request media',
+                         description, headers, **kwargs)
+
+
+# TODO(vytas): Consider supporting -charset- stuff.
+#   Does anyone use that (?)
+class BodyPart:
+    """
```
kgriffs: style: Remove linebreak after `"""`

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+#   * Better documentation
+#   * Document MultipartParseError
+
+_ALLOWED_CONTENT_HEADERS = frozenset([
+    b'content-type',
+    b'content-disposition',
+    b'content-transfer-encoding',
+])
+
+DEFAULT_SUPPORTED_CHARSETS = (
```
kgriffs: How was this list chosen? Might be worth an inline comment.

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+                if stream.peek(2) != CRLF:
+                    raise MultipartParseError('unexpected form epilogue')
+                break
+            elif separator:
+                raise MultipartParseError('unexpected form structure')
+
+            headers = {}
+            headers_block = stream.read_until(_CRLF_CRLF, max_headers_size,
+                                              MultipartParseError)
+            stream.read(4)
```
kgriffs: Looks like this consumes the double-CRLF, but it might be nice to have an inline comment to that effect.

kgriffs: Should `read_until()` support an option to consume the delimeter?

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+                            'the deprecated Content-Transfer-Encoding header '
+                            'field is unsupported')
+                    # NOTE(vytas): RFC 7578, section 4.8.
+                    #   Other header fields MUST NOT be included and MUST be
+                    #   ignored.
+                    elif name in _ALLOWED_CONTENT_HEADERS:
+                        headers[name] = value
+
+            remaining_part_count -= 1
+            if remaining_part_count == 0:
```
kgriffs: This looks like an off-by-one error. I.e., if `max_body_part_count` is `1`, `remaining_part_count` is initialized to `1`. Then, after consuming the first part, `remaining_part_count` becomes `0` and `MultipartParseError` is raised even though we have only consumed more than `max_body_part_count` yet. I haven't checked the tests yet, but there should be one for this boundary condition.

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+                raise MultipartParseError(
+                    'maximum number of form body parts exceeded')
+
+            yield BodyPart(stream.delimit(delimiter, MultipartParseError),
+                           headers, self._parse_options)
+
+        stream.exhaust()
```
kgriffs: I need to go check this, but I remember elsewhere seeing:

```py
if handler.exhaust_stream:
    self.bounded_stream.exhaust()
```

Are we potentially disregarding the developer's preference here by just always exhausting the stream regardless of the flag on the handler?

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+class MultipartFormHandler(BaseHandler):
+    """
```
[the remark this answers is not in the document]
vytas7: Oh, I actually have a habit/stylistic preference for inserting that extra line break... :open_mouth: 

I'll make the style consistent with the rest (majority) of the Falcon code.

FWIW, maybe we should document this preference in `CONTRIBUTING.md`? Although it does mention trying to be consistent.

FWIW, [PEP 257 does not mandate one or another form](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings):
> The summary line may be on the same line as the opening quotes or on the next line. The entire docstring is indented the same as the quotes at its first line (see example below).

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+        max_body_part_count (int): The maximum amount of body parts in the
+            form (default: 64).
+
+        max_body_part_buffer_size (int): The maximum amount of bytes to buffer
+            as `data` property (default: 1 MiB).
+
+        max_body_part_headers_size (int): The maximum size (in bytes) of the
+            body part headers structure (default: 8192).
+
+        media_handlers (Handlers): A dict-like object that allows you to
```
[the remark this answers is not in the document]
vytas7: I'll reword. Tbh I've just copy-pasted this from `RequestOptions` / `ResponseOptions`, I can simplify these as well.

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+    Multipart form (content type ``multipart/form-data``) media handler.
+
+    The ``multipart/form-data`` media type for HTML5 forms is defined in the
+    `RFC 7578 <https://tools.ietf.org/html/rfc7578>`_.
+
+    The multipart media type itself is defined in the
+    `RFC 2046 section 5.1 <https://tools.ietf.org/html/rfc2046#section-5.1>`_.
+
+    .. note::
+       Note that unlike many others, this handler does not consume the stream
```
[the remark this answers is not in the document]
vytas7: Oh, not 100% what I meant back then but I think I actually was talking about other handlers in the Falcon framework :man_facepalming: They all do a blank `read()`.

But I agree it doesn't hurt to signify the difference from other parsing implementations. Maybe we could highlight that in the intro part of `multipart.rst`?

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+       immediately. Rather, the serialized media object shall be iterated to
+       parse the body parts.
+    """
+
+    def __init__(self, parse_options=None):
+        self.parse_options = parse_options or MultipartParseOptions()
+
+    def deserialize(self, stream, content_type, content_length):
+        _, options = cgi.parse_header(content_type)
+        boundary = options.get('boundary')
```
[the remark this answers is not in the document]
vytas7: Agreed, EAFP suits perfectly here, particularly given that we raise another follow-up exception.

docs/api/multipart.rst
vytas7: Do you suggest removing this paragraph?

falcon/media/multipart.py
vytas7: Touché; there is a test for this boundary, but it is off-by-one from this boundary condition :man_facepalming:

falcon/media/multipart.py
vytas7: > Should `read_until()` support an option to consume the delimeter (sic)?
That makes sense :+1: I'll check if it can be easily implemented.

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+        if boundary is None:
+            raise errors.HTTPInvalidHeader(
+                'No boundary specifier found in {!r}'.format(content_type),
+                'Content-Type')
+
+        # NOTE(vytas): RFC 2046, section 5.1.
```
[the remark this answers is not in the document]
vytas7: Thanks!

docs/api/multipart.rst
vytas7: Ouch. Fixed.

falcon/media/multipart.py
vytas7: Umm... that approach would be cleaner and more straightforward, I guess.

IIRC I probably did it this way because the part stream's `read_until()` is exposed to API users, so the generic `MultipartParseError` would sound a bit off. But we could maybe define a new error type for `read_until()`, and re-raise that in the multipart handler?

falcon/media/multipart.py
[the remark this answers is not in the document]
kgriffs: Ah, good point. IIRC this was a style I picked up a long time ago on another project that had it mentioned in their style guide and I must have neglected to say anything about it.

https://github.com/falconry/falcon/pull/1573

docs/api/multipart.rst
kgriffs: TBH I think we should just decide one way or another and do it if possible for the alpha. What are the potential problems with adding it to the default handlers dict?

kgriffs: (If it is enabled by default, the docs should mention this fact.)

vytas7: Sure, let's do that :+1: 
There is one potential problem with circular imports, but I'll tape it one or another way,

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
+import functools
+import io
+
+DEFAULT_CHUNK_SIZE = io.DEFAULT_BUFFER_SIZE * 4
+
+
+class BufferedStream:
+
+    def __init__(self, read, max_stream_len, chunk_size=None):
+        self._read_func = read
```
kgriffs: It may make sense here to accept a file-like "stream" to wrap here, rather than the read method itself, ala `io.BufferedReader()`.

kgriffs:
> IIRC I probably did it this way because the part stream's read_until() is exposed to API users, so the generic MultipartParseError would sound a bit off. But we could maybe define a new error type for read_until(), and re-raise that in the multipart handler?

That could work. BufferedStream can define its own exception types and raise those as needed.

tests/test_media_multipart.py
```diff
@@ -0,0 +1,495 @@
...
+
+    handler = media.MultipartFormHandler()
+    form = handler.deserialize(
+        io.BytesIO(data), 'multipart/form-data; boundary=BOUNDARY', len(data))
+    with pytest.raises(falcon.HTTPBadRequest):
+        for part in form:
+            pass
+
+
+def test_unsupported_charset():
```
kgriffs: Could you also add tests for mismatched charsets (e.g., charset says it is one thing, but the actual data fails to decode)?

tests/test_streams.py
```diff
@@ -0,0 +1,249 @@
+import io
+
+import pytest
+
+from falcon.util import BufferedStream
+
+
+TEST_DATA = (
```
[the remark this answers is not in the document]
kgriffs: Also, let's make sure to thoroughly test boundary strings with unexpected (non-_bchars_) characters in them, missing or extraneous newlines, and any other way you can think of that the request body may be corrupted by a super evil attacker who doesn't care to follow the RFCs too closely. 😈

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
...
+            result = self._buffer
+            self._buffer_len = 0
+            self._buffer = b''
+            return result
+
+        if amount < self._buffer_len - self._buffer_pos:
+            self._buffer_pos += amount
+            return self._buffer[self._buffer_pos - amount:self._buffer_pos]
+
+        # NOTE(vytas): if amount > self._buffer_len - self._buffer_pos
```
kgriffs: Shouldn't this comment read: `if amount >=` (greater than or equal to)?

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
...
+        read_amount = amount - self._buffer_len + self._buffer_pos
```
kgriffs: Nitpick: The math seems a bit more intuitive to me (esp. given the `if amount >= self._buffer_len - self._buffer_pos` comment) when written as:

```py
amount - (self._buffer_len - self._buffer_pos)
```

kgriffs: Regardless, it seems that you would want to read in `max(read_amount, self._chunk_size)` and buffer what is not immediately needed.

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
...
+        result = self._buffer[self._buffer_pos:]
+        self._buffer_len = 0
+        self._buffer_pos = 0
+        self._buffer = b''
+        self._max_bytes_remaining -= read_amount
```
kgriffs: See also my other comment re trusting `self._read_func()` to return the requested number of bytes (it could be less).

falcon/media/multipart.py
```diff
@@ -0,0 +1,412 @@
...
+
+        return self._data
+
+    @property
+    def text(self):
+        content_type, options = cgi.parse_header(self.content_type)
+        if content_type != 'text/plain':
+            return None
+
+        charset = options.get('charset') or self._parse_options.default_charset
```
kgriffs: Any particular reason not to just do the following? 

```py
options.get('charset', self._parse_options.default_charset)
```

falcon/media/multipart.py
```diff
@@ -0,0 +1,412 @@
...
+            raise MultipartParseError(
+                'unsupported charset: {}'.format(charset))
+        return self.data.decode(charset)
+
+    @property
+    def content_type(self):
+        # NOTE(vytas): RFC 7578, section 4.4.
+        #   Each part MAY have an (optional) "Content-Type" header field, which
+        #   defaults to "text/plain".
+        value = self._headers.get(b'content-type') or b'text/plain'
```
kgriffs: Same here.

falcon/media/multipart.py
```diff
@@ -0,0 +1,412 @@
...
+            separator = stream.read_until(_CRLF, 2, MultipartParseError)
+            if separator == b'--':
+                if stream.peek(2) != _CRLF:
+                    raise MultipartParseError('unexpected form epilogue')
+                break
+            elif separator:
+                raise MultipartParseError('unexpected form structure')
+
+            headers = {}
+            headers_block = stream.read_until(_CRLF_CRLF, max_headers_size,
```
kgriffs: Is there a test with a header block that exceeds `max_headers_size`?

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
...
+                # NOTE(vytas): we have now verified that all bytes currently in
+                #   the buffer are delimiter-free, including the border of the
+                #   upcoming chunk
+                return self._finalize_read_until(
+                    amount, result, have_bytes, next_chunk=next_chunk,
+                    next_chunk_len=next_chunk_len)
+
+            have_bytes += self._buffer_len - self._buffer_pos
+            if self._buffer_pos > 0:
+                result.append(self._buffer[self._buffer_pos:])
```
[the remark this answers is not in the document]
kgriffs: I did a few quick benchmarks (CPython 3.8) using variations of the test provided in one of the answers to [this SO question](https://stackoverflow.com/questions/9099145/where-are-python-bytearrays-used). The `b''.join()` pattern is indeed more performant. When combining four 8K chunks the difference on my box was something like 1.14 µs vs. 2.43 µs. The difference seemed to increase non-linearly vs. the number of chunks.

On the other hand, `bytearray.extend()` has the advantage of letting you free the chunk immediately rather than building up a list of them and holding them all in memory. This briefly requires double the memory (after the `b''.join()` but before the list of chunks is deallocated).

falcon/util/streams.py
kgriffs: Nevermind; I see how this argument is used by `delimit()`.

docs/user/faq.rst
```diff
@@ -561,41 +561,56 @@ POSTed form parameters may also be read directly from
...
+   In further development versions of Falcon 3.0 series,
+   :class:`falcon.media.MultpartFormHandler` may be promoted to the default
+   media handlers.
+
+How can I save POSTed files (from a multipart form) directly to AWS S3?
+-----------------------------------------------------------------------
+
+As highlighted in the previous answer dealing with
+:ref:`files posted as multipart form <access_multipart_files>`,
+:class:`falcon.media.MultipartFormHandler` may be used to iterated through
```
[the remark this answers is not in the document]
vytas7: Ouch I have no clue how I managed to copy-paste this _iterator through_ all over the PR :man_facepalming:

docs/user/faq.rst
```diff
@@ -561,41 +561,56 @@ POSTed form parameters may also be read directly from
...
+    # ...
 
-    file_item = form[name]
-    if file_item.file:
-        # TODO: It's an uploaded file... read it in
-    else:
-        # TODO: Raise an error
+    for part in req.media:
+        if part.name == 'myfile':
+            s3.upload_fileobj(part.stream, 'mybucket', 'mykey')
```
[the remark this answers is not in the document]
vytas7: Thanks!

We want 'em multipart forms to fly in the cloud :cloud_with_lightning: :slightly_smiling_face:

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
...
+                read_amount = self._max_bytes_remaining
+            if read_amount == 0:
+                return self._finalize_read_until(
+                    amount, result, have_bytes, delimiter, delimiter_pos,
+                    missing_delimiter_error=missing_delimiter_error)
+
+            next_chunk = self._read_func(read_amount)
+            next_chunk_len = len(next_chunk)
+            self._max_bytes_remaining -= read_amount
+            if next_chunk_len < self._chunk_size:
```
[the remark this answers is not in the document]
vytas7: I was developing under the assumption that at least WSGI input wrappers must block and return the requested number of bytes, otherwise it is considered an EOF.
But maybe this was a false assumption... I'll reread what the PEPs say.

Also, it might make sense supporting such read functions anyway... I'll check this.

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
+import functools
+import io
+
+DEFAULT_CHUNK_SIZE = io.DEFAULT_BUFFER_SIZE * 4
```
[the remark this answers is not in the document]
vytas7: I'll add.
I still haven't settled on the best value though, but leaning towards this one :arrow_up: (32 KiB).
On one hand, we'd like to keep buffer bloat to the minimum, OTOH we'd want reasonable performance.

See also some benchmarks which I was looking at when trying to pick: https://travis-ci.org/vytas7/falcon-multipart-tests/jobs/640673269?utm_medium=notification&utm_source=github_status

falcon/media/multipart.py
```diff
@@ -0,0 +1,402 @@
...
+            body part headers structure (default: 8192).
+
+        media_handlers (Handlers): A dict-like object that allows you to
+            configure the media-types that you would like to handle.
+            By default, a handler is provided for the ``application/json``
+            media type.
+
+        supported_charsets (frozenset): The list of supported character
+            encodings that are understood by the parser. The provided charsets
+            must by provided in lowercase, and must also be understood by
```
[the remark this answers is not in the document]
vytas7: :man_facepalming:

falcon/media/multipart.py
vytas7: No, I just probably wrote that without thinking too much. We do not actually need to be too forgiving here.

docs/user/faq.rst
```diff
@@ -561,41 +561,56 @@ POSTed form parameters may also be read directly from
...
 
 .. code:: python
 
-    # TODO: Either validate that content type is multipart/form-data
-    # here, or in another hook before allowing execution to proceed.
+    for part in req.media:
+        # TODO: Do something with the body part
+        pass
+
+.. note::
```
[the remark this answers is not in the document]
vytas7: Done, I've now polished both the FAQ item and the `multipart.rst` intro.

falcon/util/misc.py
```diff
@@ -305,3 +308,25 @@ def get_http_status(status_code, default_reason='Unknown'):
...
+    Returns:
+        str: sanitized filename
+    """
+    # TODO(vytas): max_length (int): Maximum length of the returned
+    #     filename. Should the returned filename exceed this restriction, it is
+    #     truncated while attempting to preserve the extension.
+    filename = unicodedata.normalize('NFKD', filename)
+    if filename.startswith('.'):
+        filename = filename.replace('.', '_', 1)
+    return _UNSAFE_CHARS.sub('_', filename) or 'file'
```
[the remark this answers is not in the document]
vytas7: I was unsure about this whole function, and the corresponding `BodyPart.secure_filename` property, whether this is needed in Falcon at all. OTOH Werkzeug has it, and a property is a low overhead porcelain.

Agreed about that `or 'file'`. We could:
*  Just note in the docstring that this function may return an empty string
* Add a default kwarg parameter, like `fallback_filename='something'`. OTOH the property won't be able to utilize it directly, and adding this to parser options might be a bit overboard
* Keep 'or 'file'` with a more descriptive default name, suggestions welcome for that name :slightly_smiling_face:

falcon/util/misc.py
```diff
@@ -305,3 +308,25 @@ def get_http_status(status_code, default_reason='Unknown'):
...
+
+    Returns:
+        str: sanitized filename
+    """
+    # TODO(vytas): max_length (int): Maximum length of the returned
+    #     filename. Should the returned filename exceed this restriction, it is
+    #     truncated while attempting to preserve the extension.
+    filename = unicodedata.normalize('NFKD', filename)
+    if filename.startswith('.'):
+        filename = filename.replace('.', '_', 1)
```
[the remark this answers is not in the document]
vytas7: Moreover, it's a special case in a non-critical porcelain function :slightly_smiling_face:
`filename.replace('.', '_', 1)` looks more æsthetically pleasing, and it performs better on 3.8, so let's keep it.

falcon/media/multipart.py
vytas7: The bad part is I don't remember now. :grimacing: 

Maybe we should just ditch this setting, and always allow all [standard encodings allowed by Python](https://docs.python.org/3.8/library/codecs.html#standard-encodings)?

falcon/util/misc.py
```diff
@@ -305,3 +308,25 @@ def get_http_status(status_code, default_reason='Unknown'):
...
+        filename (str): Arbitrary filename input from the request, such as
+            multipart form filename field.
+
+    Returns:
+        str: sanitized filename
+    """
+    # TODO(vytas): max_length (int): Maximum length of the returned
+    #     filename. Should the returned filename exceed this restriction, it is
+    #     truncated while attempting to preserve the extension.
+    filename = unicodedata.normalize('NFKD', filename)
```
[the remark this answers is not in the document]
vytas7: I'm not sure it the choice of `NFKD` contributes anything to security. Added an alternative explanation with examples.

tests/test_streams.py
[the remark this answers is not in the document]
vytas7: Added a couple of simple random forms, a valid and an invalid one.
A hardened test suite is still WiP.

falcon/cyutil/streams.pyx
```diff
@@ -0,0 +1,312 @@
...
+                                self._read_func(read_amount))
+                self._buffer_len = len(self._buffer)
+                self._buffer_pos = 0
+            self._max_bytes_remaining -= read_amount
+
+        return self._buffer[self._buffer_pos:self._buffer_pos + amount]
+
+    cdef Py_ssize_t _normalize_size(self, amount):
+        cdef Py_ssize_t result
+        cdef Py_ssize_t max_amount = self._max_bytes_remaining + self._buffer_len
```
vytas7: This was not fixed in comparison to the Python version (missing support for `self._buffer_pos`).
I'll fix and add more tests.

falcon/util/streams.py
vytas7: This approach is ported from `BoundedStream`:
```python
        # NOTE(kgriffs): Default to reading all remaining bytes if the
        # size is not specified or is out of bounds. This behaves
        # similarly to the IO streams passed in by non-wsgiref servers.
        if (size is None or size == -1 or size > self._bytes_remaining):
            size = self._bytes_remaining

        self._bytes_remaining -= size
        return target(size)
```

falcon/util/misc.py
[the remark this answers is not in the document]
kgriffs: I vote for returning an empty string and documenting that. It's easy enough for someone to key off of that and use their own default logic in that case or raise an error. If they don't have any handling for it, when they go to save it somewhere the thing they are using should refuse to write the data...which is actually IMHO what you want since it means you got some likely malicious input. Actually, with that in mind, maybe we should actually be raising an error in this case rather than an empty string or default name?

falcon/media/multipart.py
kgriffs: I guess this is not critical. It just seems like this is a common pattern (read_until followed by consuming the delimiter), hence the suggestion.

falcon/util/streams.py
kgriffs: OK, seems like we need to fix it over there as well then.

falcon/util/streams.py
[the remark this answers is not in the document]
kgriffs: I suppose we could just make sure the docs make it clear that people should avoid reading an entire stream in one call (vs. piping it to the destination) since it could eat up a lot of memory. But some people will still ignore that or not read the docs closely. So I wonder if we could do something adaptive? Read the first few chunks in to the list, but at some threshold switch over to to extending a `bytearray`?

falcon/media/multipart.py
vytas7: I do like your suggestion, I just haven't distilled the updated design and implementation yet since the edge case behaviour will likely change as a result as well. For the good, I think.

falcon/media/multipart.py
kgriffs: Yeah, I think that would work. It should be more than sufficient, and as long as our docs state this fact, there shouldn't be any big surprises for the developer.

falcon/util/streams.py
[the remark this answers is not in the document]
vytas7: I was benchmarking a bit and `io.BytesIO` surprisingly seems to do the right thing™ bringing the best of both worlds, so I'll probably incorporate it into the solution one or another way.

falcon/util/misc.py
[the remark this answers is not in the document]
vytas7: I went on to implement the exception path, thanks!

falcon/util/streams.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class BufferedStream:
```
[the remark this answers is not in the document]
vytas7: My undercover plan was to phase out `BoundedStream` (in favour of this class) at some point, since this class is supposed to be its superset (backed by Cython on some platforms).

I can rename it to `BufferedReader`, that makes sense.

falcon/media/multipart.py
vytas7: We'll somehow need to document this away and make it clear it is a Handler preference, not a developer preference.
Because if a `BufferedReader` has been used, it has no mechanism to inject buffered data back into the WSGI input stream. So it makes no sense not to exhaust as the data state would be undefined.

I'll remove this redundant exhaust from here.

falcon/util/streams.py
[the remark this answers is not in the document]
vytas7: Done renaming.

falcon/util/streams.py
vytas7: No. This comment was correct.

The condition above, however, should have been (less than or equal to), making it quite an ugly bug.
Found that by simulating a glitchy WSGI stream in the now improved unit tests, fixed.

falcon/util/reader.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def exhaust(self):
+        self.pipe()
+
+    def delimit(self, delimiter):
+        read = functools.partial(self.read_until, delimiter)
+        return type(self)(read, self._normalize_size(None), self._chunk_size)
+
+    def readline(self, size=-1):
+        return self.read_until(b'\n', size) + self.read(1)
```
vytas7: This can get a bit funky in edge cases, clean up is WiP.

falcon/util/streams.py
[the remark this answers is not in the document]
vytas7: Added a shim using `BytesIO`.

vytas7:
@kgriffs @nZac @jmvrbanac I'm still working on hardening the unit test suite, but I don't think the review should block on that. Ready to go for the second round!!!

Furthermore, the default `BufferedReader` chunk size is still TBD. Opinions are very welcome!
Some data to assist in choosing: https://travis-ci.org/vytas7/falcon-multipart-tests/jobs/644933725
I'm a bit divided between 32 KiB (currently presented in this PR) and 64 KiB.
64 KiB seems to give the best performance, however, with Cython helping to do all that pesky offset match virtually "for free" (compared to Python statements), the difference is not huge between the two. And I'm not a fan of buffer bloat.

kgriffs:
> I'm a bit divided between 32 KiB (currently presented in this PR) and 64 KiB.

32K seems like a good default in terms of balancing speed and memory usage. I say stick with that but also document that the app developer may want to change `DEFAULT_CHUNK_SIZE` (or perhaps this should be a `RequestOptions` field?) to 64K if they frequently read large file uploads and have enough RAM capacity to handle it.

Also, should we be worried about any potentially nasty surprises if the value of `io.DEFAULT_BUFFER_SIZE` ever changes?

vytas7:
> 32K seems like a good default in terms of balancing speed and memory usage. I say stick with that but also document that the app developer may want to change `DEFAULT_CHUNK_SIZE` (or perhaps this should be a `RequestOptions` field?) to 64K if they frequently read large parts (file uploads) and have enough RAM capacity to handle it.

:+1: Sounds good.
I've stashed this changed locally so far though; I need to take a look again with a fresh head and see what's the best way to make `RequestOptions` accessible from media handlers. Atm the `req` object is not passed down to these.

> Also, should we be worried about any potentially nasty surprises if the value of `io.DEFAULT_BUFFER_SIZE` ever changes?

Good point. I changed it to an explicit number.

Issue #1194, opened 2018-01-05
Consider adding basic CORS support directly into the framework.


kgriffs:
When dealing with APIs, supporting CORS is a common use case. It may make sense to include basic CORS functionality that allows access from any origin to the entire API.

Such a feature might be implemented along the lines of the following middleware component (albeit integrated directly into ``falcon.API``):

```py
# NOTE: Not fully tested

class CORSComponent(object):
    def process_response(self, req, resp, resource, req_succeeded):
        resp.set_header('Access-Control-Allow-Origin', '*')

        if (req_succeeded
            and req.method == 'OPTIONS'
            and req.get_header('Access-Control-Request-Method')
        ):
            # NOTE(kgriffs): This is a CORS preflight request. Patch the
            #   response accordingly.

            allow = resp.get_header('Allow')
            resp.delete_header('Allow')

            allow_headers = req.get_header(
                'Access-Control-Request-Headers', 
                default='*'
            )

            resp.set_headers((
                ('Access-Control-Allow-Methods', allow),
                ('Access-Control-Allow-Headers', allow_headers),
                ('Access-Control-Max-Age', '86400'),  # 24 hours
            ))
```

santeyio:
I'll give this a shot

foresmac:
I don't think this method makes sense; you'd have to already have an `on_options` method that generates a non-405 response. I think we might have used a mixin or middleware for this on a project; let me see if I can dig up the code.

foresmac:
Ok, yeah, we used a mixin like the following:
```python
class PreflightOptions(object):

    def on_options(self, req, res, *args, **kwargs):
        res.status = falcon.HTTP_OK
```

But the CORS middleware worked on `process_resource`, so it got routed to a valid view.

Then we check the origin domain against a white list (`ALLOWED_ORIGINS`) and/or a regex (`ALLOWED_ORIGINS_REGEX`) defined elsewhere (we use a standard `app/config.py` file).

```python
def is_allowed_origin(origin):
    url = urlparse(origin)
    if url.netloc.split(':')[0] in ALLOWED_ORIGINS:
        return True
    else:
        for pattern in ALLOWED_ORIGINS_REGEX:
            if re.match(pattern, origin):
                return True
    return False
```

kgriffs:
See also: https://github.com/falconry/falcon/issues/1220

Mumrau:
Will this still be added in falcon 2.0?

vytas7:
[...]

PR #1591, opened 2019-10-25
doc(CORS): Add news fragment for issue #1194

kgriffs:
(Also updates other docs to Generalize to "all responses".)

### DIFF
--- a/docs/_newsfragments/1194-cors.feature.rst
+++ b/docs/_newsfragments/1194-cors.feature.rst
@@ -1,3 +1,3 @@
 A new kwarg, ``cors_enable``, was added to the :class:`falcon.API` initializer.
 ``cors_enable`` can be used to enable a simple blanket CORS policy for all
-routes. (See also: :ref:`cors`.)
+responses. (See also: :ref:`cors`.)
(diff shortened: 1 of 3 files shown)


docs/_newsfragments/1194-cors.feature.rst
```diff
@@ -0,0 +1,3 @@
+A new kwarg, ``cors_enable``, was added to the :class:`falcon.API` initializer.
+``cors_enable`` can be used to enable a simple blanket CORS policy for all
+routes. (See also: :ref:`cors`.)
```
vytas7: A minor nitpick. Could we possibly generalize _all routes_ to _all responses_?
Since the CORS middleware would also apply to sink output, and pretty much almost everything barring higher hierarchy `process_response` middlewares.

kgriffs: Good point.

kgriffs: Updated.

vytas7: Good catch wrt other places inside source code :+1: When suggesting the change I was totally focusing on just the towncrier note

docs/_newsfragments/1194-cors.feature.rst
```diff
@@ -0,0 +1,3 @@
+A new kwarg, ``cors_enable``, was added to the :class:`falcon.API` initializer.
```
nZac: Might want to change this to `falcon.App` depending on how #1592 lands.

vytas7: Let's handle this in #1592 IMHO (serializable workflow :slightly_smiling_face: )

PR #1598, opened 2019-10-31
doc(changelog): Add supported platforms section for 3.0.0

kgriffs:


docs/changes/3.0.0.rst
```diff
@@ -1,6 +1,16 @@
 Changelog for Falcon 3.0.0
 ==========================
 
+Changes to Supported Platforms
```
vytas7: I'm not sure how this integrates with Towncrier.

@nZac could you shed some light on this? Can we keep adding stuff to this document manually, and then somehow make `towncrier` append/insert to this document?

nZac: I will have to check but the towncrier quickstart says you can add `.. towncrier release notes start` to the document and that is where the notes will go https://pypi.org/project/towncrier/

kgriffs: Yeah, I wasn't sure about that either. But if there is a better way just let me know and I'll refactor. /me is being lazy by not looking it up himself.

vytas7: It seems we also need to remove the title "Changelog for Falcon 3.0.0" from this file, and set `title_format` in `pyproject.toml`, or somehow disable title generation, otherwise they get duplicated.

`towncrier`'s README says you can set simply set `title_format` to `false` to disable title generation, however it was exploding for me with the latest Towncrier from PyPi... That is where I stopped digging.

kgriffs: OK gents, hopefully I've addressed these comments. Please take another look.

pyproject.toml
```diff
@@ -3,3 +3,9 @@
     package_dir = ""
     filename = "docs/changes/3.0.0.rst"
     directory = "docs/_newsfragments"
+
+    # TODO(kgriffs): Set this to false when the next major version of
+    #   towncrier is released (post 19.2). For now, we just have to
+    #   remember to remove the extra newlines in the rendered changelog.
+    title_format = ""
+
```
vytas7: FWIW I filed this as https://github.com/hawkowl/towncrier/issues/164

PR #1601, opened 2019-10-31
chore(ci): promote CPython 3.8.0 to the "default" Tox/CI interpreter

vytas7:
Here we promote CPython 3.8.0 to the "default" Tox/CI interpreter where applicable.
Still left out are our benchmark Docker images.

CPython 3.8.0 is still a bit fresh at the time of writing, however, I expect it to become more widespread in the wild by the time Falcon 3.0 [final] is released.


tox.ini
```diff
@@ -150,18 +149,18 @@ commands = flake8 examples \
 # For viewing environ dicts generated by various WSGI servers
 # --------------------------------------------------------------------
 
-[testenv:py37_dump_gunicorn]
-basepython = python3.7
+[testenv:py38_dump_gunicorn]
+basepython = python3.8
```
vytas7: Note that `gunicorn` < 20.0 emits a warning upon being run with CPython 3.8.0:
```
RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
```

This should be FAPP treated as cosmetic, and a fix to silence it is under way: https://github.com/benoitc/gunicorn/issues/2091

PR #1603, opened 2019-11-02
feat(API): on exception, select most specific error handler available

csojinb:
# Summary of Changes

Rather than selecting error handlers in LIFO order of registration, select the error handler corresponding to the nearest direct ancestor of the exception type raised. So, for example, if an app only adds a custom error handler for the Python ``Exception`` class, and an ``HTTPForbidden`` error is raised, then we use the default handler for ``HTTPError`` rather than the more general ``Exception`` handler (which is the pre-existing behavior). 

This is implemented by storing error handlers on the API object as a dict rather than a list and looking them up using the method resolution order attribute (`__mro__`) on the raised exception class. 

### NOTE: 
~~This commit only includes the actual implementation and does not address testing or documentation. I am seeking implementation feedback before completing those additional changes.~~ Tests and documentation have been added. Thanks @vytas7 for implementation notes!

### BREAKING CHANGE: 
Registration of a new error handler for type `E` will no longer override previously-registered error handlers for subclasses of type `E`. Registration order will no longer matter *except* when multiple error handlers are registered for the exact same exception type, in which case the most recently registered error handler overrides the previous ones. 

# Related Issues

Addresses #1514.

~~I believe it would also allow reversion of https://github.com/falconry/falcon/pull/1599.~~ Not true. I misremembered why #1599 was needed.

# Pull Request Checklist

### DIFF
--- a/docs/_newsfragments/1603-smart-exception-handler-precedence.feature.rst
+++ b/docs/_newsfragments/1603-smart-exception-handler-precedence.feature.rst
@@ -1,0 +1,1 @@
+Exceptions are now handled by the registered handler for the most specific matching exception class, rather than in reverse order of registration. "Specificity" is determined by the method resolution order of the raised exception type. (See :meth:`~.App.add_error_handler` for more details.)
--- a/falcon/app.py
+++ b/falcon/app.py
@@ -827,10 +827,9 @@ class App:
 
                 return True
 
-        # NOTE(kgriffs): No error handlers are defined for ex
-        # and it is not one of (HTTPStatus, HTTPError), since it
-        # would have matched one of the corresponding default
-        # handlers.
+        # NOTE(kgriffs): No error handlers are defined for ex and it
+        # is not one of (HTTPStatus, HTTPError, Exception), since it
+        # would have matched one of the corresponding default handlers.
         return False
 
     # PERF(kgriffs): Moved from api_helpers since it is slightly faster
--- a/falcon/app.py
+++ b/falcon/app.py
@@ -207,7 +207,7 @@ class App:
         self._request_type = request_type
         self._response_type = response_type
 
-        self._error_handlers = []
+        self._error_handlers = {}
         self._serialize_error = helpers.default_serialize_error
 
         self.req_options = RequestOptions()
@@ -616,18 +616,11 @@ class App:
         except TypeError:
             exception_tuple = (exception, )
 
-        if all(issubclass(exc, BaseException) for exc in exception_tuple):
-            # Insert at the head of the list in case we get duplicate
-            # adds (will cause the most recently added one to win).
-            if len(exception_tuple) == 1:
-                # In this case, insert only the single exception type
-                # (not a tuple), to avoid unnnecessary overhead in the
-                # exception handling path.
-                self._error_handlers.insert(0, (exception_tuple[0], handler))
-            else:
-                self._error_handlers.insert(0, (exception_tuple, handler))
-        else:
-            raise TypeError('"exception" must be an exception type.')
+        for exc in exception_tuple:
+            if not issubclass(exc, BaseException):
+                raise TypeError('"exception" must be an exception type.')
+
+            self._error_handlers[exc] = handler
 
     def set_error_serializer(self, serializer):
         """Override the default serializer for instances of :class:`~.HTTPError`.
@@ -799,6 +792,22 @@ class App:
         self._compose_error_response(
             req, resp, falcon.HTTPInternalServerError())
 
+    def _find_error_handler(self, ex):
+        # NOTE(csojinb): The `__mro__` class attribute returns the method
+        # resolution order tuple, i.e. the complete linear inheritance chain
+        # ``(type(ex), ..., object)``. For a valid exception class, the last
+        # two entries in the tuple will always be ``BaseException``and
+        # ``object``, so here we iterate over the lineage of exception types,
+        # from most to least specific.
+
+        # PERF(csojinb): The expression ``type(ex).__mro__[:-1]`` here is not
+        # super readable, but we inline it to avoid function call overhead.
+        for exc in type(ex).__mro__[:-1]:
+            handler = self._error_handlers.get(exc)
+
+            if handler is not None:
+                return handler
+
     def _handle_exception(self, req, resp, ex, params):
         """Handle an exception raised from mw or a responder.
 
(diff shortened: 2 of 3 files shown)


falcon/api.py
```diff
@@ -856,10 +856,44 @@ def _get_body(self, resp, wsgi_file_wrapper=None):
...
 
             return iterable, None
 
         return [], 0
+
+    # ------------------------------------------------------------------------
+    # Helpers that do not require self
+    # ------------------------------------------------------------------------
+
+    # QUESTION(csojinb): Find out if this should go here or in `api_helpers` or
```
csojinb: I wasn't sure if the stuff in `api_helpers.py` is meant to be "public" or not? I don't see any other static methods here, so perhaps it is intended to collect all non-state-required helpers, even if they aren't meant to be "public". If that is the case, I can move this there.

falcon/api.py
```diff
@@ -786,22 +787,21 @@ def _handle_exception(self, req, resp, ex, params):
...
+                err_handler(req, resp, ex, params)
+            except HTTPStatus as status:
+                self._compose_status_response(req, resp, status)
+            except HTTPError as error:
+                self._compose_error_response(req, resp, error)
+
+            return True
+
+        # NOTE(kgriffs): No error handlers are defined for ex and it
+        # is not one of (HTTPStatus, HTTPError, Exception), since it
```
csojinb: I realized that this note didn't get updated in #1527. I guess it is not too late to add that change there, but I think doing so will override the approval?

vytas7: I was the only who approved, I can do it again :slightly_smiling_face: 
Not too late IMHO.

falcon/api.py
```diff
@@ -770,6 +764,13 @@ def _http_status_handler(self, req, resp, status, params):
     def _http_error_handler(self, req, resp, error, params):
         self._compose_error_response(req, resp, error)
 
+    def _find_error_handler(self, ex):
+        for exc in self._get_exception_lineage(ex):
```
vytas7: We should IMHO inline `type(ex).__mro__[:-1]` here for performance (maybe with a `PERF` note explaining why we are not reusing `get_exception_lineage`).
By saying for performance, I don't just mean avoiding an unnecessary function call which has become cheaper in CPython 3.7+, but also avoiding this check:
```python
        if not isinstance(ex, BaseException):
            raise TypeError('"exception" must be an exception type instance.')
```

It is important to check the type when adding handlers, but when handling an actual exception, we should presume it is of correct type :slightly_smiling_face: Even if in some unfathomable edge case it is not (old-style exceptions? although they were IIRC barred from Python 3), re-raising a `TypeError` would not be any more helpful.

csojinb: Sounds good. I pulled it out into a separate function for readability, but I think we can handle that fine with a `PERF` note as you suggested

falcon/api.py
csojinb: Heh. Funny how after months and months of that PR sitting there unchanged, I was too late to squeeze in one more change because I left my laptop alone for 4 hours :joy:

kgriffs: Sorry, I guess I jumped the gun. 😄

csojinb: No worries! I did say that it was ready to go.

csojinb:
Would this change warrant a towncrier entry?

falcon/api.py
```diff
@@ -532,6 +533,31 @@ def add_error_handler(self, exception, handler=None):
...
+
+            app = falcon.API()
+            app.add_error_handler(falcon.HTTPNotFound, custom_handle_not_found)
+            app.add_error_handler(falcon.HTTPError, custom_handle_http_error)
+            app.add_error_handler(Exception, custom_handle_uncaught_exception)
+            app.add_error_handler(falcon.HTTPNotFound, custom_handle_404)
+
+        QUESTION(csojinb): Do we want the usages of falcon.HTTP* below to include links to the error classes?
+
+        If an instance of ``falcon.HTTPForbidden`` is raised, it will be
```
csojinb: Open questions on the style in this section:

* Should I use single or double backticks for the handler function names? I couldn't decide whether they counted as "variables" or "inline code".
* Should the ussages of `falcon.HTTP*` in this example include `:class:` so that there's a link to the relevant part of the api doc? I didn't do that to start because the specific error classes used for the examples are fairly arbitrary.

falcon/api.py
```diff
@@ -532,18 +533,27 @@ def add_error_handler(self, exception, handler=None):
...
 
             app = falcon.API()
-            app.add_error_handler(Exception, custom_handle_uncaught_exception)
+            app.add_error_handler(falcon.HTTPNotFound, custom_handle_not_found)
             app.add_error_handler(falcon.HTTPError, custom_handle_http_error)
-            app.add_error_handler(CustomException)
+            app.add_error_handler(Exception, custom_handle_uncaught_exception)
+            app.add_error_handler(falcon.HTTPNotFound, custom_handle_404)
+
+        If an instance of ``falcon.HTTPForbidden`` is raised, it will be
```
csojinb: Open questions on the style in this section:

* Should I use single or double backticks for the handler function names? I couldn't decide whether they counted as "variables" or "inline code".
 * Should the ussages of falcon.HTTP* in this example include :class: so that there's a link to the relevant part of the api doc? I didn't do that to start because the specific error classes used for the examples are fairly arbitrary.

vytas7:
@csojinb Re Towncrier I would think, yes, we need an entry (or actually two?) since it is both a breaking change, and a new feature.

falcon/api.py
```diff
@@ -269,7 +269,8 @@ def __call__(self, env, start_response):  # noqa: C901
                     # next-hop child resource. In that case, the object
                     # being asked to dispatch to its child will raise an
                     # HTTP exception signalling the problem, e.g. a 404.
-                    responder, params, resource, req.uri_template = self._get_responder(req)
+                    responder, params, resource, req.uri_template = self._get_responder(
+                        req)
```
mikeyusko: Not a big deal, but do we need additional space for `self._get_responder`?

falcon/api.py
```diff
@@ -879,7 +898,8 @@ def _get_body(self, resp, wsgi_file_wrapper=None):
                     iterable = wsgi_file_wrapper(stream,
                                                  self._STREAM_BLOCK_SIZE)
                 else:
-                    iterable = helpers.CloseableStreamIterator(stream, self._STREAM_BLOCK_SIZE)
+                    iterable = helpers.CloseableStreamIterator(
+                        stream, self._STREAM_BLOCK_SIZE)
```
mikeyusko: And the same here:)

falcon/api.py
csojinb: You mean, move it to another line because it's past 80 characters? Or, are you asking if this change shouldn't have occurred?

I didn't intentionally make this change here, I think it must have been my auto-pep8. It looks like Falcon's pep8-checker rules allow 99-character lines, so probably I should revert this newline change. I guess my tool didn't pick up the config.

mikeyusko: Yeah, falcon allows 99-characters lines, [flake8 config](https://github.com/falconry/falcon/blob/master/tox.ini#L129).

In that case, I think would be better to revert that newline change

falcon/api.py
kgriffs: double backticks are fine, although I tend to add parens to the function name myself to help make it obvious what it is. 

Re class names, I'm fine either way TBH, since you are referencing example code here. But it couldn't hurt to add the `:class:` directive.

csojinb:
Yes, I can try to do it this weekend

csojinb:
Okay, I rebased, undid the overzealous pep8-ing, tweaked the `add_error_handler` docs, and added a towncrier fragment. I think that's everything. cc @kgriffs @vytas7

csojinb:
Oh, so it is. I guess kdiff made that decision for me and I didn't look closely enough. I'll fix it and re-push

csojinb:
@vytas7 fixed

csojinb:
Moving my copied TODO list out of the PR description into this comment:

### TODOs:
- [x] Add towncrier news fragments
- [x] Rebase on master (fix conflicts)
- [x] Tweak docstrings (optional)
- [x] Tweak line wrapping (optional)

(Turns out you can't just check off checkboxes in someone else's comment.)

kgriffs:
🎉

PR #1604, opened 2019-11-03
perf(parse_query_string): offer a faster version of parse_query_string in Cython

### LINKED ISSUE #1600: uri.parse_query_string() incorrectly parses an empty string
cf
```python
>>> from urllib import parse
>>> from falcon.util import uri
>>> parse.parse_qs('', keep_blank_values=True)
{}
>>> uri.parse_query_string('', keep_blank=True)
{'': ''}
```

The good ol' _an empty string splits to a list with one empty string item_ trick is biting us here as I understand.

vytas7:
I will provide some benchmarks showcasing that this is actually an improvement.
But the PR is hopefully otherwise ready.

I have also sneaked in a fix for #1600 to match Cython and Python URI utils behaviour:

Closes #1600


vytas7:
Note that if the approach is acceptable, we could in the future easily define `parse_query_bytes` or similar to further optimize deserialization directly from the payload bytes.

We could quickly do the ASCII check by simply verifying every quadword in the payload `& 0x8080808080808080LL` is zero or something in that vein.

falcon/cyutil/uri.pyx
```diff
@@ -0,0 +1,299 @@
...
+        #  This will be used to determine where to start decoding, if at all.
+        if partition < 0:
+            if encoded_start_key < 0:
+                encoded_start_key = pos
+        else:
+            if encoded_start_val < 0:
+                encoded_start_val = pos
+
+    if length > start:
+        if partition >= 0:
```
kgriffs: I know that it would probably require passing a long list of values to factor this out to a function, or passing in a struct-like state object, so it may not be worth DRYing this up. However, it may be worth prefixing this section and the section above with a comment reminding the reader to keep the code in the two locations in sync.

falcon/cyutil/uri.pyx
```diff
@@ -0,0 +1,299 @@
...
+    cdef Py_ssize_t pos
+    cdef int decoded
+
+    result = <unsigned char*> PyMem_Malloc(end - start)
+    if not result:
+        raise MemoryError()
+
+    try:
+        for pos in range(encoded_start, end):
+            if data[pos] != b'%':
```
kgriffs: Is the "no plus" specialization worth the performance increase? If so, it may be a good idea to add some inline comments that point out the differences so that if these functions are ever modified, it makes it more obvious what the developer has to do to specialize  ``cy_decode()``.

I'm all for violating DRY to get a significant speed boost as long as we are careful about the maintenance burden this might introduce _and_ we have comprehensive tests for all paths and a wide variety of inputs. That being said, I can't imagine this code changing that much over time.

falcon/util/uri.py
```diff
@@ -394,3 +404,19 @@ def unquote_string(quoted):
...
                           for q in tmp_quoted.split(r'\\')])
+
+
+# TODO(vytas): Restructure this in favour of a cleaner way to hoist the pure
+# Cython functions into this module.
+decode = _cy_decode or decode  # NOQA
+parse_query_string = _cy_parse_query_string or parse_query_string  # NOQA
+
+
+__all__ = [
```
kgriffs: +1 for defining ``__all__``; we need to start doing this in every module.

falcon/util/uri.py
```diff
@@ -264,7 +274,7 @@ def parse_query_string(query_string, keep_blank=False, csv=True):
     # both short and long query strings. Tested on CPython 3.4.
     for field in query_string.split('&'):
         k, _, v = field.partition('=')
-        if not (v or keep_blank):
+        if not v and (not keep_blank or not k):
```
kgriffs: Nice catch! One suggestion, though, would be to refactor the statement to reduce the number of boolean operations, e.g.:

```py
if not (v or (keep_blank and k)):
    continue
```

tests/test_utils.py
```diff
@@ -281,6 +290,19 @@ def test_parse_query_string(self):
...
         assert result['é'] == 'a=b'
 
+    @pytest.mark.parametrize('query_string,keep_blank,expected', [
+        ('', True, {}),
+        ('', False, {}),
+        ('flag1&&&&&flag2&&&', True, {'flag1': '', 'flag2': ''}),
+        ('flag1&&&&&flag2&&&', False, {}),
+        ('malformed=%FG%1%Hi%%%a', False, {'malformed': '%FG%1%Hi%%%a'}),
+        ('spade=♠&spade=♠', False, {'spade': ['♠', '♠']}),  # Unicode query
+    ])
```
kgriffs: Could you also test cases with strings that contain 2+ `=` characters in a row? Also, test strings that contain a field that starts with an `=` character (key is missing), either at the beginning of the string or immediately following an `&`. Besides these, feel free to dream up any other strange inputs a vulnerability scanner might throw at us.

falcon/cyutil/uri.pyx
vytas7: No, I don't think it's worth the performance increase (if any). I think it was actually the first version that I implemented, and then I had realized I also need to support the default "unquote plus" case too. Human workflow artefact, one can say :thinking:
Simple boolean switches and prediction branching check-ups are dirty fast on modern CPUs, particularly on non-RISC ones thanks to parallel execution pipelines. Sometimes even [too smart for their own good :slightly_smiling_face:](https://en.wikipedia.org/wiki/Spectre_(security_vulnerability)). I think given our default is to unquote `+`-es, I'll be able to DRY this without any noticeable impact.

falcon/cyutil/uri.pyx
```diff
@@ -0,0 +1,299 @@
...
+cdef EMPTY_STRING = u''
+
+
+cdef inline int cy_decode_hex(unsigned char nibble1, unsigned char nibble2):
+    return HEX_CHARS[(nibble1 << 8) | nibble2]
+
+
+cdef unicode cy_decode_no_plus(unsigned char* data, Py_ssize_t start,
+                               Py_ssize_t end, Py_ssize_t encoded_start):
+    if encoded_start < 0:
```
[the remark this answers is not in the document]
vytas7: Will improve.

falcon/cyutil/uri.pyx
vytas7: Agreed, this block grew a bit copypaste-ish. I'll see what I can do about it. In the worst case I'll add some comments indeed.

falcon/util/uri.py
kgriffs: (not a big deal; if the version as-is seems to be slightly faster in the common case I'm not opposed to keeping it)

vytas7: I'm admittedly a bit divided here... As discussed on Gitter, estimating that having a value is the most common case, my version actually seems to be outperforming anything else (CPython 3.8):
```python


>>> import timeit
>>> k = 'key'; v = 'value'; keep_blank = True
>>> timeit.timeit(lambda: not (v or (keep_blank and k)), number=int(1e8))
5.702389151003445
>>> timeit.timeit(lambda: not (v or (keep_blank and k)), number=int(1e8))
5.684955962999084
>>> timeit.timeit(lambda: not (v or (keep_blank and k)), number=int(1e8))
5.685243817999435
>>> timeit.timeit(lambda: not v and (not keep_blank or not k), number=int(1e8))
5.500417819006543
>>> timeit.timeit(lambda: not v and (not keep_blank or not k), number=int(1e8))
5.490288328001043
>>> timeit.timeit(lambda: not v and (not keep_blank or not k), number=int(1e8))
5.471835395997914
```

tests/test_utils.py
vytas7: Will do.
Indeed, it makes sense to have a very solid unit test selection here, due to the gravity of operating on raw pointers. If there is some shady edge case in Python code for one request in a million, no one is probably going to care, however getting a `SIGSEGV` in a web framework worker process is unlikely to go unnoticed :slightly_smiling_face:

vytas7:
@kgriffs Hopefully ready for another review now!

I haven't done anything about boolean operation grouping in the pure Python fix though. I'm thinking to myself it's largely irrelevant for CPython users since they should look to the presented cythonized variant anyway; so I'll benchmark performance characteristics on PyPy3 and make a decision based on that.

PR #1558, opened 2019-06-29
docs(tutorial): fix typos and failing test in the tutorial

cravindra:
docs(tutorial): fix typos and test in the tutorial

# Summary of Changes
There was a failing test because of breaking changes in `2.0.0` - check for an instance of `wsgiref.validate.InputWrapper` instead of `falcon.request_helpers.BoundedStream`.  Minor typo fix in the tutorial.

# Related Issues
[Test described in tutorial fails #1554](https://github.com/falconry/falcon/issues/1554)

# Pull Request Checklist

This is just a reminder about the most common mistakes.  Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md) at least once; it will save you a few review cycles!

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing to do.

- [x] Added **tests** for changed code.
- [x] Prefixed code comments with GitHub nick and an appropriate prefix.
- [x] Coding style is consistent with the rest of the framework.
- [x] Updated **documentation** for changed code.
    - [x] Added docstrings for any new classes, functions, or modules.
    - [x] Updated docstrings for any modifications to existing code.
    - [x] Added references to new classes, functions, or modules to the relevant RST file under `docs/`.
    - [x] Updated all relevant supporting documentation files under `docs/`.
    - [x] A copyright notice is included at the top of any new modules (using your own name or the name of your organization).
    - [x] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/directives.html?highlight=versionadded#directive-versionadded).
- [x] Changes (and possible deprecations) have [towncrier](https://pypi.org/project/towncrier/) news fragments under `docs/_newsfragments/`. (Run `towncrier --draft` to ensure it renders correctly.)

PR #1611, opened 2019-11-20
chore(tutorial): clean up & automate look tests

vytas7:
This PR builds further upon the _look_ tutorial improvements delivered by @cravindra  in https://github.com/falconry/falcon/pull/1558 , and aims to run these tests in Travis to make sure our tutorial stays valid.


examples/look/requirements/test
```diff
@@ -1,3 +1,5 @@
+gunicorn==19.9.0
```
vytas7: Note that our recommended way to run the _look_ example with `gunicorn` does not work on Gunicorn 20.0.0.

However, if I interpret https://github.com/benoitc/gunicorn/issues/2175 correctly, a fix may be already in `master` or under way.

jmvrbanac: For what it's worth, I generally import and use gunicorn directly in my apps when wrapping gunicorn. We might investigate doing that here as well.

vytas7: Yeah, that's generally a good approach. I have adopted it at work for some development servers (we use Nginx+uWSGI on prod though) as it's easy to toss in a static route, inject some dependencies or w/e needed for the dev environment.

However, I'm not sure that's a good approach here in this "look" tutorial as we recommend running gunicorn [like that](https://falcon.readthedocs.io/en/stable/user/tutorial.html#testing-your-application) and [like that](https://falcon.readthedocs.io/en/stable/user/tutorial.html#refactoring-for-testability). We could of course improve that, but IMHO running gunicorn in the command line is arguably easier for new users. Plus, it's much easier to swap out command-line unicorn to something else, for instance, waitress for Windows folks etc.

Issue #1519, opened 2019-05-06
Filter test suite warnings output by wsgiref to non-standard HTTP methods.


kgriffs:

vytas7:
Resolved fixed in https://github.com/falconry/falcon/pull/1618

PR #1618, opened 2019-12-02
test: Clean up pytest warnings and fix missing comma in WEBDAV_METHODS

kgriffs:
# Summary of Changes

This patch cleans up the less-than-useful warnings that are typically reported by pytest runs whenever there are any test failures. Along the way, I discovered that the `WEBDAV_METHODS` list was missing a comma.

### DIFF
--- a/falcon/constants.py
+++ b/falcon/constants.py
@@ -25,7 +25,7 @@ WEBDAV_METHODS = [
     'PROPPATCH',
     'REPORT',
     'UNCHECKIN',
-    'UNLOCK'
+    'UNLOCK',
     'UPDATE',
     'VERSION-CONTROL',
 ]
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,4 +8,6 @@ universal = 1
 test=pytest
 
 [tool:pytest]
-addopts = --ignore falcon --ignore examples
+filterwarnings =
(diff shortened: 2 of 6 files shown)


falcon/constants.py
```diff
@@ -25,7 +25,7 @@
     'PROPPATCH',
     'REPORT',
     'UNCHECKIN',
-    'UNLOCK'
+    'UNLOCK',
```
vytas7: Ouch, so it means what we were supporting before was actually `UNLOCKUPDATE`? :thinking:

setup.cfg
```diff
@@ -8,4 +8,6 @@ universal = 1
 test=pytest
 
 [tool:pytest]
-addopts = --ignore falcon --ignore examples
+filterwarnings =
+    ignore:Unknown REQUEST_METHOD. '(CONNECT|DELETE|GET|HEAD|OPTIONS|PATCH|POST|PUT|TRACE|CHECKIN|CHECKOUT|COPY|LOCK|MKCOL|MOVE|PROPFIND|PROPPATCH|REPORT|UNCHECKIN|UNLOCK|UPDATE|VERSION-CONTROL)':wsgiref.validate.WSGIWarning
```
kgriffs: As it happens, pytest does its own warnings collection and so our `warnings.filterwarnings()` call doesn't help in this case.

falcon/constants.py
kgriffs: Yep. 💩

vytas7:
Fixes #1519

Issue #1546, opened 2019-05-26
In `Response.set_headers` method, when receiving mappings, check for interface compliance rather than concrete class


tribals:
There is plenty of ways to implement `dict`-like behavior. In general, it is not necessary to subclass `dict` directly.

We must give the developers stable interface if he/she wants to pass their own mappings. For example, famous `requests` implements their own case insensitive `dict`-like mappnig for headers. If I pass such headers to `set_headers`, code breaks because of this check:

https://github.com/falconry/falcon/blob/3b8ca25a78790b343d1c875fe9fd5e7e5742a8f9/falcon/response.py#L625-L626

Better approach is to check for interface compliance:

```
from collections.abc import Mapping

if isinstance(headers, Mapping):
    headers = headers.items()
```

kgriffs:
That makes sense. In fact, we could be even more lenient and just check whether the object implements `items()`. Regardless, the docs also need to be updated. We should also grep the code for `isinstance()` and see if there are other places that could use a similar treatment.

mikeyusko:
@tribals @kgriffs 

Hm, that's a common use case? in order to satisfy `abc.Mapping` interface, a user should to do something like that:
```
class MyMapping(abc.Mapping):
    def __getitem__(self, item):
        ... code

    def __iter__(self):
        ... code 

    def __len__(self):
        .... code
```
If that makes sense, I can send a PR.

vytas7:
@mikeyusko What to you think of @kgriffs' idea to check if the object quacks `items()`?

mikeyusko:
@vytas7 Yeah, if it is a use-case it would be better to implement own falcon `mapping`  class.

e.g
```
class Mapping:
      ... some other stuff

     def items():
          ... code
------------------------------------------
import falcon

class CustomHeaders(falcon.Mapping):
    def items():
        ... code
```

mikeyusko:
@vytas7 I'll work on it.

tribals:
@mikeyusko I think, the idea is exactly the opposite: we shouldn't implement our own dict-like class. There are too many of them already. In fact, it seems like every HTTP framework implements its own.

But there is completely no need.

What we want here, is just to iterate over key-value pairs. And this is exactly what .items() do. It's a very stable interface. So, IMO, we can depend on it.

So, code should look like this:

    try:
        for k, v in headers.items():
            # do what need to be done
    except AttributeError as e:
        raise ValueError('Headers must be iterable') from e
That will be enough.

[...]

PR #1630, opened 2020-01-04
Rewrites behaviour of Response.set_headers() method

mikeyusko:
# Summary of Changes

Changed behavior in `Response.set_headers()` the method, now we're checking for interface compliance, instead of check whether an object is an instance of ``dict``.

Also, new changes allow passing a custom class which will be setting headers for a response.
e.g
```python
class CustomHeaders:
    # ... custom flow etc.
    def items(self):
        return [('header', 'value')]  # or return {'header': 'value'}


class Resource:
    def on_get(self, req, resp):
        custom_headers = CustomHeaders()

        resp.set_headers(custom_headers)
```

# Related Issues

https://github.com/falconry/falcon/issues/1546

# Pull Request Checklist

- [ ] Changes (and possible deprecations) have [towncrier](https://pypi.org/project/towncrier/) news fragments under `docs/_newsfragments/`. (Run `towncrier --draft` to ensure it renders correctly.)

### DIFF
--- a/docs/_newsfragments/1630-set-response.feature.rst
+++ b/docs/_newsfragments/1630-set-response.feature.rst
@@ -1,0 +1,2 @@
+The :meth:`~.Response.set_headers` method now can retrieve an instance of an object
+which implements ``items()`` method.
--- a/docs/_newsfragments/1630-set-response.feature.rst
+++ b/docs/_newsfragments/1630-set-response.feature.rst
@@ -1,2 +1,2 @@
-The :meth:`~.Response.set_headers` method now can retrieve an instance of an object
-which implements ``items()`` method.
+The :meth:`~.Response.set_headers` method now accepts an instance of any dict-like
+object that implements an ``items()`` method.
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -662,8 +662,8 @@ class Response:
 
         header_items = getattr(headers, 'items', None)
 
-        if header_items and callable(header_items):
-            headers = headers.items()
+        if callable(header_items):
+            headers = header_items()
 
         # NOTE(kgriffs): We can't use dict.update because we have to
         # normalize the header names.
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -635,21 +635,25 @@ class Response:
             :meth:`~.append_header` or :meth:`~.set_cookie`.
 
         Args:
-            headers (dict or list): A dictionary of header names and values
+            headers (dict or list or object): A dictionary of header names and values
                 to set, or a ``list`` of (*name*, *value*) tuples. Both
                 *name* and *value* must be of type ``str`` and
-                contain only US-ASCII characters.
+                contain only US-ASCII characters, also an instance of an object
+                which implements ``items()`` method can be passed, the method should
+                return a ``dict`` or ``list`` of ``tuple``.
 
                 Note:
                     Falcon can process a list of tuples slightly faster
                     than a dict.
 
         Raises:
-            ValueError: `headers` was not a ``dict`` or ``list`` of ``tuple``.
-
+            ValueError: `headers` was not a ``dict`` or ``list`` of ``tuple`
+                         or an instance of an object which implements ``items()`` method.
         """
 
-        if isinstance(headers, dict):
+        header_items = getattr(headers, 'items', None)
+
+        if header_items and callable(header_items):
             headers = headers.items()
 
         # NOTE(kgriffs): We can't use dict.update because we have to
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -635,20 +635,18 @@ class Response:
             :meth:`~.append_header` or :meth:`~.set_cookie`.
 
         Args:
-            headers (dict or list or object): A dictionary of header names and values
-                to set, or a ``list`` of (*name*, *value*) tuples. Both
-                *name* and *value* must be of type ``str`` and
-                contain only US-ASCII characters, also an instance of an object
-                which implements ``items()`` method can be passed, the method should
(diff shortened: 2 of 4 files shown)


falcon/response.py
```diff
@@ -635,21 +635,25 @@ def set_headers(self, headers):
             :meth:`~.append_header` or :meth:`~.set_cookie`.
 
         Args:
-            headers (dict or list): A dictionary of header names and values
+            headers (dict or list or object): A dictionary of header names and values
```
kgriffs: I think we can update the docstring to be more general. The user may pass either a dict-like object or an iterable of `[name, value]` two-member iterables. I.e.:

```
headers (Iterable[[str, str]]): An iterable of ``[name, value]`` two-member 
    iterables, or a dict-like object that implements an ``items()`` method...
```

falcon/response.py
```diff
@@ -635,21 +635,25 @@ def set_headers(self, headers):
...
                 to set, or a ``list`` of (*name*, *value*) tuples. Both
                 *name* and *value* must be of type ``str`` and
-                contain only US-ASCII characters.
+                contain only US-ASCII characters, also an instance of an object
+                which implements ``items()`` method can be passed, the method should
+                return a ``dict`` or ``list`` of ``tuple``.
 
                 Note:
                     Falcon can process a list of tuples slightly faster
```
kgriffs: Change to:

> Falcon can process an iterable of tuples slightly faster...

docs/_newsfragments/1630-set-response.feature.rst
```diff
@@ -0,0 +1,2 @@
+The :meth:`~.Response.set_headers` method now can retrieve an instance of an object
```
kgriffs: Suggested edit:

```
The :meth:`~.Response.set_headers` method now accepts an instance of any dict-like object that implements an ``items()`` method.
```

falcon/response.py
```diff
@@ -635,21 +635,25 @@ def set_headers(self, headers):
...
-            ValueError: `headers` was not a ``dict`` or ``list`` of ``tuple``.
-
+            ValueError: `headers` was not a ``dict`` or ``list`` of ``tuple`
+                         or an instance of an object which implements ``items()`` method.
         """
 
-        if isinstance(headers, dict):
+        header_items = getattr(headers, 'items', None)
+
+        if header_items and callable(header_items):
```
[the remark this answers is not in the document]
mikeyusko: Yeah, I'll add.

falcon/response.py
mikeyusko: Nice suggestion, I'll rewrite the docstring.

falcon/response.py
mikeyusko: Fixed.

docs/_newsfragments/1630-set-response.feature.rst
mikeyusko: Fixed.

falcon/response.py
```diff
@@ -646,21 +646,23 @@ def set_headers(self, headers):
...
-            ValueError: `headers` was not a ``dict`` or ``list`` of ``tuple``.
-
+            ValueError: `headers` was not a ``dict`` or ``list`` of ``tuple``
+                         or ``Iterable[[str, str]]``.
         """
 
-        if isinstance(headers, dict):
+        header_items = getattr(headers, 'items', None)
+
+        if header_items and callable(header_items):
```
vytas7: A less obvious further optimization for the mapping case, could we potentially just check `if callable(header_items):`?
```python
>>> callable(None)
False
```
This approach would perform slightly worse when just an iterable is provided though.
We would also need to add a `NOTE` explaining the above reasoning if we go this route.

OTOH maybe just checking for callable is also more correct? Consider the following example:
```python
>>> class HeaderFactory:
...     def __call__(self):
...         return (('X-one', '1'), ('X-two', '2'))
...     def __bool__(self):
...         return False
... 
>>> class Example:
...     def __init__(self):
...         self.items = HeaderFactory()
... 
>>> example = Example()
>>> example.items()
(('X-one', '1'), ('X-two', '2'))
>>> if example.items and callable(example.items):
...     raise NotImplementedError("I'm unreachable.")
... 
>>> callable(example.items)
True
```

If we keep the current solution we should IMHO document that `items()` is not called unless it is a truthy value.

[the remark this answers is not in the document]
mikeyusko: Yeah, for sure, it will save nanoseconds, I'll change it.

mikeyusko: I think the approach with `items()` is a bit cleaner, and won't confused people, also the approach is straightforward and easy to understand, I prefer things that are easy for understanding.

Also, if we will switch to the suggested approach, I'm not sure that we'll take a lot of advantages.

Also, we will do the same thing with `getattr()`, since we're supporting `list of tuples` it also will require to check if the object has the attribute `items()` and etc.

So, I vote for the current approach;)

vytas7: @mikeyusko I think you might have misunderstood me. Probably my fault too, I got a bit carried away by those edge cases :thinking:

I was trying to compare the two following alternatives:
(current)
```python
if header_items and callable(header_items):
    # get items()
```
(proposed for discussion)
```python
if callable(header_items):
    # get items()
```

mikeyusko: @vytas7 Sorry for the misunderstanding:)

Yes, I also agree here, we can easy check `callable` and will be enough for our case, thanks!

I'll push the proposed changes tomorrow.

vytas7: @kgriffs any preference on `if header_items and callable(header_items):` vs `if callable(header_items):`?

kgriffs: I'm fine with `if callable(header_items)`.

PR #1644, opened 2020-01-21
fix(testing.helpers): flip_coin() sometimes leads to coverage errors

kgriffs:


kgriffs:
> LGTM 👍
> 
> Going forward, we should probably consider mentioning this switching in the `ASGIRequestEventEmitter` docs though so that our users know what to expect.

Good idea. I added note to my docs "to do" list.

PR #1646, opened 2020-01-22
feat(testing): Accept a content_type kwarg when simulating requests

kgriffs:
# Summary of Changes

It is often the case that the only reason to pass the headers kwarg to
simulate_request() is to set the content type. This patch adds a
convenience kwarg for this use case.

# Related Issues

N/A

# Pull Request Checklist

### DIFF
--- a/falcon/testing/client.py
+++ b/falcon/testing/client.py
@@ -254,9 +254,9 @@ class Result:
 #   appears to be "hanging", which might indicates that the app is
 #   not handling the reception of events correctly.
 def simulate_request(app, method='GET', path='/', query_string=None,
-                     headers=None, body=None, json=None, file_wrapper=None,
-                     wsgierrors=None, params=None, params_csv=True,
-                     protocol='http', host=helpers.DEFAULT_HOST,
+                     headers=None, content_type=None, body=None, json=None,
+                     file_wrapper=None, wsgierrors=None, params=None,
+                     params_csv=True, protocol='http', host=helpers.DEFAULT_HOST,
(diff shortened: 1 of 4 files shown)


falcon/testing/client.py
```diff
@@ -314,8 +319,8 @@ def simulate_request(app, method='GET', path='/', query_string=None,
             may be passed, in which case it will be used as-is.
         json(JSON serializable): A JSON document to serialize as the
             body of the request (default: ``None``). If specified,
-            overrides `body` and the Content-Type header in
-            `headers`.
+            overrides `body` and forces the Content-Type header to
```
vytas7: The `json` docstring clarification should be replicated to other `simulate_*` friends (where applicable).

Furthermore, would it be helpful to explicitly state that by forcing we mean it also takes precedence even over the `content-type` parameter?

kgriffs: done

falcon/testing/client.py
```diff
@@ -371,6 +377,10 @@ def simulate_request(app, method='GET', path='/', query_string=None,
             prefix=False,
         )
 
+    if content_type is not None:
```
mikeyusko: `if content_type is not None` -> `if content_type`?

kgriffs: The caller probably won't ever need to set the header to an empty string, but it may make sense to allow it just in case.

mikeyusko: Understood, in that case, it make sense 👍

PR #1647, opened 2020-01-22
feat(responders): Raise a specialized subclass of HTTPNotFound

kgriffs:
When no route matches a request, raise a subclass of HTTPNotFound so that
a custom error handler can distinguish that specific case if desired.

(This was inspired by a recent chat with a user in the community.)

### DIFF
--- a/docs/api/errors.rst
+++ b/docs/api/errors.rst
@@ -87,20 +87,107 @@ Mixins
 Predefined Errors
 -----------------
 
-.. automodule:: falcon
-    :noindex:
-    :members: HTTPBadRequest,
-        HTTPInvalidHeader, HTTPMissingHeader,
-        HTTPInvalidParam, HTTPMissingParam,
-        HTTPUnauthorized, HTTPForbidden, HTTPNotFound, HTTPMethodNotAllowed,
-        HTTPNotAcceptable, HTTPConflict, HTTPGone, HTTPLengthRequired,
-        HTTPPreconditionFailed, HTTPPayloadTooLarge, HTTPUriTooLong,
-        HTTPUnsupportedMediaType, HTTPRangeNotSatisfiable,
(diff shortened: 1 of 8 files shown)


docs/api/errors.rst
```diff
@@ -87,20 +87,107 @@ Mixins
 Predefined Errors
 -----------------
 
-.. automodule:: falcon
```
kgriffs: This was preventing other docs from being able to create reference links to the error class docs.

kgriffs:
Good ideas, I'll work on those.

kgriffs:
> * We should adapt tests to verify the new behaviour, i.e. that the specialized subclass is raised, ideally in a scenario where the responder, if routed, would raise a plain `HTTPNotFound`.
> * `docs/api/routing.rst` needs to be revised
> * `docs/user/faq.rst` needs to be revised (`How can I write a custom handler for 404 and 500 pages in falcon?`)

@vytas7  I've attempted to address these points. Let me know what you think!

docs/user/faq.rst
```diff
@@ -436,22 +436,23 @@ around:
...
-responder that simply raises an instance of :attr:`falcon.HTTPNotFound`. You
-can use :meth:`falcon.App.add_error_handler` to register a custom error handler
-for this exception type. Alternatively, you may be able to configure your web
-server to transform the response for you (e.g., using Nginx's ``error_page``
-directive).
+responder that simply raises an instance of :class:`~.HTTPRouteNotFound`, which
+the framework will in turn render as a 404 response. You can use
+:meth:`falcon.App.add_error_handler` to override the default handler for this
+exception type (or for its parent type,  :class:`~.HTTPNotFound`).
+Alternatively, you may be able to  configure your web server to transform the
```
vytas7: Not that it matter much in reStructured Text, but there in an extraneous space after `able to`.

kgriffs: fixed

Issue #777, opened 2016-05-10
Make error initializers more consistent


kgriffs:
This issue involves reworking error class initializer method signatures to be more consistent and predictable. Each `__init__` should order args and kwargs as follows:
1. Required, specialized args. Most error classes will not have any required args. 
2. `title=None`
3. `description=None`
4. `**kwargs`

kgriffs:
Moving this to 3.0 in the interest of getting 2.0 out sooner.

CaselIT:
I'll try to work on this

PR #1621, opened 2019-12-15
doc(changelog): reorganize towncrier

csojinb:
# Summary of Changes

Reorganize towncrier to match previous changelogs

# Related Issues

N/A

# Pull Request Checklist

This is just a reminder about the most common mistakes.  Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md) at least once; it will save you a few review cycles!

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing to do.

- [x] Added **tests** for changed code.
- [x] Prefixed code comments with GitHub nick and an appropriate prefix.
- [x] Coding style is consistent with the rest of the framework.
- [x] Updated **documentation** for changed code.
    - [x] Added docstrings for any new classes, functions, or modules.
    - [x] Updated docstrings for any modifications to existing code.
    - [x] Added references to new classes, functions, or modules to the relevant RST file under `docs/`.
    - [x] Updated all relevant supporting documentation files under `docs/`.
    - [x] A copyright notice is included at the top of any new modules (using your own name or the name of your organization).
    - [x] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/directives.html?highlight=versionadded#directive-versionadded).
- [x] Changes (and possible deprecations) have [towncrier](https://pypi.org/project/towncrier/) news fragments under `docs/_newsfragments/`. (Run `towncrier --draft` to ensure it renders correctly.)

If you have *any* questions to *any* of the points above, just **submit and ask**! This checklist is here to *help* you, not to deter you from contributing!

*PR template inspired by the attrs project.*

PR #1655, opened 2020-01-25
Fix: Make error initializers more consistent

CaselIT:
# Summary of Changes

Normalized the interface of the http errors with common title and description kwargs.
Compared to the issue I've left the `headers` kw as is, since it's used in all errors

# Related Issues

Fixes:  #777

# Pull Request Checklist

### DIFF
--- a/docs/_newsfragments/777.removal.rst
+++ b/docs/_newsfragments/777.removal.rst
@@ -1,0 +1,4 @@
+The class :class:`~.falcon.http_error.NoRepresentation` was removed since
+all :class:`~.falcon.HTTPError` have optional representation.
+Deprecate the use of positional arguments for the optional kw args of
+the HTTPError subclasses
--- a/docs/_newsfragments/777.removal.rst
+++ b/docs/_newsfragments/777.removal.rst
@@ -1,4 +1,2 @@
 The class :class:`~.falcon.http_error.NoRepresentation` was removed since
 all :class:`~.falcon.HTTPError` have optional representation.
-Deprecate the use of positional arguments for the optional kw args of
-the HTTPError subclasses
--- a/falcon/errors.py
+++ b/falcon/errors.py
@@ -1692,8 +1692,7 @@ class HTTPInvalidHeader(HTTPBadRequest):
     """
 
     def __init__(self, msg, header_name, headers=None, **kwargs):
-        description = ('The value provided for the {0} header is '
-                       'invalid. {1}')
+        description = 'The value provided for the "{0}" header is invalid. {1}'
         description = description.format(header_name, msg)
 
         super().__init__('Invalid header value', description, headers, **kwargs)
@@ -1735,7 +1734,7 @@ class HTTPMissingHeader(HTTPBadRequest):
     """
 
     def __init__(self, header_name, headers=None, **kwargs):
-        description = 'The {0} header is required.'
+        description = 'The "{0}" header is required.'
         description = description.format(header_name)
 
         super().__init__('Missing header value', description, headers, **kwargs)
--- a/falcon/errors.py
+++ b/falcon/errors.py
@@ -153,6 +153,10 @@ class HTTPUnauthorized(HTTPError):
             challenges to use as the value of the WWW-Authenticate header in
             the response.
 
+            Note:
+                The existing value of the WWW-Authenticate in headers will be
+                overridden by this value
+
             (See also: RFC 7235, Section 2.1)
         href (str): A URL someone can visit to find out more information
             (default ``None``). Unicode characters are percent-encoded.
@@ -362,6 +366,10 @@ class HTTPMethodNotAllowed(OptionalRepresentation, HTTPError):
     Args:
         allowed_methods (list of str): Allowed HTTP methods for this
             resource (e.g., ``['GET', 'POST', 'HEAD']``).
+
+            Note:
+                The existing valuessss of the Allow in headers will be
+                overridden by this value.
 
     Keyword Args:
         title (str): Human-friendly error title. If not provided, and
@@ -719,11 +727,6 @@ class HTTPPayloadTooLarge(HTTPError):
 
         description (str): Human-friendly description of the error, along with
             a helpful suggestion or two.
-
-        retry_after (datetime or int): Value for the Retry-After
-            header. If a ``datetime`` object, will serialize as an HTTP date.
-            Otherwise, a non-negative ``int`` is expected, representing the
-            number of seconds to wait.
         headers (dict or list): A ``dict`` of header names and values
             to set, or a ``list`` of (*name*, *value*) tuples. Both *name* and
             *value* must be of type ``str`` or ``StringType``, and only
@@ -740,6 +743,15 @@ class HTTPPayloadTooLarge(HTTPError):
             Note:
                 Falcon can process a list of ``tuple`` slightly faster
                 than a ``dict``.
+        retry_after (datetime or int): Value for the Retry-After
+            header. If a ``datetime`` object, will serialize as an HTTP date.
+            Otherwise, a non-negative ``int`` is expected, representing the
+            number of seconds to wait.
+
+            Note:
+                The existing value of the Retry-After in headers will be
+                overridden by this value
+
         href (str): A URL someone can visit to find out more information
             (default ``None``). Unicode characters are percent-encoded.
         href_text (str): If href is given, use this as the friendly
@@ -893,6 +905,10 @@ class HTTPRangeNotSatisfiable(OptionalRepresentation, HTTPError):
     Args:
         resource_length: The maximum value for the last-byte-pos of a range
             request. Used to set the Content-Range header.
+
+            Note:
+                The existing value of the Content-Range in headers will be
+                overridden by this value
 
     Keyword Args:
         title (str): Error title (default '416 Range Not Satisfiable').
@@ -1188,6 +1204,11 @@ class HTTPTooManyRequests(HTTPError):
             header. If a ``datetime`` object, will serialize as an HTTP date.
             Otherwise, a non-negative ``int`` is expected, representing the
             number of seconds to wait.
+
+            Note:
+                The existing value of the Retry-After in headers will be
+                overridden by this value
+
         href (str): A URL someone can visit to find out more information
             (default ``None``). Unicode characters are percent-encoded.
         href_text (str): If href is given, use this as the friendly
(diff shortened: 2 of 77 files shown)


CaselIT:
I've a couple of questions:

- should the kwargument be kw only? Using `def __init__(self, *, title=None, ...)`?
- some types have additional keyword args, that are placed in `headers`, like `challenges` in `HTTPUnauthorized`. I've not changed this. Should we change it?

kgriffs:
> should the kw argument be kw only? Using PEP 3102 def __init__(self, *, title=None, ...)?

Yes, I think that would be a Good Thing™. As an aside, we should consider doing this in other places as well so we don't have to worry about a simple reordering of kwargs down the road causing a breaking change. @vytas7 thoughts?

kgriffs:
> some types have additional keyword args, that are placed in headers, like challenges in HTTPUnauthorized. I've not changed this. Should we change it?

It would not hurt to normalize the ordering of extra params such that they always follow _headers_ (although it would only be cosmetic if PEP 3102 is used). Regardless, we will need to make sure we have test coverage for things like that overriding any values also set in _headers_.

falcon/errors.py
```diff
@@ -377,11 +372,9 @@ class HTTPMethodNotAllowed(OptionalRepresentation, HTTPError):
             base articles related to this error (default ``None``).
     """
 
-    def __init__(self, allowed_methods, headers=None, **kwargs):
+    def __init__(self, allowed_methods, title=None, description=None, headers=None, **kwargs):
```
kgriffs: We'll need a breaking change news fragment since people may have been passing _headers_ positionally.

CaselIT: I'll add it to the existing one regarding the `NoRepresentation`. Also the kw only change

CaselIT:
> we will need to make sure we have test coverage for things like that overriding any values also set in headers.

I think the current implementation overrides in all of them. I can check if that is the case and check if there are tests for it. Sadly the code coverage does not help in this case

falcon/errors.py
CaselIT: btw `towncrier` does not properly render the misc changelog

CaselIT:
I'm changing to kwonly.

There seem to be a lot of changes required, both in code and in documentations.
Just letting you know

CaselIT:
I've notices that for the headers it says: `headers (dict or list)`

But the list case is not handled by all the errors that add something to the headers, like `HTTPUnauthorized`:

https://github.com/falconry/falcon/blob/c817a42d183b89250e633969a1b9a3e23d95aef5/falcon/errors.py#L165-L170

CaselIT:
should be almost done other than https://github.com/falconry/falcon/pull/1655#issuecomment-578503132

CaselIT:
> should the kw argument be kw only? Using PEP 3102 def __init__(self, *, title=None, ...)?

I know I'm the one who suggested it, and I've already implemented it, but thinking a bit more about it and also judging by the amount of refactoring that this has required in falcon, I'm not sure if it's worth the trouble. By it I mead that it will probably cause a lot of refactoring on the users who use falcon, without a clear improvement.
I think we can keep the changes in falcon, but I would revert the actual requirement for kw_only. Or at least move the `*` after the `title` and `description`.

What do you thing @kgriffs @vytas7

vytas7:
> I know I'm the one who suggested it, and I've already implemented it, but thinking a bit more about it and also judging by the amount of refactoring that this has required in falcon, I'm not sure if it's worth the trouble. By it I mead that it will probably cause a lot of refactoring on the users who use falcon, without a clear improvement.

I'd need to read through signatures of existing `HTTPError`s & `HTTPStatus`es more carefully to provide a more informed opinion.

A quick reaction off the top of my head would be that it indeed sounds like a bit too abrupt breaking change without any compatibility shims or deprecation warnings, even given that 3.0 will be a major release.

kgriffs:
Good points. It does look like that will cause a bigger burden on app developers than we want to impose this time around. We may have to save this for a 4.0, since there are already enough other things in 3.0 for people to worry about.

CaselIT:
I agree.
Should I actually deprecate them with a warning, or will a line in the documentation suffice?
Regarding the deprecation, I'm not sure if there is something better than something like

```py
def deprecate_arg(allowed=0, fn):
  def wraps(self, *args, **kwargs):
    if len(args)> allowed:
      # warn here
    return fn(self, *args, **kwargs)
  return wraps

class HttpBadRequest(...):
  @deprecate_arg
  def __init__(self, title=None, description=None, headers=None, **kwargs):
    ...
```

CaselIT:
The only thing missing is to decide if we want to raise a warning regarding the kwonly args, or if we should just add a note in the docs. (or maybe nothing at all)

kgriffs:
IMO we should at least add a warning section to the relevant docstrings. As for outputting a deprecation warning, I'm inclined to say yes as well, since this raising errors is typically not super performance sensitive, and the extra nudge to the app developer can't hurt. @vytas7 thoughts?

vytas7:
Agreed re the deprecation warning, not many are going to care or even notice about a line or two in the docs. It is ideal to emit them if possible, and agreed re the performance impact being lesser for exceptions.

CaselIT:
ok, so something like the `deprecate_arg` in the snippet

CaselIT:
Ok I've replaced the kwonly with a warning and a note in the documents. (not sure if it should be a warning)

I think it's ready for review

falcon/util/misc.py
```diff
@@ -428,3 +429,34 @@ def code_to_http_status(code):
...
+          ``self``. Defaults to True.
+    """
+
+    template = (
+        'Calls with{} positional args are deprecated.'
+        ' Specify them as keyword arguments instead'
+    )
+    text = ' more than {}'.format(allowed_positional) if allowed_positional else ''
+    warn_text = template.format(text)
+    if is_method:
```
kgriffs: Coverage is complaining about this line because the condition is never false.

falcon/util/misc.py
```diff
@@ -428,3 +429,34 @@ def code_to_http_status(code):
         return getattr(status_codes, 'HTTP_' + str(code))
     except AttributeError:
         return str(code)
+
+
+def deprecated_args(allowed_positional, is_method=True):
+    """Flags a method call with positional args as deprecate
```
kgriffs: "deprecate" ==> "deprecated"

falcon/util/misc.py
```diff
@@ -428,3 +429,34 @@ def code_to_http_status(code):
...
+    Args:
+        allowed_positional (int): Number of allowed positional arguments
+        is_method (bool, optional): The decorated function is a method. Will
+          add one to the number of allowed positional args to account for
+          ``self``. Defaults to True.
+    """
+
+    template = (
+        'Calls with{} positional args are deprecated.'
+        ' Specify them as keyword arguments instead'
```
kgriffs: The second statement probably needs a period. We may also want to add a "please" in there.

falcon/util/misc.py
```diff
@@ -428,3 +429,34 @@ def code_to_http_status(code):
...
+
+    Args:
+        allowed_positional (int): Number of allowed positional arguments
+        is_method (bool, optional): The decorated function is a method. Will
+          add one to the number of allowed positional args to account for
+          ``self``. Defaults to True.
+    """
+
+    template = (
+        'Calls with{} positional args are deprecated.'
```
kgriffs: Looks like this is missing a space before the `{}` field.

falcon/util/misc.py
```diff
@@ -428,3 +429,34 @@ def code_to_http_status(code):
         return getattr(status_codes, 'HTTP_' + str(code))
     except AttributeError:
         return str(code)
+
+
+def deprecated_args(allowed_positional, is_method=True):
```
kgriffs: Heh, perhaps the decorator should employ PEP 3102 itself?

falcon/util/misc.py
CaselIT: I'm adding a test

falcon/util/misc.py
CaselIT: It's on line 448
```
text = ' more than {}'.format(allowed_positional) if allowed_positional else ''
warn_text = template.format(text)
```

CaselIT:
I've noticed that the changelog needs updating.

docs/_newsfragments/777.removal.rst
```diff
@@ -0,0 +1,4 @@
+The class :class:`~.falcon.http_error.NoRepresentation` was removed since
+all :class:`~.falcon.HTTPError` have optional representation.
+Deprecate the use of positional arguments for the optional kw args of
```
kgriffs: Seems like this deprecation could be its own news fragment?

kgriffs:
> I've noticed that the changelog needs updating.

Nice catch! I think we are all trying to get used to making news fragments part of the PRs.

docs/_newsfragments/777.removal.rst
CaselIT: I'm not sure how to make them. They should be 777 since tha is the issue, ans I think are both removal?

kgriffs: You could make the other one "misc" since it hasn't been removed (as a breaking change) yet, *per se*. Also, I think it is reasonable to add a few (hyphenated) words of description after the issue number (as has been done with other fragments.

@csojinb @vytas7 thoughts? Related: https://github.com/falconry/falcon/pull/1621#discussion_r358016645

CaselIT: I've used only the numbers as the names since I though that we wanted to use towncrier to automatically add the link to the issue for every changelog file.
I can make a new misc later

kgriffs: Ahh, I see. I just tried that and indeed it drops a `(#{number})` reference in there, although it doesn't generate a GH link (perhaps a sphinx extension could be employed for this). That is nice, but on the other hand it seems like there should be a way to (1) have towncrier allow multiple entries for the same issue and (2) to better support items that don't have an associated issue, perhaps by simply not appending the `(name-of-news-fragment)`. Perhaps we need a post-processing script?

CaselIT: There is a configuration parameter for towncrier to make it apply the gh link. We were trying it in gitter with @vytas7 

I can just change the name for now, and then the names of the newsitems will be changed after the configuration for towncrier is figured out

CaselIT: the config was `issue_format = "https://github.com/falconry/falcon/issues/{issue}"`

PR #1657, opened 2020-01-28
fix(asgi): Errors caused by not being able to detect cythonized coroutine functions

kgriffs:
# Summary of Changes

This patch provides workarounds for the fact that cythonized coroutine functions
can not be detected.

In several places in the framework, we rely on being able to introspect a passed
callable in order to determine how to wrap it or call it later. As a workaround,
this patch gives the caller a way to explicitly indicate that the framework may
assume a coroutine function.

Also, we relax our type checking so that we don't raise an error on Cython
functions; the app developer will simply have to be trusted to pass the correct
type of function if they are using Cython.

# Pull Request Checklist

### DIFF
--- a/falcon/asgi/response.py
+++ b/falcon/asgi/response.py
@@ -161,7 +161,7 @@ class Response(falcon.response.Response):
         scheduled to run on the event loop as soon as possible.
 
         The callback will be invoked without arguments. Use
-        :py:meth`functools.partial` to pass arguments to the callback as needed.
+        :py:meth:`functools.partial` to pass arguments to the callback as needed.
 
         Note:
             If an unhandled exception is raised while processing the request,
--- a/falcon/asgi/response.py
+++ b/falcon/asgi/response.py
@@ -152,7 +152,7 @@ class Response(falcon.response.Response):
         return data
 
     def schedule(self, callback):
-        """Schedules an async callback to run soon after sending the HTTP response.
+        """Schedule an async callback to run soon after sending the HTTP response.
 
         This method can be used to execute a background job after the response
         has been returned to the client.
@@ -207,7 +207,7 @@ class Response(falcon.response.Response):
             self._registered_callbacks.append(rc)
 
     def schedule_sync(self, callback):
-        """Schedules a synchronous callback to run soon after sending the HTTP response.
+        """Schedule a synchronous callback to run soon after sending the HTTP response.
 
         This method can be used to execute a background job after the
         response has been returned to the client.
@@ -218,7 +218,7 @@ class Response(falcon.response.Response):
         :py:meth:`asyncio.AbstractEventLoop.set_default_executor`).
 
         The callback will be invoked without arguments. Use
-        :py:meth`functools.partial` to pass arguments to the callback
+        :py:meth:`functools.partial` to pass arguments to the callback
         as needed.
 
         Note:
--- a/falcon/asgi/response.py
+++ b/falcon/asgi/response.py
@@ -15,10 +15,11 @@
 """ASGI Response class."""
 
 from asyncio.coroutines import CoroWrapper
-from inspect import iscoroutine
+from inspect import iscoroutine, iscoroutinefunction
 
 from falcon import _UNSET
 import falcon.response
+from falcon.util.misc import is_python_func
 
 __all__ = ['Response']
 
@@ -151,20 +152,16 @@ class Response(falcon.response.Response):
         return data
 
     def schedule(self, callback):
-        """Schedules a callback to run soon after sending the HTTP response.
+        """Schedules an async callback to run soon after sending the HTTP response.
 
-        This method can be used to execute a background job after the
-        response has been returned to the client.
+        This method can be used to execute a background job after the response
+        has been returned to the client.
 
-        If the callback is an async coroutine function, it will be scheduled
-        to run on the event loop as soon as possible. Alternatively, if a
-        synchronous callable is passed, it will be run on the event loop's
-        default ``Executor`` (which can be overridden via
-        :py:meth:`asyncio.AbstractEventLoop.set_default_executor`).
+        The callback is assumed to be an async coroutine function. It will be
+        scheduled to run on the event loop as soon as possible.
 
         The callback will be invoked without arguments. Use
-        :py:meth`functools.partial` to pass arguments to the callback
-        as needed.
+        :py:meth`functools.partial` to pass arguments to the callback as needed.
 
         Note:
             If an unhandled exception is raised while processing the request,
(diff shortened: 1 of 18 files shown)


falcon/hooks.py
```diff
@@ -27,7 +27,7 @@
     '|'.join(method.lower() for method in COMBINED_METHODS)))
 
 
-def before(action, *args, **kwargs):
+def before(action, *args, is_async=None, **kwargs):
```
kgriffs: I could also add a `falcon.asgi.hooks()` that just always sets this argument to `True`, but I'm not certain how helpful that would be. Thoughts?

tox.ini
```diff
@@ -100,29 +100,36 @@ setenv =
     FALCON_ASGI_WRAP_NON_COROUTINES=Y
     FALCON_TESTING_SESSION=Y
     PYTHONASYNCIODEBUG=1
+install_command = python -m pip install --no-build-isolation {opts} {packages}
```
kgriffs: I found that the `test_cython.test_imported_from_c_modules` was failing on my box. It seemed to be related to the presence of the pyproject.toml file triggering pip to enable build isolation, and therefore it wasn't finding cython in the environment. If that's true, we'll need to update our docs that talk about using cython, and we'll probably need to look into migrating wholesale over to PEP 517 at some point.

vytas7: I had something like that, but then `git clean -fdx`ed the source tree and it passed.

falcon/util/misc.py
```diff
@@ -77,6 +78,22 @@ def decorator(func):
...
+        func: The function object to check.
+    Returns:
+        bool: ``True`` if the function is a Cython function or method;
+        ``False`` otherwise.
+
+    """
+    if inspect.ismethod(func):
+        func = func.__func__
+
+    return 'cython' in type(func).__name__
```
vytas7: I'm not getting this heuristic to work for functions (as opposed to cythonized class methods), they just show up as builtins:
```python
>>> import falcon
>>> falcon.uri.parse_query_string
<built-in function parse_query_string>
>>> falcon.API
<class 'falcon.app.API'>
>>> falcon.API.__init__
<cyfunction API.__init__ at 0x7f8818024608>
```

tests/asgi/test_cythonized_asgi.py
```diff
@@ -0,0 +1,82 @@
...
+
+    client.simulate_get()
+    time.sleep(0.5)
+    assert resource.counter['backround:on_get:async'] == 2
+    assert resource.counter['backround:on_get:sync'] == 40
+
+
+@pytest.mark.skipif(not pyximport, reason='Cython not installed')
+@pytest.mark.skipif(
+    sys.version_info.minor < 7,
```
vytas7: This won't work as expected on Python 4.0 :smiling_imp:

falcon/util/misc.py
kgriffs: Hmm, I could have sworn that was working at some point but I must have been mistaken. I'll take a look.

tests/asgi/test_cythonized_asgi.py
kgriffs: ah, good point. Might as well future-proof it.

falcon/util/misc.py
kgriffs: I was able to repro this after adding some more tests to `test_cythonized_asgi`. Working on a fix.

falcon/hooks.py
```diff
@@ -82,14 +95,14 @@ def let(responder=responder):
...
-            do_before_one = _wrap_with_before(responder, action, args, kwargs)
+            do_before_one = _wrap_with_before(responder, action, args, kwargs, is_async)
 
             return do_before_one
 
     return _before
 
 
-def after(action, *args, **kwargs):
+def after(action, *args, is_async=None, **kwargs):
```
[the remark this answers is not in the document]
kgriffs: I will say, however, that this can break at least one calling pattern when the default is not `False` (not a problem in this case). If the caller pre-composes a kwargs dict, for example, and wants to just pass through an option `bar` from upstream that does default to `None` for whatever reason, they can not always just set {'foo': bar}, but will have to add additional logic to avoid including the kwarg in the dict in order to be sure they trigger the desired default value for the downstream invocation.

kgriffs:
@vytas7 I attempted to address all of your feedback with this update.

falcon/hooks.py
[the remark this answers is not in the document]
vytas7: Yeah -- sometimes "the argument is not provided" is not the same as "the boolean `false` is provided" :thinking:

falcon/asgi/response.py
```diff
@@ -180,6 +177,58 @@ def schedule(self, callback):
...
+        This method can be used to execute a background job after the
+        response has been returned to the client.
+
+        The callback is assumed to be a synchronous (non-coroutine) function.
+        It will be scheduled on the event loop's default ``Executor`` (which
+        can be overridden via
+        :py:meth:`asyncio.AbstractEventLoop.set_default_executor`).
+
+        The callback will be invoked without arguments. Use
+        :py:meth`functools.partial` to pass arguments to the callback
```
vytas7: Probably a missing colon (after `:py:meth`)

falcon/hooks.py
```diff
@@ -102,6 +115,13 @@ def after(action, *args, **kwargs):
             order given, immediately following the *req*, *resp*, *resource*,
             and *params* arguments.
 
+    Keyword Args:
+        is_async (bool): Set to ``True`` for ASGI apps to provide a hint that
+            the decorated responder is a coroutine function (i.e., that it
+            is defined with ``async def``). This is only necessary for
```
vytas7: Hmm... maybe we could tone down language here somewhat :thinking: 
I'm thinking of the example that Cython's Stefan provided in a related discussion, something along the lines of
```python
>>> import asyncio
>>> async def test():
...     print('Hello!')
...     await asyncio.sleep(1)
... 
>>> def coro_factory():
...     return test()
... 
>>> async def main():
...     await coro_factory()
... 
>>> asyncio.run(main())
Hello!
>>> import inspect
>>> inspect.iscoroutinefunction(test)
True
>>> inspect.iscoroutinefunction(coro_factory)
False
```
TL;DR you can never _really_ know whether a function is a coroutine function (except, of course, trying to call it and see).

falcon/asgi/response.py
```diff
@@ -151,20 +152,16 @@ def data(self, value):
...
-        synchronous callable is passed, it will be run on the event loop's
-        default ``Executor`` (which can be overridden via
-        :py:meth:`asyncio.AbstractEventLoop.set_default_executor`).
+        The callback is assumed to be an async coroutine function. It will be
+        scheduled to run on the event loop as soon as possible.
 
         The callback will be invoked without arguments. Use
-        :py:meth`functools.partial` to pass arguments to the callback
-        as needed.
+        :py:meth`functools.partial` to pass arguments to the callback as needed.
```
vytas7: missing colon here as well

falcon/asgi/response.py
```diff
@@ -180,6 +177,58 @@ def schedule(self, callback):
...
+
+        rc = (callback, True)
+
+        if not self._registered_callbacks:
+            self._registered_callbacks = [rc]
+        else:
+            self._registered_callbacks.append(rc)
+
+    def schedule_sync(self, callback):
+        """Schedules a synchronous callback to run soon after sending the HTTP response.
```
vytas7: It would probably better to word this as an imperative, i.e., "Schedule a ..."

falcon/hooks.py
kgriffs: While technically it is true that you can't know without calling in all cases, usually you are really talking about something that was declared via async def, so you could argue that what we really want to know is "was this function declared with async def" (i.e., you are making an equivalency argument between coroutine functions and the syntax typically used to declare them).

kgriffs: I suppose the factory pattern is probably the most likely exception to this rule... Let me see about reworking these is_async docstrings.

vytas7: Ah, that's true, you are actually defining it that way.
In any case, what I wanted to say was that this `is_async` flag might be useful in any kind of edge case, not just cythonized functions.

falcon/asgi/response.py
kgriffs: Good catch. We need to be consistent about that going forward.

kgriffs: Done.

falcon/asgi/response.py
kgriffs: Done.

falcon/hooks.py
kgriffs: Done. Let me know what you think.

falcon/asgi/response.py
kgriffs: Done.

PR #1658, opened 2020-01-28
Fix tests on windows

### LINKED ISSUE #1656: Test failure on windows
Tests on windows have some failures and errors.

The summary is `47 failed, 2094 passed, 79 skipped, 1 warning, 16 errors in 1885.17s (0:31:25)`

The errors output is here:
https://gist.github.com/CaselIT/16792866e7dd2c545b186207976e060e

Some failues are related to asgi, while other are simple stuff like `/` vs `\` in the path.
Probably a skip on the incompatible tests and the use of `os.path.normpath` would solve most of these

This is mainly an issue to let you know this, I don't think it's really necessary fixing these, since the rests run so slow on windows (31 minutes) that it would not be practical to develop on it.

CaselIT:
# Summary of Changes

Make the test work on windows. This mainly involves changes to the `tests` files.
The serve static constructor has been updated to use `os.path.normpath` on the `directory` and the `fallback_filename`
I've also created ids for pytest fixtures so that the logging is a bit more readable, in particular to the fixtures that tests long strings and the `asgi` fixture so that the tests read `[asgi]` and `[wsgi]` instead of true/false (maybe async/sync is better?)

# Related Issues

Closes #1656 

# Pull Request Checklist

This  is still a work in progress. Currently I've fixed the test related to the path errors. There are still errors in the async part.

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -1069,5 +1069,6 @@ class ResponseOptions:
         self.default_media_type = DEFAULT_MEDIA_TYPE
         self.media_handlers = Handlers()
 
-        mimetypes.init()
+        if not mimetypes.inited:
+            mimetypes.init()
         self.static_media_types = mimetypes.types_map
--- a/falcon/routing/static.py
+++ b/falcon/routing/static.py
@@ -49,13 +49,16 @@ class StaticRoute:
         if not prefix.startswith('/'):
             raise ValueError("prefix must start with '/'")
 
-        if not os.path.isabs(directory):
+        self._directory = os.path.normpath(directory)
+        if not os.path.isabs(self._directory):
             raise ValueError('directory must be an absolute path')
 
         if fallback_filename is None:
             self._fallback_filename = None
         else:
-            self._fallback_filename = os.path.normpath(os.path.join(directory, fallback_filename))
+            self._fallback_filename = os.path.normpath(
+                os.path.join(self._directory, fallback_filename)
+            )
             if not os.path.isfile(self._fallback_filename):
                 raise ValueError('fallback_filename is not a file')
 
@@ -66,7 +69,6 @@ class StaticRoute:
             prefix += '/'
 
         self._prefix = prefix
-        self._directory = os.path.normpath(directory)
         self._downloadable = downloadable
 
     def match(self, path):
--- a/falcon/routing/static.py
+++ b/falcon/routing/static.py
@@ -55,7 +55,7 @@ class StaticRoute:
         if fallback_filename is None:
             self._fallback_filename = None
         else:
-            self._fallback_filename = os.path.join(directory, fallback_filename)
+            self._fallback_filename = os.path.normpath(os.path.join(directory, fallback_filename))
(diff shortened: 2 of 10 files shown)


CaselIT:
Currently on windows is
`2146 passed, 87 skipped, 1 warning in 414.10s (0:06:54)`

falcon/routing/static.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, prefix, directory, downloadable=False, fallback_filename=None
             prefix += '/'
 
         self._prefix = prefix
-        self._directory = directory
+        self._directory = os.path.normpath(directory)
```
CaselIT: Maybe this normpath should be before the check on line 52?
https://github.com/falconry/falcon/blob/e0459ebf7f85a65c1cafa968330985d60cbd036c/falcon/routing/static.py#L52-L53

CaselIT:
Upon further investigation, most (about 80%) of the time on windows  is lost in the `mimetypes.init()` call.

checking to see if mimetypes has been inited before rerunning the function brings the time down to less than 2 minutes

CaselIT:
I think it's ready for review.
After the change to the mimetypes init the status on windows is `2158 passed, 75 skipped, 1 warning in 92.66s (0:01:32)`

falcon/routing/static.py
kgriffs: Yeah, I think that would make sense. From the `isabs()` docs:

> Return True if path is an absolute pathname. On Unix, that means it begins with a slash, on Windows that it begins with a (back)slash after chopping off a potential drive letter.

CaselIT: fixed in the last commit

CaselIT:
I'm not familiar with travis so I cannot offer any insight on how it works for windows.

Seems that codecov has thrown a fit in the last commit

falcon/response.py
```diff
@@ -1069,5 +1069,6 @@ def __init__(self):
         self.default_media_type = DEFAULT_MEDIA_TYPE
         self.media_handlers = Handlers()
 
-        mimetypes.init()
+        if not mimetypes.inited:
```
kgriffs: Curiously, when I run coverage on my box this is the only line that is reported as missing. Apparently the condition is never true (as far as coverage can detect). But I was able to manually verify that the branch does get hit by tests in `asgi.test_asgi_servers`. So something must be beating us to the punch for the in-process tests.

CaselIT: Maybe coverage inits the mimetypes? Should me maybe to a reload in the conftest to ensure that we get a clean instance of mimetypes?

Running normally this is hit every time an App is instantiated, so it would be hundreds of times in the falcon tests

This is a workaround for slow windows tests, it's not very important at runtime since the application is not continually instantiated. So we could place an `is_windows and falcon_tests` here.

CaselIT: slow windows test may happen on users test suite, since it is a similar situation regarding instantiating the app many times

CaselIT: I've added a test for it in the last commit

PR #1665, opened 2020-02-05
Delay compilation until fist use of the router

### LINKED ISSUE #1550: Adding many routes to CompiledRouter is slow (takes O(n^2) time)
i have about 400 api route to add
falcon compiles method “find” each time and slower as the count grows
it slows down hug booting
does anyone else have same problem？
i made a patch to fix it，but not very elegent，
i wish there will be a better solution

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

CaselIT:
# Summary of Changes

Delay the router compilation until the first actual use of the router.

~~By default each route is compiled in a standalone instance of the router, to check that it will actually compile. This behavior is controlled by the `verify_route_on_add` flag in the compiled router option.
By default this is set to true since it's impact is negligible unless thousands of routes are added.~~

The method `add_route` accepts a new keyword to make the router eagerly compile. This can be used to finalize the router when the last route is added, to avoid incurring a penalty on the first routing

The performance app setup to first response is this (in s)

| routes | master | delayed router | ~~verify_route_on_add=True~~|
|--------|--------|-----------------------------|----------------------------|
|  9     |0.0034604|0.0014578|0.0043392|
|  81    |0.1205411|0.0094007|0.0329528|
|  729   |11.238887|0.0883184|0.3974814|
|  6561  |1144.1065|0.9569258|3.5786625|
|  59049 |  slow   |10.385946|36.764869|

with the simple test
```py
pathMatrix = [["foo", "bar", "baz"], ["oof", "rab", "zab"]]
paths = ["/" + "/".join(p) for p in itertools.product(*(pathMatrix * mult))]
start = time.perf_counter()
for i, p in enumerate(paths, 1):
    app.add_route(p, res)
T.TestClient(app).simulate_get(paths[0])
```

Edit: updated with new updated data that no longer has `verify_route_on_add`

# Related Issues

fixes #1550

# Pull Request Checklist


CaselIT:
I've added to the table the times with a very naive stand alone compile for every added route on a new router to check if the route it can actually compile
this is the check:
```py
        if kwargs.get("_check") is not True:
            temp = CompiledRouter()
            temp._options = self._options
            temp.add_route(uri_template, resource, _check=True, **kwargs)
```

CaselIT:
I've had another look at the compile functions of the route and I think that safest way to ensure that every route can compile on its own is to have another instance of the router that we use for this standalone compile. Doing otherwise may have unintended side effects of the router instance, but I'm not very familiar with the compiled router so I may be missing an easier / cleaner solution.
I'm not sure if this should be configurable or not. I guess an option in the router options would not hurt

Regarding the implementation, I think the delayed compilation could be the default and `add_route` could have a new kwarg `compile` that when set compiles the router, retaining the current behavior.
This allows users to compile the router when they add the last route if he prefer to avoid the penalty on the first request.

@kgriffs @vytas7 what do you think?

CaselIT:
@kgriffs @vytas7 This should be ready for review.
I'm not sure about the name of the option `verify_route_on_add`. I'm usually not very good at naming things. If you came up with a better name I would be happy to change it.

I've updated the inital post table

falcon/routing/compiled.py
```diff
@@ -531,6 +568,23 @@ def _instantiate_converter(self, klass, argstr=None):
...
+        """Compiles the router, sets the `_find` attribute and returns its result
+
+        This method is set to the `_find` attribute to delay the compilation of the
+        router until its used for the first time. subsequent calls to `_find` will
+        be processed by the actual routing function.
+        This method must have the same signature as the function returned by the
+        :meth:`.CompiledRouter._compile`.
+        """
+        # NOTE(caselit): replace the find with the actual method
+        self._find = self._compile()
```
CaselIT: should we have a lock of some kind here to support multi-threading/async invocations?

kgriffs:
> Regarding the implementation, I think the delayed compilation could be the default and `add_route` could have a new kwarg `compile` that when set compiles the router, retaining the current behavior.
> This allows users to compile the router when they add the last route if they prefer to avoid the penalty on the first request.

I'm cool with this approach. We just need the docs to mention the reason someone might use the param, as you outlined above.

falcon/routing/compiled.py
kgriffs: I don't think we need a lock here, assuming `_compile()` itself is threadsafe. Inside `_compile()`, the only potential issue would be one thread mutating the set of routes while another thread is attempting to generate the AST. But there aren't any points in there where the GIL or a green thread or coroutine might yield execution, so we should be safe.

vytas7: Reposting my somewhat similar comments from Gitter

> @CaselIT you could probably rely on atomicity of Python structures guarded by the GIL, but then it is IMHO important to be crystal clear about that in the comments and/or docs.

> E.g., setting a dictionary or object key/value is atomic under the GIL. So just compiling and replacing the routing method should work fine from that perspective.

> One aspect that might be worth considering in this approach is, however, an API facing high load directly when from the second 0. That may trigger a large number of parallel compilations. However, this is only an issue when using threads, and not async/green threads, and APIs are normally brought online by means of an LB check or similar, that could be used as an opportunity to compile.

CaselIT: I'm not sure if compile is thread safe. I mean, repeated invocations should produce the same output, but since it mutates the state of the router (it sets return value and patters) it may happen that an invocation of find fails/returna the wrong result since it happens with the wrong values?

CaselIT: I guess a simple
```py
with self._compile_lock:
    if self._find == self._compile_and_find:
        self._find = self._compile()
```
would not have a lot of downsides and we avoid multiple compilation in multi threading envs

CaselIT:
I've added a lock to guard the router compile. While the compile itself may be thread safe, and python has the gil, this avoids wasting time compiling multiple times the router.

I've also update the documentation. 

I'm not sure if we should mention that the router is "lazy" here https://github.com/falconry/falcon/blob/0d419ef49f26a41a335b67d5eeeed29365c35308/docs/api/routing.rst#default-router
Something along the lines of 
```rst
By default the decision three is compiled only when the router receives the first request.
See :class:`.CompiledRouter` for more information on this
```

falcon/routing/compiled.py
```diff
@@ -206,7 +241,19 @@ def insert(nodes, path_index=0):
...
-        self._find = self._compile()
+
+        if kwargs.get('compile', False):
+            self._find = self._compile()
+        else:
+            if self._options.verify_route_on_add:
+                if self._verify_route_router is None:
+                    self._verify_route_router = CompiledRouter(_options=self._options)
+                kwargs['compile'] = True
+                self._verify_route_router._roots.clear()
```
kgriffs: Why does the tree need to be reset each time?

kgriffs: nevermind, I see that it is to speed things up since we are compiling. Would it be worth modifying the router class to do verification without compilation? Some errors will not be detected unless the entire tree is present (e.g., route conflicts).

CaselIT: if we do not clear the roots we end up with the same behaviour as now, ie every time we will compile all the routes.

> Some errors will not be detected unless the entire tree is present (e.g., route conflicts).

I think insert will raise in this case here https://github.com/falconry/falcon/blob/c3e91875f1030c68d17efa27cf8df3d18f337761/falcon/routing/compiled.py#L186
and we are calling

I think that the route compile will raise if it does not find the converter for an id. I'm not sure if there are other cases.
Do you have an example of what we are missing not compiling all the routes?

>  Would it be worth modifying the router class to do verification without compilation?

Yes, I think we could do it, but I'm not sure if it would be worth it. The current times are excellent with the compilation step of the single route, and an user could also decide to remove it once he is sure that the routes work

CaselIT: > I think that the route compile will raise if it does not find the converter for an id. I'm not sure if there are other cases.

I've had another look at it, and it seems that this is checked by the add_route function here
https://github.com/falconry/falcon/blob/c3e91875f1030c68d17efa27cf8df3d18f337761/falcon/routing/compiled.py#L169

So I'm not sure what we gain by compiling the routes one by one as It's done now by the `_verify_route_router`.
Maybe we could just omit it?

CaselIT: I've tried to set the `verify_route_on_add` to false by default and the only failure in the tests are the one raised when the converter are instantiated, ie these ones

https://github.com/falconry/falcon/blob/c3e91875f1030c68d17efa27cf8df3d18f337761/tests/test_default_router.py#L454-L465

and I think that is because we are only instantiating the converter when compiling.

If we try to instantiate them in `_validate_template_segment`, just to test that it works and then throw the instance away, I think we could just omit the compile step.

CaselIT:
Moving here from the review comment.

I've had another look and I think we could avoid the standalone compilation step of every route, by trying to also instantiate the field converters in the `_validate_template_segment`. 
https://github.com/falconry/falcon/blob/c3e91875f1030c68d17efa27cf8df3d18f337761/falcon/routing/compiled.py#L282
This is the only check that is not currently performed by the `add_route` that is left to the compile step.

I think with this the implementation would be a lot cleaner, since there would be no need for the internal router instance, the new router options, etc

CaselIT:
I've updated the `_validate_template_segment` function to also verify that a converter can be instantiated with the provided params

docs/api/routing.rst
```diff
@@ -57,6 +57,8 @@ Default Router
 
 Falcon's default routing engine is based on a decision tree that is
 first compiled into Python code, and then evaluated by the runtime.
+By default the decision tree is compiled only when the router handles
```
vytas7: I would put a comma here:
> `By default, the decision tree <...>`

falcon/routing/compiled.py
```diff
@@ -142,6 +164,20 @@ class can use suffixed responders to distinguish requests
                 Another class might use a suffixed responder to handle
                 a shortlink route in addition to the regular route for the
                 resource.
+            compile (bool): Optional flag that can be used to compile the
+                routing logic on this call. By default :class:`.CompiledRouter`
```
vytas7: `By default,` as well

falcon/routing/compiled.py
```diff
@@ -142,6 +164,20 @@ class can use suffixed responders to distinguish requests
...
+                delays compilation until the first request is routed. This may
+                introduce a noticeable amount of latency when handling the first
+                request, especially when the application implements a large
+                number of routes. Setting `compile` to ``True`` when the last
+                route is added ensures that the the first request will not be
```
vytas7: _the ~the~ first request_ (sic)

falcon/routing/compiled.py
```diff
@@ -531,6 +578,25 @@ def _instantiate_converter(self, klass, argstr=None):
         src = '{0}({1})'.format(klass.__name__, argstr)
         return eval(src, {klass.__name__: klass})
 
+    def _compile_and_find(self, path, _return_values, _patterns, _converters, params):
+        """Compiles the router, sets the `_find` attribute and returns its result
```
vytas7: Since this is a private method, probably it doesn't even need a perfect docstring.

But otherwise it would preferably need to start with an imperative (i.e. _Compile_), and the sentence should end with a full stop (period).

tests/test_compiled_router.py
```diff
@@ -0,0 +1,116 @@
...
+    num_threads = 3
+    barrier = Barrier(num_threads)
+
+    def find():
+        nonlocal calls
+        barrier.wait()
+        assert router.find('/foo') is None
+        calls += 1
+
+    threads = [Thread(target=find) for i in range(num_threads)]
```
vytas7: Cool multithreaded tests :100:

falcon/routing/compiled.py
CaselIT: I added it since it must keep the same interface as `_find` so a bit of documentation is useful, even if only to highlight among other private methods

CaselIT:
Thanks for the review, 
I've updated the documentation with the suggestions.

> I would suggest to mention the new compile keyword argument in App.add_route(...) docs,

I've added a short documentation there, linking to the more complete one on the router implementation

falcon/app.py
```diff
@@ -470,6 +470,10 @@ class can use suffixed responders to distinguish requests
                 Another class might use a suffixed responder to handle
                 a shortlink route in addition to the regular route for the
                 resource.
+            compile (bool): Optional flag that can be provided when using the default
+                :class:`.CompiledRouter` to compile the routing logic on this call,
+                since it will otherwise delays compilation until the first request
```
kgriffs: `delays` ➡️ `delay`

CaselIT:
I think that restarting the failed job should clear the failure

PR #1667, opened 2020-02-06
doc(asgi): Another round of doc updates, this time centered around Request/Response

kgriffs:
# Summary of Changes

Here's another round of ASGI-related doc improvements. This one centers mostly on things related to `Request` and `Response`.

# Related Issues

#1358

### DIFF
--- a/docs/api/cookies.rst
+++ b/docs/api/cookies.rst
@@ -9,10 +9,11 @@ Getting Cookies
 ~~~~~~~~~~~~~~~
 
 Cookies can be read from a request either via the
-:py:meth:`~.Request.get_cookie_values` method or the :py:attr:`~.Request.cookies`
-attribute on the :py:class:`~.Request` object. Generally speaking, the
-:py:meth:`~.Request.get_cookie_values` method should be used unless you need a
-collection of all the cookies in the request.
+:py:meth:`~.falcon.Request.get_cookie_values` method or the
+:py:attr:`~.falcon.Request.cookies` attribute on the
+:py:class:`~.falcon.Request` object. Generally speaking, the
+:py:meth:`~.falcon.Request.get_cookie_values` method should be used unless you
+need a collection of all the cookies in the request.
 
 .. code:: python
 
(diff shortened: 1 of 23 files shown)


kgriffs:
I still need to double-check a few things but this is almost ready for review...

falcon/asgi/stream.py
```diff
@@ -53,6 +108,7 @@ def __init__(self, receive, content_length=None):
         self._pos = 0
 
     def __aiter__(self):
+        """Hello"""
```
CaselIT: I think you missed this

docs/api/cookies.rst
```diff
@@ -9,10 +9,11 @@ Getting Cookies
 ~~~~~~~~~~~~~~~
 
 Cookies can be read from a request either via the
-:py:meth:`~.Request.get_cookie_values` method or the :py:attr:`~.Request.cookies`
-attribute on the :py:class:`~.Request` object. Generally speaking, the
-:py:meth:`~.Request.get_cookie_values` method should be used unless you need a
-collection of all the cookies in the request.
+:py:meth:`~.falcon.Request.get_cookie_values` method or the
```
kgriffs: I just fixed these up for now to resolve doc warnings, but I plan to circle back in all these places and sort out how to also point people to ASGI Request/Response docs.

falcon/asgi/structures.py
```diff
@@ -5,6 +5,60 @@
 
 
 class SSEvent:
```
vytas7: A finding not directly related to the proposed changeset, but there seems to be some asymmetry between parameter validation in the constructor and the `serialize()`-ation function.

Passing a parameter that does not reduce to truth in a boolean test can go unnoticed in the up front check, e.g.:
```python
>>> from falcon.asgi.structures import SSEvent
>>> event = SSEvent(text=0.0)
>>> event.serialize()
  File "/home/vytas/progs/falcon/falcon/asgi/structures.py", line 85, in serialize
    block += 'data: ' + self.text + '\n'
TypeError: can only concatenate str (not "float") to str
```

kgriffs: Nice catch. I'll address this in a followup PR.

PR #1670, opened 2020-02-11
test: Enable Windows gate on Travis CI

kgriffs:
# Summary of Changes

Enable Windows gate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,17 @@ matrix:
         - python: 3.8
           env: TOXENV=check_vendored
 
+        - name: Python 3.8 (Windows)
+          env: TOXENV=py38_nocover
+          os: windows
+          language: bash
+          before_install:
+            - choco install python3 --params "/InstallDir:C:\\Python"
+            - export PATH="/c/Python:/c/Python/Scripts:$PATH"
+            - python -m pip install --upgrade pip wheel coverage
+
+          script: tox
+
 script: tox -- -v
 
 notifications:
--- a/tests/asgi/test_asgi_servers.py
+++ b/tests/asgi/test_asgi_servers.py
@@ -104,25 +104,53 @@ def _run_server_isolated(process_factory, host, port):
     yield server
 
     if _WIN32:
+        # NOTE(kgriffs): Calling server.terminate() is equivalent to
+        #   server.kill() on Windows. We don't want to do the this;
+        #   forcefully killing a proc causes the Travis job to fail,
+        #   regardless of the tox/pytest exit code. """
+        #
+        #   Instead, we send CTRL+C. This does require that the handler be
+        #   enabled via SetConsoleCtrlHandler() in _uvicorn_factory()
+        #   below. Alternatively, we could send CTRL+BREAK and allow
+        #   the process exit code to be 3221225786.
+        #
         import signal
-        print('\n[Sending CTRL_C_EVENT to server process...]')
+        print('\n[Sending CTRL+C (SIGINT) to server process...]')
         server.send_signal(signal.CTRL_C_EVENT)
         try:
             server.wait(timeout=10)
-        except (KeyboardInterrupt, subprocess.TimeoutExpired):
+        except KeyboardInterrupt:
             pass
+        except subprocess.TimeoutExpired:
+            print('\n[Killing stubborn server process...]')
+            server.kill()
+            server.communicate()
+    else:
+        print('\n[Sending SIGTERM to server process...]')
+        server.terminate()
 
-    print('\n[Sending SIGTERM to server process...]')
-    server.terminate()
-
-    try:
-        server.communicate(timeout=10)
-    except subprocess.TimeoutExpired:
-        server.kill()
-        server.communicate()
+        try:
+            server.communicate(timeout=10)
+        except subprocess.TimeoutExpired:
+            print('\n[Killing stubborn server process...]')
+            server.kill()
+            server.communicate()
 
 
 def _uvicorn_factory(host, port):
+    if _WIN32:
+        script = f"""
+import uvicorn
+import ctypes
+ctypes.windll.kernel32.SetConsoleCtrlHandler(None, 0)
+uvicorn.run('_asgi_test_app:application', host='{host}', port={port})
+"""
+        return subprocess.Popen(
+            ('python', '-c', script),
+            cwd=_MODULE_DIR,
+            creationflags=subprocess.CREATE_NEW_PROCESS_GROUP,
+        )
+
     # NOTE(vytas): uvicorn+uvloop is not (well) supported on PyPy at the time
     #   of writing.
     loop_options = ('--http', 'h11', '--loop', 'asyncio') if _PYPY else ()
@@ -133,9 +161,12 @@ def _uvicorn_factory(host, port):
         '_asgi_test_app:application'
     )
 
+    creationflags = subprocess.CREATE_NEW_PROCESS_GROUP if _WIN32 else 0
+
     return subprocess.Popen(
         ('uvicorn',) + loop_options + options,
         cwd=_MODULE_DIR,
+        creationflags=creationflags,
     )
 
 
@@ -189,6 +220,7 @@ def server_base_url(request):
             yield base_url
 
         assert server.returncode == 0
+
         break
 
     else:
--- a/tox.ini
+++ b/tox.ini
@@ -57,6 +57,12 @@ deps = {[testenv]deps}
 whitelist_externals = {[with-coverage]whitelist_externals}
 commands = {[with-coverage]commands}
 
+[testenv:py38_nocover]
+deps = {[testenv]deps}
+       pytest-randomly
+       jsonschema
+commands = pytest tests []
+
 # --------------------------------------------------------------------
 # Additional test suite environments
 # --------------------------------------------------------------------


CaselIT:
This was giving problems also to me in the other branch, but I thought to have solved it with the ctrl_c signal. It seems that according to the shell used that may not work.
I'll try messing around with it a bit more to see if we stop this.
The main problem is that the call to terminate does nothing, so the process is killed after a timeout.

CaselIT:
I've tried to find a more reliable way of starting/stipping the process in windows, and it's a mess.
The best solution I came up with is to use a process that manually creates an uvcorn server. 
This process also starts a thread what waits for a signal (a value in a queue) form the test process to set `should_exit` of the uvicorn server to true. Something like

```py
server = None
def killThread(queue):
    queue.get()
    server.should_exit = True

def serverProcess(queue: multiprocessing.SimpleQueue):
    global server
    import uvicorn
    t = threading.Thread(target=killThread, args=(queue,))
    t.start()
    conf = uvicorn.Config('tests.asgi._asgi_test_app:application', host='127.0.0.1', port=3333)
    server = uvicorn.Server(conf)
    server.run()
    t.join()

# then it's started with
q = multiprocessing.SimpleQueue()
proc = multiprocessing.Process(target=serverProcess, args=(q,))
proc.start()

# and terminates with
q.put(None)
```

This correctly terminates the uvicorn server and the process, with exit code 0. But at this point I honestly think we can just kill the process on windows and be done with it.

It shouldn't be this hard...

kgriffs:
> It shouldn't be this hard...

It really shouldn't, heh. I guess for now maybe you are right that we just don't bother. If something goes wrong with the server otherwise, our `requests`-based tests will fail anyway.

CaselIT:
Indeed.
While investigating this, I've noticed that the solution that I added, the `send_signal` is not supposed to work, as per documentation https://docs.python.org/3/library/signal.html#signal.CTRL_C_EVENT but it seems to work.

Maybe we should just remove the logic to kill the server on windows, and just kill it and accept that the exit code will be 1..

kgriffs:
So... turns out that killing the unvicorn process may be causing travis to decide the build failed. When I change the script to this:

`tox -- -k 'not test_asgi_servers'`

Suddenly Travis starts to flag the build as succeeded.

tests/asgi/test_asgi_servers.py
```diff
@@ -109,17 +109,20 @@ def _run_server_isolated(process_factory, host, port):
         server.send_signal(signal.CTRL_C_EVENT)
         try:
             server.wait(timeout=10)
-        except (KeyboardInterrupt, subprocess.TimeoutExpired):
+        except KeyboardInterrupt
             pass
+        except subprocess.TimeoutExpired:
+            server.kill()
```
kgriffs: It seems like it makes sense to just go straight for the kill in this case. According to the docs, `kill()` is just an alias for `terminate()` on Windows anyway.

tests/asgi/test_asgi_servers.py
```diff
@@ -109,17 +109,20 @@ def _run_server_isolated(process_factory, host, port):
         server.send_signal(signal.CTRL_C_EVENT)
         try:
             server.wait(timeout=10)
-        except (KeyboardInterrupt, subprocess.TimeoutExpired):
+        except KeyboardInterrupt
             pass
```
kgriffs: I wasn't sure about this case. When KeyboardInterrupt is raised, does that mean the process got the CTRL+C and exited cleanly?

tests/asgi/test_asgi_servers.py
```diff
@@ -109,17 +109,20 @@ def _run_server_isolated(process_factory, host, port):
         server.send_signal(signal.CTRL_C_EVENT)
```
kgriffs: Apparently this is supported when "processes started with a creationflags parameter which includes CREATE_NEW_PROCESS_GROUP". I added that flag, but it doesn't seem to make any difference on Travis.

https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal

tests/asgi/test_asgi_servers.py
```diff
@@ -133,9 +136,12 @@ def _uvicorn_factory(host, port):
         '_asgi_test_app:application'
     )
 
+    creationflags = subprocess.CREATE_NEW_PROCESS_GROUP if _WIN32 else 0
```
kgriffs: See comment above.

.travis.yml
```diff
@@ -9,48 +9,68 @@ cache:
...
+          before_install:
+            - choco install python3 --params "/InstallDir:C:\\Python"
+            - export PATH="/c/Python:/c/Python/Scripts:$PATH"
+            - python -m pip install --upgrade pip wheel
+
+          # TODO: Figure out why Travis CI fails the build when a spawned
+          #   uvicorn process does not terminate cleanly and/or figure out
+          #   a way to cleanly shut down the uvicorn process in the first
+          #   place.
+          script: tox -- -k 'not test_asgi_servers'
```
kgriffs: I guess we could move forward with enabling the gate like this until we sort out the root cause?

tests/asgi/test_asgi_servers.py
CaselIT: From my tests, kill is not an alias for terminate, since terminate seems to be completely ignored on windows, while kill does kill the process

tests/asgi/test_asgi_servers.py
CaselIT: It's a bit strange. I think there are bugs in how python handles this.
I think that this ctrl_c signal is sent to the whole process tree, not only the process you send it to, so the caller recieves a ctrl_c event, and that does not make any sense

tests/asgi/test_asgi_servers.py
CaselIT: I tried it, and if irrc when starting a process in a new process group the signals are ignored because the current process can only send signals to its group. So what eneded up happening is that only the caller recieves the ctrl_c

.travis.yml
CaselIT: The whole behaviour is very strange on windows. The docs seem to say something, but what works is something else thats not supposed to work.

I'm not sure if it's uvicorn that does something strange that makes the process behave this way or it's simply buggy the python implementation on windows. (Maybe it's just windows that's biggy)

CaselIT:
> Suddenly Travis starts to flag the build as succeeded.

Have you tried removing the ctrl_c logic and just kill the process on windows?
Maybe it's this that makes Travis fail?

tests/asgi/test_asgi_servers.py
CaselIT: ctrl_c is ignored by default, unless the child process explicitly enables support for it, ctrl_break works, but the exit code is not 0

kgriffs: Thanks! That info was what I was missing.

kgriffs:
Seems to be working now!

CaselIT:
I may have found a way to cleanly exit with 0, that should not be too hard to do.

```py
def _uvicorn_factory(host, port):
    if _WIN32:
        commnad = """
import uvicorn
import ctypes
ctypes.windll.kernel32.SetConsoleCtrlHandler(None, 0)
uvicorn.run('_asgi_test_app:application', host='{}', port={})
"""
        return = subprocess.Popen(
            ("python", "-c", commnad.format("127.0.0.1", 3333)),
            cwd=_MODULE_DIR,
            creationflags=subprocess.CREATE_NEW_PROCESS_GROUP,
        )
```

The `ctypes.windll.kernel32.SetConsoleCtrlHandler(None, 0)` enables the crtl_c handler. So we would send ctr_c and the exti code should be 0

tests/asgi/test_asgi_servers.py
```diff
@@ -188,7 +201,11 @@ def server_base_url(request):
 
             yield base_url
 
-        assert server.returncode == 0
+        # NOTE: In the Travis CI environment, sending CTRL+BREAK works but
+        #   results in a non-zero return code (3221225786 observed?).
```
CaselIT: That number is the normal exit code after a ctrl break

CaselIT:
> 
> 
> I may have found a way to cleanly exit with 0, that should not be too hard to do.
> 
> ```python
> def _uvicorn_factory(host, port):
>     if _WIN32:
>         commnad = """
> import uvicorn
> import ctypes
> ctypes.windll.kernel32.SetConsoleCtrlHandler(None, 0)
> uvicorn.run('_asgi_test_app:application', host='{}', port={})
> """
>         return = subprocess.Popen(
>             ("python", "-c", commnad.format("127.0.0.1", 3333)),
>             cwd=_MODULE_DIR,
>             creationflags=subprocess.CREATE_NEW_PROCESS_GROUP,
>         )
> ```
> 
> The `ctypes.windll.kernel32.SetConsoleCtrlHandler(None, 0)` enables the crtl_c handler. So we would send ctr_c and the exti code should be 0

I've tried this and it seems to work for travis
https://github.com/CaselIT/falcon/commit/928da0ec9d97b90fbfc5f65993d79b20f64f12bb
https://travis-ci.org/CaselIT/falcon/builds/650774332

kgriffs:
OK, I have applied the suggestion from @CaselIT - seems to work!

tests/asgi/test_asgi_servers.py
vytas7: The exit code does not need to be 0 on Unix-like either.
It is normal that, for instance, on Linux , terminating a program with a signal `N` makes it return an exit code `-N`.

PR #1682, opened 2020-02-23
perf(Request): short-circuit path encode/decode when possible

### LINKED ISSUE #1492: perf(Request): Short-circuit path encode/decode when possible
See if we can quickly determine if the following is required for a given path and short-circuit it when it is not (from `Request.__init__`):

```py
# PEP 3333 specifies that PATH_INFO variable are always
# "bytes tunneled as latin-1" and must be encoded back
path = path.encode('latin1').decode('utf-8', 'replace')
```

See also this section from `falcon/testing/helpers.py`:

```py
# NOTE(kgriffs): The decoded path may contain UTF-8 characters.
# But according to the WSGI spec, no strings can contain chars
# outside ISO-8859-1. Therefore, to reconcile the URI
# encoding standard that allows UTF-8 with the WSGI spec
# that does not, WSGI servers tunnel the string via
# ISO-8859-1. falcon.testing.create_environ() mimics this
# behavior, e.g.:
#
#   tunnelled_path = path.encode('utf-8').decode('iso-8859-1')
#
# falcon.Request does the following to reverse the process:
#
#   path = tunnelled_path.encode('iso-8859-1').decode('utf-8', 'replace')
#
path = path.encode('utf-8').decode('iso-8859-1')
```

vytas7:
Closes #1492 

# Summary of Changes

Improve performance of WSGI path decoding by short-circuiting it to return the same path if it only contains ASCII characters.

# Testing:
The suite:
```python
import timeit

from falcon.util.misc import isascii


print(isascii.__doc__)

PATH = '/data/items/something?query=apples%20and%20oranges'


def before():
    path = PATH
    path = path.encode('iso-8859-1').decode('utf-8', 'replace')
    return path


def after():
    path = PATH
    if not isascii(path):
        path = path.encode('iso-8859-1').decode('utf-8', 'replace')
    return path


def test_timeit_before():
    timeit.timeit(before, number=int(1e7))
    print([timeit.timeit(before, number=int(1e7)) for _ in range(3)])


def test_timeit_optimized():
    timeit.timeit(after, number=int(1e7))
    print([timeit.timeit(after, number=int(1e7)) for _ in range(3)])
```

## CPython 3.5.9 (Docker image: `python:3.5`, no Cython)
```
test.py::test_timeit_before [4.975278416000037, 4.959721629999876, 4.948355078000532]
PASSED
test.py::test_timeit_optimized [4.783882353000081, 4.784126579000258, 4.789863130999947]
PASSED
```

## CPython 3.5.9 (Docker image: `python:3.5`, `cythonize -i misc.pyx`)
```
test.py::test_timeit_before [5.003832942999907, 5.0074363920002725, 5.0109554970003956]
PASSED
test.py::test_timeit_optimized [1.842615818000013, 1.842888543999834, 1.8508781330001511]
PASSED
```

## CPython 3.8.2 (Docker image: `python:3`)
```
test.py::test_timeit_before [3.779901597999924, 3.7834596579996287, 3.787558049999461]
PASSED
test.py::test_timeit_optimized [1.8798802580004121, 1.8772937459998502, 1.8789225440004884]
PASSED
```

## PyPy 3.6.9 (`pypy-7.3.0`) (Docker image: `pypy:3`)

I had to modify the suite for PyPy since the JIT was "cheating" by caching constants:
```python
import timeit

from falcon.util.misc import isascii


print(isascii.__doc__)

PATH = '/data/items/something?query=apples%20and%20oranges'

counter = 0


def before():
    global counter
    counter += 1

    path = PATH + str(counter)
    path = path.encode('iso-8859-1').decode('utf-8', 'replace')
    return path


def after():
    global counter
    counter += 1

    path = PATH + str(counter)
    if not isascii(path):
        path = path.encode('iso-8859-1').decode('utf-8', 'replace')
    return path


def test_timeit_before():
    global counter
    counter = 0

    timeit.timeit(before, number=int(1e7))
    print([timeit.timeit(before, number=int(1e7)) for _ in range(3)])


def test_timeit_optimized():
    global counter
    counter = 0

    timeit.timeit(after, number=int(1e7))
    print([timeit.timeit(after, number=int(1e7)) for _ in range(3)])
```
```
test.py::test_timeit_before [0.6101447539995206, 0.6059662850002496, 0.6076860740004122]
PASSED
test.py::test_timeit_optimized [0.42552743699980056, 0.4231114679996608, 0.42348902499998076]
PASSED
```

PyPy is inherently faster here because it uses UTF-8 like (single-byte) representation for Unicode internally, which requires less memory throughput to check the same things.

# Related Issues

# Pull Request Checklist

Not done yet, waiting to sync with https://github.com/falconry/falcon/pull/1621:


kgriffs:
For noteworthy performance improvements, would it make sense to include a news fragment?

vytas7:
@kgriffs It does make sense, I was just unsure what we settled on wrt the naming schema & extension categories.
But I'll add a tentative one, so we can rename later.

kgriffs:
What is the impact of this change on performance without Cython? I would check 3.6 as well as 3.8.

falcon/util/uri.py
```diff
@@ -406,9 +408,31 @@ def unquote_string(quoted):
...
+        str: decoded string
+    """
+    try:
+        # PERF(vytas): Use EAFP here to quickly check if there are any
+        #   non-ASCII bytes in the string, and return the input verbatim if
+        #   possible.
+        tunneled.encode('ascii')
+        return tunneled
+    except ValueError:
+        return tunneled.encode('iso-8859-1').decode('utf-8', 'replace')
```
kgriffs: How does performance of this approach compare to [str.isascii()](https://docs.python.org/3/library/stdtypes.html?highlight=isascii#str.isascii)?

vytas7: OMG :man_facepalming: 

I managed to overlook the existence of this method, and reimplemented it. I'll check...

kgriffs: Heh, no worries. We'll still need a solution for 3.5 and 3.6 either way.

vytas7: Oh, I got scared it had existed since Stone Age, and I was reimplementing it in Cython :slightly_smiling_face:

vytas7:
I was actually rewriting this from the isascii perspective. The built-in one has better perf than Cython allows me to achieve (it allows semi-efficient unicode char iteration, but I cannot access the string memory in a compatible, non deprecated way. So I cannot process 64-bit or more per cycle).

vytas7:
@kgriffs Ready for-review now, I hope!

I'm still testing that builtin `str.isascii`. It looked like it was outperforming the cythonized version slightly, but 3.5-cythonized actually outdid 3.8 in the results presented above.
I'll add a test comparing those for 3.8.

Either way, the difference should be very insignificant, and IMHO it makes more sense to just use the builtin function. Thoughts?

PR #1686, opened 2020-02-27
doc(multipart): Edit note re how the request body is consumed

kgriffs:
# Summary of Changes

In this patch, I reordered the clauses in the first sentence of the first note in the multipart forms overview and made a few other minor edits for clarity and correctness.

# Related Issues

#953

### DIFF
--- a/docs/api/multipart.rst
+++ b/docs/api/multipart.rst
@@ -30,12 +30,13 @@ default, allowing you to use ``req.media`` to iterate over the
             # TODO: Do something else
 
 .. note::
-   The request stream is only consumed along iteration over the body parts in a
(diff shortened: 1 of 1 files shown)


vytas7:
I have also tried skimming through those multipart docs from the user perspective (impossible that is probably, but still), and I realized we should also add one or sentences of prose to the [MultipartFormHandler](https://falcon.readthedocs.io/en/3.0.0a1/api/media.html#falcon.media.MultipartFormHandler) class docstring, as well as linking to "See also: [Multipart Forms](https://falcon.readthedocs.io/en/3.0.0a1/api/multipart.html)".

PR #1687, opened 2020-02-27
refactor(asgi.Request): Change the media property to an alias of get_media()

kgriffs:
# Summary of Changes

This patch adds back support for the media property to help make the ASGI
and WSGI interfaces as similar as possible.

# Related Issues

#1358


vytas7:
This looks good, I'm thinking to adopt the same thinking for multipart media, the async flavour.

PR #1698, opened 2020-03-15
chore(mypy): fix mypy warnings & re-enable the mypy gate in Travis

vytas7:

### DIFF
--- a/falcon/asgi/response.py
+++ b/falcon/asgi/response.py
@@ -14,7 +14,7 @@
 
 """ASGI Response class."""
 
-from asyncio.coroutines import CoroWrapper
+from asyncio.coroutines import CoroWrapper  # type: ignore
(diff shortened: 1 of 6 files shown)


falcon/asgi/response.py
```diff
@@ -14,7 +14,7 @@
 
 """ASGI Response class."""
 
-from asyncio.coroutines import CoroWrapper
+from asyncio.coroutines import CoroWrapper  # type: ignore
```
vytas7: An alternative fix is to do `from asyncio.coroutines import CoroWrapper as CoroWrapper`, but that also looks awkward.
Thoughts?

kgriffs: I think it's OK as-is since it makes it clear this is a mypy workaround.

Issue #1589, opened 2019-10-25
Media handlers do not get a chance to deserialize zero-length payload


vytas7:
At the time of writing, `Request.media` property implementation starts this way:
```python
if self._media is not None or self.bounded_stream.is_exhausted:
    return self._media
```

If media in question is serialized to an empty (byte)string, i.e. the request `Content-Length` is zero, the `bounded_stream` is automatically considered exhausted. Hence `Request.media` automatically reduces to `None` without even resolving media handlers for the submitted content type.

While this is the correct behaviour for certain types of media, such as YAML, and for many other media types such as JSON, Msgpack, most audio/video/picture types etc it looks like an acceptable (albeit debatable since deserialization of empty input for these is undefined or defined to be malformed -- see also the discussion in https://github.com/falconry/falcon/issues/1216 ) workaround, there might be cases where an empty string could be expected to be deserialized to another value, such as in the `application/x-www-form-urlencoded` case:
```python
>>> import urllib.parse
>>> urllib.parse.parse_qs(b'')
{}
```

kgriffs:
I agree that now would be a good time to address this long-standing issue while we can make some breaking changes for 3.0.

Based on my understanding of the discussion so far, this is what I propose in terms of requirements:

- [ ] Provide a more explicit way for the call chain and the app to distinguish between an empty body and a failure by a handler to parse the body. See also the next requirement.
- [ ] Provide a way for the app to override behavior when a parse error is encountered vs. just always returning with HTTPBadRequest (I like [this suggestion](https://github.com/falconry/falcon/issues/1216#issuecomment-448619858)).
- [ ] Allow different media handlers to treat an empty body in whatever way makes the most sense for the media type in question (and point this out in the docs for media handler implementers and in the implemented media handler docstrings). I don't think we need to have them raise an EmptyBodyError as suggested elsewhere. For example, `URLEncodedFormHandler` could return {} in that case, while `JSONHandler` could opt to return `None`.
- [ ] The `media` property should distinguish between the body having already been consumed vs. an empty body, and in the former case return whatever was parsed previously (if anything was parsed previously; otherwise if something else consumed the body before we had a chance at it, fall back to same behavior as an empty body?). This will require initializing `self._media` to some kind of special value rather than `None`.

kgriffs:
[...]

PR #1701, opened 2020-03-21
Improve error and status handling

### LINKED ISSUE #452: Error representation strangeness
Hello,

I'm using falcon 0.2.0rc2, and I've noticed some behaviour regarding error representation that I don't understand:

1) An instance of `HTTPNotFound` without a description claims to have no representation even if it does have a title. This differs from the behaviour of `HTTPError`:

``` python
>>> import falcon
>>> falcon.HTTPError(falcon.HTTP_404, title='Not found.').has_representation
True
>>> falcon.HTTPNotFound(title='Not found.').has_representation
False
```

2) An error that claims to have no representation does not result in an empty response body even though the documentation states otherwise (http://falcon.readthedocs.org/en/0.2.0rc2/api/errors.html#HTTPError.has_representation). This means that a response body that is set earlier is sent to the client (instead of an empty body)

``` python
#!/usr/bin/env python3
import falcon
from wsgiref.simple_server import make_server

class Foo:
    def on_get(self, req, resp):
        resp.body = 'I get sent to the client.'
        raise falcon.HTTPNotFound(title='Not found')  # has_representation is False

app = falcon.API()
app.add_route('/foo', Foo())

httpd = make_server('', 5000, app)
httpd.serve_forever()
```

I'm not sure if those are bugs, but they seemed unintuitive enough to me to report them here.

[edit: corrected wrong error code in example 1]

CaselIT:
# Summary of Changes

Improve error and status handling:
- all falcon errors have a representation
- remove `has_representation` and `OptionalRepresentation`
- ensure that falcon clears the previous body when handling an `HttpStatus` or an `Error`
- fix `HttpError.to_xml` return type docs and set it directly to `Request.data`

# Related Issues

Fixes #452

# Pull Request Checklist


CaselIT:
not sure why codecov misreported the coverage

kgriffs:
Blocked on: https://github.com/pytest-dev/pytest-randomly/issues/246

If a new version of pytest-randomly is not immediately forthcoming, we'll have to pin the version.

CaselIT:
that was fast. already released https://github.com/pytest-dev/pytest-randomly/issues/246#issuecomment-614183962

tests/test_httperror.py
```diff
@@ -583,15 +583,15 @@ def test_405_without_body(self, client):
 
         response = client.simulate_request(path='/405')
         assert response.status == falcon.HTTP_405
-        assert not response.content
+        assert response.text == falcon.HTTPMethodNotAllowed(['PUT']).to_json()
```
vytas7, line 586: Do we have assertions against the actual response content?
The above approach is cool and DRY, but if there was a bug in `to_json()`, this would pass. In other words, this test is just verifying that the `to_json()` method is called, and the call result is transferred to the response body.

docs/_newsfragments/777.breakingchange.rst
```diff
@@ -1,2 +1,2 @@
-The class :class:`~.falcon.http_error.NoRepresentation` was removed since
-all :class:`~.falcon.HTTPError` have optional representation.
\ No newline at end of file
+The class :class:`~.falcon.http_error.NoRepresentation` was removed. All
```
vytas7, line 1: Same as above, the breaking change is too abrupt IMHO.

falcon/app.py
```diff
@@ -880,6 +870,9 @@ def _compose_status_response(self, req, resp, http_status):
         # NOTE(kgriffs): If http_status.body is None, that's OK because
         # it's acceptable to set resp.body to None (to indicate no body).
         resp.body = http_status.body
+        # NOTE(caselit): Reset the data and media, so that the body is
+        # sent in all cases, even when it's None
```
vytas7, line 874: Could explain the "even when it's `None`" part? How can the body be sent if it is `None`?

docs/_newsfragments/777.breakingchange.rst
CaselIT: This was removed in https://github.com/falconry/falcon/pull/1655

falcon/app.py
CaselIT: I'll reword it.
What I meant it that we should not use data or media of body is `None`

falcon/http_error.py
```diff
@@ -136,10 +123,6 @@ def __init__(self, status, title=None, description=None, headers=None,
     def __repr__(self):
         return '<%s: %s>' % (self.__class__.__name__, self.status)
 
-    @property
```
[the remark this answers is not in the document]
CaselIT: Ok.

falcon/http_error.py
```diff
@@ -208,26 +191,3 @@ def to_xml(self):
 
         return (b'<?xml version="1.0" encoding="UTF-8"?>' +
                 et.tostring(error_element, encoding='utf-8'))
-
-
-class OptionalRepresentation:
```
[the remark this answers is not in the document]
CaselIT: In https://github.com/falconry/falcon/pull/1655 `NoRepresentation` was removed completely. I think that we should keep both or remove both

tests/test_httperror.py
CaselIT: Yes, this test does only that. It assumes that `.to_json()` is correct.
`.to_json()` should be already tested somewhere else. 

In case it's missing I can add that test, but I would not retest it here, or this test (all the other similar ones) become overly broad

vytas7: As long as `to_json()` is tested elsewhere for `HTTPMethodNotAllowed`, it's fine.

falcon/http_error.py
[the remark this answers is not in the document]
vytas7: Agreed that we should be consistent.

My vote, as also explained in other comments, is on keeping both, as well as keeping `has_representation`.
@kgriffs thoughts?

tests/test_httperror.py
CaselIT: It seems not. I'll test in one tests

CaselIT:
I've update the code after the review. The only thing left to decide is what to do regarding
`OptionalRepresentation ` and `NoRepresentation `

falcon/http_error.py
```diff
@@ -138,6 +127,8 @@ def __repr__(self):
 
     @property
     def has_representation(self):
+        warn = 'has_representation is deprecated and is currently unused by falcon'
+        warnings.warn(warn, DeprecatedWarning)
```
vytas7, line 131: In the default CPython configuration, these are unfortunately oftentimes suppressed.

Since under the hood a property is a function, maybe you could just reuse our [`deprecated`](https://github.com/falconry/falcon/blob/master/falcon/util/misc.py#L123) utility?

CaselIT: I use `DeprecatedWarning` of falcon.

I'm not using deprecated since it adds `Call to deprecated function has_representation(...)` before the specified warning, and I thought that it may create confusion

kgriffs:
I agree that we should either remove both classes or keep both. That being said, if we leave them around and just say they are deprecated (and the framework itself no longer makes use of them; that could still be a breaking change that we leave in and note in the news fragment, since it should have minimal impact), we would have to put back the checks for `has_representation` in case someone's custom error class is using one of the deprecated mixins. 

By itself, removing the classes altogether would probably not cause a lot of issues for existing apps, but I can appreciate @vytas7 "thousand paper cuts" argument. Perhaps it would be a good idea to avoid introducing additional breaking changes going forward for the remainder of the 3.x work.

CaselIT:
While I agree that breaking changes should be minimized, I don't think there are drastic changes for version 3 so I don't think those will cause a lot of issues.
And in the current milestone for v3 of breaking changes there is only this one and #1589.

That said, I'm cool either way

> we would have to put back the checks for has_representation in case someone's custom error class is using one of the deprecated mixins.

I don't think we should restore it even if we re-add the mixin classes.
It's easy to provide a custom serializer with the previous logic if one is needed:

```py
def mySerializer(req, resp, exception):
    if not exception.has_representation:
        return
    return falcon.app_helpers.default_serialize_error(req,resp,expeption)
```

vytas7:
Good points @kgriffs @CaselIT ... Right, this is a bit tricky, because it's hard to make the change without actually making a change :thinking:

I'm not too opinionated about this.
Just off the top of my head, I would probably do something along the lines of
* Resurrect `NoRepresentation`, keep `OptionalRepresentation`. Document them as deprecated.
* Keep the `has_representation` property, but document it as deprecated. Referencing emits a warning.
* **Breaking change** (but very low impact). No standard Falcon classes any longer use these mixins.
* **Breaking change**. The default error serializer no longer respects the `has_representation` property.

CaselIT:
If we all agree can update this with the proposed changes in the latest message.

Regarding NoRepresentation and OptionalRepresentation, since they are mixins, should I add a warning? The best place for this, would probably be [__init_subclass__](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__) since they don't have __init__ and it's probably best to warn when a subclass is defined on application load, but that's only py3.6+.
I can fix it with a metaclass, but I'm not sure it's worth the hassle for a deprecated feature...

Suggestions?

vytas7:
I don't think we need any deprecation warnings on mixins, as there is no clean way indeed.
The only meaningful logic these mixins provide are centered on the `has_representation` property anyway, so as long as referencing it emits a warning, we should be fine IMHO.

CaselIT:
Ok, I will warn in `has_representation`

CaselIT:
I've updated the pull request.

I've added an attribute to deprecated for when it's used on properties. I've also removed the exception on warning when falcon testing is the env, since I think we should test that the deprecated stuff actually raises the deprecation warning in the tests.

CaselIT:
MyPy fails in a strange way.
> falcon/http_error.py:126: error: Decorated property not supported
> falcon/http_error.py:227: error: Decorated property not supported
> falcon/http_error.py:259: error: Decorated property not supported
> Found 3 errors in 1 file (checked 77 source files)

is there a way of skipping these checks?

kgriffs:
> MyPy fails in a strange way.
> 
> > falcon/http_error.py:126: error: Decorated property not supported
> > falcon/http_error.py:227: error: Decorated property not supported
> > falcon/http_error.py:259: error: Decorated property not supported
> > Found 3 errors in 1 file (checked 77 source files)
> 
> is there a way of skipping these checks?

You can try appending `# type: ignore` to the lines in question.

CaselIT:
Thansk, that did it

CaselIT:
Anything missing from this one?

docs/_newsfragments/452.breakingchange.rst
```diff
@@ -0,0 +1,6 @@
+The class :class:`~.falcon.http_error.OptionalRepresentation` and the attribute
+:attr:`~.falcon.HTTPError.has_representation` were removed. The default error
+serializer now generates a representation for every error type that derives from
+:class:`falcon.HTTPError`. In addition, Falcon now ensures that any previously
+set response body is cleared before handling any subclass of
+:class:`falcon.HTTPStatus` or any other raised exception.
```
vytas7, line 6: Not sure the latter part of the statement is universally true.

Consider, for instance, the following app:
```python
import falcon


class Resource:

    def on_get(self, req, resp):
        resp.media = {'message': 'Hello, World!'}
        resp.set_header('X-Value', str(0/0))


def handle_zero_division(req, resp, ex, params):
    print('resp.media ==', resp.media)
    print('This went wrong:', ex)


app = falcon.App()
app.add_error_handler(ZeroDivisionError, handle_zero_division)
app.add_route('/test', Resource())
```
```
resp.media == {'message': 'Hello, World!'}
This went wrong: division by zero
```
I.e., it seems that when handling _any other_ raised exception, the response body is not necessarily cleared.

falcon/http_error.py
```diff
@@ -210,6 +201,39 @@ def to_xml(self):
...
+    a body (as dictated by HTTP standards or common practice), or in the
+    case that a detailed error response may leak information to an attacker.
+
+    Note:
+        This mixin class must appear before ``HTTPError`` in the base class
+        list when defining the child; otherwise, it will not override the
+        `has_representation` property as expected.
+
+    Warning:
+        As of Falcon 3.0, this mixin class is no longer used since all falcon
```
vytas7, line 222: _Falcon_ and _falcon_ -- different capitalization in the same line.

docs/_newsfragments/452.breakingchange.rst
CaselIT: Nice catch. I'll fix that, since my intention was to do as in the note :)

docs/_newsfragments/777.breakingchange.rst
CaselIT: corrected to deprecated, since it was readded

CaselIT:
Updated the documentation and fixed the bug on custom exceptions.
@vytas7 can you take another look

falcon/app.py
```diff
@@ -936,6 +926,8 @@ def _handle_exception(self, req, resp, ex, params):
         """
         err_handler = self._find_error_handler(ex)
 
+        # NOTE(caselit): Reset body, data and media before calling the handler
+        resp.body = resp.data = resp.media = None
```
vytas7, line 930: Shouldn't we move these statements inside under the `if err_handler is not None`?
In which scenario is it beneficial to reset the body if we haven't found any error handler?

CaselIT: Yes, there should be no difference, since if we do not have an err_handler falcon does not manage the response and instead makes the wsgi serve handle it.
That should only happen on exception that do not inherit from `Exception` now, like `KeyboardInterrupt` or `SystemExit`, since we add an handler for `Exception`. This case should be very rare, so it's probably the same to have it inside or outside the if.

PR #1705, opened 2020-03-30
doc(install): Add --no-build-isolation to examples

kgriffs:


CaselIT:
You can avoid pep517 until falcon does not actually use it, and with it it's build isolation rules, by omitting `pyproject.toml` from the distribution by changing `manifest.ini`.
Sqlalchemy is currently using this behavior: https://github.com/sqlalchemy/sqlalchemy/blob/master/MANIFEST.in#L17

You can check this since Sqlalchemy version 1.13.15 had pyproject in the distribution and therefore uses pep517 vs 1.13.16 removed it, and uses the old method.

Issue #1640, opened 2020-01-18
Add cookies parameter to simulate_request


vytas7:
Add a new parameter, _cookies_, to our test client's `simulate_request` & its specialized friends `simulate_*`, where applicable.

We can either accept just a mapping object, or also support fallback to an iterable over pairs if the provided mapping doesn't quack `items()`? Compared to just supporting a mapping, this would enable sending multiple cookies with the same name, which is rather uncommon but apparently not forbidden: [RFC 6265, Section 4.2.2](https://tools.ietf.org/html/rfc6265#section-4.2.2). :arrow_left: @kgriffs thoughts?
Or should we even supporting sending a full-fledged morsel jar, [like `requests` does](https://requests.readthedocs.io/en/master/user/quickstart/#cookies)? :thinking:

Regardless of the interface chosen, the provided cookie key-value pairs should be serialized per [RFC 6265, Section 5.4](https://tools.ietf.org/html/rfc6265#section-5.4).

- [ ] Implement a new parameter _cookies_ in `simulate_request`
- [ ] Add the new parameter to the specialized `simulate_*` methods, where applicable (check, maybe `OPTIONS` doesn't need _cookies_ as browsers to not expose them in preflight IIRC)
- [ ] Revise the FAQ item from https://github.com/falconry/falcon/pull/1632, if merged

mikeyusko:
@vytas7 Hm, I'll take a look into it.

myuz:
@vytas7 the issue is still relevant? if yes, I'll work on it, and sorry for an extreme delay.

vytas7:
@myuz Sure, feel free to! :+1: 

And yes, it is still very relevant.

vytas7:
This was resolved fixed in
https://github.com/falconry/falcon/pull/1715
* https://github.com/falconry/falcon/pull/1776

PR #1715, opened 2020-04-19
Add cookies parameter to simulate_* family

myuz:
# Summary of Changes

Added `cookies` parameter to` simulate_ * `family methods.

That's PR is not completed yet, and I'm also looking to check with the community whether I'm going in the right direction, and correctly building `cookies` for the` simulate_request` method.

For now, it's working fine and `cookies` passed fine to` Request` object.

Example:
```
from falcon import App


class CustomTestCookies:
    def items(self):
        return [('cls', 'class cookie')]


class TestResource:
    def on_post(self, req, resp):
        print(req.cookies)

    def on_get(self, req, resp):
        print(req.cookies)

app = App()
app.add_route('/', TestResource())

test_client = TestClient(app)

# print ->>> {'cookie': 'one', 'cookie-two': 'two'}
response = test_client.simulate_post('/', cookies={'cookie': 'one', 'cookie-two': 'two'})

# print ->>> {'cls': 'class cookie'}
second_response = test_client.simulate_get('/', cookies=CustomTestCookies())
```
Also, one more point -  Right now I see only one way how we can test it.
we should test [create_environ](https://github.com/falconry/falcon/blob/master/falcon/testing/helpers.py#L492) method and check whether it include `HTTP_COOKIE` key.
 
Thoughts?

# Related Issues

https://github.com/falconry/falcon/issues/1640

# Pull Request Checklist

- [] Prefixed code comments with GitHub nick and an appropriate prefix.
- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/index.html) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)


falcon/testing/helpers.py
```diff
@@ -631,6 +631,12 @@ def create_environ(path='/', query_string='', http_version='1.1',
     if content_length != 0:
         env['CONTENT_LENGTH'] = str(content_length)
 
+    if cookies is not None and method != 'OPTIONS':
```
CaselIT, line 634: why the filter on the method?

falcon/testing/client.py
```diff
@@ -258,7 +258,7 @@ def simulate_request(app, method='GET', path='/', query_string=None,
                      file_wrapper=None, wsgierrors=None, params=None,
                      params_csv=True, protocol='http', host=helpers.DEFAULT_HOST,
                      remote_addr=None, extras=None, http_version='1.1',
-                     port=None, root_path=None, asgi_chunk_size=4096,
+                     port=None, root_path=None, cookies=None, asgi_chunk_size=4096,
```
CaselIT, line 261: cookies should be documented

falcon/testing/helpers.py
myuz: `OPTIONS` method discards `Set-Cookie` header, and that's the reason why we are checking the method here.

falcon/testing/client.py
myuz: yeah, sure

myuz:
@CaselIT I think the PR is ready for review.

FYI; I also changed these two lines [1](https://github.com/falconry/falcon/pull/1715/files#diff-ae00b1cb2ba4b3b3028ab05d2a2b4e6aR1) and [2](https://github.com/falconry/falcon/pull/1715/files#diff-ae00b1cb2ba4b3b3028ab05d2a2b4e6aR1) I've worked on it, and to be consistence decided to change my old nickname to a new one;)

myuz:
Ok, `py3.5` tests were failed, and I guess it related to `insertion-order `in `dict`, since `Python >=2.7 and <3.6` doesn't support `insertion-order`. Should we split these two tests, or just use `OrderedDict`?

@kgriffs @CaselIT thoughts?

CaselIT:
I think you could use an orderdict, or even  compare with `in ('foo=bar; baz=foo', 'baz=foo; foo=bar')`

vytas7:
@myuz Have you had a chance to look into supporting not only key-value dicts, but also :cookie: jars, for instance, in the same vein as [`requests` do](https://requests.readthedocs.io/en/master/user/quickstart/#cookies)?

myuz:
@vytas7 Hey!

Be honest, I don't see any advantages of `jars` for `testing` purposes, in case if we need it, for general use, maybe you are right.

vytas7:
@myuz As I also wrote on Gitter, regarding the cookie jar functionality, I was thinking it could be useful to be able to simulate authentication sessions/browser by sending back [`Result.cookies`](https://falcon.readthedocs.io/en/stable/api/testing.html#falcon.testing.Result.cookies) to the next simulated request.

Would it be possible out of the box in your approach?

myuz:
@vytas7 The current approach doesn't support it, I also agree with you that it will be a useful thing.

OK, I'll add support for `jars`.

myuz:
@vytas7 @CaselIT  PR is ready for review.

Also added support for `jars`, in the example under I providing code snippet how it looks.

```
import falcon
import json
from falcon import testing


class Foo:
    def on_get(self, req, resp):
        resp.set_cookie('has_permission', 'true')

    def on_post(self, req, resp):
        if req.cookies['has_permission'] == 'true':
            resp.body = json.dumps({'has_permission': 'Yes!'})


app = falcon.App()
app.add_route('/jars', Foo())

client = testing.TestClient(app)

response_one = client.simulate_get('/jars')
response_two = client.simulate_post('/jars', cookies=response_one.cookies)

assert response_two.json == {'has_permission': 'Yes!'}

```

myuz:
Yeah, I missed it, I plan to update the documentation today.

tests/test_testing.py
```diff
@@ -192,3 +197,40 @@ def on_post(self, req, resp):
...
+def test_create_environ_cookies_options_method():
+    environ = testing.create_environ(method='OPTIONS', cookies={'foo': 'bar'})
+
+    assert 'HTTP_COOKIE' not in environ
+
+
+def test_cookies_jar():
+    class Foo:
+        def on_get(self, req, resp):
+            resp.set_cookie('has_permission', 'true')
```
kgriffs, line 221: This is the only test for a single cookie (vs. multiple). I think it would be helpful for posterity to mention that here in a comment so that this particular input variation does not get lost if/when this test is ever modified. Alternatively, a separate test could be added that explicitly tests setting a single cookie in the request.

tests/test_testing.py
```diff
@@ -192,3 +197,40 @@ def on_post(self, req, resp):
...
+                resp.status = falcon.HTTP_200
+            else:
+                resp.status = falcon.HTTP_403
+    app = App()
+    app.add_route('/jars', Foo())
+
+    client = testing.TestClient(app)
+
+    response_one = client.simulate_get('/jars')
+    response_two = client.simulate_post('/jars', cookies=response_one.cookies)
```
kgriffs, line 234: It would be great to surface this pattern in the docs. At a minimum, we should mention in the `Result` docstring that the `cookies` property can be passed directly as the cookies kwarg when simulating subsequent requests.

myuz: Yes, nice catch, I'll update the docstring.

tests/test_testing.py
myuz: I prefer to add a comment to the test, to avoid duplication of code(actually what will be).

myuz:
@kgriffs Originally I've planned to do it only for synchronous version in the PR, for ASGI I'll do in a follow-up PR, sounds good?

kgriffs:
> @kgriffs Originally I've planned to do it only for synchronous version in the PR, for ASGI I'll do in a follow-up PR, sounds good?

Sounds like a plan!

docs/user/faq.rst
```diff
@@ -959,12 +959,41 @@ the tutorial in the docs provides an excellent introduction to
...
+
+
+    def test_cookies(client):
+        resp = client.simulate_get('/cookies', cookies={'cookie': 'cookie value'})
+
+        assert resp.json == {'cookies': {'cookie': 'cookie value'}}
+
+
+Alternatively, you can set the Cookie header directly, as showing in the
+following example.
```
vytas7, line 995: Not a hard requirement, but maybe we don't need to duplicate the whole file again, and just illustrate the difference in `test_cookies(client)`.

falcon/testing/helpers.py
vytas7: Should we then add a `NOTE` explaining this for the source code reader?

kgriffs: IMHO a `NOTE` would be helpful.

docs/user/faq.rst
kgriffs: Perhaps something like:

> Alternatively, you can set the Cookie header directly as demonstrated in this version of `test_cookies()`:

...and then only show the altered method in the code snippet?

PR #1722, opened 2020-05-16
Improve cors middleware.

### LINKED ISSUE #1719: Additional options to the internal cors middleware
The current cors middleware is a bit too limited for use in anything other than the most simple setups.

As discussed on gitter, I think we could add a bit more options, without complicating it too much.
In particular I think that the `App.cors_enable` flag should still add the middleware configured as is, but when manually adding it to the application more options should be exposed.
Options:
- `Access-Control-Allow-Origin`: default to '*', a str or iterable of strings. 
- `Access-Control-Expose-Headers`: default to not set, a str or iterable of strings.
- ` Access-Control-Allow-Credentials`: default to not set, True to always enable, an iterable of strings to match against the `Origin`.

Changes to the processing, ref https://www.w3.org/TR/cors/#resource-processing-model:
- Do nothing if the `Origin` header is not present.
- if  `Access-Control-Allow-Origin` is not '*', check if the `Origin` is in it. If not terminate
- If adding credential, set `Access-Control-Allow-Origin` to the origin
- Add `Access-Control-Expose-Headers` if defined

CaselIT:
Allows customization of the list of allowed origins, the list of
origins that support credential and the additional headers to expose

# Related Issues

Closes #1719 

# Pull Request Checklist

### DIFF
--- a/docs/api/cors.rst
+++ b/docs/api/cors.rst
@@ -16,11 +16,16 @@ By default, Falcon's built-in CORS support is disabled, so that any cross-origin
 requests will be blocked by the browser. Passing ``cors_enable=True`` will
 cause the framework to include the necessary response headers to allow access
 from any origin to any route in the app. Individual responders may override this
-behavior by setting the Access-Control-Allow-Origin header explicitly.
+behavior by setting the ``Access-Control-Allow-Origin`` header explicitly.
 
 Whether or not you implement a CORS policy, we recommend also putting a robust
 AuthN/Z layer in place to authorize individual clients, as needed, to protect
 sensitive resources.
+
+Directly passing the :class:`falcon.CORSMiddleware` middleware to the application
+allows customization of the CORS policy applied. The middleware allows customizing
+the allowed origins, if credentials should be allowed and if additional headers
+that can be exposed.
 
 Usage
 -----
@@ -36,6 +41,10 @@ Usage
             # Enable a simple CORS policy for all responses
             app = falcon.App(cors_enable=True)
 
+            # Enable CORS policy for example.com and allows credentials
+            app = falcon.App(middleware=falcon.CORSMiddleware(
+                allow_origin='example.com', allow_credentials=True))
+
     .. tab:: ASGI
 
         .. code:: python
@@ -44,3 +53,12 @@ Usage
 
             # Enable a simple CORS policy for all responses
             app = falcon.asgi.App(cors_enable=True)
+
+            # Enable CORS policy for example.com and allows credentials
+            app = falcon.asgi.App(middleware=falcon.CORSMiddleware(
+                allow_origin='example.com', allow_credentials=True))
+
+CORSMiddleware
+--------------
+
+.. autoclass:: falcon.CORSMiddleware
--- a/docs/api/cors.rst
+++ b/docs/api/cors.rst
@@ -25,7 +25,7 @@ sensitive resources.
 Directly passing the :class:`falcon.CORSMiddleware` middleware to the application
 allows customization of the CORS policy applied. The middleware allows customizing
 the allowed origins, if credentials should be allowed and if additional headers
-that can be exposed.
+can be exposed.
 
 Usage
 -----
--- a/docs/api/cors.rst
+++ b/docs/api/cors.rst
@@ -43,7 +43,7 @@ Usage
 
             # Enable CORS policy for example.com and allows credentials
             app = falcon.App(middleware=falcon.CORSMiddleware(
-                allow_origin='example.com', allow_credentials=True))
+                allow_origins='example.com', allow_credentials=True))
 
     .. tab:: ASGI
 
@@ -56,7 +56,7 @@ Usage
 
             # Enable CORS policy for example.com and allows credentials
             app = falcon.asgi.App(middleware=falcon.CORSMiddleware(
-                allow_origin='example.com', allow_credentials=True))
+                allow_origins='example.com', allow_credentials=True))
 
 CORSMiddleware
 --------------
--- a/docs/api/cors.rst
+++ b/docs/api/cors.rst
@@ -43,7 +43,7 @@ Usage
 
             # Enable CORS policy for example.com and allows credentials
             app = falcon.App(middleware=falcon.CORSMiddleware(
-                allow_origins='example.com', allow_credentials=True))
+                allow_origins='example.com', allow_credentials='*'))
 
     .. tab:: ASGI
 
@@ -56,7 +56,7 @@ Usage
 
             # Enable CORS policy for example.com and allows credentials
             app = falcon.asgi.App(middleware=falcon.CORSMiddleware(
-                allow_origins='example.com', allow_credentials=True))
+                allow_origins='example.com', allow_credentials='*'))
 
 CORSMiddleware
 --------------
--- a/falcon/app.py
+++ b/falcon/app.py
@@ -153,7 +153,7 @@ class App:
         cors_enable (bool): Set this flag to ``True`` to enable a simple
             CORS policy for all responses, including support for preflighted
             requests. An instance of :py:class:`~.CORSMiddleware` can instead be
-            passet to the middleware argument to customize its behaviour.
+            passed to the middleware argument to customize its behaviour.
             (default ``False``).
             (See also: :ref:`CORS <cors>`)
 
--- a/falcon/app.py
+++ b/falcon/app.py
@@ -152,7 +152,9 @@ class App:
 
         cors_enable (bool): Set this flag to ``True`` to enable a simple
             CORS policy for all responses, including support for preflighted
-            requests (default ``False``).
+            requests. An instance of :py:class:`~.CORSMiddleware` can instead be
+            passet to the middleware argument to customize its behaviour.
+            (default ``False``).
             (See also: :ref:`CORS <cors>`)
 
     Attributes:
--- a/falcon/app.py
+++ b/falcon/app.py
@@ -362,7 +362,7 @@ class App:
                 default_media_type = None
             elif (
                 length is not None and
-                req.method == 'HEAD'and
+                req.method == 'HEAD' and
                 resp_status not in _BODILESS_STATUS_CODES and
                 'content-length' not in resp._headers
             ):
--- a/falcon/asgi/app.py
+++ b/falcon/asgi/app.py
@@ -281,7 +281,7 @@ class App(falcon.app.App):
 
                 if not spec_version.startswith('1.') and not spec_version.startswith('2.'):
                     raise UnsupportedScopeError(
-                        f'Only versions 1.x and 2.x of the ASGI "lifespan" scope are supported.'
+                        'Only versions 1.x and 2.x of the ASGI "lifespan" scope are supported.'
                     )
 
                 await self._call_lifespan_handlers(spec_version, scope, receive, send)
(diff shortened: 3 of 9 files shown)


falcon/middlewares.py
```diff
@@ -1,16 +1,80 @@
...
+    See also:
+
+    * https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+    * https://www.w3.org/TR/cors/#resource-processing-model
+
+    Keyword Arguments:
+        allow_origin (Union[str, Iterable[str]]): List of origins to allow (case
+            sensitive). The string ``'*'`` acts as a wildcard, matching every origin.
+            (default ``'*'``).
+        expose_headers (Optional[Union[str, Iterable[str]]]): List of additional headers to
```
vytas7, line 22: IMHO it might be worth adding a note that we are setting the [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) header, and also mention which headers are [whitelisted by default](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header) (if we're defining as _additional_).

falcon/middlewares.py
```diff
@@ -1,16 +1,86 @@
...
+            expose via the ``Access-Control-Expose-Headers`` header. These headers are in addition
+            to the CORS-safelisted ones: ``Cache-Control``, ``Content-Language``,
+            ``Content-Length``, ``Content-Type``, ``Expires``, ``Last-Modified``, ``Pragma``.
+            (default ``None``).
+
+            See also:
+            https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
+        allow_credentials (Union[bool, Iterable[str]]): List of origins to allow credentials (case
```
vytas7, line 30: It's totally a matter of taste, but I don't think we use this pattern much elsewhere in Falcon... ("Polymorphic" union parameters which can be ``False``, ``True`` and something else.)

Have you considered splitting into explicit parameters, probably something along the lines of how [`falcon-cors` does it](https://falcon-cors.readthedocs.io/en/master/falcon_cors.html)?

falcon/middlewares.py
```diff
@@ -1,16 +1,86 @@
...
+            sensitive). When ``True`` allows credentials for each allowed origin. This has effect
+            only if the origin is allowed by the ``allow_origin`` argument. (default ``False``).
+    """
+    def __init__(
+        self,
+        allow_origin: Union[str, Iterable[str]] = '*',
+        expose_headers: Optional[Union[str, Iterable[str]]] = None,
+        allow_credentials: Union[bool, Iterable[str]] = False
```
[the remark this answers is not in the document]
CaselIT: usually black adds them for me :)

falcon/middlewares.py
CaselIT: I did think about splitting it, but I think it may be more convenient this way. After parsing it has only two states: True or a set of string (empty when this argument is false), so I don't think it overly complicates matters.

Also with two parameters we would need to decide how they interact, who takes precedence and how the bool parameter is used. Does True enable all origins or when used with a list of string the list takes precedence? Does False disable all origins or the list takes precedence? A single parameter provides a cleaner interface in my opinion.

CaselIT: I wanted to update this. I still prefer the single argument, but I can change this to enable two arguments.
@vytas7 do you have suggestions on what interaction should the two arguments have? In falcon-cors the `allow_credentials_all_origins` takes precedence if `True`, else `allow_credentials_origins_list` is used if the `allow_credentials_all_origins` was `False`.

@kgriffs do you have preferences?

kgriffs: The single-argument approach is reminiscent of how things often work in the Node.js community; it can be a bit strange until you get used to it. But I can't really think of an alternative which would be better since it may actually be harder to reason about two params, and it would be more of a burden to pass a dict, or set of tuples, to `allow_origin` that explicitly provides `True`/`False` for each origin. 

So I am leaning toward just keeping it as-is, but perhaps making the docstring a little more clear, for example:

```rst
List of origins for which to allow credentials (case sensitive). Alternatively, a boolean
value may be passed to specify that credentials should be allowed for all allowed 
origins (``True``) or disallowed for all (``False``). This parameter takes effect only 
if the origin is allowed by the ``allow_origin`` argument. (Default ``False``).
```

vytas7: Since this is just some batteries included, and not a part of the core API, I'm fine with it either way.

I'm still a bit uneasy with the values this union parameter accepts though, particularly in the light of `expose_headers` accepting `None` (and it is even the default value).
Regardless of how we document it, this may be confusing: `expose_headers=None` -- none -- no headers to expose -- sounds logical, but passing `None` to `allow_credentials` (allone none -- allow no credentials one may think) would explode while trying to make a `frozenset` out of `None`?

With that in mind, would it be worth making the first check just `not allow_credentials` or more explicitly allow `is None`?

falcon/app.py
```diff
@@ -152,7 +152,9 @@ def process_response(self, req, resp, resource, req_succeeded)
 
         cors_enable (bool): Set this flag to ``True`` to enable a simple
             CORS policy for all responses, including support for preflighted
-            requests (default ``False``).
+            requests. An instance of :py:class:`~.CORSMiddleware` can instead be
+            passet to the middleware argument to customize its behaviour.
```
CaselIT, line 156: passet -> passed

falcon/asgi/app.py
```diff
@@ -212,7 +212,9 @@ async def process_response(self, req, resp, resource, req_succeeded)
 
         cors_enable (bool): Set this flag to ``True`` to enable a simple
             CORS policy for all responses, including support for preflighted
-            requests (default ``False``).
+            requests. An instance of :py:class:`..CORSMiddleware` can instead be
+            passet to the middleware argument to customize its behaviour.
```
CaselIT, line 216: passet -> passed

falcon/middlewares.py
CaselIT: Sure we can make that change. I'll have another look and try to normalize the parameters a bit incorporating the suggestions.

CaselIT:
I've updated the pr:
- improved the docs for allow_credentials
- rename allow_origin parameter to allow_origins, making it plural like the other parameters
- allow_credentials now can accept None (behaves like false) and a str (behaves like a list of one element), similar to the other parameters.

falcon/middlewares.py
```diff
@@ -1,16 +1,92 @@
...
+            (default ``'*'``).
+        expose_headers (Optional[Union[str, Iterable[str]]]): List of additional headers
+            to expose via the ``Access-Control-Expose-Headers`` header.
+            These headers are in addition to the CORS-safelisted ones:
+            ``Cache-Control``, ``Content-Language``, ``Content-Length``, ``Content-Type``,
+            ``Expires``, ``Last-Modified``, ``Pragma``. (default ``None``).
+
+            See also:
+            https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
+        allow_credentials (Optional[Union[bool, str, Iterable[str]]]): List of origins for
```
vytas7, line 30: Looking great now :+1: 

Tbh looking at this again I got another idea to un-union the interface :thinking: 

Could we possibly unify in the following way:
* `allow_origins` (unchanged, maybe we could add support for `None`?) -- list of origins, `*` means allow all
* `expose_headers` (unchanged) -- list of headers, `None` means expose none extra
* `allow_credentials` -- list of origins, `*` means allow all allowed by `allow_origins`, `None` means allow none

Thoughts?

CaselIT: You know, I actually like this proposal more than the current one. 
The only thing we have to make clear is that `allow_credentials=='*'` only works for allowed origins, but that was the same as `True` before.

CaselIT: I've updated the pr applying this suggestion

docs/api/cors.rst
```diff
@@ -16,12 +16,17 @@ By default, Falcon's built-in CORS support is disabled, so that any cross-origin
...
-behavior by setting the Access-Control-Allow-Origin header explicitly.
+behavior by setting the ``Access-Control-Allow-Origin`` header explicitly.
 
 Whether or not you implement a CORS policy, we recommend also putting a robust
 AuthN/Z layer in place to authorize individual clients, as needed, to protect
 sensitive resources.
 
+Directly passing the :class:`falcon.CORSMiddleware` middleware to the application
+allows customization of the CORS policy applied. The middleware allows customizing
+the allowed origins, if credentials should be allowed and if additional headers
```
vytas7, line 27: The sentence structure got a bit broken here (unless I'm reading it totally wrong):
> and if additional headers ~~that~~ can be exposed

falcon/middlewares.py
```diff
@@ -1,16 +1,101 @@
...
+    See also:
+
+    * https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+    * https://www.w3.org/TR/cors/#resource-processing-model
+
+    Keyword Arguments:
+        allow_origins (Union[str, Iterable[str]]): List of origins to allow (case
+            sensitive). The string ``'*'`` acts as a wildcard, matching every origin.
+            (default ``'*'``).
+        expose_headers (Optional[Union[str, Iterable[str]]]): List of additional headers
```
vytas7, line 22: IMHO we could vastly improve clarity by explaining it is response headers, i.e.
> List of additional **response** headers to expose via...

falcon/middlewares.py
```diff
@@ -1,16 +1,101 @@
...
+            to expose via the ``Access-Control-Expose-Headers`` header.
+            These headers are in addition to the CORS-safelisted ones:
+            ``Cache-Control``, ``Content-Language``, ``Content-Length``, ``Content-Type``,
+            ``Expires``, ``Last-Modified``, ``Pragma``. (default ``None``).
+
+            See also:
+            https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
+        allow_credentials (Optional[Union[str, Iterable[str]]]): List of origins (case sensitive)
+            for which to allow credentials via the ``Access-Control-Allow-Credentials`` header.
+            The string ``'*'`` acts as a wildcard, matching every all allowed origins. Set to
```
vytas7, line 32: > matching every all allowed origins

Something got mixed up here, :arrow_right: _matching every ~~all~~ allowed origin_ or _matching ~~every~~ all allowed origins_

falcon/middlewares.py
```diff
@@ -1,16 +1,101 @@
...
+            ``None`` to disallowed all origin. This parameter takes effect only
```
vytas7, line 33: To disallow~~ed~~ (sic)

vytas7, line 33: to disallow~~ed~~ (sic) all origin**s**

falcon/middlewares.py
CaselIT: must have forgot to re-read that sentence. I've gone with every

CaselIT:
Hopefully I've not added other typos. Need to go towards the 🛌

vytas7:
As someone also situated the CEST timezone, I have to agree wrt heading towards the :sleeping_bed: Same here.

falcon/middlewares.py
```diff
@@ -1,16 +1,101 @@
...
+        allow_credentials: Optional[Union[str, Iterable[str]]] = None,
+    ):
+        if allow_origins == '*':
+            self.allow_origins = allow_origins
+        else:
+            if isinstance(allow_origins, str):
+                allow_origins = [allow_origins]
+            self.allow_origins = frozenset(allow_origins)
+            if '*' in self.allow_origins:
+                raise ValueError(
```
kgriffs, line 50: /me likes this check

kgriffs:
Nice work!

PR #1728, opened 2020-06-08
feat(ASGI): async multipart/form-data media handler

### LINKED ISSUE #1674: ASGI+multipart/form-data
Add ASGI support to [MultipartFormHandler](https://falcon.readthedocs.io/en/latest/api/media.html#falcon.media.MultipartFormHandler).

The parsing interface could look similar to the [WSGI flavour](https://falcon.readthedocs.io/en/latest/api/multipart.html), just async:
```python
form = await req.get_media()

async for part in form:
    async with aiofiles.open('mydest', 'wb') as dest:
        await part.stream.pipe(dest)
```

There's an open question what to do with properties that would become async, such as [`part.text`](https://falcon.readthedocs.io/en/latest/api/multipart.html#falcon.media.multipart.BodyPart.text). We may need to refactor them to `async def BodyPart.get_text()` etc. Suggestions welcome!

vytas7:
Closes #1674

Interface is largely the same as for the WSGI counterpart, with the following changes (made them for WSGI too):
* [`BodyPart.data`](https://falcon.readthedocs.io/en/latest/api/multipart.html#falcon.media.multipart.BodyPart.data) becomes (primarily) `BodyPart.get_data()`, but the property exists as a convenience alias
* [`BodyPart.media`](https://falcon.readthedocs.io/en/latest/api/multipart.html#falcon.media.multipart.BodyPart.media) becomes (primarily) `BodyPart.get_media()`, but the property exists as a convenience alias
* [`BodyPart.text`](https://falcon.readthedocs.io/en/latest/api/multipart.html#falcon.media.multipart.BodyPart.text) becomes (primarily) `BodyPart.get_text()`, but the property exists as a convenience alias

Furthermore, the ASGI `BodyPart.stream` variant can be iterated in the same fashion as [`Request.stream`](https://falcon.readthedocs.io/en/latest/api/request_and_response_asgi.html#falcon.asgi.Request.stream), and it is the most efficient way to consume the body part content.

Usage example: https://gist.github.com/vytas7/657915918327a7b2fe7a1286b7553512

Work in progress:
- [x] Better support for small wrapped `read()`s without reallocationg the buffer, as it is done for WSGI
- [x] Add a "real" multipart test with `uvicorn`
- [x] Create a new Towncrier or just update `docs/_newsfragments/953.newandimproved.rst`?
- [x] "Iteration started" errors as in [`Request.stream`](https://falcon.readthedocs.io/en/latest/api/request_and_response_asgi.html#falcon.asgi.Request.stream)?
- [x] One more round of docs cleanup/improvements
- [x] Investigate why, compared to `master`, coverage disappears for the `Request.stream` `http.disconnect` branch, even though this Draft PR is (only?) adding lots of code


falcon/testing/helpers.py
```diff
@@ -316,6 +316,33 @@ def invoke_coroutine_sync(coroutine, *args, **kwargs):
     )
 
 
+def runs_sync(coroutine):
```
kgriffs, line 319: Should this reside in `falcon/util/sync.py` instead?

tests/asgi/_asgi_test_app.py
```diff
@@ -111,6 +112,24 @@ async def emit():
         resp.sse = emit()
 
 
+class Multipart:
+    async def on_post(self, req, resp):
+        parts = {}
+
+        form = await req.get_media()
+        async for part in form:
+            sha1 = hashlib.sha1()
```
CaselIT, line 121: Another nitpick. Probably we could use a not discouraged hashing algorithm? Like sha256/sha3_256 (not sure if sha3 is in py35)

tests/asgi/test_buffered_reader.py
```diff
@@ -0,0 +1,387 @@
...
+    b'World!',
+    b'\n',
+    b'Jus',
+    b't tes',
+    b'ting some iterato',
+    b'r goodne',
+    b'',
+    b'ss.',
+    b'',
+    b'\n',
```
CaselIT, line 56: Aren't we missing non printable / non achi chars tests here?

tests/asgi/_asgi_test_app.py
CaselIT: sha3 is py3.6+ only. sha256 is probably ok

vytas7: We are not claiming any cryptographic context here though.
SHA1 has excellent collision properties for other purposes. For instance, Git [_treeish_]( https://git-scm.com/docs/gitglossary#def_SHA1) (commits, and basically any type of reference in Git) is also SHA1, and I don't think that's an issue.

tests/asgi/test_buffered_reader.py
vytas7: We are.
But, OTOH, multipart tests include these, including parsing completely random binary forms. Both in "unit" tests, and against a real `uvicorn` server.

tests/asgi/_asgi_test_app.py
CaselIT: Git is old though :). And they are in the process of migrating to a newer / better hash (to sha256 in fact I think).

vytas7: I still don't find think SHA1 is an issue in this example, but maybe I could add a comment stressing that this has nothing to do with cryptographic hashing.

For instance, storage systems use even simpler hashes such CRC, Murmur, MD5 etc for integrity checking.
Hash tables also use performant hashes to spread keys, including Python's `dict` and the `hash()` machinery.

SHA1 is by no means a _discouraged algorithm_. It is discouraged for cryptographic purposes.

docs/api/multipart.rst
```diff
@@ -11,25 +11,52 @@ Falcon features easy and efficient access to submitted multipart forms by using
...
+.. tabs::
+
+    .. group-tab:: WSGI
+
+        .. code:: python
+
+            form = req.get_media()
+            for part in form:
+                if part.content_type == 'application/json':
+                    # TODO: Body part is a JSON document, do something useful with it
```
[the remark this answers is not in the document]
vytas7: I see your point, but OTOH this pattern is used throughout the Falcon docs, presenting snippets from the user perspective where `TODO` reminds users to actually flesh out snippets that they pasted from the docs into their own code. See, for instance also in [How can I pass data from a hook to a responder, and between hooks?](https://falcon.readthedocs.io/en/latest/user/faq.html#how-can-i-pass-data-from-a-hook-to-a-responder-and-between-hooks) and many other places throughout the docs.

In that light, I'm inclined to leave it as-is, unless you vehemently object.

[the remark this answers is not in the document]
CaselIT: I was comparing it with the other lines. 
Like `# Do something with the uploaded data (file)` and `# Store this body part in a file.`
Maybe they could use a `TODO:` instead to uniform things?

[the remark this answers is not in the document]
vytas7: Aha, that's a good point. I think I was prefixing with `TODO` where it is just a placeholder, and no `TODO` where we pretend to do something, like calling an imaginary `feed_data()` etc.

falcon/testing/helpers.py
vytas7: Good question. It is a thin wrapper around `invoke_coroutine_sync`, and `invoke_coroutine_sync` is actually defined here. Maybe we should move both then?

docs/_newsfragments/1674.newandimproved.rst
```diff
@@ -0,0 +1,3 @@
+:class:`~falcon.media.MultipartFormHandler` was extended for asynchronous
```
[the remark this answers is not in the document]
vytas7: No, it's just a typical time warp... We ought to release more often.

falcon/errors.py
```diff
@@ -64,6 +64,17 @@ class OperationNotAllowed(ValueError):
     """The requested operation is not allowed."""
 
 
+# TODO(vytas): Discuss the DelimiterError base class in comparison to
```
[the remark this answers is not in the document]
vytas7: Tbh I don't remember what exactly I meant. This was a note to self :thinking:
Maybe I was wondering how base classes compared (`ValueError` vs `IOError`); but your comment answers that.
In any case, we have now discussed it, no longer needed _to do_ this. I'll simply remove the note.

falcon/errors.py
```diff
@@ -64,6 +64,17 @@ class OperationNotAllowed(ValueError):
...
+#   OperationNotAllowed.
+
+# TODO(vytas): Clean up the import hierarchy so that DelimiterError is defined
```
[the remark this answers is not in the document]
vytas7: I'll see what I can do about it. The import hierarchy got a bit screwed up here between parallel tracks for WSGI Python, Cython and ASGI :grimacing:

falcon/media/multipart.py
```diff
@@ -152,8 +208,24 @@ def __init__(self, stream, headers, parse_options):
         self._headers = headers
         self._parse_options = parse_options
 
-    @property
-    def data(self):
+    def get_data(self):
+        """Returns the body part content bytes.
+
+        The maximum number of bytes that could be read is configurable via
+        :class:`MultipartParseOptions`, and a :class:`.MultipartParseError` is
```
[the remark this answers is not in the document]
vytas7: Added a sentence reasoning about the limit.

docs/api/multipart.rst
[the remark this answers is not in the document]
vytas7: Cleaned up now.

falcon/testing/helpers.py
kgriffs: After thinking about this some more, I do like the idea of just colocating all the sync/async adapters in util/sync, even if some of them are really only intended for use in tests. And we may be able to make some of these adapters more efficient later on using greenletio.

falcon/errors.py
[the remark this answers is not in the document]
vytas7: :cold_sweat: I've attempted to factor out deprecation utilities into a separate module until `util`, and that seems to work in terms of breaking circular module init dependencies.

falcon/testing/helpers.py
vytas7: That makes sense. I'll move both into `falcon/util/sync.py`.

vytas7: Sed and done.

PR #1729, opened 2020-06-10
Work around sphinx 3.1.0 bug causing parallel build to fail on Linux

kgriffs:


vytas7:
FWIW, this seems to be flagged as a high-priority bug on the Sphinx side, so hopefully we can restore parallel builds in the future.

Ref: https://github.com/sphinx-doc/sphinx/issues/7802

PR #1739, opened 2020-07-07
docs(recipes): discuss prettifying JSON responses

vytas7:
We've been debating this with @CaselIT on Gitter, so I decided to just go ahead and package my answers into a recipe.

### DIFF
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -92,6 +92,8 @@ media type that JSON also supports (i.e. dicts, lists, etc).
 If JSON Schema does not meet your needs, a custom validator may be
 implemented in a similar manner to the one above.
 
+.. _content-type-negotiaton:
+
 Content-Type Negotiation
 ------------------------
 
@@ -184,6 +186,8 @@ Supported Handler Types
 
 .. autoclass:: falcon.media.URLEncodedFormHandler
     :no-members:
+
+.. _custom-media-handler-type:
 
 Custom Handler Type
 -------------------
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -200,6 +200,15 @@ provided by Falcon:
     :members:
     :member-order: bysource
 
+.. tip::
+    In order to use your custom media handler in a :ref:`Falcon app <app>`,
+    you'll have to add an instance of your class to the app's media handlers
+    (specified in :attr:`RequestOptions <falcon.RequestOptions.media_handlers>`
+    and :attr:`ResponseOptions<falcon.ResponseOptions.media_handlers>`,
+    respectively).
+
+    See also: :ref:`custom_media_handlers`.
+
 
 Handlers Mapping
 ----------------
--- a/docs/user/recipes/pretty-json.rst
+++ b/docs/user/recipes/pretty-json.rst
@@ -1,0 +1,95 @@
+Prettifying JSON Responses
+==========================
+
+To make JSON responses even more human readable, it might be desirable to
+prettify the output indendation. By default, Falcon's :class:`JSONHandler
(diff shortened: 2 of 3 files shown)


docs/user/recipes/pretty-json.rst
```diff
@@ -0,0 +1,95 @@
...
+                except ValueError:
+                    # TODO: Handle invalid params?
+                    indent = None
+
+            result = json.dumps(media, indent=indent, sort_keys=bool(indent))
+            return result.encode()
+
+Furthermore, we'll need to implement content-type negotiation to accept the
+indented JSON content type for response serialization. The bare minimum
+example using middleware is described here: :ref:`content-type-negotiaton`.
```
CaselIT, line 73: I think that the `content-type-negotiaton` section could benefit from an example, similar to the one provided here https://falcon.readthedocs.io/en/stable/api/media.html#replacing-the-default-handlers

```py
extra_handlers = {
    'application/my-media-format: MyMediaFormatHandler()
}

app = falcon.App()

app.req_options.media_handlers.update(extra_handlers)
app.resp_options.media_handlers.update(extra_handlers)
```

vytas7: It does have [an example](https://falcon.readthedocs.io/en/stable/api/media.html#content-type-negotiation) though? :thinking: 
Or do you suggest copying that example into this recipe too? (I actually was thinking of that, but in the end just left the link).

CaselIT: Sorry, I must have misplaced my comment. I actually meant the `custom-media-handler-type` ones.
I'll move it there

docs/api/media.rst
```diff
@@ -185,6 +187,8 @@ Supported Handler Types
 .. autoclass:: falcon.media.URLEncodedFormHandler
     :no-members:
 
+.. _custom-media-handler-type:
+
```
CaselIT, line 191: This was the correct place. The docs for the `custom-media-handler-type` could probably use an example similar to https://falcon.readthedocs.io/en/stable/api/media.html#replacing-the-default-handlers or a link to them.
An example from the previous comment:
```py
extra_handlers = {
    'application/my-media-format: MyMediaFormatHandler()
}

app = falcon.App()

app.req_options.media_handlers.update(extra_handlers)
app.resp_options.media_handlers.update(extra_handlers)
```

Probably a link is better since we avoid repetition and maybe we can also mention it in the examples there that custom media handlers can be added that way.

It's not really the focus of this pr, but I was reading it while reviewing it and I think it could be improved by this change. If an user is linked to it directly without having read the previous part, it does not know how to use the custom handler.

vytas7: Ah, that's a fair point.
I'll link to [Replacing the Default Handlers](https://falcon.readthedocs.io/en/latest/api/media.html#replacing-the-default-handlers) both from this tutorial (i.e., explaining what to actually do with that `json_handler`) and from the custom handler type.

vytas7: Done.

docs/user/recipes/pretty-json.rst
```diff
@@ -0,0 +1,100 @@
...
+        def deserialize(self, stream, content_type, content_length):
+            data = stream.read()
+            return json.loads(data.decode())
+
+        def serialize(self, media, content_type):
+            _, params = cgi.parse_header(content_type)
+            indent = params.get('indent')
+            if indent is not None:
+                try:
+                    indent = int(indent)
```
kgriffs, line 68: It may be nice to include a check here for very large indentation values that should be disallowed to help guard against a DoS.

vytas7: Nice catch; improved.

Issue #321, opened 2014-10-02
ASGI+WebSocket


kerwin:
Is there a way to support websocket on falcon?

joranbeasley:
I would also like to see something like FlaskWebsockets  for falcon ...

kgriffs:
We've been toying with the idea of expanding the framework to support interfaces other than WSGI, such as twisted.web and aiohttp. If we start going down that path, WebSocket support in Falcon may make sense after all.

s-trooper:
just as info, it seems django going to solve wsgi related problems this way: http://channels.readthedocs.org/en/latest/asgi.html

kgriffs:
Related: #582

ulgens:
Any progress?

jrgp:
I'm interested in using Websockets with Falcon as well.

pushshift:
+1 For any type of websocket functionality.  Would be great if we could find something workable.

jrgp:
Definitely. I have a use case for websockets + falcon I'd implement if support was there. I haven't gone the node route as I don't want to maintain two codebases for the same project.

kgriffs:
Hi everyone, I've started work on #1055 and so far [uvicorn](http://www.uvicorn.org) looks like a very promising (no pun intended!) avenue for adding WebSocket support to Falcon. I'd like to gather a few concrete use cases to help inform the design. If you can, please provide an overview of how you are using--or plan to use--WebSocket connections, including subprotocols (either IANA-registered or bespoke). Thanks!

pushshift:
@kgriffs,  

Awesome -- thanks!  Quick question for you.  I understand from earlier in the conversation that you said Websockets uses the upgrade header in order for it to work correctly.  My question is whether SSE (Server Side Events) support within Falcon is doable currently?  My understanding is that SSE is more of a one-way channel from the server to the browser.

Does Falcon currently support SSE and if not, would implementing Websockets also implement SSE support?

Thanks again!

kgriffs:
@pushshift The WSGI spec isn't really designed for things like SSE, but it should be possible with uvicorn/ASGI (see also: http://www.uvicorn.org/#streaming-responses). This would probably be an extension of #1055 since it is based on regular HTTP vs. WebSocket.

kgriffs:
(We'll probably want to label this as "experimental" in the first iteration since there will likely be feedback and perhaps immature dependencies.)

Camilo-HG:
Hey, hi! 

[...]

PR #1741, opened 2020-07-14
feat(asgi): support the ASGI WebSocket connection scope

kgriffs:
# Summary of Changes

This patch adds support for the ASGI WebSocket connection scope. The primary code paths have been tested via uvicorn/daphne and the feature is considered alpha quality until full test coverage and documentation are added. Note also that there is currently no support in the testing module for simulating WebSocket connections.

# Related Issues

#321

# Pull Request Checklist

### DIFF
--- a/docs/api/websocket.rst
+++ b/docs/api/websocket.rst
@@ -1,0 +1,435 @@
+.. _ws:
+
+WebSocket (ASGI Only)
+=====================
+
+.. contents:: :local:
+
+Falcon builds upon the
+`ASGI WebSocket Specification <https://asgi.readthedocs.io/en/latest/specs/www.html#websocket>`_
+to provide a simple, no-nonsense WebSocket server implementation.
+
+With support for both
+`WebSocket <https://tools.ietf.org/html/rfc6455>`_ and
+`Server-Sent Events <https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events>`_ (SSE), Falcon facilitates real-time, event-oriented communication between an ASGI
+application and a web browser, mobile app, or other client application.
+
+.. note::
+
+    See also :attr:`falcon.asgi.Response.sse` to learn more about Falcon's
+    Server-Sent Event (SSE) support,
+
+Usage
+-----
+
+With Falcon you can easily add WebSocket support to any route in your ASGI
+app, simply by implementing an ``on_websocket()`` responder in the
+resource class for that route. As with regular
+HTTP requests, WebSocket flows can be augmented with middleware
+components and media handlers.
+
+When a WebSocket handshake arrives (via a standard HTTP request), Falcon will
+first route it as usual to a specific resource class instance. Along the way,
+the following middleware methods will be invoked, if implemented on any
+middleware objects configured for the app:
+
+.. code:: python
+
+    class SomeMiddleware:
+        async def process_request_ws(self, req, ws):
+            """Process a WebSocket handshake request before routing it.
+
+            Note:
+                Because Falcon routes each request based on req.path, a
+                request can be effectively re-routed by setting that
+                attribute to a new value from within process_request().
+
+            Args:
+                req: Request object that will eventually be
+                    passed into an on_websocket() responder method.
+                ws: The WebSocket object that will be passed into
+                    on_websocket() after routing.
+            """
+
+        async def process_resource_ws(self, req, ws, resource, params):
+            """Process a WebSocket handshake request after routing.
+
+            Note:
+                This method is only called when the request matches
+                a route to a resource.
+
+            Args:
+                req: Request object that will be passed to the
+                    routed responder.
+                ws: WebSocket object that will be passed to the
+                    routed responder.
+                resource: Resource object to which the request was
+                    routed.
+                params: A dict-like object representing any additional
+                    params derived from the route's URI template fields,
+                    that will be passed to the resource's responder
+                    method as keyword arguments.
+            """
+
+If a route is found for the requested path, the framework will then check for
+a responder coroutine named ``on_websocket()`` on the target resource. If the
+responder is found, it is invoked in a similar manner to a regular
+``on_get()`` responder, except that a :class:`falcon.asgi.WebSocket` object
+is passed in, instead of an object of type :class:`falcon.asgi.Response`.
+
+For example, given a route that includes an ``account_id`` path parameter, the
+framework would expect an ``on_websocket()`` responder similar to this:
+
+.. code:: python
+
+    async def on_websocket(self, req: Request, ws: WebSocket, account_id: str):
+        pass
+
+If no route matches the path requested in the WebSocket handshake, control then
+passes to a default responder that simply raises an instance of
+:class:`~.HTTPRouteNotFound`. By default, this error will be rendered as a 403
+response with a 3404 close code. This behavior can be modified by adding a
+custom error handler (see also: :meth:`~falcon.asgi.App.add_error_handler`).
+
+Similarly, if a route exists but the target resource does not implement an
+``on_websocket()`` responder, the framework invokes a default responder that
+raises an instance of :class:`~.HTTPMethodNotAllowed`. This class will be
+rendered by default as a 403 response with a 3405 close code.
+
+.. _ws_media_handlers:
+
+Media Handlers
+--------------
+
+By default, :meth:`~.falcon.asgi.WebSocket.send_media` and
+:meth:`~.falcon.asgi.WebSocket.receive_media` will serialize to (and
+deserialize from) JSON for a TEXT payload, and to/from MessagePack
+for a BINARY payload (see also: :ref:`bimh`).
+
+.. note::
+
+    In order to use the default MessagePack handler, the extra ``msgpack``
+    package (version 0.5.2 or higher) must be installed in addition
+    to ``falcon`` from PyPI:
+
+    .. code::
+
+        $ pip install msgpack
+
+WebSocket media handling can be customized by using
+:attr:`falcon.asgi.App.ws_options` to specify an alternative handler for
+one or both payload types, as in the following example.
+
+.. code:: python
+
+    # Let's say we want to use a faster JSON library. You could also use this
+    #   pattern to add serialization support for custom types that aren't
+    #   normally JSON-serializable out of the box.
+    class RapidJSONHandler(falcon.media.TextBaseHandlerWS):
+        def serialize(self, media: object) -> str:
+            return rapidjson.dumps(media, ensure_ascii=False)
+
+        # The raw TEXT payload will be passed as a Unicode string
+        def deserialize(self, payload: str) -> object:
+            return rapidjson.loads(payload)
+
+
+    # And/or for binary mode we want to use CBOR:
+    class CBORHandler(media.BinaryBaseHandlerWS):
+        def serialize(self, media: object) -> bytes:
+            return cbor2.dumps(media)
+
+        # The raw BINARY payload will be passed as a byte string
+        def deserialize(self, payload: bytes) -> object:
+            return cbor2.loads(payload)
+
+    app = falcon.asgi.App()
+
+    # Expected to (de)serialize from/to str
+    json_handler = RapidJSONHandler()
+    app.ws_options.media_handlers[falcon.WebSocketPayloadType.TEXT] = json_handler
+
+    # Expected to (de)serialize from/to bytes, bytearray, or memoryview
+    cbor_handler = ProtocolBuffersHandler()
+    app.ws_options.media_handlers[falcon.WebSocketPayloadType.BINARY] = cbor_handler
+
+The ``falcon`` module defines the following :py:class:`~enum.Enum` values for
+specifying the WebSocket payload type:
+
+.. code:: python
+
+    falcon.WebSocketPayloadType.TEXT
+    falcon.WebSocketPayloadType.BINARY
+
+Extended Example
+----------------
+
+Here is a more comprehensive (albeit rather contrived) example that illustrates
+some of the different ways an application can interact with a WebSocket
+connection. This example also introduces some common WebSocket errors raised
+by the framework.
+
+.. code:: python
+
+    import falcon.asgi
+    import falcon.media
+
+
+    class SomeResource:
+
+        # Get a paginated list of events via a regular HTTP request.
+        #
+        #   For small-scale, all-in-one apps, it may make sense to support
+        #   both a regular HTTP interface and one based on WebSocket
+        #   side-by-side in the same deployment. However, these two
+        #   interaction models have very different performance characteristics,
+        #   and so larger scale-out deployments may wish to specifically
+        #   designate instance groups for one type of traffic vs. the
+        #   other (although the actual applications may still be capable
+        #   of handling both modes).
+        #
+        async def on_get(self, req: Request, account_id: str):
+            pass
+
+        # Push event stream to client. Note that the framework will pass
+        #   parameters defined in the URI template as with HTTP method
+        #   responders.
+        async def on_websocket(self, req: Request, ws: WebSocket, account_id: str):
+
+            # The HTTP request used to initiate the WebSocket handshake can be
+            #   examined as needed.
+            some_header_value = req.get_header('Some-Header')
+
+            # Reject it?
+            if some_condition:
+                # If close() is called before accept() the code kwarg is
+                #   ignored, if present, and the server returns a 403
+                #   HTTP response without upgrading the connection.
+                await ws.close()
+                return
+
+            # Examine subprotocols advertised by the client. Here let's just
+            #   assume we only support wamp, so if the client doesn't advertise
+            #   it we reject the connection.
+            if 'wamp' not in ws.subprotocols:
+                # If close() is not called explicitly, the framework will
+                #   take care of it automatically with the default code (1000).
+                return
+
+            # If, after examining the connection info, you would like to accept
+            #   it, simply call accept() as follows:
+            try:
+                await ws.accept(subprotocol='wamp')
+            except WebSocketDisconnected:
+                return
+
+            # Simply start sending messages to the client if this is an event
+            #   feed endpoint.
+            while True:
+                try:
+                    event = await my_next_event()
+
+                    # Send an instance of str as a WebSocket TEXT (0x01) payload
+                    await ws.send_text(event)
+
+                    # Send an instance of bytes, bytearray, or memoryview as a
+                    #   WebSocket BINARY (0x02) payload.
+                    await ws.send_data(event)
+
+                    # Or if you want it to be serialized to JSON (by default; can
+                    #   be customized via app.ws_options.media_handlers):
+                    await ws.send_media(event)  # Defaults to WebSocketPayloadType.TEXT
+                except WebSocketDisconnected:
+                    # Do any necessary cleanup, then bail out
+                    return
+
+            # ...or loop like this to implement a simple request-response protocol
+            while True:
+                try:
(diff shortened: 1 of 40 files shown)


tests/asgi/test_asgi_servers.py
```diff
@@ -93,6 +96,246 @@ def test_sse(self, server_base_url):
...
+            async with websockets.connect(
+                server_url_events_ws,
+                extra_headers=extra_headers,
+            ) as ws:
+                got_message = False
+
+                while True:
+                    try:
+                        # TODO: Why is this failing to decode on the other side? (raises an error)
+                        # TODO: Why does this cause Daphne to hang?
```
vytas7, line 124: FWIW, I was testing this (`websockets`) module with our uWSGI-based WebSocket (which, FWIW, works flawlessly on all major browsers), and it was hanging in a strange way, particularly if receiving a fair amount of data without sending.

falcon/asgi/ws.py
```diff
@@ -0,0 +1,340 @@
...
+            )
+
+        return self._mh_bin_deserialize(data)
+
+    async def _send(self, msg: dict):
+        try:
+            await self._asgi_send(msg)
+        except Exception as ex:
+            # NOTE(kgriffs): likely uvicorn (or some other project using the
+            #   the the "websockets" Python library. Daphne, at least, doesn
```
vytas7, line 288: A generous amount of `the`s in a row :slightly_smiling_face:

tests/asgi/test_asgi_servers.py
```diff
@@ -93,6 +96,246 @@ def test_sse(self, server_base_url):
...
+
+    def test_missing_responder(self, server_url_events_ws):
+        server_url_events_ws += '/404'
+
+        async def do_test():
+            try:
+                async with websockets.connect(server_url_events_ws):
+                    pass
+
+                pytest.fail('no error raised')
```
vytas7, line 177: Could it be worth adding a `NOTE` why we aren't just using [`pytest.raises`](https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions)?

kgriffs: 👍  Refactored all but one instance of this pattern, and added a note to the remaining instance.

falcon/asgi/ws.py
kgriffs: reworded

kgriffs:
This should be test complete now...just need to finish up the docs!

CaselIT (changes requested):
Looking great. I've left some mostly nitpicks.

Also, http 3 is already in use. Should we support it?

CaselIT:
Regarding event vs future, I've tried to do a test to see if there is actually some difference. The test (that is quite pointless)
```py
import asyncio
import random
import time

async def with_event(num):
    async def wait():
        await evt.wait()

    async def setter():
        for i in range(10):
            pass
        evt.set()

    for i in range(num):
        evt = asyncio.Event()
        await asyncio.gather(wait(), setter())

async def with_event2(num):
    async def wait():
        await evt.wait()

    async def setter():
        for i in range(10):
            pass
        evt.set()

    evt = asyncio.Event()
    for i in range(num):
        await asyncio.gather(wait(), setter())
        evt.clear()

async def with_future(num):
    async def wait():
        await fut

    async def setter():
        for i in range(10):
            pass
        fut.set_result(None)

    for i in range(num):
        fut = asyncio.Future()
        await asyncio.gather(wait(), setter())

count = 100_000

operations = [with_event, with_event2, with_future]
random.shuffle(operations)
for op in operations:
    start = time.perf_counter()
    asyncio.run(op(count))
    print(op.__name__, time.perf_counter() - start)
```
On windows with python 3.8.5 I get there results
```
with_event 4.5359089
with_future 4.216959699999999
with_event2 4.491865800000001
```
so there is a bit of difference, but it's not very large ~6%

CaselIT (commented):
A couple of typos on the tests. looking good otherwise

falcon/asgi/ws.py
```diff
@@ -0,0 +1,562 @@
...
+
+        # Internal Error
+        #
+        #   See also: https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
+        #
+        self.error_close_code = WSCloseCode.SERVER_ERROR
+
+        # NOTE(kgriffs): The websockets library itself will buffer, so we keep
+        #   this value fairly small to reduce memory usage by default. On the
+        #   other hand, AFAICT, autobahn-based servers such as Daphne do not
```
vytas7, line 441: It might be just me, but I'm not sure I'm following the reasoning here for the Daphne part.
Why does it help to increase the value on the Falcon side? Wouldn't it add even more buffer bloat to the system?
Or does it make sense by accepting Daphne's reality in order to avoid queue full errors when Daphne is trying to spill a huge buffer onto our app? (Although it seems we are instead deferring work to a future?)

vytas7 (changes requested):
This is looking really good :100:

I would need to do a more hands-on testing to get a better feeling of the whole flow, particularly given that the ASGI spec, itself, lacks clarity wrt some error handling aspects and edge cases.

falcon/asgi/app.py
```diff
@@ -777,36 +890,66 @@ def _prepare_middleware(self, middleware=None, independent_middleware=False):
...
 
-    async def _python_error_handler(self, req, resp, error, params):
-        falcon._logger.error('Unhandled exception in ASGI app', exc_info=error)
-        self._compose_error_response(req, resp, falcon.HTTPInternalServerError())
+        if ws:
+            await self._ws_cleanup_on_error(ws)
 
-    async def _handle_exception(self, req, resp, ex, params):
+    async def _ws_disconnected_error_handler(self, req, resp, error, params, ws):
+        falcon._logger.warning(f'[FALCON] WebSocket client disconnected with code {error.code}')
```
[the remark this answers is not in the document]
CaselIT: The main advantage of using `%s` is that the formatting is only done if it's actually needed. But I agree that in this case it would not change much.

falcon/asgi/app.py
```diff
@@ -293,6 +309,30 @@ async def __call__(self, scope, receive, send):  # noqa: C901
...
+                    raise UnsupportedScopeError(
+                        'Only versions 2.x of the ASGI "websocket" scope are supported.'
+                    )
+
+                try:
+                    http_version = scope['http_version']
+                except KeyError:
+                    http_version = '1.1'
+
+                if http_version not in {'1.1', '2'}:
```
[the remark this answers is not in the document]
kgriffs: It's not explicitly allowed in the ASGI spec, but I suppose it wouldn't hurt to include `'3'` here.

falcon/asgi/app.py
```diff
@@ -767,7 +820,67 @@ async def _call_lifespan_handlers(self, ver, scope, receive, send):
...
+            # None when a middleware method already set
+            # resp.complete to True.
+            if resource:
+                for process_resource_ws in resource_mw:
+                    await process_resource_ws(req, resource, params)
+
+            try:
+                await on_websocket(req, web_socket, **params)
+            except HTTPNotFound:
+                falcon._logger.warning(
```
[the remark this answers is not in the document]
kgriffs: Good catch. I have removed this try...except since it essentially duplicates ``_python_error_handler()``, and have switched to using ``debug()`` with `%` based formatting.

falcon/asgi/app.py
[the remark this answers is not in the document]
kgriffs: This will probably be a common occurrence, esp. if the app developer is lazy and does not catch ``WebSocketDisconnected`` within the responder. With that in mind, I switched to `debug()` (to avoid spamming the log) and `%` formatting (even though perf isn't as critical here, might as well).

falcon/asgi/app.py
```diff
@@ -819,3 +962,21 @@ async def _handle_exception(self, req, resp, ex, params):
...
+            # NOTE(kgriffs): This can be raised by Daphne. We also
+            #   may raise it ourselves for errors codes < 1000, but in that
+            #   case we just include this string in the exception message
+            #   to make it easier to verify test coverage of the following.
+            if 'invalid close code' in str(ex).lower():
+                await ws.close(_FALLBACK_WS_ERROR_CODE)
+            else:
+                falcon._logger.warning(
+                    '[FALCON] Attempt to close web connection cleanly failed '
+                    f'due to raised error: {traceback.format_exc()}'
```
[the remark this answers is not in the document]
kgriffs: Good catch. Fixed.

falcon/media/json.py
```diff
@@ -110,3 +110,72 @@ async def serialize_async(self, media, content_type):
...
+        from falcon import media
+
+        import rapidjson
+
+        json_handler = media.JSONHandlerWS(
+            dumps=rapidjson.dumps,
+            loads=rapidjson.loads,
+        )
+
+        app = falcon.App()
```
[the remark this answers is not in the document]
kgriffs: fixed

tests/asgi/test_asgi_servers.py
```diff
@@ -135,6 +138,230 @@ def test_sse_client_disconnects_early(self, server_base_url):
...
+    async def test_close_code_on_unhandled_http_error(self, server_url_events_ws):
+        extra_headers = {'x-raise-error': 'http'}
+
+        async with websockets.connect(server_url_events_ws, extra_headers=extra_headers) as ws:
+            await ws.wait_closed()
+
+        assert ws.close_code == 3400
+
+    # @pytest.mark.parametrize('mismatch', ['send', 'recv'])
+    # @pytest.mark.parametrize('mismatch_type', ['text', 'data'])
```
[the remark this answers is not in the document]
kgriffs: 🤦 I forgot to put this back after some debugging.

vytas7:
I left [this FAQ item](https://falcon.readthedocs.io/en/latest/user/faq.html#does-falcon-support-websocket) out of https://github.com/falconry/falcon/pull/1782 in favour of this PR, but we could tackle it on that one too :slightly_smiling_face:

falcon/testing/helpers.py
```diff
@@ -331,6 +334,318 @@ async def collect(self, event: Dict[str, Any]):
...
+    #   in a future code revision. It also makes it more consistent
+    #   with the other methods.
+    async def close(self, code: Optional[int] = None):
+        # NOTE(kgriffs): Give our collector a chance in case the
+        #   server is trying to close at the same time (e.g., there was an
+        #   unhandled error and the server wants to disconnect with an error
+        #   code.) We await a few times to let the server app settle across
+        #   multiple of its own await's.
+        for __ in range(3):
+            await asyncio.sleep(0)
```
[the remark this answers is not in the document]
CaselIT: yes

CaselIT (commented):
There are some comments left in the test_ws file.

Other than it looks ok!

falcon/asgi/ws.py
kgriffs: Actually, you are right. I'm not sure what I was thinking when I wrote this. Even if we block on our side, the WebSocket library will continue to build up its own buffer, so back pressure is not really something we can affect. Supposing the app sporadically slows down its consumption and the Daphne buffer builds up, when the app speeds up, having a larger Falcon buffer could potentially reduce the number of pauses in the future running the pump as it absorbs the backlog. But its hard to guess if such pauses would materially impact performance, and even so it would require comprehensive workload/app-specific benchmarking.

falcon/asgi/ws.py
```diff
@@ -0,0 +1,671 @@
...
+        self.subprotocols = tuple(scope.get('subprotocols', []))
+
+        # TODO(kgriffs): Should we make the use of _BufferedReceiver
+        #   configurable? For example, if the developer knows that
+        #   they will be interleaving receives and sends, then they
+        #   will be able to find out about the 'websocket.disconnect'
+        #   event via one of their receive() calls, and there is no
+        #   need for the added overhead.
+        #
+        #   On the other hand, using _BufferedReceiver may still be
```
kgriffs, line 86: ``_BufferedReceiver`` probably can't assist with back pressure after all, since Falcon is just acting in the stead of the server app, and when Falcon stops receiving messages due to the queue being full, the WebSocket server will just keep receiving from the client in a separate async task.

tests/asgi/test_ws.py
```diff
@@ -0,0 +1,1036 @@
...
+    app.add_middleware(MiddlewareB())
+
+    app.add_route('/{p1}/{p2}', resource)
+
+    async with testing.ASGIConductor(app) as c:
+        async with c.simulate_ws('/v1/v2', headers={}) as ws:
+            assert (await ws.receive_text()) == 'v1:v2:hello:42'
+
+            for i in range(10):
+                message = str(i) or ''  # Test round-tripping the empty string as well
```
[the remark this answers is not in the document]
kgriffs: 🤦 fixed

tests/asgi/test_ws.py
```diff
@@ -0,0 +1,1036 @@
...
+
+                for i in range(100):
+                    await ws.send_text('ignore')
+
+                # NOTE(kgriffs): Cause the buffered receive() on the
+                #   server side to wait occasionally on a new message.
+                await asyncio.sleep(0.001)
```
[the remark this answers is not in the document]
kgriffs: Good point. Actually, it may make sense to make this dynamic so that part of the time it is slower and part of the time it is faster.

falcon/asgi/ws.py
kgriffs: Perhaps I should document this after all, and mention that setting it to zero will disable buffering?

vytas7: +1 for making buffering configurable, and allowing to disable it altogether.

Not sure what the default value should be... Gut feeling says, given that most of ASGI stuff seems to err on the side of excessive buffer bloat, we might want to keep this disabled by default? I would need to test how this feels in practical use tho.

vytas7:
From my perspective, the only pending item is the back pressure discussion and message buffering refinements.
I am otherwise ready to approve :+1: :shipit:

falcon/asgi/ws.py
kgriffs: Having it enabled leads to a better developer experience IMO since it abstracts away subtleties re detecting when the client disconnects.

kgriffs: I suppose we could disable it if we detect uvicorn but the danger is that they change their behavior at some point to more strictly observe the ASGI spec.

kgriffs:
OK gents, I have attempted to address the latest round of feedback. Let me know what you think regarding enabling the receive queue by default.

falcon/asgi/ws.py
vytas7: Ah, yeah, I completely forgot about that aspect of the spec 🤔
Agreed that we must handle it on our side because of that.

falcon/constants.py
```diff
@@ -38,7 +39,16 @@
     if method.strip() != ''
 ]
 
-COMBINED_METHODS = HTTP_METHODS + WEBDAV_METHODS + FALCON_CUSTOM_HTTP_METHODS
+_META_METHODS = [
+    'WEBSOCKET',
```
vytas7, line 43: This might have some weird ramifications for WSGI apps... not sure if it is practically significant to warrant a change though.

Consider the following scenario (tested against this branch):
```python
import falcon


class WeirdStuff:

    def on_websocket(self, req, resp):
        resp.media = {'message': 'Hello'}


app = falcon.App()
app.add_route('/test', WeirdStuff())
```
:arrow_right: 
```
$ curl -X WEBSOCKET http://localhost:8000/test
{"message": "Hello"}
```

vytas7 (changes requested):
I took a fresh look at this... and here come more nitpicks :smiling_imp:

falcon/asgi/app.py
```diff
@@ -650,27 +689,53 @@ def add_error_handler(self, exception, handler=None):
...
-                If not specified explicitly, the handler will default to
-                ``exception.handle``, where ``exception`` is the error
-                type specified above, and ``handle`` is a static method
-                (i.e., decorated with ``@staticmethod``) that accepts
-                the same params just described. For example::
+        Keyword Args:
+            handler (callable): A coroutine function taking the
+                form::
+
+                    async func(req, resp, ex, params, ws=None):
```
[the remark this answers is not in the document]
kgriffs: whoops. fixed.

docs/api/websocket.rst
```diff
@@ -0,0 +1,468 @@
...
+                    #   not already called by the app, the framework will take
+                    #   care of it.
+
+                    # NOTE: If you do not handle this exception, it will be
+                    #   bubbled up to a default error handler that simply
+                    #   logs the message as a warning and then closes the
+                    #   server side of the connection. This handler can be
+                    #   overwritten as with any other error handler for the app.
+
+                    # NOTE: If the error handler accepts an 'ws' kwarg, the framework
```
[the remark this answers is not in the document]
kgriffs: I opted to simply remove this section and add an "Error Handling" section to `websocket.rst` to explain things in more detail. Let me know what you think!

falcon/constants.py
kgriffs: Good catch. I went ahead and explicitly disallowed it just in case.

vytas7 (approved):
It feels that there's room for microoptimization (at least for the WSGI App) by refactoring away the new `WEBSOCKET` check, as well as referencing the `.is_websocket` property. But we are talking nanoseconds, and we could do this microtuning before/after the first beta.

Excellent work otherwise! :100:

falcon/asgi/ws.py
```diff
@@ -506,12 +500,25 @@ class WebSocketOptions:
         media_handlers (dict): A dict-like object for configuring media handlers
             according to the WebSocket payload type (TEXT vs. BINARY) of a
             given message. See also: :ref:`ws_media_handlers`.
+        max_receive_queue (int): The maximum number of incoming messages to
+            enqueue if the reception rate exceeds the consumption rate of the
+            application (default ``4``). When this limit is reached, the
```
CaselIT, line 505: any particular reason for 4? why not 3 or 5?

docs/api/websocket.rst
```diff
@@ -0,0 +1,505 @@
...
+surfaces this event by raising an instance of :class:`~.WebSocketDisconnected`
+to the caller.
+
+On the other hand, the ASGI spec requires the ASGI server to silently consume
+messages sent by the app after the connection has been lost (i.e.,  it should
+not be considered an error). Therefore, an endpoint that primarily streams
+outbound events to the client might continue consuming resources unnecessarily
+for some time after the connection is lost.
+
+As a workaround, Falcon implements a small incoming message queue that is used
```
CaselIT, line 115: I'm not really sold by workaround in this context. Maybe mitigation could be better?

CaselIT (approved):
I've left a couple of minor comments, but I think we can also merge as is

docs/api/websocket.rst
vytas7: IMHO, _workaround_ is fine. _Mitigation_ would suit better if we could mitigate buffer bloat, but we simply can not.

_Workaround_ here means that the ASGI spec does not afford uniform signalling of client disconnect events in all cases, but we work around that issue with an own layer of logic. I would perceive _mitigation_ as more connected to patching security holes, reducing harmful effects.

kgriffs:
> It feels that there's room for microoptimization (at least for the WSGI App) by refactoring away the new `WEBSOCKET` check, as well as referencing the `.is_websocket` property. But we are talking nanoseconds, and we could do this microtuning before/after the first beta.

True enough; I opted for correctness initially but made a note to circle back on this in terms of optimization, bandwidth permitting.

docs/api/websocket.rst
kgriffs: I like "mitigation".

PR #1745, opened 2020-07-26
docs(recipes): discuss parsing nested multipart forms

vytas7:
When it comes to nested multipart forms, most other Python frameworks just slam a fat `wontfix`.

I don't think it should be something we would actively support, but our parser architecture actually happens to support some limited cases out of the box. So I went ahead and added a recipe explaining how one can enable the functionality.


vytas7:
This PR also adds one small functional improvement: the deprecated `Content-Transfer-Encoding` header is accepted if its value is `binary`.

docs/user/recipes/multipart-mixed.rst
```diff
@@ -0,0 +1,96 @@
...
+
+    parser = falcon.media.MultipartFormHandler()
+    parser.parse_options.media_handlers['multipart/mixed'] = (
+        falcon.media.MultipartFormHandler())
+
+.. note::
+    Here we create a new parser (with default options) for nested parts,
+    effectively disallowing further recursion.
+
+    If traversing into even deeper multipart form hierarchy was desired, we
```
kgriffs, line 38: This sentence would better agree with the preceding one if it were converted to present tense, e.g.:

> If traversing into even deeper multipart form hierarchy is desired, we can just reuse the same parser.

docs/user/recipes/multipart-mixed.rst
```diff
@@ -0,0 +1,96 @@
...
+    could just reuse the same parser.
+
+Let us now use the nested form enabled parser in an app:
```
kgriffs, line 41: Using three modifiers for "parser" makes the sentence a little difficult to read. Perhaps reword to "nesting-aware parser" or similar?

docs/user/recipes/multipart-mixed.rst
```diff
@@ -0,0 +1,96 @@
...
+
+    parser = falcon.media.MultipartFormHandler()
+    parser.parse_options.media_handlers['multipart/mixed'] = (
+        falcon.media.MultipartFormHandler())
+
+    app = falcon.App()
+    app.req_options.media_handlers[falcon.MEDIA_MULTIPART] = parser
+    app.add_route('/forms', Forms())
+
+We should now be able to consume a legacy, the now obsolete
```
kgriffs, line 67: Seems like this sentence somehow became convoluted.

docs/user/recipes/multipart-mixed.rst
vytas7: Indeed, I've just happened to glue those parts without rereading the paragraph.

PR #1751, opened 2020-08-02
fix(CompiledRouter): support URI templates with a trailing slash

### LINKED ISSUE #1544: Routes with trailing slash do not work under default API Options
Hello,

It seems when using the current default for api.req_options.strip_url_path_trailing_slash which is False, adding a slash at the end of a route doesn't work.

I think I found out what the problem is.

Here is the code I tried:
```python
import falcon

class StaticResource(object):

    def on_get(self, _: falcon.Request, resp: falcon.Response) -> None:
        resp.status = falcon.HTTP_200
        resp.content_type = 'text/plain'
        resp.body = 'static'

def create_app() -> falcon.API:
    api = falcon.API()
    api.req_options.strip_url_path_trailing_slash = False
    api.add_route('/static_route/', StaticResource())
    api.add_route('/static_route2', StaticResource())
    print(api._router._finder_src)
    return api
```

The finder_src compiled function is :
```python
def find(path, return_values, patterns, converters, params):
    path_len = len(path)
    if path_len > 0:
        if path[0] == 'static_route':
            if path_len == 1:
                return return_values[0]
            return None
        if path[0] == 'static_route2':
            if path_len == 1:
                return return_values[1]
            return None
        return None
    return None
```
We see that the code generated for both routes is the same.
The path passed to this function when calling uri: "http://127.0.0.1:8000/static_route/" is : ['static_route', ''] which has a length of 2, leading the function to return None and therefore an HTTPError 404.
When calling uri: "http://127.0.0.1:8000/static_route", the function returns the correct route and leads to HTTP 200.

Therefore using the default option leads to an impossibility to have routes ending with a trailing slash and even worse, when using routes ending with a trailing slash, the route without it works...

vytas7:
### Summary of Changes

Previously, the default [`CompiledRouter`](https://falcon.readthedocs.io/en/latest/api/routing.html#falcon.routing.CompiledRouter) was erroneously stripping trailing slashes from URI templates.
This has been fixed so that it is now possible to add two different routes for a path with and without a trailing forward slash.

### Related Issues

Closes #1544 

### Open Questions

I'm a bit divided what to do with `add_route()` in case [`RequestOptions.strip_url_path_trailing_slash`](https://falcon.readthedocs.io/en/latest/api/app.html#falcon.RequestOptions.strip_url_path_trailing_slash) is set to `True`, and the provided template ends with a trailing slash?
* Do nothing special, effectively adding a non-functional route that is never hit (this may be a breaking change is someone had such a route).
* Raise a `ValueError` asking to only add routes without a trailing slash (this may be a breaking change is someone had such a route).
* Strip the trailing slash from the template before submitting it to the router's `add_route()`.
* Strip the trailing slash from the template before submitting it to the router's `add_route()`, and emit a deprecation warning.
* Something else?

I have now proceeded with the first option, documenting this as a breaking change, and sprinkling some warnings throughout the docs. Thanks @CaselIT for input!


CaselIT:
Regarding the question, the problem I see is that we cannot assume that an user has set the options before adding the routes. Also the options may be dynamically updated.
Taking this into consideration, I don't think a warning should be raised, we should just document, maybe with a warning, the behaviour of the router wrt that request option.

vytas7:
Ah, good point wrt the configuration order @CaselIT !
[`RequestOptions.strip_url_path_trailing_slash`](https://falcon.readthedocs.io/en/latest/api/app.html#falcon.RequestOptions.strip_url_path_trailing_slash) is by default `False`, so if the user hasn't set that to `True` yet, we obviously cannot do anything about it. But if the user has already overridden to `True`, I am inclined to emit a warning though, that might be helpful in debugging... :thinking:

CaselIT:
The documentation looks good 👍

vytas7:
@kgriffs good point; I have now added a dedicated test to better exercise this scenario.

CaselIT (approved):

Issue #1281, opened 2018-05-14
Make using Cython with Falcon easier


kgriffs:
Ideas:

* Specify cython with setuptools extras
* Ship manylinux wheel

kgriffs:
(Re manylinux we can port the build script from the 2.0 branch.)

koobs:
While we're leveling up cython/packaging with regard to falcon, some additional notes, quoted from Cython documentation [1]

 * It is strongly recommended that you **distribute the generated .c files** as well as your Cython sources, so that users can install your module without needing to have Cython available.

 * It is also recommended that Cython compilation not be enabled by default in the version you distribute.  Even if the user has Cython installed ...

[1] http://docs.cython.org/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules

vytas7:
Discussion from https://github.com/falconry/falcon/pull/1753 (see also other comments therein)

@CaselIT said
> I think all of the supported manylinux images. https://github.com/RalfG/python-wheels-manylinux-build/tags
> Manylinux 2014 does support s390x: https://github.com/pypa/manylinux#manylinux2014
> 
> When creating wheels for sqlalchemy, I've read that is still suggested to use manylinux2010 (or manylinux1 if support for older linux is required). I'm using the 2010 in sqlalchemy, but it manages to automatically create also the manylinux1 version.
> Both are uploaded to pypi and pip chooses the appropriate one for the system

vytas7:
Thanks for your feedback @koobs !
We'll think of shipping the generated `.c` files; that's an interesting idea.

When it comes to not enabling Cython by default, I'm admittedly not sure I'm following the reasoning, particularly if we follow the `.c` file advice.
Regardless, this issue is mostly about binary wheels, and that is where you obviously do want to compile whatever possible.

CaselIT:
I can take a look at adapting the release workflow used in sqlalchemy for falcon, if the plan is to migrate to use github actions

vytas7:
Re shipping `.c` files, I would be reluctant littering Git with them, and more importantly, having the burden of keeping them in sync with `.py` & `.pyx`.

Which means, if we pursue this, we'd probably have to
* Accommodate `setup.py` to be able to use these `.c` files, if present
* Generate them in GH release Actions (similarly to building wheels) when generating `sdist`, before uploading `sdist` to PyPy.

CaselIT:
So the plan could be:
- The action runs when a release is created

### For wheels:
[...]

PR #1753, opened 2020-08-05
chore(CI): migrate Travis CI ⇒ GitHub Actions

vytas7:
We are getting Windows and Mac support with zero effort.
And runtimes are really quick, around 6-7 minutes for the whole suite.

Things to do:
- [x] Update `on` hooks to reflect Falcon workflow
- [x] Codecov integration
- [x] CPython 3.5.2 gate (easiest to probably just `apt-get` on 16.04)
- [x] Update `README.rst` banners
- [x] Add a big endian architecture (could be done in QEMU, @CaselIT has an ARM recipe from a SQLAlchemy PR); had to do this in a separate manual action unfortunately, taking ~40 minutes. There is optimization potential, but bringing that to the vicinity of 6-7 minutes does not sound realistic at this point.

A workflow run example on my fork:
https://github.com/vytas7/falcon/actions/runs/198951620


CaselIT:
Question: do we want to have also arm tests?

vytas7:
No, I don't think that is a high priority tbh.
But maybe we could later build a separate workflow for building wheels, and that should probably include ARM (w/e architecture Ubuntu and Amazon are using, it's called just `arm64` I think).

CaselIT:
For building wheels you can use the manylinux image and or this image to automate it
- https://github.com/pypa/manylinux
- https://github.com/RalfG/python-wheels-manylinux-build

vytas7:
Which architectures does that action do? Is `s390x` included, which ARM flavours are included?

CaselIT:
I think all of the supported manylinux images. https://github.com/RalfG/python-wheels-manylinux-build/tags
Manylinux 2014 does support s390x: https://github.com/pypa/manylinux#manylinux2014
When creating wheels for sqlalchemy, I've read that is still suggested to use manylinux2010 (or manylinux1 if support for older linux is required). I'm using the 2010 in sqlalchemy, but it manages to automatically create also the manylinux1 version. Both are uploaded and pip chooses the appropriate one for the installed system

vytas7:
Aha, that's awesome. We'll probably have to skip 32 bit architectures though, because we don't test enough on them. For instance, WSGI cythonized buffered reader is disabled by design on 32-bit, falling back to pure Python.

CaselIT:
32bit is really only windows at this point.

vytas7:
`i686` is well 32 bit?

CaselIT:
Yes, but who is still using 32bit linux at this point?

vytas7:
True; that's also why I'm saying we should drop them. All major distributions have now dropped 32 bits, or demoted that to community support.

vytas7:
And thanks @CaselIT for your insights, that's really useful.

I _think_ we may decide to cut straight to `manylinux2014` ([PEP 599]()), since we dropped Python 2, <= Python 3.4 support, so you can probably expect people running Python 3.5+ on `manylinux2014` compatible platforms.

CaselIT:
The only problem may be pip. Note that if it works like in sqlalchemy, 2014 may also generate 2010 and 1 wheels.
Pip can then download the appropriate one

CaselIT:
Following up from https://github.com/falconry/falcon/issues/1281. I think there should also be a job that test that the readme is accepted by pypi using `twine check`, since I don't think that the create wheel action is the best place to do that check.

Incidentally the current readme in master fails the test :)

vytas7:
Agreed; to that end, we ought to add a `tox` environment for that.

tox.ini
```diff
@@ -276,6 +283,20 @@ basepython = python3.8
 deps =
 commands = {toxinidir}/tools/check-vendored.sh
 
+# --------------------------------------------------------------------
+# Package sanity check with twine
+# --------------------------------------------------------------------
+
+[testenv:twine_check]
```
kgriffs, line 290: +1

Issue #1652, opened 2020-01-25
Improve test readme. Without setting the proper env variable 300+ failure occur


CaselIT:
If I follow the readme to run the test

> You can manually test changes to the Falcon framework by switching to the directory of the cloned repo and then running pytest:

> $ cd falcon
> $ pip install -r requirements/tests
> $ pytest tests

I get the following result 
` 261 failed, 1819 passed, 20 skipped, 40 warnings, 128 errors`

I've tried with tox and it works correctly.

I've traced the issue to the missing env variable `FALCON_ASGI_WRAP_NON_COROUTINES=Y`

I think the readme should mention it

CaselIT:
maybe the variable `FALCON_ASGI_WRAP_NON_COROUTINES` should just be set in `conftest` since it is clearly required by the test?

vytas7:
We ought to clean this up the 3.0 beta release.

CaselIT:
This can be solverd by placing in `setup.cfg` under the `[tool:pytest]` key
```
env =
    FALCON_ASGI_WRAP_NON_COROUTINES = Y
    FALCON_TESTING_SESSION = Y
    PYTHONASYNCIODEBUG = 1
```

vytas7:
Reopening this, as it apparently requires the [`pytest-env`](https://pypi.org/project/pytest-env/) plugin that we haven't documented.

Alternatively, we could just look at what the said plugin does, and vendor it in `conftest.py` (considering it hasn't been updated in 3 years+).

PR #1754, opened 2020-08-06
Action to create, test and publish wheels

CaselIT:
# Summary of Changes

Action to generate the wheels on release. Fixes https://github.com/falconry/falcon/issues/1281

Track prograss:
- https://github.com/CaselIT/falcon/releases
- https://github.com/CaselIT/falcon/actions/runs/201295934
- https://test.pypi.org/project/falcon-caselit/3.0.0.dev5/#files

TODO:
- [ ] remove various todos
- [x] create upload sdist
- [x] create c files to place in sdist
- [x] crate wheels for linux (manylinux1,2010,2014)
- [x] create wheels for windows & osx
- [x] create wheels for other supported platforms
- [x] secrets must be added by the falcon maintainers to allow pushing to pypi (also on the test pypi I would think?)


CaselIT:
Regarding the compilation of cython files with cython, are the generated c files always the same regardless of the python 3 version?
I generate them with `cythonize -3 falcon/**/*.pyx`, but I'm not sure if it is ok

CaselIT:
For arm64, there are problems testing for python 3.5, since it tries to install `maturin` from source, and it requires rust to be installed. I think we could not provide a wheel for py 3.5 arm64. Or we skip testing it?

setup.cfg
```diff
@@ -1,16 +1,22 @@
...
+#     ignore:Using or importing the ABCs:DeprecationWarning
+#     ignore:cannot collect test class 'TestClient':pytest.PytestCollectionWarning
+
+env =
+    FALCON_ASGI_WRAP_NON_COROUTINES = Y
+    FALCON_TESTING_SESSION = Y
+    PYTHONASYNCIODEBUG = 1
+
+[flake8]
+max-line-length = 100
```
CaselIT, line 22: This was a mistake in stashing. it will be removed once I revert the `tag_build`

kgriffs:
This is looking good! I added a token to the project secrets for the real index. Re the test index, I am waiting on an inquiry to see if we can take over https://test.pypi.org/project/falcon/ from the current owner.

kgriffs:
Question: Is there a convenient way to run the action in "test mode" which would flip it to point to the test index and disable the upload of the release assets? I am admittedly not familiar with GH actions to know if it would be easy to toggle this...

CaselIT:
> 
> 
> Question: Is there a convenient way to run the action in "test mode" which would flip it to point to the test index and disable the upload of the release assets? I am admittedly not familiar with GH actions to know if it would be easy to toggle this...

Not currently. I think it could be done by supporting a manual trigger and do the steps that upload the wheel to github and to pypi only if the triggering event is a release. 

Regarding changing the test pypi I don't know, since you need to both change the url and the token, since pypi and the test pipy are completely separate installation and do not share the accounts.
I guess that the `TWINE_REPOSITORY_URL` could be set to a github secret, so that the configuration would be:
- to use the official pypi: official token, remove the secret connected to the url.
- to use the test pypi: test token, create a secret connected to the `TWINE_REPOSITORY_URL` with the test url as value.

Another alternative (but I'm not sure if it's possible) is to add an if to the upload to pypi step to check for a particular message in the release message to disable it on demand?

kgriffs:
[Added TEST_PYPI_TOKEN secret]

CaselIT:
Ok, I'll do the update to the pr

CaselIT:
Updated. IIRC to test this this workflow should be on the default branch, so it's probably best to test on a fork, instead of the actual repo here.
The action trigger when a release (not a tag) is created.

I think we could keep the wheel creation targeted to test pypi for 1 or 2 of the alpha/beta release, check that everything works, then manually upload the wheel to the live pypi for that couple of test run. Then a follow on pr (maybe before the last rc) can change it target the actual pypi.

setup.cfg
```diff
@@ -13,4 +11,4 @@ filterwarnings =
     ignore:Unknown REQUEST_METHOD. '(FOO|BAR|BREW|SETECASTRONOMY)':wsgiref.validate.WSGIWarning
     ignore:"@coroutine" decorator is deprecated:DeprecationWarning
     ignore:Using or importing the ABCs:DeprecationWarning
-    ignore:cannot collect test class 'TestClient':pytest.PytestCollectionWarning
+    ignore:cannot collect test class 'TestClient':pytest.PytestCollectionWarning
```
vytas7, line 14: No newline at the end of file.

vytas7 (changes requested):
As far as I am qualified (I haven't been involved in the Falcon release process too much), this looks great :+1: Bar one missing newline at the end of file, this is ready to go if @kgriffs is happy.

Btw, do `setup.cfg` additions also address https://github.com/falconry/falcon/issues/1652?

setup.cfg
CaselIT: done.

CaselIT:
> Btw, do setup.cfg additions also address #1652? nvm I misread, you only added these to the respective envs.

That was the setup.cfg I use locally, since I like using pytest directly wrt tox. They indeed fix #1652 
They where published because I mistakenly stages all the file instead of only the relevant lines

vytas7:
> They where published because I mistakenly stages all the file instead of only the relevant lines.

That explains. Because I swear I saw them! :slightly_smiling_face:

vytas7 (approved):
:100:

CaselIT:
It would be great if we could review and merge this, so we can tweak it in the next pre-releases

vytas7 (approved):

kgriffs (approved):

CaselIT:
Thanks for merging. I've made a follow up at #1787 that fixes an error with linux build that has come up after I created this

PR #1756, opened 2020-08-07
chore: add ASGI classifier, along with the standard generic WWW/HTTP classifier

kgriffs:


CaselIT:
There seem to be a problem. From [this run](https://github.com/CaselIT/falcon/runs/960225092?check_suite_focus=true):
```
HTTPError: 400 Client Error: Invalid value for classifiers. Error: Classifier 'Topic :: Internet :: WWW/HTTP :: ASGI' is not a valid classifier. for url: https://test.pypi.org/legacy/
```

I'm not sure if it is a problem of the test pypi or not

CaselIT:
Here https://pypi.org/classifiers/ the asgi one is not listed

CaselIT:
It may be a problem only of the test pypi, since the tag a2 has them and it was published correctly https://github.com/falconry/falcon/blob/3.0.0a2/setup.py

PR #1757, opened 2020-08-09
Add arm build. Improve workflow for emulated workflow

CaselIT:
Improve the workflow action for s390x, adding caches for the pip packages. This allows the run time to be about 25 minutes . This is an example of a run with cache https://github.com/CaselIT/falcon/actions/runs/201298435

It also adds a test for the arm architectore


.github/workflows/tests-emulated.yaml
```diff
@@ -0,0 +1,74 @@
...
+            chown -R $(whoami) $PIP_CACHE_DIR &&
+            pip install -U pip tox &&
+            python --version &&
+            pip --version &&
+            tox --version &&
+            tox -e py38_cython"
+
+      - name: Run tox arm64v8
+        if: ${{ matrix.architecture == 'arm64v8' }}
+        uses: docker://arm64v8/python:3.8-buster
```
vytas7, line 56: You could probably just template these on `matrix.architecture`, but skipping `if` and the copypasted blocks altogether?

CaselIT: nope, I've tried and if fails if you template uses. both for actions and for docker. See this try [here](https://github.com/CaselIT/falcon/actions/runs/200257229) (scroll down)
That would be too easy I guess.

There is also no support for yaml anchors as of yet.
We could add an `sh` script, but I'm not sure where the best for it could be.

vytas7: Aha, nvm then, it works as-is.

CaselIT: We may be able to use this new feature though https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action

Probably best to experiment on a follow on pr though

CaselIT: But I guess you still cannot template `uses`, so it may be a moot point

PR #1758, opened 2020-08-09
fix(media): let media handlers deserialize empty media

### LINKED ISSUE #1216: Processing empty req.media results in 400

### LINKED ISSUE #1589: Media handlers do not get a chance to deserialize zero-length payload

CaselIT:
# Summary of Changes

Let media handlers deserialize an empty media.

NOTE: as is this change creates a possible large breaking change for WSGI application since it changes empty json requests from `None` to an 400 error.
As implemented is probably the cleanest version, but I think we may scale it back a bit by:
- if the media handler returns an error, the stream was empty and the user has not supplied a fallback value, warn the user that next version will propagate the exception
- return `None`

Thought?

# Related Issues

Fixes #1589
Fixes #1216

# Pull Request Checklist


CaselIT:
> As implemented is probably the cleanest version, but I think we may scale it back a bit by:
> 
>    * if the media handler returns an error, the stream was empty and the user has not supplied a fallback value, warn the user that next version will propagate the exception
> 
>    * return `None`

The problem with this is that there is no way of not getting the warning if one prefers the new behavior.

docs/_newsfragments/1589.breakingchange.rst
```diff
@@ -0,0 +1,3 @@
+Deserializing an empty JSON body on a request will return a BadRequest.
+Previously the ``None`` was returned. To retain the current behaviour
```
vytas7, line 2: Suggested edit: _Previously, ``None`` was returned._

docs/_newsfragments/1589.breakingchange.rst
```diff
@@ -0,0 +1,3 @@
+Deserializing an empty JSON body on a request will return a BadRequest.
```
vytas7, line 1: Should we probably be more specific that this deserialization pertains to `req.media` / `req.get_media()`?

docs/_newsfragments/1589.breakingchange.rst
vytas7, line 2: Probably you meant: :arrow_right: _To retain the previous behavior, please <...>_?

falcon/asgi/request.py
```diff
@@ -693,16 +695,31 @@ async def get_media(self):
...
         Warning:
             This operation will consume the request stream the first time
             it's called and cache the results. Follow-up calls will just
             retrieve a cached version of the object.
 
+        Args:
+            when_empty_fallback: Fallback value to return when there is no body
+                in the request and the media handler raises an error
+                (like in the case of the default JSON media handler).
+                By default falcon use the value returned by the media handler
```
vytas7, line 715: Not critical, but I would capitalize Falcon everywhere unless referring to the module name. I am aware the current docs are a bit of a mix tho.

vytas7, line 715: Suggested edit: _By default, Falcon uses the value returned by the media handler or propagates the raised exception, if any_.

falcon/asgi/request.py
```diff
@@ -716,6 +733,10 @@ async def get_media(self):
                 self.content_type,
                 self.content_length
             )
+        except Exception:
+            if when_empty_fallback is _UNSET or self.stream.tell() != 0:
```
vytas7, line 737: Is there a test for ASGI without the `Content-Length` header? It might be that our test framework currently even does not offer good support for that...

But it would be good to add tests for failing media deserialization, where the `Content-Length` header is unset, both when there is a streamed body, and when not.

docs/_newsfragments/1589.breakingchange.rst
vytas7, line 1: will return a BadRequest :arrow_right: will raise an instance of `HTTPBadRequest`.

CaselIT:
Thanks for the review.

> Your fallback machinery is undeniably neat, but for existing users that would be almost as abrupt breaking as adding try... excepts everywhere.

That was my thinking as well, as is now it's probably the cleanest behaviour but it is possibly a very large change. All in all, I'm not sure I would be in favour of merging it as is.

> 
> 
> > As implemented is probably the cleanest version, but I think we may scale it back a bit by:
> > 
> > * if the media handler returns an error, the stream was empty and the user has not supplied a fallback value, warn the user that next version will propagate the exception
> > * return `None`
> 
> The problem with this is that there is no way of not getting the warning if one prefers the new behavior.

Thinking a bit more on this, I think we need to provide a constant that an user may use to opt for the new behaviour, something like `MEDIA_PROPAGATE_EMPTY_EXCEPTION` (possibly shorter), so that the warning may be silenced opting for the new behaviour.
Then in falcon 4 we may set this value by default in `get_media`.
Enum are not used by falcon atm (I think) but maybe one could be used for this? Not sure

falcon/asgi/request.py
CaselIT: I'm not sure, but I think tell returns the number of read bytes, regardless of it the body was streamed or not. After trying to de-serialize the media it should be `> 0` if there was a body.

I can look into trying to omit content-length

vytas7:
A side note re `enum`s. No, they are not used by Falcon per se yet, but we have now added support for setting `resp.status` to an instance of [`http.HTTPStatus`](https://docs.python.org/3/library/http.html#http.HTTPStatus).

CaselIT:
Note: this should maybe wait for https://github.com/falconry/falcon/pull/1741 since it changes handlers a bit

kgriffs:
> I think we need to provide a constant that an user may use to opt for the new behaviour

Just as a reminder, don't forget to create an issue and assign it to the appropriate milestone so this work doesn't fall through the cracks.

> Note: this should maybe wait for #1741 since it changes handlers a bit

Will you have some time to circle back on this PR over the next few days? Thanks!

CaselIT:
> Will you have some time to circle back on this PR over the next few days? Thanks!

Sure, I can complete this now that we have also the websocket done.

But I'm not sure if passing a value to `get_media` is the best way. I mean it makes sense for json, but as @vytas7 pointed out, it may not be great for other media handler, like the multiplar one

I guess the alternative is to have something on the handler, like a method we can call or a constant (but the method is more flexible) or have an option like the above suggested `MEDIA_PROPAGATE_EMPTY_EXCEPTION ` suggested above

kgriffs:
TBH I think we need to also update the media handlers themselves to afford a more specific error handling protocol. Otherwise, catching any `Exception` type could potentially mask some issues from the app developer, making said issues difficult to diagnose.

I also think we should bubble up more context to the app developer re what actually happened, allowing them to choose their own adventure as needed. The more affordances the better.

With that in mind, I would propose something along these lines:

```py
def deserialize(self, stream, content_type, content_length):
    data = stream.read()

    # BREAKING: Previously we would raise json.JSONDecodeError in this case;
    #   we should probably subclass ValueError since json.JSONDecodeError also
    #   subclasses the same.
    #
    #   We might also want to install a default error handler in the app
    #   instance for this type. Perhaps it could just raise HTTPBadRequest.
    if not data:
        raise MediaNotFoundError()

    try:
        return self.loads(data.decode())
    except ValueError as ex:
        # Could be either json.JSONDecodeError or UnicodeDecodeError, since
        #   both subclass ValueError.

        # BREAKING: Previously we would just raise errors.HTTPBadRequest in
        #   this case. To mitigate the impact of this change, a default error
        #   handler can be added to the App instance that will raise
        #   an instance of HTTPBadRequest.
        raise MediaMalformedError(ex)  # set inner exception
```

The `get_media()` method would not catch these exceptions by default, but let them bubble up to the caller. This is similar to what happens already. We could still accept a kwarg to specify a default value (more on this a bit further down below).

We would note in the docs that custom media handlers should normally follow this pattern (especially if they are releasing it in a library for third-party consumption), although they might choose to return a default value in lieu of raising an instance of MediaNotFoundError.

In terms of customizing the default behavior, the framework provides at least four affordances: (a) implement your own custom media handler, (b) catch the exception in the responder, and (c) add custom exception handlers to the App instance, (d) implement a custom Request subclass that overrides `get_media()`. It might be useful to explicitly note these options in the media handling docs.

(b) could look something like this:

```py
try:
    resource_representation = self.media

except MediaNotFoundError:  # Empty request body
    # Depending on the interface specification, the app might just opt for a
    # default value here, or return a 400, or something else.
    resource_representation = {}

except MediaMalformedError as ex: # Bad syntax or incomplete body
    ex.inner  # Can be referenced to determine the root cause
    pass

except Exception as ex:
    # Some other weird thing might have happend; deal with it here
    # or omit this except block and let a registered error handler
    # deal with it.
    pass
```

We could also keep the kwarg for `get_media()` for returning a default value when `MediaNotFoundError` or `MediaMalformedError` is raised (other exceptions would not be caught). This kind of makes sense in the context of what we afford for headers and what the stdlib affords for `dict`. Anyway, this is essentially a new feature and people could switch from using the `req.media` to opt into using the feature if they find it useful. The default value might very well need to be different depending on the API design and the resource being manipulated. For now we default to just raising the errors since that is similar to what we already do, and if people want a default value they can opt in as just suggested.

In addition, we might accept a kwarg in the media handler initializer for setting a universal default that the handler would return in lieu or raising an error, but I'm not yet convinced that is better than `try...except` in the responders.

On the other hand, for multipart a default value doesn't make a lot of sense. If someone wants to specify one as a kwarg to `get_media()` I guess that's their choice and there isn't any reason for the framework to explicitly disallow this pattern. But we shouldn't do anything in the design that requires all media handlers to return a default value instead of raising an error

vytas7:
That's a great summary @kgriffs.

Adding to this, as I wrote on the issue, I would also suggest adding the default value fallback as an initializer option to certain handlers we ship, namely `JSONHandler` (and probably `MessagePackHandler`).
This way, people could easily customize their APIs to match 2.0 behaviour.

vytas7:
Tbh repeated access to `req.media` is problematic regardless, so maybe we should be explicitly discouraging this pattern?

If a responder exploded with a `MediaMalformedError`, or chose to suppress it, for middleware processing this response, it would look as either some default value, or a `MediaNotFoundError` would be raised, since the request stream could have been exhausted at this point.

But then we should probably be suggesting alternatives to this pattern in our FAQ.

vytas7:
@kgriffs just a clarification :arrow_up_down: 
```python
    # BREAKING: Previously we would raise json.JSONDecodeError in this case;
    #   we should probably subclass ValueError since json.JSONDecodeError also
    #   subclasses the same.
    #
    #   We might also want to install a default error handler in the app
    #   instance for this type. Perhaps it could just raise HTTPBadRequest.
    if not data:
        raise MediaNotFoundError()
```
While this is true for an isolated instance of `JSONHandler`, the current 2.0 behaviour is that [`req.media` simply returns `None`](https://github.com/falconry/falcon/blob/2.0/falcon/request.py#L961).

CaselIT:
Thanks for the summary.

I think that it makes sense. We should probably have a section in the media docs that explains the rationale and how custom handlers are best implemented.

I'll try to update this a bit to see if we can move it in the right direction. (I think I'll first update the implementation and ask for feedback, then once we have nailed it update the testing)

Regarding the exceptions, should they be subclasses of HTTPBadRequest?
If not, I think we should provide a default handler for them that sends a 400 http error, otherwise by default the generic Exception handler would be called resulting in a 500 error that may be unexpected to most users.

vytas7:
> Regarding the exceptions, should they be subclasses of HTTPBadRequest?
If not, I think we should provide a default handler for them that sends a 400 http error, otherwise by default the generic Exception handler would be called resulting in a 500 error that may be unexpected to most users.

Agreed @CaselIT .

If I'm just using the basic media functionality like `req.get_media()`, I expect the framework to return an [`HTTP 400`](https://tools.ietf.org/html/rfc7231#section-6.5.1) to the client out of the box, provided the handler in question is conforming to our protocol (which the standard `JSONHandler` obviously must do).

CaselIT:
One thing to decide if is if the websockets handlers should follow the same behaviour

CaselIT:
I've updated the implementation to see how it could look like.

I've gone for the route of having the media errors be subclasses of `HTTPBadRequest`. In the end we have tree errors:
- a generic `MediaError` subclass of `HTTPBadRequest`
- `MediaMalformedError` subclass of `MediaError`
- `MediaNotFoundError` subclass of `MediaError`

The generic `MediaError` is raised by the `Request` class if any of error was raised while de-serializing the request the first time. This is to avoid ambiguous behaviors, like raising a `MediaMalformedError` the first time and then `MediaNotFoundError` on following accesses.

Could you check if this implementation may make sense? It's not ready to be merged yet, since it's missing tests and docs.

kgriffs:
> One thing to decide is if the websockets handlers should follow the same behaviour

Let's address this post 3.0

kgriffs:
> The generic `MediaError` is raised by the `Request` class if any of error was raised while de-serializing the request the first time. This is to avoid ambiguous behaviors, like raising a `MediaMalformedError` the first time and then `MediaNotFoundError` on following accesses.

Good catch. This is a subtle, yet important issue.

I've been thinking about this, and IMO the ideal developer experience is probably to consistently get `MediaMalformedError` or `MediaNotFoundError` both the first time and the Nth time. To me this is conceptually more consistent with the idea that we "cache" the result of parsing the body into a a media result, whether it be a successful result or an error result. Furthermore, raising a generic error is less transparent and may make things that much harder to debug.

CaselIT:
> 
> 
> > The generic `MediaError` is raised by the `Request` class if any of error was raised while de-serializing the request the first time. This is to avoid ambiguous behaviors, like raising a `MediaMalformedError` the first time and then `MediaNotFoundError` on following accesses.
> 
> Good catch. This is a subtle, yet important issue.
> 
> I've been thinking about this, and IMO the ideal developer experience is probably to consistently get `MediaMalformedError` or `MediaNotFoundError` both the first time and the Nth time. To me this is conceptually more consistent with the idea that we "cache" the result of parsing the body into a a media result, whether it be a successful result or an error result. Furthermore, raising a generic error is less transparent and may make things that much harder to debug.

I also had the same thought, and I agree that it would be best. The only problem with it is maybe that we need to keep a reference of the exception until the request get gced, but I think it would not cause too much hassle, since it should be a fairly short time in any case.

I'll do that and remove the generic `MediaError`.

CaselIT:
@vytas7 @kgriffs I've updated. It should be ready for review.

The only thing to decide if we should somehow raise a `MediaNotFoundError` in the multipart handler

falcon/request.py
```diff
@@ -990,12 +994,32 @@ def get_media(self):
...
+                (like in the case of the default JSON media handler).
+                By default, Falcon uses the value returned by the media handler
+                or propagates the raised exception, if any.
+
+        Returns:
+            media (object): The deserialized media representation.
         """
-        if self._media is not None or self.bounded_stream.eof:
+        if self._media_error is not None:
+            raise self._media_error
```
kgriffs, line 1021: Check `when_empty_fallback` here as well?

kgriffs (changes requested):

falcon/asgi/request.py
```diff
@@ -712,16 +715,33 @@ async def get_media(self):
...
+                By default, Falcon uses the value returned by the media handler
+                or propagates the raised exception, if any.
+
         Returns:
             media (object): The deserialized media representation.
         """
 
-        if self._media is not None or self.stream.eof:
+        if self._media_error is not None:
+            raise self._media_error
```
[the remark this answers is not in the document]
CaselIT: My reasoning for not checking it is that after the first call, this function replays the previous call.

But I guess it may be strange

falcon/errors.py
```diff
@@ -2288,6 +2288,100 @@ def __init__(self, param_name, headers=None, **kwargs):
...
+
+class MediaMalformedError(HTTPBadRequest):
+    """400 Bad Request.
+
+    Exception raised by a media handler when trying to parse a malformed body.
+    The actual exception is stored in the ``source_error`` attribute.
+
+    Args:
+        media_type (str): The media type that was raised this exception.
+        source_error (Exception or None): The source exception that was the cause
```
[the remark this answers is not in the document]
CaselIT: Sure, I'm already using `raise from` in the handler

falcon/asgi/request.py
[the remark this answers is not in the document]
CaselIT: on second thought, I think I agree with you. It makes better sense to consider it also on following calls, otherwise it's not a very useful parameter

falcon/errors.py
```diff
@@ -2288,6 +2288,100 @@ def __init__(self, param_name, headers=None, **kwargs):
...
+    """400 Bad Request.
+
+    Exception raised by a media handler when trying to parse an empty body.
+
+    Note:
+        Some media handler, like the URL-encoded form one, allow an empty body.
+        In these cases this exception will not be raised.
+
+    Args:
+        media_type (str): The media type that was raised this exception.
```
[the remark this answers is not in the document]
CaselIT: I think we should use the header values also, since it would probably be better wrt "JSON" or "messagepack"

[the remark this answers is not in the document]
CaselIT: we don't seem to store it anywhere, so I guess it's the same

CaselIT:
ready for another round

vytas7:
@CaselIT let me take a look at the multipart stuff tomorrow.

It shouldn't be too hard to raise that specialized error in the parsing process.
Not sure, however, if it brings much benefit in the multipart case. If the supplied `Content-Type` provides a valid `boundary` parameter, it could be treated as part of the form, i.e., at least it shouldn't be sent at random together witth `GET` requests.
But it probably still has value for some generalized middleware (?)

kgriffs:
@vytas7 What do you think we need to do with multipart for 3.0 vs. what we can push to a later release?

vytas7:
On consideration, it's probably fine to leave it as-is.

Providing `Content-Type` set to `multipart/form-data` with a valid boundary is enough to construct and return the iterator over parts... So maybe we could do without that specialized empty body error for now (3.0).

vytas7 (changes requested):
Doh, managed to write this comment on a wrong PR...

Meant it here: we also need to restore 100% coverage before merging this. As I understand, a branch is not fully covered here: https://codecov.io/gh/falconry/falcon/pull/1758/diff#D4-2385

CaselIT:
Yes I noticed, I'll test it explicitly

CaselIT:
I think it;s ready for final review

docs/_newsfragments/1589.breakingchange.rst
```diff
@@ -0,0 +1,17 @@
...
+   Previously ``None`` was returned in all cases without calling the handler.
+*  Exceptions raised by the handlers are wrapped as
+   :class:`falcon.MediaMalformedError`, and return a
+   ``400 Bad Request`` response.
+*  Subsequent calls to :meth:`Request.get_media` or :attr:`Request.media` will
+   re-raise the same exception, if the fist call ended in an error, unless the
+   exception was a :class:`falcon.MediaNotFoundError` and a default value is passed.
+   Previously ``None`` was returned.
+
+External handlers should update their logic to align to the internal falcon handlers.
```
vytas7, line 17: Nitpick: I would capitalize Falcon everywhere unless referring to `falcon` as a module.

docs/api/media.rst
```diff
@@ -128,6 +128,30 @@ middleware. Here is an example of how this can be done:
...
+   :class:`falcon.MediaMalformedError` should be raised. The original exception, if any,
+   is stored in the ``__cause__`` attribute of the raised instance. By default, this
+   error will return a  ``400 Bad Request`` response to the client.
+
+If any exception was raised by the handler while parsing the body, all subsequent invocations
+of :meth:`Request.get_media` or :attr:`Request.media` will result in a re-raise of the same
+exception, unless the exception was a :class:`falcon.MediaNotFoundError` and a default value
+is passed in ``when_empty_fallback``.
+
+External handlers should update their logic to align to the internal falcon handlers.
```
vytas7, line 153: Same here, I would prefer _Falcon_.

falcon/media/urlencoded.py
```diff
@@ -9,7 +10,11 @@ class URLEncodedFormHandler(BaseHandler):
     URL-encoded form data handler.
 
     This handler parses ``application/x-www-form-urlencoded`` HTML forms to a
-    ``dict`` in a similar way that URL query parameters are parsed.
+    ``dict`` in a similar way that URL query parameters are parsed. An empty body
+    will be parsed as an empty dict.
+
+    This handler will raise :class:`falcon.MediaMalformedError` in case of errors,
+    usually only if the body cannot be parsed as ASCII.
```
vytas7, line 17: Could it be worth to clarify:
> _When deserializing,_ this handler will...

In addition, not a big deal, but we could probably be slightly more specific than that _usually_ here.
AFAICT there are only two cases where this might raise:
* The request payload is not ASCII as you describe it here.
* Any of the URL-encoded strings is not valid UTF-8.

vytas7 (changes requested):
IMHO this collective design effort has really shaped up nicely :100: Let's hope our users will agree :slightly_smiling_face: 

I only had a couple of cosmetic documentation nitpicks, but it wouldn't be the end of the world merging this as-is in favour of getting `b1` out the door quicker.

falcon/asgi/request.py
```diff
@@ -712,16 +715,38 @@ async def get_media(self):
...
         """
 
-        if self._media is not None or self.stream.eof:
+        if self._media_error is not None:
+            if when_empty_fallback is _UNSET or not isinstance(
+                self._media_error, errors.MediaNotFoundError
+            ):
+                raise self._media_error
+            self._media = when_empty_fallback
+            self._media_error = None
```
kgriffs, line 748: It seems like we might want to avoid introducing potentially surprising side effects when the empty fallback is provided. As an app developer I would personally expect that if I called the method without passing `when_empty_fallback`, an error would be raised consistently on an empty body, even if somewhere else (say in a middleware component) the method was invoked with a fallback value. My mental model is more like `dict.get()` vs. `dict.setdefault()` (albeit that the analogy is a bit of a stretch).

falcon/media/msgpack.py
```diff
@@ -9,10 +9,14 @@
...
     """Handler built using the :py:mod:`msgpack` module.
 
-    This handler uses ``msgpack.unpackb()`` and ``msgpack.packb()``. The
+    This handler uses ``msgpack.unpackb()`` and ``msgpack.Packer().pack()``. The
     MessagePack ``bin`` type is used to distinguish between Unicode strings
     (of type ``str``) and byte strings (of type ``bytes``).
 
+    This handler will raise a :class:`falcon.MediaNotFoundError` when attempting
+    to parse an empty body; it will raise a :class:`falcon.MediaMalformedError`
+    when if an error happens while parsing the body.
```
kgriffs, line 18: ~~when~~

kgriffs (changes requested):

falcon/asgi/request.py
CaselIT: that was done deliberately, but I see your point.

the though was that `when_empty_fallback` takes the place of the data returned by the handler. Instead you are suggesting we need to supply it the every time since we don't cache it. So also here https://github.com/falconry/falcon/blob/6d0d2bd67ac8a34d256e932e24639dfcab544f07/falcon/asgi/request.py#L763-L767 we would not cache it?

Because otherwise it would not consistent the first invocation vrt the subsequent ones

vytas7: I'm with @kgriffs on this one. :arrow_up: If we support this machinery for `Request.get_media()` (as the `when_empty_fallback` _kwarg_), that is.

kgriffs: So what I was suggesting was that we would not cache `when_empty_fallback` but just use it if supplied, similar to how the `default` kwarg works for `get_header()` et al. (note also that taking this approach would make `get_media()` more consistent with the rest of the interface).

I.e.:

```py
# This is the first time get_media is accessed. We try to read in the request body
#   and discover it is empty. We remember this fact. But then we see that
#   a value was given for when_empty_fallback, so we return that value.
a = req.get_media(when_empty_fallback=None)
assert a == None

# This time around we find that the body was previously determined to be empty.
#   So we don't try to deserialize it, and just return the fallback instead.
b = req.get_media(when_empty_fallback={})
assert b == {}

# Yet again we find that the body is known to be empty. But this time we do
#   not have a fallback value, so we need to raise an error to the app. The
#   app can then decide what to do about this situation (or just let the default
#   400 error bubble up) since by not providing a fallback value, it is implied 
#   that the client violated the API contract (a body is required).
#
# I would not expect {} to be returned here. Granted, it is an edge case to
#   have an app that is reading from req.get_media() multiple times in different 
#   ways, but when it comes to side effects, it is often prudent to be better 
#   safe than sorry.
with pytest.raises(errors.MediaNotFoundError):
    req.get_media()
```

vytas7: Reposting my thoughts from Gitter where I was basically stating the same, it feels:
> If we supported some sort of fallback on the handler level, then it would make sense that it would be materialized for subsequent invocations without explicit parameters.
> But if it is just a `req.get_media()` customization, it should be altering `get_media()`'s behaviour for that invocation, not inscribing any side effects for the remaining lifetime of the request-response cycle.

CaselIT: makes sense

CaselIT:
Updated

docs/_newsfragments/1589.breakingchange.rst
```diff
@@ -0,0 +1,17 @@
+General refactoring of internal media handler:
+
+*  Deserializing an empty body with an handler that does not support it will
```
vytas7, line 3: Oh, sorry, a late grammar nitpick... A handler. H is pronounced here.

docs/_newsfragments/1589.breakingchange.rst
```diff
@@ -0,0 +1,17 @@
...
+   raise :class:`falcon.MediaNotFoundError`, and return a
+   ``400 Bad Request`` response. This error may be suppressed by passing
+   a default value to ``get_media`` to be used in case of empty body.
+   See also :meth:`Request.get_media` for details.
+   Previously ``None`` was returned in all cases without calling the handler.
+*  Exceptions raised by the handlers are wrapped as
+   :class:`falcon.MediaMalformedError`, and return a
+   ``400 Bad Request`` response.
+*  Subsequent calls to :meth:`Request.get_media` or :attr:`Request.media` will
+   re-raise the same exception, if the fist call ended in an error, unless the
```
vytas7, line 13: _the fist (sic) call_ :arrow_right: the first call

docs/_newsfragments/1589.breakingchange.rst
```diff
@@ -0,0 +1,17 @@
...
+   exception was a :class:`falcon.MediaNotFoundError` and a default value is passed.
```
vytas7, line 14: I am not 100%, but I would add a comma here: exception was a <class>, and a default value is passed.

docs/api/media.rst
```diff
@@ -128,6 +128,30 @@ middleware. Here is an example of how this can be done:
...
+   that don't allow empty bodies as a valid value, such as ``JSON``, an instance of
+   :class:`falcon.MediaNotFoundError` should be raised. By default, this error
+   will return a  ``400 Bad Request`` response to the client.
+   This exception may be suppressed by passing a value to the ``default_when_empty``
+   argument when calling :meth:`Request.get_media`. In this case, this value will
+   be returned by the call. The same value is also returned by all following invocations.
+*  If an handler encounters an error while parsing a non-empty body, an instance of
+   :class:`falcon.MediaMalformedError` should be raised. The original exception, if any,
+   is stored in the ``__cause__`` attribute of the raised instance. By default, this
+   error will return a  ``400 Bad Request`` response to the client.
```
vytas7, line 146: This is fine, but "this error will be rendered as" would sound slightly more accurate IMHO.

docs/api/media.rst
```diff
@@ -128,6 +128,30 @@ middleware. Here is an example of how this can be done:
...
+Version 3 of Falcon updated how the handling of exceptions raised by handlers behaves:
+
+*  Falcon lets the media handler try to deserialized an empty body. For the media types
+   that don't allow empty bodies as a valid value, such as ``JSON``, an instance of
+   :class:`falcon.MediaNotFoundError` should be raised. By default, this error
+   will return a  ``400 Bad Request`` response to the client.
+   This exception may be suppressed by passing a value to the ``default_when_empty``
+   argument when calling :meth:`Request.get_media`. In this case, this value will
+   be returned by the call. The same value is also returned by all following invocations.
+*  If an handler encounters an error while parsing a non-empty body, an instance of
```
vytas7, line 143: A handler

vytas7 (changes requested):
Just a couple of grammar nitpicks to merely prove I've read this again :smiling_imp:

docs/_newsfragments/1589.breakingchange.rst
CaselIT: I think not, since I want to connect the exception with the default being passed. Alone it makes sense without: "unless the exception was a :class:`falcon.MediaNotFoundError` and a default value is passed."

vytas7: Either way is absolutely fine for me. _methinks_ though it is a compound sentence with two clauses: the **exception was** a class, and a default **value is** passed.

CaselIT: I view the compound as the before part: _if the fist call ended in an error_

vytas7: It could very well be that you are right. I normally just let @kgriffs fix stuff like this for me. :smiling_imp:

tests/test_media_urlencoded.py
```diff
@@ -37,7 +38,7 @@ def test_urlencoded_form_handler_serialize(data, expected):
 class MediaMirror:
 
     def on_post(self, req, resp):
-        resp.media = req.media
```
vytas7, line 40: I haven't checked, but have we left enough places where `media` is referenced as a property?
If not, we should IMHO, since that behaviour is still supported in parallel to `get_media()`.

tests/test_request_media.py
```diff
@@ -199,17 +211,138 @@ def test_complete_consumption(asgi):
...
+            util.async_to_sync(_check_error, req, False)
+
+    class ResAsync:
+        async def on_get(self, req, resp):
+            await _check_error(req, True)
+
+    client = create_client(asgi, resource=ResAsync() if asgi else Res())
+
+    res = client.simulate_get('/', body='')
+    assert res.status == falcon.HTTP_749
```
vytas7, line 348: This is probably fine as-is, but note that `HTTP 749` is very unlikely to work on a real ASGI server (in ASGI, the response status is provided as an integer status code).

vytas7 (approved):
:+1:

kgriffs (approved):
LGTM

tests/test_media_urlencoded.py
CaselIT: Yes, there are some tests that are specific to this in test_request_media

tests/test_request_media.py
CaselIT: that makes me sad 😂

vytas7: Without a doubt, that is a glaring deficiency in that spec.

Issue #1607, opened 2019-11-11
Errors in response serialization are not handled


vytas7:
I found this when testing @kgriffs' ASGI branch, but this is an issue in all (recent) Falcon version, including the (at the time of writing) stable 2.0.

Errors in response serialization are not handled, for example:
* Missing media handler for the given [response content-type](https://falcon.readthedocs.io/en/stable/api/request_and_response.html#falcon.Response.content_type) results in an [HTTPUnsupportedMediaType](https://falcon.readthedocs.io/en/stable/api/errors.html#falcon.HTTPUnsupportedMediaType).
* Any exceptions raised in [media serialization](https://falcon.readthedocs.io/en/stable/api/media.html#falcon.media.BaseHandler.serialize).

This is somewhat more important now with the recently merged https://github.com/falconry/falcon/issues/1507. The user might expect that at least the generic handler catches the cases above. They would be instead propagated to the WSGI server.

kgriffs:
It does seem like `HTTPUnsupportedMediaType` should be converted to an appropriate 415 response, and in the general case should be caught by the generic handler. I think we ought to fix this for 3.0 or 3.1.

vytas7:
Scheduling this tentatively for 3.0 then, although we may need to push it to 3.1 or 4.0.

vytas7:
Adding **blocked-by-release** while blocking on the ASGI branch (it makes sense to sync this change for both App flavours).

vytas7:
I cannot reproduce this bug, it looks like it was fixed at some point (~2020 Jan).
I'll just add a couple of more explicit regression tests as per the issue's description.

PR #1768, opened 2020-08-29
Normalize json dumps

### LINKED ISSUE #1767: Configure json dumps in a single place
Falcon lets the user customize the json serializer used to serialize the media property of a response, but HttpError and SSEvent do not use the user provided serializer, using instead the default python json one.

CaselIT:
# Summary of Changes
Always use the configured media handler for json when loading or serializing it.

# Related Issues
Closes #1767

# Pull Request Checklist


falcon/app_helpers.py
```diff
@@ -183,7 +184,8 @@ def default_serialize_error(req, resp, exception):
 
     if preferred is not None:
         if preferred == 'application/json':
-            resp.body = exception.to_json()
+            dumps = getattr(resp.options.media_handlers.get(MEDIA_JSON), 'dumps', None)
```
vytas7, line 187: This is a fair bit of dark magic that IMHO would benefit from a docstring or a `NOTE` explaining how this works in different scenarios: `MEDIA_JSON` is not in `media_handlers`, `MEDIA_JSON` is in media handlers, but it is a custom handler not employing `dumps` etc.

Furthermore, this can potentially lead to hard-to-debug scenarios where the handler does define `dumps`, but the signature is not as we expect here, for instance, it requires an extra indentation level parameter, or what not. We should at minimum discuss this (the special meaning of `dumps`) in the custom handlers docs.

Alternatively, we could be checking the handler type, i.e. if it is strictly of type `JSONHandler`, or a subclass of it (?).

docs/_newsfragments/1767.misc.rst
```diff
@@ -0,0 +1,4 @@
+Use ``Response`` media handlers when serializing to json ``HTTPError``
```
vytas7, line 1: Towncrier newsfragments are not subject to the docstring conventions; they should IMHO be worded as news items, telling what has happened.

vytas7 (changes requested):
Admittedly, I have some misgiving about the proposed approach.

From a comment above:
> Furthermore, this can potentially lead to hard-to-debug scenarios where the handler does define `dumps`, but the signature is not as we expect here, for instance, it requires an extra indentation level parameter, or what not. We should at minimum discuss this (the special meaning of `dumps`) in the custom handlers docs.

What is more, https://github.com/falconry/falcon/pull/1741 is proposing to add even more ways to handle media, directly (de)serializing from(to) a bytestring as opposed to parsing the request stream.

That distinction (stream/binary string) is probably worth keeping for performance reasons. But maybe we could somehow allow using bytestring handlers for requests too, shimming `stream.read()` where applicable?
The bytestring kind of handler could be also applied not only to WebSockets, but also to SSE, error serialization etc, as also prototyped in this PR.

These are just random thoughts off the top of my head; unfortunately I don't have a one design proposal to rule them all (at least not yet :smiling_imp: )

The way to deprecate stuff at the module level on 3.7+ is cool though, we could actually use it at more places if/when needed.

CaselIT:
> The way to deprecate stuff at the module level on 3.7+ is cool though, we could actually use it at more places if/when needed.

Indeed. It was one of the reason why it was added. [PEP562](https://www.python.org/dev/peps/pep-0562/)

Regarding the rest, I guess we should wait for websocket to land and the update this one.

falcon/app_helpers.py
CaselIT: I think we could just document this in media, specifying that loads and dumps of the json handler if exits and has that attributes are used by the application to handler (de)serialization of json.

vytas7:
One more random thought on unifying serialization. Ideally, the default error serializer should also use response media handlers; so if one is, for instance, building a YAML or Msgpack API, the error structures should by default also pick the default media type set for the API.

Some open questions:
* How could this look for XML?
* How would an error while serializing an exception look like? (See also: https://github.com/falconry/falcon/issues/1607 and [`HTTP 776`](https://github.com/joho/7XX-rfc) )

CaselIT:
This should maybe wait for https://github.com/falconry/falcon/pull/1741

falcon/app_helpers.py
kgriffs: > but the signature is not as we expect here

There are at least three edge cases to address, and to ensure are documented for the developer:

1. A JSON handler is not defined at all. 
2. A JSON handler is defined using ``JSONHandler`` but the dumps/loads require parameters. (We should be OK on this one already since the docs talk about it and recommend ``functools.partial``).
3. A JSON handler is defined but it does not derive from ``JSONHandler``.  In this case it is possible that dumps/loads attributes are defined but not as expected. (The code already falls back to stdlib if dumps/loads are not found... alternatively would it make sense to use the `serialize()` method, although that would require extra branches in the logic since the return type is different?)

TBH I think it would be a more prudent design in the abstract to stick with the ``BaseHandler`` serialize/deserialize interface methods if it doesn't make things too convoluted. That would prevent us from coupling too tightly with the specific ``JSONHandler`` implementation.

docs/_newsfragments/1767.misc.rst
kgriffs: Capitalize JSON

falcon/app_helpers.py
```diff
@@ -183,7 +184,8 @@ def default_serialize_error(req, resp, exception):
...
+            resp.body = exception.to_json(dumps)
         else:
             # NOTE(caselit): to_xml already returns bytes
             resp.data = exception.to_xml()
```
kgriffs, line 191: Just a reminder to make sure the docs point out that this behavior only applies to JSON.

kgriffs (changes requested):

falcon/app_helpers.py
CaselIT: In the end I've opted to use the handler directly

falcon/__init__.py
```diff
@@ -50,7 +50,7 @@
 from falcon.response import Response, ResponseOptions  # NOQA
 
 
-ASGI_SUPPORTED = _sys.version_info.minor > 5
+ASGI_SUPPORTED = _sys.version_info >= (3, 6)
```
CaselIT, line 53: let's support python 4 🚀

falcon/asgi/structures.py
```diff
@@ -138,7 +149,10 @@ def serialize(self):
         elif self.text is not None:
             block += 'data: ' + self.text + '\n'
         elif self.json is not None:
-            block += 'data: ' + json_dumps(self.json, ensure_ascii=False) + '\n'
+            if handler is None:
+                handler = _DEFAULT_JSON_HANDLER
+            string = handler.serialize(self.json, MEDIA_JSON).decode()
```
CaselIT, line 154: this is a bit of side effect. encore -> decode -> encode again

falcon/http_error.py
```diff
@@ -204,6 +212,11 @@ def to_xml(self):
                 et.tostring(error_element, encoding='utf-8'))
 
 
+# NOTE: initialized in falcon.media.json, that is always imported since Request/Respose
+# are imported by falcon init.
+_DEFAULT_JSON_HANDLER = None
```
CaselIT, line 217: circular import workaround

falcon/request.py
```diff
@@ -1739,9 +1741,17 @@ def get_param_as_json(self, name, required=False, store=None, default=None):
...
+            MEDIA_JSON, MEDIA_JSON, raise_not_found=False
+        )
+        if handler is None:
+            handler = _DEFAULT_JSON_HANDLER
+
         try:
-            val = json.loads(param_value)
-        except ValueError:
+            val = handler.deserialize(
+                BytesIO(param_value.encode('utf-8')), MEDIA_JSON, len(param_value)
```
CaselIT, line 1752: sadly this is a bit akward

tests/test_media_handlers.py
```diff
@@ -205,11 +205,11 @@ async def on_post(self, req, resp):
...
 
-    result = testing.simulate_get(app, '/')
-    assert result.status_code == 500
+    with pytest.raises(NotImplementedError):
+        testing.simulate_get(app, '/')
 
-    result = testing.simulate_post(app, '/', json={})
-    assert result.status_code == 500
+    with pytest.raises(NotImplementedError):
+        testing.simulate_post(app, '/', json={})
```
CaselIT, line 212: I'm not sure if error.to_json should better guard for this case.

CaselIT (commented):
I've removed the magic, opting to use the handler directly. But it's has some awkward cases

tests/test_media_handlers.py
```diff
@@ -205,10 +205,11 @@ async def on_post(self, req, resp):
 
     app.add_route('/', ResourceAsync() if asgi else Resource())
 
-    result = testing.simulate_get(app, '/')
+    # NOTE(caselit): force serialization in xml, since error.to_json uses the faulty handler
+    result = testing.simulate_get(app, '/', headers={'Accept': 'text/xml'})
```
CaselIT, line 209: I'm not sure if error.to_json should better guard for this case.

falcon/__init__.py
vytas7: Lol I already had this type of comment elsewhere :)

tests/test_media_handlers.py
CaselIT: This was already pointed out by @vytas7 here https://github.com/falconry/falcon/pull/1768#issuecomment-683310866

falcon/app.py
```diff
@@ -766,16 +766,13 @@ def set_error_serializer(self, serializer):
             def my_serializer(req, resp, exception):
                 representation = None
 
-                preferred = req.client_prefers(('application/x-yaml',
-                                                'application/json'))
+                preferred = req.client_prefers((MEDIA_YAML, MEDIAL_JSON))
```
vytas7, line 769: I cannot find any definition for `MEDIAL_JSON`. Is this a typo of `MEDIA_JSON`?

vytas7, line 769: It might also make sense to refer to these constants using the `falcon.` namespace, as otherwise we are implying that the reader has to `from falcon import *`.

tests/test_utils.py
```diff
@@ -1070,3 +1072,12 @@ def a_function(a=1, b=2):
         assert len(recwarn) == 0
         a_function(1, b=2)
         assert len(recwarn) == 1
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason='module __getattr__ requires python 3.7')
```
vytas7, line 1077: Great to have extensive testing here. :+1:

tests/test_media_handlers.py
```diff
@@ -262,3 +263,25 @@ async def on_post(self, req, resp):
...
+
+    class Resource:
+        def on_get(self, req, resp):
+            raise falcon.HTTPForbidden()
+
+    class ResourceAsync:
+        async def on_get(self, req, resp):
+            raise falcon.HTTPForbidden()
+
+    app.add_route('/', ResourceAsync() if asgi else Resource())
```
vytas7, line 283: We would automatically wrap sync resources for async routes, as explained on another PR.
So we could DRY this a bit.

falcon/asgi/structures.py
```diff
@@ -111,7 +112,17 @@ def __init__(
...
         self.comment = comment
 
-    def serialize(self):
+    def serialize(self, handler=None):
+        """Serialize this event to string.
+
+        Args:
+            handler: Handler object that will be used to serialize the ``json`` attribute to
+                string. When not provided a default handler using the python builtin
+                json library will be used (default ``None``).
```
vytas7, line 121: Capitalize JSON here (and elsewhere).

falcon/asgi/structures.py
```diff
@@ -111,7 +112,17 @@ def __init__(
 
         self.comment = comment
 
-    def serialize(self):
+    def serialize(self, handler=None):
+        """Serialize this event to string.
+
+        Args:
+            handler: Handler object that will be used to serialize the ``json`` attribute to
+                string. When not provided a default handler using the python builtin
```
vytas7, line 120: Capitalize Python, or just remove it altogether IMHO. The reader will know which language we are targeting here.

vytas7, line 120: Suggested punctuation:
> When not provided, <...>

vytas7 (changes requested):
Some quick documentation nitpicks off the top of my head.
Still trying to wrap my head around all the design implications.

tests/asgi/test_sse.py
```diff
@@ -182,6 +183,49 @@ async def _test():
...
+        result = client.simulate_get()
+        assert result.text == (
+            'data: {"FOO": "BAR"}\n'
+            '\n'
+            'data: {"BAR": "BAZ"}\n'
+            '\n'
+        )
+
+    def test_no_json_media_handler(self, client):
+        client.app.resp_options.media_handlers.pop(falcon.MEDIA_JSON)
```
vytas7, line 218: I find it hard to reason about the effects of this `pop` before https://github.com/falconry/falcon/pull/1785 is merged.
But given the test passes, we can probably leave it as-is.

vytas7 (changes requested):

falcon/asgi/structures.py
CaselIT: I meant the library name, but I'll replace with JSON

falcon/http_error.py
```diff
@@ -165,16 +166,23 @@ def to_dict(self, obj_type=dict):
...
 
         Returns:
-            str: A JSON document for the error.
+            bytes: A JSON document for the error.
 
         """
 
         obj = self.to_dict(OrderedDict)
-        return json.dumps(obj, ensure_ascii=False)
+        if not handler:
```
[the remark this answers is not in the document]
CaselIT: not sure if it's faster or not. but it's probably easier to read

.gitignore
```diff
@@ -62,3 +62,4 @@ dash
 #py Environment
 /venv
 /env
+docs/_built
```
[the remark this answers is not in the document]
CaselIT: That's was not supposed to be committed

falcon/http_error.py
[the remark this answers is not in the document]
CaselIT: nevermind, I prefer the if version since it's easier to see if both branches are taken. coverage does not consider the or one a branch iirc

falcon/asgi/structures.py
CaselIT: used return early in this case

falcon/http_error.py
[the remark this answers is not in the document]
vytas7: On consideration, I agree.
FWIW, I find the `or` pattern equally easier to read, and more performant. **However,** if a custom media handler is defined as _falsey_ for some strange reason, it would then unexpectedly fallback to the default handler, and it's a surprise we want to avoid.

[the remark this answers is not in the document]
CaselIT: regarding performance there is actually no apparent difference. it's basically a wash.
regarding the falsey it's the same with the if, I'm just doing `not handler`, not `handler is not None` 

personally I don't think it's worth another change :)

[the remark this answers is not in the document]
vytas7: I.e. I prefer `if handler is None` due the falsey object issue.

tests/test_media_handlers.py
CaselIT: so this would work also without the async resource?

falcon/http_error.py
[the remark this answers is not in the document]
CaselIT: ok I guess. A lot of places do use `if obj` or `if not obj` or `obj or other` so with this in mind maybe a more general refactoring is in order?

tests/test_media_handlers.py
CaselIT: This probably should be solved in #1607.

falcon/http_error.py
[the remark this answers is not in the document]
vytas7: It depends what that `obj` is. If it is something we can control or at least expect a specific type (e.g., primitive built-ins or types defined by us), a boolean test is fine.

vytas7 (changes requested):
Thanks @CaselIT, this is a great effort to unify these scattered handlers, enabling reuse of the same media handler interface.

Unfortunately, it seems that we need to rethink our designs here before we can move forward.

Currently, both your HTTP error serializer and `SSEvent` JSON serializer would simply try to call `handler.serialize(<data>, MEDIA_JSON)`.
However, the ASGI flavour of the framework primarily relies not on `serialize()`, but instead, [`serialize_async()`](https://falcon.readthedocs.io/en/latest/api/media.html#falcon.media.BaseHandler.serialize_async) is `await`ed (although the latter can piggyback on the former, if not defined explicitly).

vytas7:
For reference, reposting some random thoughts of the top of my head from [Gitter](https://gitter.im/falconry/dev?at=5fc29b6cf7e4510fcef788c5):
> I was basically thinking that we would have two types of handlers:
> * Stream handlers, much like the current BaseHandler; the ASGI variant awaits serialization, the WSGI counterpart obviously does not.
> * Binary data handlers, not dissimilar to the current BinaryBaseHandlerWS. Always sync.
>
> The latter would handle WebSocket media, SSE, and error serialization. Error serialization probably being the most questionable one whether it is enough.
> However, in case more complex logic is needed, one could override the awaitable custom error serializer (is it awaitable now?..).

CaselIT:
> 
> 
> Thanks @CaselIT, this is a great effort to unify these scattered handlers, enabling reuse of the same media handler interface.
> 
> Unfortunately, it seems that we need to rethink our designs here before we can move forward.
> 
> Currently, both your HTTP error serializer and `SSEvent` JSON serializer would simply try to call `handler.serialize(<data>, MEDIA_JSON)`.
> However, the ASGI flavour of the framework primarily relies not on `serialize()`, but instead, [`serialize_async()`](https://falcon.readthedocs.io/en/latest/api/media.html#falcon.media.BaseHandler.serialize_async) is `await`ed (although the latter can piggyback on the former, if not defined explicitly).

I've updated to docs to require that the json handler have the sync implementation even for asgi application.

While inconvenient, I don't think it would cause to much of an hassle to users, considering that we also provide a falcon implementation.

kgriffs:
I agree that this approach shouldn't be a big burden on the developer since the differences between the two methods will me small and easy to maintain. Also, most people will just use the default media handlers and not even think about it. 

Noting this requirement in the docs is a good idea. We should also make sure that we raise a user-friendly error if the method is not implemented on the handler.

kgriffs (changes requested):
Just some doc nits. I still need to review the implementation.

CaselIT:
documentation updated @kgriffs

vytas7:
> Noting this requirement in the docs is a good idea. We should also make sure that we raise a user-friendly error if the method is not implemented on the handler.

Could we maybe fallback to the default handler upon `NotImplementedError`, and document this behaviour? It ain't pretty but FAPP it would be quite robust.

CaselIT:
> 
> 
> > Noting this requirement in the docs is a good idea. We should also make sure that we raise a user-friendly error if the method is not implemented on the handler.
> 
> Could we maybe fall back to the default handler upon `NotImplementedError`, and document this behaviour? It ain't pretty but FAPP it would be quite robust.

that may also be an option

falcon/media/base.py
```diff
@@ -16,14 +18,26 @@ def serialize(self, media, content_type):
...
             content_type (str): Type of response content.
 
         Returns:
             bytes: The resulting serialized bytes from the input object.
         """
-        raise NotImplementedError()
+        if MEDIA_JSON in content_type:
+            raise NotImplementedError(
+                'The JSON media handler requires the sync interface to be implemented even in '
+                "ASGI applications, because it's used internally by the falcon."
```
kgriffs, line 37: missing "framework" at the end of this sentence?

kgriffs:
> Could we maybe fall back to the default handler upon NotImplementedError

I kind of think we should just go ahead and raise it to the app just to be sure the developer isn't getting a result they don't expect (i.e., they thought we were calling their thing when all along we weren't but we were silently falling back to something else).

falcon/request.py
CaselIT: @kgriffs should I remove the 'uft-8' here to make it in-line with the other pr?

vytas7:
That's a good point @kgriffs .
I had this as an idea to offer improved damage control, but after all, it's an edge case that shouldn't be seen to often. As such, it's probably better to be explicit indeed.

falcon/request.py
CaselIT: i did

tests/test_media_handlers.py
CaselIT: i guess that this is what's the env variable `FALCON_ASGI_WRAP_NON_COROUTINES` is for

CaselIT:
updated

falcon/request.py
```diff
@@ -1739,9 +1741,19 @@ def get_param_as_json(self, name, required=False, store=None, default=None):
...
+        handler = self.options.media_handlers.find_by_media_type(
+            MEDIA_JSON, MEDIA_JSON, raise_not_found=False
+        )
+        if handler is None:
+            handler = _DEFAULT_JSON_HANDLER
+
         try:
-            val = json.loads(param_value)
-        except ValueError:
+            # TODO(CaselIT): find a way to avoid encode + BytesIO if handlers
```
kgriffs, line 1751: Yes, this will ding performance, but hopefully it isn't used often enough and the overhead is not great enough to pose a serious problem. But maybe down the road if we unity handling interfaces with the WS handlers by including a deserialize from `bytes` this can be resolved.

kgriffs (changes requested):
LGTM sans that error message nitpick.

falcon/asgi/structures.py
```diff
@@ -134,11 +145,15 @@ def serialize(self):
...
-            block += 'data: ' + self.data.decode() + '\n'
+            block += f'data: {self.data.decode()}\n'
         elif self.text is not None:
-            block += 'data: ' + self.text + '\n'
+            block += f'data: {self.text}\n'
         elif self.json is not None:
-            block += 'data: ' + json_dumps(self.json, ensure_ascii=False) + '\n'
+            if handler is None:
+                handler = _DEFAULT_JSON_HANDLER
+            string = handler.serialize(self.json, MEDIA_JSON)
```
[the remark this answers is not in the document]
CaselIT: ah indeed, since they are bytes :)

falcon/request.py
CaselIT: yes, that's an option. I guess this is one thing to consider for v4.

CaselIT:
Updated, I think ready for final review

falcon/request.py
vytas7: I'll try to channel my thoughts into a proposal for 3.x or 4.0: https://github.com/falconry/falcon/pull/1768#issuecomment-738896304

kgriffs (approved):

falcon/media/base.py
```diff
@@ -71,7 +90,13 @@ def deserialize(self, stream, content_type, content_length):
         Returns:
             object: A deserialized object.
         """
-        raise NotImplementedError()
+        if MEDIA_JSON in content_type:
+            raise NotImplementedError(
```
vytas7, line 94: That's actually a neat way to provide a meaningful hint. :+1:

vytas7 (changes requested):
LGTM with one theoretical edge case. Not sure if it is too exotic, or it should be documented.
But if one has such a custom JSON handler that features very sparse output possibly containing `\n\n`, that might break the SSE encapsulation.

Regardless, we also need to restore 100% coverage before merging this. As I understand, a branch is not fully covered here: https://codecov.io/gh/falconry/falcon/pull/1758/diff#D4-2385

vytas7 (approved):
:+1: (bar the edge case comment above)

CaselIT:
> 
> 
> LGTM with one theoretical edge case. Not sure if it is too exotic, or it should be documented.
> But if one has such a custom JSON handler that features very sparse output possibly containing `\n\n`, that might break the SSE encapsulation.

oh, that's quite an edge case. `print(json.dumps({'x':'a'}, indent='\n\n\n'))`
Should we strip `\n\n` from the returned object?

vytas7:
Maybe a documentation note is enough?

PR #1770, opened 2020-08-29
chore: allow running pytest directly without tox

### LINKED ISSUE #1652: Improve test readme. Without setting the proper env variable 300+ failure occur

CaselIT:
# Summary of Changes

Add `FALCON_ASGI_WRAP_NON_COROUTINES` and `FALCON_TESTING_SESSION` env variables to the setput.cfg configuration of pytest, so that the test can be run also by calling `pytest` directly without tox

# Related Issues

closes  #1652


setup.cfg
```diff
@@ -14,3 +14,9 @@ filterwarnings =
     ignore:"@coroutine" decorator is deprecated:DeprecationWarning
     ignore:Using or importing the ABCs:DeprecationWarning
     ignore:cannot collect test class 'TestClient':pytest.PytestCollectionWarning
+
+env =
+    FALCON_ASGI_WRAP_NON_COROUTINES = Y
+    FALCON_TESTING_SESSION = Y
+    # PYTHONASYNCIODEBUG is optional
+    PYTHONASYNCIODEBUG = 1
```
vytas7, line 22: If `PYTHONASYNCIODEBUG` is optional, maybe we could resist the temptation to set it by default?
It adds horrible overhead (50x and more in some cases) to asyncio tests.

vytas7 (dismissed):
:+1: 

I just had one open question regarding `PYTHONASYNCIODEBUG`, but I do not have a strong preference myself.

setup.cfg
CaselIT: I left it here because it is set in tox.ini. But it's optional in the sense that I use it without it and everything works.
I can comment it

vytas7: I'm fine with it either way.
But it probably does make sense to comment it out because just running `pytest` does not necessarily do other things `tox` does either. It just does what it does -- runs `pytest`.

vytas7 (approved):
:+1:

Issue #1749, opened 2020-08-01
Response.downloadable_as does not follow RFC 6266 recommendations for UTF-8


vytas7:
At the time of writing, [Response.downloadable_as](https://falcon.readthedocs.io/en/stable/api/request_and_response.html#falcon.Response.downloadable_as) always uses the `attachment; filename="filename.ext"` form regardless of the provided filename. However, [RFC 6266](https://tools.ietf.org/html/rfc6266) specifies that a quoted `filename` shall be interpreted as ISO-8859-1 (although some agents are able to sniff UTF-8), and recommends using `filename*` whenever `filename` cannot be faithfully represented with US-ASCII.

CaselIT:
I believe we have to escape a comma `,` because chrome does not like this.

The following example breaks on chrome, (but works on firefox)
```py
import falcon

class R:
    def on_get(self, req,res):
        res.text = '123'
        res.downloadable_as = '≥,.txt'

app = falcon.App()
app.add_route('/try', R())
```

PR #1775, opened 2020-09-14
fix(Response.downloadable_as): properly handle comma in unicode filenames

CaselIT:
When `dowloadable_as` was updated to handle unicode in filenames in https://github.com/falconry/falcon/pull/1750 a bug was introduced when the file has unicode character and a has comma in them.

This fixes the bug by escaping also the commas.

Closes https://github.com/falconry/falcon/issues/1749 (again)


vytas7 (approved):

vytas7:
Nice catch! :+1:

PR #1792, opened 2020-12-06
Try fixing wheel builds

CaselIT:
# Summary of Changes

This pr should fix the errors in https://github.com/falconry/falcon/actions/runs/393277790

In particular:
- install setuptools >= 44 everywhere
- do not install daphne on windows since it's not used and twisted seem not to compile on py39
- pin `cryptography<3.2` to avoid errors due to old version of open-ssl

### DIFF
--- a/requirements/tests
+++ b/requirements/tests
@@ -15,7 +15,7 @@ uvicorn; python_version >= '3.6'
 websockets; python_version >= '3.6'
 
 # Handler Specific
-cbor2
+cbor2; python_version >= '3.6'
 msgpack
 mujson
 ujson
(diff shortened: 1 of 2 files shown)


requirements/tests
```diff
@@ -9,7 +9,7 @@ pytest-asyncio
 httpx; python_version >= '3.6'
 uvicorn; python_version >= '3.6'
 aiofiles; python_version >= '3.6'
-daphne; python_version >= '3.6'
+daphne; python_version >= '3.6' and sys_platform != 'win32'
```
vytas7, line 12: We should also add `python_version >= '3.6'` to the `cbor` requirement below, IIRC it is only tested in the WebSocket context.

vytas7 (approved):
Thanks for this :+1:

vytas7 (approved):

kgriffs (approved):

CaselIT:
I think we miss disabling daphne in s390x since it causes some strange problems.

PR #1796, opened 2020-12-07
chore: use tox to run wheel tests. make daphne optional when testing

### LINKED ISSUE #1795: Do not test daphne by default
Daphne is not too reliable in the test, and causes frequent failures to the ci.

A custom tox environment could be used to thest daphne only on one environment

CaselIT:
# Summary of Changes

- Move daphne to a tox env
- Use tox to check the wheels reducing duplication

# Related Issues

fixes #1795

### DIFF
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -65,11 +65,8 @@ jobs:
           FALCON_ASGI_WRAP_NON_COROUTINES: Y
           FALCON_TESTING_SESSION: Y
         run: |
-          pip install -f dist --no-index falcon
-          python -c 'import sys; sys.path.pop(0); from falcon.cyutil import misc, reader, uri'
-          pip install 'setuptools>=44'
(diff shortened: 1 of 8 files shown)


CaselIT:
This is the updated run of the pubhish wheel: https://github.com/CaselIT/falcon/runs/1513947887

All will fail at publish, since it does not have the correct token, the important bit is that they arrived at the publish step

CaselIT:
Making the x390x gate happy was not easy, but I think I managed it.

vytas7 (approved):
> the workaround is to first install ujson by itself with pip, then use tox so that it uses the cached wheel.

If the mountain will not come to Muhammad... :see_no_evil:

CaselIT:
> > the workaround is to first install ujson by itself with pip, then use tox so that it uses the cached wheel.
> 
> If the mountain will not come to Muhammad... :see_no_evil: 

My next option was to remove the mountain...

kgriffs (approved):

PR #1814, opened 2020-12-13
fix(ws): msgpack import error not handled

kgriffs:
# Summary of Changes

This patch modifies the default websocket binary media handler to not require msgpack upon instantiation.


falcon/media/msgpack.py
```diff
@@ -83,18 +83,33 @@ class MessagePackHandlerWS(BinaryBaseHandlerWS):
...
+            raise RuntimeError(
+                'The default WebSocket media handler for BINARY payloads requires '
+                'the msgpack package, which is not installed.'
+            )
+
         return self.packer.pack(media)
 
     def deserialize(self, payload: bytes) -> object:
+        if not self.msgpack:  #
+            # TODO(kgriffs): There is probably a more elegant way to handle this situation
```
kgriffs, line 107: We can revisit this later; hopefully this is "good enough" for 3.0

vytas7 (changes requested):
LGTM, but we need to actually submit coverage from a msgpack-free environment in _Actions_ too (or else we wouldn't get 100% coverage) before merging.

CaselIT:
I'm not too convinced by this approach. Why don't we create a fake handler that is used if there is no msgpack?
This fake handler would just raise an error.
When creating the default handlers we would then install this fake handler if there is no msg pack.

Something like:

```py

class MissingRequirementHandler:
  def __init__(self, library: str):
    self. library = library
  def _raise(self, *args, **kwargs):
    raise RuntimeError(f'Missing required library {self.library}')
  serialize = deserialize = _raise
```

kgriffs:
> LGTM, but we need to actually submit coverage from a msgpack-free environment in _Actions_ too (or else we wouldn't get 100% coverage) before merging.

Nice catch, I forgot to update GH actions. Will do.

kgriffs:
>  we would then install this fake handler 

That would be the more elegant approach. 👍

CaselIT:
Btw if we want to target 3.1 for the nicer approach, since it would not be a breaking change, I'm fine with this implementation for now :)

vytas7 (approved):
:+1:

CaselIT (approved):

PR #1825, opened 2020-12-20
[PoC] [WiP Draft] Performance gate prototype

### LINKED ISSUE #1450: Add basic performance testing to the gate
Create a basic performance test based on historical trends and add it to as a Travis job. We should consider failing the build in the face of a performance regression.

Our first pass at this can be very basic; we can always improve it over time.

vytas7:
This is an early prototype of how implementing the https://github.com/pythonspeed/cachegrind-benchmarking approach could look like for Falcon.

The prototype builds upon this excellent article by @itamarst: https://pythonspeed.com/articles/consistent-benchmarking-in-ci/ (thanks @njsmith for kindly pointing to this idea).

Closes #1450

To do if we want to proceed with this:
- [x] ASGI performance metric
- [x] Barebones "Hello, World!" performance metric
- [x] Media performance metric
- [ ] Routing performance metric
- [x] URL params and headers performance metric
- [ ] Clean up `tox` environments, requirements etc
- [ ] Add Cython support
- [ ] Run Cython gates for master only (?)
- [ ] Add PyPy support, master only (?); probably only informational?

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Set up Python
         uses: actions/setup-python@v2.1.4
-        if: ${{ matrix.toxenv != 'py35' && matrix.toxenv != 'perf_hello' && matrix.toxenv != 'perf_media' }}
+        if: ${{ matrix.toxenv != 'py35' && matrix.toxenv != 'perf_asgi' && matrix.toxenv != 'perf_hello' && matrix.toxenv != 'perf_media' && matrix.toxenv != 'perf_query' }}
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -103,7 +103,7 @@ jobs:
           python-version: 3.8
 
       - name: Set up Python 3.8 (Ubuntu 20.04 build)
-        if: ${{ matrix.toxenv == 'perf_hello' || matrix.toxenv == 'perf_media' }}
+        if: ${{ matrix.toxenv == 'perf_asgi' || matrix.toxenv == 'perf_hello' || matrix.toxenv == 'perf_media' || matrix.toxenv == 'perf_query' }}
         run: |
           sudo apt-get update
           sudo apt-get install -y build-essential curl python3.8 python3.8-dev python3.8-distutils python-is-python3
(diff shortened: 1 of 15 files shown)


itamarst:
Excited to see this, and to see if it turns out to be useful. I'm going to be setting up something similar for one of my projects, so will share what I come up with if this is still in progress, or maybe steal from you, depending :grinning:

itamarst:
Note that I've found a bug in the cachegrind.py script calculation, so you'll want to pull a new version once I've updated it (tomorrow hopefully).

vytas7:
Heh, thanks for heads up @itamarst !
I did experience that the `cachegrind.py` metric was a lot (1-2 orders of magnitude) noisier than the instruction count from `valgrind`, if that is what would be revised.
But maybe it _is_ a bit noisier in its nature, so I didn't pay to much attention to it.

vytas7:
@itamarst Btw, another thing I was really surprised when toying with this approach, was how much "rogue" Python hash seeds can affect performance :slightly_smiling_face:

itamarst:
Yeah, you really want to set a fixed PYTHONHASHSEED for consistency.

The issue was that I was counting L3 hits wrong, if something hit RAM I also counted this as hitting L3 (i.e. L3 hits were too high). I am not sure if this will have much impact on the noisiness though.

itamarst:
OK, https://github.com/pythonspeed/cachegrind-benchmarking has been updated.

vytas7:
Thanks @itamarst , I'll check that out!

vytas7:
@itamarst thanks again for the update.
The noisiness is gone, and the variation of the least-squares linear regression fitting error is now low beyond belief :100: .
In fact, I'm now getting exactly the same cost of one iteration in two different CI runs (within at least 9 significant digits!).

itamarst:
I got benchmarks working. Beyond what is in original repository:

1. I store results as pretty-printed, sorted JSON file on disk. Expectation is that developer runs this locally and checks in result, but I'm single-person project.
2. On every PR, I run the benchmarks again and add diff as GitHub comment: https://github.com/pythonspeed/filprofiler/blob/master/.github/workflows/main.yml#L120
3. In addition to setting `PYTHONHASHSEED` and figuring out equivalent fixed-seed for Rust, I ended up using Conda environments to reduce noise between my machine and GitHub Actions machines (so e.g. it's the same Python binary, rather than different dot-versions or different compilers etc.). The result is quite consistent on my machine, and a little noisy between my machine and VMs, but better than it would be without Conda.

Example output: https://github.com/pythonspeed/filprofiler/pull/110

vytas7:
This has been rotting for so long due to the lack of my bandwidth, that I'm thinking to wait a couple of weeks more and then migrate this to Ubuntu 22.04 + CPython 3.10, to have the same gauge for a longer time.

PR #1832, opened 2020-12-22
doc(request): clarify use of csv option

kgriffs:


CaselIT (dismissed):

falcon/request.py
```diff
@@ -1606,12 +1617,19 @@ def get_param_as_list(self, name, transform=None,
 
         Returns:
             list: The value of the param if it is found. Otherwise, returns
-            ``None`` unless required is True. Empty list elements will be
-            discarded. For example, the following query strings would
-            both result in `['1', '3']`::
+            ``None`` unless *required* is ``True``. Empty list elements will be
+            discarded by default, but this can be configured by setting the
+            :attr:`~falcon.RequestOptions.keep_blank_qs_values` option. For
```
vytas7, line 1622: AFAICT this is not true:
```python
>>> import falcon
>>> import falcon.testing
>>> app = falcon.App()
>>> def debug(req, resp):
...     resp.media = req.get_param_as_list('ids')
... 
>>> app.req_options.auto_parse_qs_csv = True
>>> app.add_sink(debug, '/test')
>>> 
>>> client = falcon.testing.TestClient(app)
>>> resp = client.simulate_get('/test/instances?active=true&ids=1000,,,100,123456')
>>> resp.json
['1000', '', '', '100', '123456']
```

The default value of [`falcon.RequestOptions.keep_blank_qs_values`](https://falcon.readthedocs.io/en/latest/api/app.html#falcon.RequestOptions.keep_blank_qs_values) is `True`; empty list elements will be kept by default.

vytas7 (changes requested):
:+1: 

Looks good, barring one factual error that needs to be addressed before merging.

falcon/request.py
kgriffs: Oh, nice catch. Maybe we should consider changing the default? Seems like normally you would not want those and `parse_query_string` defaults to ``False``.

kgriffs: Fixed the docstring to be correct.

CaselIT (approved):

vytas7 (approved):
👍

PR #1777, opened 2020-10-12
docs(Request): clarify auto_parse_qs_csv behavior for splitting

danilito19:
# Summary of Changes

Clarify documentation for `auto_parse_qs_csv` to indicate that only the first set of values for a given key are split by comma and subsequent values are left as is.

# Related Issues

Closes #1597

# Pull Request Checklist

This is just a reminder about the most common mistakes.  Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](../CONTRIBUTING.md) at least once; it will save you a few review cycles!

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing to do.

- [ ] Applied changes to both WSGI and ASGI code paths and interfaces (where applicable).
- [ ] Added **tests** for changed code.
- [ ] Prefixed code comments with GitHub nick and an appropriate prefix.
- [ ] Coding style is consistent with the rest of the framework.
- [ ] Updated **documentation** for changed code.
    - [ ] Added docstrings for any new classes, functions, or modules.
    - [ ] Updated docstrings for any modifications to existing code.
    - [ ] Updated both WSGI and ASGI docs (where applicable).
    - [ ] Added references to new classes, functions, or modules to the relevant RST file under `docs/`.
    - [ ] Updated all relevant supporting documentation files under `docs/`.
    - [ ] A copyright notice is included at the top of any new modules (using your own name or the name of your organization).
    - [ ] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/directives.html?highlight=versionadded#directive-versionadded).
- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/index.html) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)

If you have *any* questions to *any* of the points above, just **submit and ask**! This checklist is here to *help* you, not to deter you from contributing!

*PR template inspired by the attrs project.*

PR #1833, opened 2020-12-22
fix(parse_query_string): only the first set of values for the same key are are split

kgriffs:
Address #1777

To Do:

- [x] Decide if this is a good idea. 😄 
- [x] Add news fragment(s)
- [x] Port fix to pyx implementation
- [x] Ensure full test coverage

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1946,8 +1946,8 @@ class RequestOptions:
 
             When `auto_parse_qs_csv` is set to ``True``, the query string value
             is also split on non-percent-encoded commas and these items
-            are added to the final list (i.e. ``t=1,2,3&t=4``
-            becomes ``['1', '2', '3', '4']``).
+            are added to the final list (i.e. ``t=1,2,3&t=4,5``
+            becomes ``['1', '2', '3', '4', '5']``).
 
             Warning:
                 Enabling this option will cause the framework to misinterpret
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1937,13 +1937,16 @@ class RequestOptions:
                 also http://goo.gl/6rlcux).
 
         auto_parse_qs_csv: Set to ``True`` to split query string values on
-            any non-percent-encoded commas (default ``False``). When ``False``,
+            any non-percent-encoded commas (default ``False``).
+
+            When ``False``,
             values containing commas are left as-is. In this mode, list items
             are taken only from multiples of the same parameter name within the
-            query string (i.e. ``/?t=1,2,3&t=4`` becomes ``['1,2,3', '4']``).
+            query string (i.e. ``t=1,2,3&t=4`` becomes ``['1,2,3', '4']``).
+
             When `auto_parse_qs_csv` is set to ``True``, the query string value
             is also split on non-percent-encoded commas and these items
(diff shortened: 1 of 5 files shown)


kgriffs:
@vytas7 if this PR seems like a good idea, do you want to contribute the cython fix?

vytas7:
Sure. It should be fairly simple since that code path is not fully cythonized.

And I do think it is a good idea, I'm not sure I understood the original intent tbh. It could have been an optimization, but the case itself seems to be quite uncommon: normally you would choose either or.

vytas7:
@kgriffs Done!
While at it, I've also removed one roundtrip via Python call, but I haven't benchmarked much. It _should_ be a slight improvement, although the whole csv path isn't particularly optimized, it's rather cythonized Python code.

kgriffs:
>  but the case itself seems to be quite uncommon: normally you would choose either or.

Agreed. However we have a reputation to uphold. 😸

falcon/request.py
```diff
@@ -1937,13 +1937,16 @@ class RequestOptions:
...
+            When ``False``,
             values containing commas are left as-is. In this mode, list items
             are taken only from multiples of the same parameter name within the
-            query string (i.e. ``/?t=1,2,3&t=4`` becomes ``['1,2,3', '4']``).
+            query string (i.e. ``t=1,2,3&t=4`` becomes ``['1,2,3', '4']``).
+
             When `auto_parse_qs_csv` is set to ``True``, the query string value
             is also split on non-percent-encoded commas and these items
-            are added to the final list (i.e. ``/?t=1,2,3&t=4``
+            are added to the final list (i.e. ``t=1,2,3&t=4``
```
CaselIT, line 1949: we could use as example `t=1,2,3&t=4,5` so it's clear what the behaviour is

tests/test_query_params.py
```diff
@@ -182,6 +182,67 @@ def test_option_auto_parse_qs_csv_simple_true(self, simulate_request, client, re
...
+        query_string = qs
+        simulate_request(client=client, path='/', query_string=query_string)
+
+        req = resource.captured_req
+
+        # Old Behavior
+        #
+        # assert req.params['t'] == ['1', '2', '3,4']
+        # assert req.get_param('t') in ['1', '2', '3,4']
+        # assert req.get_param_as_list('t') == ['1', '2', '3,4']
```
CaselIT, line 239: I think this can be cleaned up

CaselIT (changes requested):
Looks ok, just a documentation improvement

tests/test_query_params.py
kgriffs: Good catch. I had that in there for purposes of discussion.

kgriffs:
@vytas7 thanks for your help on this!

CaselIT (approved):

vytas7 (approved):
LGTM 👍 

(I hope that piece of Cython was carefully reviewed by someone else than me, too 🙂)

CaselIT:
> LGTM 👍 
> 
> (I hope that piece of Cython was carefully reviewed by someone else than me, too 🙂)  

It looked ok to me

PR #1842, opened 2020-12-26
docs(install,tutorial): improve app server installation docs + a couple of touch-ups for ASGI tutorial

vytas7:

### DIFF
--- a/docs/user/install.rst
+++ b/docs/user/install.rst
@@ -138,7 +138,8 @@ For a more in-depth overview of available servers, see also:
     By default, the ``uvicorn`` package comes only with a minimal set of
     pure-Python dependencies.
     For CPython-based production deployments, you can install Uvicorn along
-    with faster alternatives such as ``uvloop`` and ``httptools``::
+    with more optimized alternatives such as ``uvloop`` (a faster event loop),
+    ``httptools`` (a faster HTTP protocol implementation) etc::
 
         $ pip install uvicorn[standard]
 
--- a/docs/user/install.rst
+++ b/docs/user/install.rst
@@ -101,9 +101,9 @@ Falcon speaks WSGI, and so in order to serve a Falcon app, you will
 need a WSGI server. Gunicorn and uWSGI are some of the more popular
 ones out there, but anything that can load a WSGI app will do.
 
-All Windows developers can use Waitress production-quality pure-Python WSGI server with very acceptable performance.
-Unfortunately Gunicorn is still not working on Windows and uWSGI need to have Cygwin on Windows installed.
-Waitress can be good alternative for Windows users if they want quick start using Falcon on it.
+Windows users can try Waitress, a production-quality, pure-Python WSGI server.
+Other alternatives on Windows include running Gunicorn and uWSGI via WSL,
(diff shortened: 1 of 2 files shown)


docs/user/install.rst
```diff
@@ -101,9 +101,9 @@ Falcon speaks WSGI, and so in order to serve a Falcon app, you will
 need a WSGI server. Gunicorn and uWSGI are some of the more popular
 ones out there, but anything that can load a WSGI app will do.
 
-All Windows developers can use Waitress production-quality pure-Python WSGI server with very acceptable performance.
-Unfortunately Gunicorn is still not working on Windows and uWSGI need to have Cygwin on Windows installed.
-Waitress can be good alternative for Windows users if they want quick start using Falcon on it.
+Windows users can try Waitress, a production-quality, pure-Python WSGI server.
+Other alternatives on Windows include running Gunicorn and uWSGI via WSL,
+as well as inside Linux Docker containers via Hyper-V.
```
CaselIT, line 106: actually docker can also use wsl2 now, that technically is still hyprer-v, but it's maybe less clear.

I would just drop the   ` via Hyper-V` altogether since it does not add much information.

vytas7: Thanks, dropped.

docs/user/install.rst
```diff
@@ -133,6 +133,18 @@ For a more in-depth overview of available servers, see also:
 
     $ pip install [uvicorn|daphne|hypercorn]
 
+.. note::
+
+    By default, the ``uvicorn`` package comes only with a minimal set of
+    pure-Python dependencies.
+    For CPython-based production deployments, you can install Uvicorn along
+    with faster alternatives such as ``uvloop`` and ``httptools``::
```
CaselIT, line 141: Faster alternative of what? I think we should specify that we are talking about the asyncio loop here

vytas7: Thanks, I'll try rewording (unless you have a good suggestion).

docs/user/install.rst
```diff
@@ -133,6 +133,18 @@ For a more in-depth overview of available servers, see also:
...
+
+        $ pip install uvicorn[standard]
```
vytas7, line 143: This is an open question whether we should extend our test dependencies to include `[standard]`.
This seems to be a rather recent change, I swear `unicorn` used to pull in `uvloop` by default before.

CaselIT: I think that a pipeline job may be useful, like the daphne and hypercorn ones.

vytas7: Hm, maybe we should add more environments in another _Action_ which would be run only upon merge to master (similar to _"tests emulated"_)? Like `uvicorn[standard]`, Windows+Cython, MacOS+Cython, Windows+Hypercorn etc. Fuller perf test suites could also be run in that fashion.

CaselIT: Sure, that's an option, but it would duplicate the configuration somewhat since iirc you cannot share the yaml code.

vytas7: To that end, I'll make my upcoming ASGI E2E test suite use `uvicorn[standard]`.

CaselIT (approved):

PR #1843, opened 2020-12-27
test(ASGI): include hypercorn in ASGI server tests

### LINKED ISSUE #1836: Include hypercorn in ASGI server tests

vytas7:
Closes #1836


CaselIT (approved):
looks ok. It also works on windows

vytas7:
Thanks for checking.
I was thinking to include it by default first, or group with Daphne, but it's dog slow (slower even than Daphne... although seemingly more stable).

PR #1844, opened 2020-12-28
perf(ASGI): cythonize falcon.asgi module tree

vytas7:


falcon/asgi/app.py
```diff
@@ -279,7 +280,7 @@ def __init__(self, *args, request_type=Request, response_type=Response, **kwargs
 
         self.add_error_handler(WebSocketDisconnected, self._ws_disconnected_error_handler)
 
-    async def __call__(self, scope, receive, send):  # noqa: C901
+    async def _call_impl(self, scope, receive, send):  # noqa: C901
```
vytas7, line 283: Another alternative is simply overwriting this with `__call__ = _wrap_asgi_coroutine_func(__call__)`. Thoughts?

CaselIT: why not just use _wrap_asgi_coroutine_func as a decorator?

CaselIT (approved):
Looks ok to me.

As a side note, maybe we could make the pipeline test cython in windows and osx, since id currently does not seem to be tested

falcon/asgi/app.py
vytas7: Doh, right :man_facepalming: I'll try that.
You can tell I was exploring completely different solutions first. :grimacing:

CaselIT (dismissed):
ok

kgriffs (approved):
LGTM

CaselIT (approved):

vytas7:
Merging, as it is overall a clear improvement (albeit a modest one), although we might need to remove that decorator, and skip cythonization of App instead... I realized what the main limitation of cythonizing coro funcs is. Other files in asgi/ are still good to keep cythonized.

PR #1845, opened 2020-12-29
perf(ASGI): tune ASGI request-response cycle performance

vytas7:
- [x] Speed up ASGI scope validation (or make concessions and be less strict)
- [x] ~~Speed up `Response` headers (I won't have the time to complete this _right now_)~~
- [x] Avoid unnecessary function calls (where reasonable)
- [x] Avoid unnecessary awaits (where reasonable)
- [x] Speed up `Request.__init__()`
- [x] Restore 100% coverage

### DIFF
--- a/falcon/asgi/_asgi_helpers.py
+++ b/falcon/asgi/_asgi_helpers.py
@@ -1,0 +1,63 @@
+# Copyright 2020 by Vytautas Liuolia.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from falcon.errors import UnsupportedError, UnsupportedScopeError
+from falcon.util.misc import _lru_cache_safe
+
+
+@_lru_cache_safe(maxsize=16)
+def _validate_asgi_scope(asgi_version, scope_type, spec_version, http_version):
+    if not asgi_version.startswith('3.'):
+        raise UnsupportedScopeError(
+            f'Falcon requires ASGI version 3.x. Detected: {asgi_version}'
+        )
+
+    if scope_type == 'http':
+        spec_version = spec_version or '2.0'
+        if not spec_version.startswith('2.'):
+            raise UnsupportedScopeError(
+                f'The ASGI "http" scope version {spec_version} is not supported.'
+            )
+        if http_version not in {'1.0', '1.1', '2', '3'}:
+            raise UnsupportedError(
+                f'The ASGI "http" scope does not support HTTP version {http_version}.'
+            )
+        return spec_version
+
+    if scope_type == 'websocket':
+        spec_version = spec_version or '2.0'
+        if not spec_version.startswith('2.'):
+            raise UnsupportedScopeError(
+                'Only versions 2.x of the ASGI "websocket" scope are supported.'
+            )
+        if http_version not in {'1.1', '2', '3'}:
+            raise UnsupportedError(
+                f'The ASGI "websocket" scope does not support HTTP version {http_version}.'
+            )
+        return spec_version
+
+    if scope_type == 'lifespan':
+        spec_version = spec_version or '1.0'
+        if not spec_version.startswith('1.') and not spec_version.startswith('2.'):
+            raise UnsupportedScopeError(
+                'Only versions 1.x and 2.x of the ASGI "lifespan" scope are supported.'
+            )
+        return spec_version
+
+    # NOTE(kgriffs): According to the ASGI spec: "Applications should
+    #   actively reject any protocol that they do not understand with
+    #   an Exception (of any type)."
+    raise UnsupportedScopeError(
+        f'The ASGI "{scope_type}" scope type is not supported.'
+    )
--- a/falcon/asgi/_asgi_helpers.py
+++ b/falcon/asgi/_asgi_helpers.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
+
 from falcon.errors import UnsupportedError, UnsupportedScopeError
 from falcon.util.misc import _lru_cache_safe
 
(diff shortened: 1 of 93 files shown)


kgriffs:
Some test cases and results regarding request header decoding (see my comment on the gist): https://gist.github.com/kgriffs/fa31baf256d81ef6e4bb1b401b4e3dae

falcon/asgi/app.py
```diff
@@ -521,11 +477,17 @@ async def __call__(self, scope, receive, send):  # noqa: C901
             })
 
             await send(_EVT_RESP_EOF)
-            self._schedule_callbacks(resp)
+
+            # PERF(vytas): Check resp._registered_callbacks directly to shave
+            #   off a function call since this is a hot/critical code path.
+            if resp._registered_callbacks:
```
CaselIT, line 483: can't we cache `_registered_callbacks`? it's used multiple times so I think creating a local variable may be worth it

falcon/asgi/app.py
```diff
@@ -521,11 +477,17 @@ async def __call__(self, scope, receive, send):  # noqa: C901
...
+                self._schedule_callbacks(resp)
```
CaselIT, line 484: can't we just pass the cached `_registered_callbacks` to the function? it does not otherwise uses resf if I'm not mistaken

falcon/asgi/request.py
```diff
@@ -345,13 +345,28 @@ class Request(falcon.request.Request):
...
-        '_asgi_server_cached',
+        # '_asgi_server_cached',
         '_first_event',
         '_receive',
-        '_stream',
+        # '_stream',
         'scope',
     ]
 
+    # PERF(vytas): Dirty hack to fall back to class attributes when unset;
```
CaselIT, line 355: Not sure I would classify this as an hack

falcon/asgi/request.py
```diff
@@ -382,18 +397,21 @@ def __init__(self, scope, receive, first_event=None, options=None):
         #  Misc.
         # =====================================================================
 
-        self._asgi_server_cached = None  # Lazy
+        # PERF(vytas): Fall back to class variable(s) when unset.
+        # self._asgi_server_cached = None  # Lazy
```
CaselIT, line 401: Since have `__dict__`  can't we just use [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property)?
It's easy to backport, but I'm have not done benchmarks to see if it's faster than out current solution.

an alternative is to remove `__dict__`

CaselIT (commented):
just some other comments

falcon/asgi/request.py
vytas7: `cached_property` is an interesting idea, but it is 3.8+ only... Worth benchmarking tho anyway.

falcon/asgi/request.py
vytas7: I would tune down the language regardless, if we go forward with the approach.

falcon/asgi/app.py
vytas7: It is never assumed to be hit in the same path. (I.e., we should not be calling `self._schedule_callbacks(...)` twice.)

falcon/asgi/request.py
CaselIT: > 
> 
> `cached_property` is an interesting idea, but it is 3.8+ only... Worth benchmarking tho anyway.
> Assuming it is implemented in C like that nifty `lru_cache`, otherwise it won't be much better than our stuff.

No, it's a pure python implementation and it can be backported on previous versions without problems (using it on 3.7 at work without issues).
Regarding performance vs a classic `@property`, it should be faster after the first access, since it's just a normal attribute access. I'm not sure regarding the first access though.
Given out use case, I think that most properties are used a low number of times (and I would hazard only once in a fair number of cases) so first access performance is fairly important.

falcon/asgi/app.py
CaselIT: ok, then I guess if we replace the argument of `_schedule_callbacks` it may be useful, otherwise maybe just a readability improvement

kgriffs:
I updated my gist with some additional findings: https://gist.github.com/kgriffs/fa31baf256d81ef6e4bb1b401b4e3dae

falcon/asgi/app.py
vytas7: I don't know if it is safe, I want to wrap this up somehow.

Like what happens if the `resp.stream` generator chooses to add schedule more callbacks along the way?
It's not trivial when to acquire the value to cache.

falcon/asgi/request.py
vytas7: We can take a look at faster ways to cache properties in 3.1.

However, if (as you point out) this `cached_property` is a generic recipe implemented in pure Python, it is unlikely to provide any improvement over our implementation.

falcon/asgi/request.py
vytas7: Reworded to a more marketable version.

tests/asgi/test_sse.py
```diff
@@ -304,7 +309,3 @@ async def emitter():
 
     with pytest.raises(TypeError):
         client.simulate_get()
-
-
-# TODO: Test with uvicorn
-# TODO: Test in browser with JavaScript
```
vytas7, line 310: Done as part of https://github.com/falconry/falcon/pull/1869 .

vytas7:
I'm looking to attempt inlining `Response.render_body()` in order to shave off 1 `await`, but this should be otherwise ready to review.
This PR improves about performance about 10-15% compared to `master`.
And makes us faster than Starlette in a "Hello, world" type of benchmark in an isolated pseudo-async loop.
However, on real Uvicorn, Starlette is still somehow magically faster... :thinking:

falcon/asgi/_asgi_helpers.py
```diff
@@ -14,6 +14,51 @@
 
 import inspect
 
+from falcon.errors import UnsupportedError, UnsupportedScopeError
+from falcon.util.misc import _lru_cache_safe
+
+
+@_lru_cache_safe(maxsize=16)
```
CaselIT, line 21: I don't remember what happens with the lru cache if the function emits an error. is the error cached or the function is called again each time?

tests/asgi/test_response_media_asgi.py
```diff
@@ -184,7 +205,7 @@ def _resolve(media_type, default, raise_not_found=True):
     assert result.json == doc
 
 
-def runTest(test_fn):
+def run_test(test_fn):
```
CaselIT, line 208: ops sorry. I use camel case @ work

tests/asgi/test_scope.py
```diff
@@ -24,6 +24,15 @@ def test_missing_asgi_version():
     assert resource.captured_req.scope['asgi']['version'] == '2.0'
 
 
+@pytest.mark.parametrize('http_version', ['0.9', '1.9', '4.0', '1337'])
```
CaselIT, line 27: I feel like this is missing a 42 :)

CaselIT (approved):
Looks nice, I'm not sure if the in-lining of the event types is noticeable, but I guess everything helps?

falcon/asgi/_asgi_helpers.py
vytas7: Errors are not cached:
```python
>>> import time
>>> import functools
>>> @functools.lru_cache
... def test(n):
...     if n == 42:
...         raise RuntimeError(f'[{time.time()}] {n} is unsupported')
...     return n
... 
>>> test(1)
1
>>> test(2)
2
>>> test(42)
  File "<stdin>", line 4, in test
RuntimeError: [1614715125.4782972] 42 is unsupported
>>> test(42)
  File "<stdin>", line 4, in test
RuntimeError: [1614715126.8562453] 42 is unsupported
```

OTOH both behaviours are acceptable for us I guess?

tests/asgi/test_scope.py
vytas7: Yes, but why not 85? :thinking:

kgriffs (approved):
Sweet, nice work.

Issue #1822, opened 2020-12-18
3.0 Benchmarking and Performance Tuning


kgriffs:
- [ ] Run benchmarks on 3.0 vs. 2.0 and take care of any performance regressions we find.
- [ ] Also, we will need to create some baseline ASGI performance tests.
- [ ] Update falconframework.org with new benchmarks

KhDenys:
@kgriffs Will you add django 3.x with wsgi and asgi modes to the benchmarks?

vytas7:
Thanks for the reminder @KhDenys ! Yes, we'll try to bring Django 3 into the equation.

kgriffs:
Created: https://github.com/falconry/falcon/issues/1881

kgriffs:
> Run benchmarks on 3.0 vs. 2.0 and take care of any performance regressions we find.

A quick test on my MBP shows that we lost 4-7 μs/req with non-cythonized wheels. With the wheels, we are a few μs faster.

kgriffs:
OK, so I figured out that I was accidentally cythonizing the 2.0.0 wheel that I built from source. 🤦‍♂️

No significant regression after all, so I think we are good.

kgriffs:
I just created new images:

https://hub.docker.com/repository/registry-1.docker.io/falconry/falcon-bench/tags?page=1&ordering=last_updated&name=3.0.1

E.g.,:

docker run -it falconry/falcon-bench:py

```
Python Version:
==================
Python 3.8.10

Installed Packages:
==================
Package      Version
------------ -------
asgiref      3.3.4
bottle       0.12.19
click        8.0.0
Django       3.2.3
falcon       3.0.1
Flask        2.0.0
itsdangerous 2.0.1
Jinja2       3.0.0
MarkupSafe   2.0.0
pip          21.1.1
pytz         2021.1
setuptools   56.2.0
sqlparse     0.4.1
Werkzeug     2.0.1
wheel        0.36.2
```

PR #1846, opened 2020-12-30
perf(asgi): Optimize media (de)serialization

kgriffs:
# Summary of Changes

This patch optimizes the media (de)serialization paths for ASGI apps by
avoiding an extra await and slightly flattening the call stack.

BREAKING CHANGE: Private member variables for the default media handlers were
    renamed. This should only affect subclasses, as these variables were never
    documented, and were not intended to be part of the public interface.

# Related Issues

#1822

# Pull Request Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -67,8 +67,9 @@ extensions = [
     'sphinx_tabs.tabs',
 
     # Falcon-specific extensions
-    'ext.rfc',
     'ext.doorway',
+    'ext.private_args',
+    'ext.rfc',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/falcon/__init__.py
+++ b/falcon/__init__.py
@@ -25,7 +25,6 @@ the framework's classes, functions, and variables::
 """
 
 import logging as _logging
-import sys as _sys
 
 # Hoist classes and functions into the falcon namespace
 from falcon.version import __version__  # NOQA
@@ -49,10 +48,6 @@ from falcon.util import *  # NOQA
 from falcon.hooks import before, after  # NOQA
 from falcon.request import Request, RequestOptions, Forwarded  # NOQA
 from falcon.response import Response, ResponseOptions  # NOQA
-
-
-ASGI_SUPPORTED = _sys.version_info >= (3, 6)
-"""Set to ``True`` when ASGI is supported for the current Python version."""
 
 
 # NOTE(kgriffs): Only to be used internally on the rare occasion that we
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -700,7 +700,7 @@ class Request(falcon.request.Request):
 
         return netloc_value
 
-    async def get_media(self, default_when_empty=_UNSET, _deserialize_cache={}):
+    async def get_media(self, default_when_empty=_UNSET):
         """Return a deserialized form of the request stream.
 
         The first time this method is called, the request stream will be
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -751,21 +751,10 @@ class Request(falcon.request.Request):
                 return default_when_empty
             raise self._media_error
 
-        handler = self.options.media_handlers.find_by_media_type(
+        handler, _, deserialize_sync = self.options.media_handlers.find_by_media_type(
             self.content_type,
             self.options.default_media_type
         )
-
-        # PERF(kgriffs): Avoid using an additional await and flatten the call
-        #   stack when possible.
-        try:
-            deserialize_sync = _deserialize_cache[handler]
-        except KeyError:
-            # PERF(kgriffs): Do not use EAFP, but rather check and
-            #   cache since we don't want a significant penalty if
-            #   a custom handler does not subclass the ABC.
-            deserialize_sync = getattr(handler, '_deserialize_sync', None)
-            _deserialize_cache[handler] = deserialize_sync
 
         try:
             if deserialize_sync:
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -759,17 +759,17 @@ class Request(falcon.request.Request):
         # PERF(kgriffs): Avoid using an additional await and flatten the call
         #   stack when possible.
         try:
-            __deserialize_sync__ = _deserialize_cache[handler]
+            deserialize_sync = _deserialize_cache[handler]
         except KeyError:
             # PERF(kgriffs): Do not use EAFP, but rather check and
             #   cache since we don't want a significant penalty if
             #   a custom handler does not subclass the ABC.
-            __deserialize_sync__ = getattr(handler, '__deserialize_sync__', None)
-            _deserialize_cache[handler] = __deserialize_sync__
+            deserialize_sync = getattr(handler, '_deserialize_sync', None)
+            _deserialize_cache[handler] = deserialize_sync
 
         try:
-            if __deserialize_sync__:
-                self._media = __deserialize_sync__(await self.stream.read())
+            if deserialize_sync:
+                self._media = deserialize_sync(await self.stream.read())
             else:
                 self._media = await handler.deserialize_async(
                     self.stream,
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -700,7 +700,7 @@ class Request(falcon.request.Request):
 
         return netloc_value
 
-    async def get_media(self, default_when_empty=_UNSET, _shortcut_cache={}):
+    async def get_media(self, default_when_empty=_UNSET, _deserialize_cache={}):
         """Return a deserialized form of the request stream.
 
         The first time this method is called, the request stream will be
@@ -759,19 +759,17 @@ class Request(falcon.request.Request):
         # PERF(kgriffs): Avoid using an additional await and flatten the call
         #   stack when possible.
         try:
-            use_shortcut = _shortcut_cache[handler]
+            __deserialize_sync__ = _deserialize_cache[handler]
         except KeyError:
-            # NOTE(kgriffs): Do not use isinstance() because we can't be sure if
-            #   we can use our shortcut for subclasses.
-            use_shortcut = type(handler) in (
-                falcon.media.JSONHandler,
-                falcon.media.MessagePackHandler
-            )
-            _shortcut_cache[handler] = use_shortcut
+            # PERF(kgriffs): Do not use EAFP, but rather check and
+            #   cache since we don't want a significant penalty if
+            #   a custom handler does not subclass the ABC.
+            __deserialize_sync__ = getattr(handler, '__deserialize_sync__', None)
+            _deserialize_cache[handler] = __deserialize_sync__
 
         try:
-            if use_shortcut:
-                self._media = handler._deserialize(await self.stream.read())
+            if __deserialize_sync__:
+                self._media = __deserialize_sync__(await self.stream.read())
             else:
                 self._media = await handler.deserialize_async(
                     self.stream,
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -751,7 +751,7 @@ class Request(falcon.request.Request):
                 return default_when_empty
             raise self._media_error
 
-        handler, _, deserialize_sync = self.options.media_handlers.find_by_media_type(
+        handler, _, deserialize_sync = self.options.media_handlers._resolve(
             self.content_type,
             self.options.default_media_type
         )
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -19,6 +19,7 @@ from falcon import request_helpers as helpers  # NOQA: Required by fixed up WSGI
 from falcon.constants import SINGLETON_HEADERS
 from falcon.forwarded import _parse_forwarded_header  # NOQA: Req. by fixed up WSGI Request attrs
 from falcon.forwarded import Forwarded  # NOQA
+import falcon.media
 import falcon.request
 from falcon.util.uri import parse_host, parse_query_string
 from . import _request_helpers as asgi_helpers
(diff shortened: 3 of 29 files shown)


falcon/asgi/request.py
```diff
@@ -699,7 +700,7 @@ def netloc(self):
 
         return netloc_value
 
-    async def get_media(self, default_when_empty=_UNSET):
+    async def get_media(self, default_when_empty=_UNSET, _shortcut_cache={}):
```
kgriffs, line 703: Giving this pattern a try. A sphinx extension is used to filter it out from the docs.

kgriffs:
In my tests, this optimization shaves off ~5-10 µs. It would be interesting to see what cachegrind says, though.

falcon/asgi/response.py
```diff
@@ -267,10 +268,30 @@ async def render_body(self):
...
+                    #   the call stack when possible.
+                    try:
+                        serialize = _serialize_cache[handler]
+                    except KeyError:
+                        # NOTE(kgriffs): Do not use isinstance() because we can't be sure if
+                        #   we can use our shortcut for subclasses.
+                        t = type(handler)
+                        if t is falcon.media.JSONHandler:
+                            serialize = handler.serialize
+                        elif t is falcon.media.MessagePackHandler:
```
vytas7, line 281: [`URLEncodedFormHandler`](https://falcon.readthedocs.io/en/latest/api/media.html#falcon.media.URLEncodedFormHandler) is also a good candidate to follow the pattern.

falcon/asgi/request.py
```diff
@@ -755,12 +756,29 @@ async def get_media(self, default_when_empty=_UNSET):
...
         )
 
+        # PERF(kgriffs): Avoid using an additional await and flatten the call
+        #   stack when possible.
         try:
-            self._media = await handler.deserialize_async(
-                self.stream,
-                self.content_type,
-                self.content_length
+            use_shortcut = _shortcut_cache[handler]
```
vytas7, line 762: Maybe the whole shortcut cache could instead be a method returning the synchronous handler, or `None`?
And we would of course plop an `lru_cache` on top of it.

kgriffs: That would be an option, but I think we would have to give up some cycles. With this simple local dict pattern I was aiming for the least overhead possible.

falcon/asgi/response.py
kgriffs: Good catch. I will circle back on that after we finish the first review pass on the PR.

falcon/asgi/request.py
vytas7: Thoughts on ditching EAFP in favour of the below (CPython 3.8):
```python
>>> XML = object()
>>> SOAP = object()
>>> 
>>> _shortcut_cache1 = {XML: True, SOAP: False}
>>> 
>>> _shortcut_cache2 = {XML}
>>> 
>>> timeit.timeit(lambda: _shortcut_cache1[XML], number=int(1e8))
6.385023718999946
>>> timeit.timeit(lambda: _shortcut_cache1[SOAP], number=int(1e8))
6.587060416000895
>>>
>>> timeit.timeit(lambda: XML in _shortcut_cache2, number=int(1e8))
6.124332429000788
>>> timeit.timeit(lambda: SOAP in _shortcut_cache2, number=int(1e8))
6.186602156001754
```

vytas7: Maybe we could instead just have a private attribute on our own handlers?
And we would set it in the constructor, using `type(self) in our_handlers`, or just `type(self) is JSONHandler` etc.

CaselIT:
This is interesting, but is not very scalable. 
I think we could rethink the handler interface a bit to support this pattern, since it should be the handler that should indicate if it support fast access to this.

Also it does not take into consideration that the handler may change after application startup

kgriffs:
In order to improve cohesion and reduce coupling, I have refactored the implementation to use a protocol-based approach, where the handlers themselves can opt-in to the optimization and specify the sync callables to be used. This simplifies the logic in the Request and Response classes while also avoiding a brittle dependency on handler internals

falcon/media/urlencoded.py
```diff
@@ -25,13 +29,21 @@ class URLEncodedFormHandler(BaseHandler):
...
+            self._deserialize_sync = self._deserialize
 
-    def serialize(self, media, content_type):
+    # NOTE(kgriffs): Make content_type a kwarg to support the
+    #   Request.render_body() shortcut optimization.
+    def serialize(self, media, content_type=None) -> bytes:
         # NOTE(vytas): Setting doseq to True to mirror the parse_query_string
         # behaviour.
-        return urlencode(media, doseq=True)
+        return urlencode(media, doseq=True).encode()
```
kgriffs, line 46: During testing I discovered and fixed this bug.

CaselIT:
I think we could do the normalization on the Handler class. I'll look into it

CaselIT:
Take a look at https://github.com/CaselIT/falcon/commit/257a820141961a88ef3ca977a88c30f812da75f4

cc @vytas7

kgriffs:
I pulled in the patches from @CaselIT and then refactored to use lru_cache(). This seems to have done the trick, but please let me know how it benchmarks on your systems.

falcon/media/handlers.py
```diff
@@ -54,27 +109,27 @@ def _resolve_media_type(self, media_type, all_media_types):
 
         return resolved
 
+    # NOTE(kgriffs): Most apps will probably only use one or two media handlers,
+    #   but we use maxsize=64 to give us some wiggle room just in case someone
+    #   is using versioned media types or something, and to cover various
+    #   combinations of the method args. We may need to tune this later.
+    @functools.lru_cache(maxsize=64)
     def find_by_media_type(self, media_type, default, raise_not_found=True):
```
vytas7, line 117: As discussed on Gitter, we should IMHO simply define new methods (`_find_async_handler` or similar etc) that we need, and make them private. And leave this one as-is.

Otherwise we need to mention the updated return signature as a breaking change, since this was not an underscored method, and the class itself is public. (NB we would have not given any release inbetween to ease this breaking change either.)

vytas7 (changes requested):
We need to decide what to do about a potential breaking change in  `find_by_media_type`.

falcon/media/urlencoded.py
vytas7: :man_facepalming:

docs/conf.py
```diff
@@ -67,8 +67,9 @@
     'sphinx_tabs.tabs',
 
     # Falcon-specific extensions
-    'ext.rfc',
     'ext.doorway',
+    'ext.private_args',
```
vytas7, line 71: Do we still need this extension after refactoring in favour of the LRU approach?
(But maybe it's still useful to keep so that we don't accidentally surface private args? :thinking: )

falcon/media/handlers.py
kgriffs: I actually agree, it's just something I forgot to do.

falcon/media/handlers.py
```diff
@@ -39,6 +42,58 @@ def __init__(self, initial=None):
...
+
+    def _rehash(self):
+        # NOTE(kgriffs): We could have done something simpler, but it would
+        #   have been a leaky abstraction, and it is probably better to avoid
+        #   any potential suprises from instances of Handlers not appearing to
+        #   be logically equivalent.
+        # NOTE(kgriffs): The generator must be wrapped in tuple() in order to
+        #   get a consistent hash value, because otherwise we end up hashing
+        #   the generator object itself.
+        self._hash = hash(tuple((k, v) for k, v in self.data.items()))
```
vytas7, line 67: How would this behave in the face of a custom unhashable handler object?

kgriffs: Good catch. I did think about unhashable handlers but then it slipped my mind to address it. That being said, a cached closure may do the trick.

kgriffs:
Incorporated additional excellent feedback from @vytas7 and @CaselIT.

docs/conf.py
kgriffs: I suppose we can leave it in since it isn't hurting anything, but I don't have a strong preference either way.

falcon/media/handlers.py
CaselIT: To avoid repetition, I think this method could call the new one and just return the first argument, so we avoid having two implementations that do the same thing.

CaselIT (dismissed):
I've left a suggestion, otherwise it's ok for me

falcon/media/handlers.py
kgriffs: That's a good point; doing so would work as the implementation of `resolve()` currently stands. I took a "safety first", non-DRY approach to give us more freedom to evolve the new private function as desired without having to worry about accidentally introducing a subtle breaking change in the logic, if not the return type. And the `find_by_media_type()` would be frozen going forward, so there really isn't much of an additional maintenance burden to speak of.

That being said, we may not do much with `resolve()` going forward, either. Which would remove the pros from my argument. 😄

CaselIT: I guess it would not hurt leaving it as is, since it's on the chopping blocks for v4

falcon/media/handlers.py
```diff
@@ -39,21 +43,75 @@ def __init__(self, initial=None):
         # Also, this results in self.update(...) being called.
         UserDict.__init__(self, handlers)
 
-    def _resolve_media_type(self, media_type, all_media_types):
-        resolved = None
+    def __setitem__(self, key, item):
```
vytas7, line 46: Nitpick: it's more canonical to name `__setitem__` arguments `key` and `value`.

falcon/util/misc.py
```diff
@@ -79,17 +80,21 @@ def decorator(func):
 if (
     # NOTE(kgriffs): https://bugs.python.org/issue28969
     (sys.version_info.minor == 5 and sys.version_info.micro < 4) or
-    (sys.version_info.minor == 6 and sys.version_info.micro < 1) or
-
-    # PERF(kgriffs): Using lru_cache is slower on pypy when the wrapped
-    #   function is just doing a few non-IO operations.
-    (sys.implementation.name == 'pypy')
+    (sys.version_info.minor == 6 and sys.version_info.micro < 1)
```
vytas7, line 83: This would benefit from Python 4 awareness too.

vytas7 (changes requested):
LGTM :+1: 
(Well, almost; I just had a couple of cosmetic nitpicks.)

falcon/media/handlers.py
kgriffs: True. I guess I wasn't looking to closely. This came from https://github.com/CaselIT/falcon/commit/257a820141961a88ef3ca977a88c30f812da75f4#diff-52784144e74b667a1069b60e8cf6a59f805f05c269e3ce7cccea884167ea5f2fR43

kgriffs: Fixed

falcon/util/misc.py
kgriffs: Do you mean 3.4? If so, I didn't put this in because we only support 3.5+ for the framework as a whole. Or perhaps you are trolling me about Python 4.0 😛

vytas7: Yes, I did mean 4.0 🙂 (I was semi-serious, why not compare the whole tuple.)

kgriffs: Heh, OK. I guess it couldn't hurt.

kgriffs:
Regarding performance testing, on my laptop I was able to verify improvements on py36 and py38 with or without cythonization.

Under PyPy I found that performance decreased by about 120 ns since the lru cache was that much slower as compared to the removed simple `self.data` optimization shown below.

```py
        try:
            return self.data[media_type]
        except KeyError:
            pass
```

However, when the simple check fails, the lru provides a large performance boost (roughly going from 4 µs to 440 ns). Therefore, given that the penalty in the common case is negligible, I am inclined to keep using the lru under PyPy.

vytas7:
Hm, isn't that simple case very common though? I.e. simply using `application/json` is probably what 99% of JSON API users are going to do...

I don't know how much it is worth optimizing this, but maybe we could detect PyPy and shim `_resolve` with smth along the lines of
```python
try:
    return self.data[media_type]
except KeyError:
    return self._cached_resolve(...)
```

vytas7 (approved):
:+1:

CaselIT:
> Hm, isn't that simple case very common though? I.e. simply using `application/json` is probably what 99% of JSON API users are going to do...
> 
> I don't know how much it is worth optimizing this, but maybe we could detect PyPy and shim `_resolve` with smth along the lines of
> ```python
> try:
>     return self.data[media_type]
> except KeyError:
>     return self._cached_resolve(...)
> ```
> 
> 

I think we could just have this logic for all implementations, since with the lru cache the function is not called often

kgriffs:
I went ahead and attempted to better improve PyPy performance since discovering that the penalty was a little worse under the newer PyPy 7.3.3 release, and because I pride myself on making poor life choices...

If you guys have time, please benchmark this last batch of changes to keep me honest. At this point I think we have reached the point of diminishing returns, so hopefully we can tie a bow on this PR soon.

falcon/__init__.py
```diff
@@ -51,10 +50,6 @@
 from falcon.response import Response, ResponseOptions  # NOQA
 
 
-ASGI_SUPPORTED = _sys.version_info >= (3, 6)
```
kgriffs, line 54: Moved this for the sake of cohesion (keeping "constants" all together)

falcon/constants.py
```diff
@@ -1,5 +1,13 @@
 from enum import Enum
 import os
+import sys
+
+
+PYPY = (sys.implementation.name == 'pypy')
```
kgriffs, line 6: I went ahead and factored out this constant since we were using it in multiple places.

CaselIT (approved):

vytas7 (approved):
:+1:

PR #1848, opened 2020-12-31
[PoC] perf(ASGI): swap out App.__call__ for a potentially more optimized one after app startup

vytas7:
This is a prototype. Things to do:

- [ ] Decide whether this is a good idea
- [ ] Reason about `App` subclassing patterns and issues we might run into
- [ ] Restore 100% coverage
- [ ] Add `NOTE`s (and potentially docs) describing the effect of this hack

### DIFF
--- a/falcon/asgi/app.py
+++ b/falcon/asgi/app.py
@@ -17,6 +17,7 @@
 import asyncio
 from inspect import isasyncgenfunction, iscoroutinefunction
 import traceback
+import weakref
 
 import falcon.app
 from falcon.app_helpers import prepare_middleware, prepare_middleware_ws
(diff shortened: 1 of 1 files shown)


kgriffs:
Can we quantify the performance gain from doing this?

falcon/asgi/app.py
```diff
@@ -270,18 +271,44 @@ async def process_response(self, req, resp, resource, req_succeeded)
...
+            if app:
+                if app._lifespan_status != EventType.LIFESPAN_STARTUP_COMPLETE:
+                    all_started = False
+            else:
+                disposed.add(ref)
+
+        cls._known_instances.difference_update(disposed)
+
+        if all_started:
+            cls.__call__ = cls._call_impl
```
kgriffs, line 297: > Reason about App subclassing patterns and issues we might run into

Indeed, we will need to add test cases to cover these patterns.

vytas7:
Re performance gain, we are shaving off one compatibility-`await` (and the real gain is for some strange reason slightly even more :thinking: ).

vytas7:
There is unfortunately some puzzling, unsolved issue with cythonized stuff and "real" ASGI servers that I don't understand.

Directly involving the cythonized app seems to produce a noticeable speedup, but wrapping that in a pure Python coroutine function negates that advantange, and even turns it into a slowdown. It seems that at least `uvicorn` always invokes the app in the way that it is unfortunately slowed down.

vytas7:
I'll kill this for now until the above is demystified.

I was testing Cython 3.0 alpha, and while it is also disappointing, it performed slightly better, so it might well be an inherent limitation of cythonized generators for now :disappointed:

PR #1849, opened 2021-01-04
chore(env): set the required Falcon envvars in conftest.py

### LINKED ISSUE #1652: Improve test readme. Without setting the proper env variable 300+ failure occur

vytas7:
Closes #1652

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,16 @@
+import os
+
 import pytest
 
 import falcon
+
+
+_FALCON_TEST_ENV = (
+    ('FALCON_ASGI_WRAP_NON_COROUTINES', 'Y'),
+    ('FALCON_TESTING_SESSION', 'Y'),
+    # NOTE: PYTHONASYNCIODEBUG is optional (set in tox.ini).
+    # ('PYTHONASYNCIODEBUG', '1'),
+)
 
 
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -28,6 +38,10 @@ def pytest_configure(config):
         plugin.mypy_argv.append('--ignore-missing-imports')
 
 
+def pytest_sessionstart(session):
+    os.environ.update(_FALCON_TEST_ENV)
```
CaselIT, line 42: I think `setdefault`  would be better, since it allows controlling the variables from the extern

CaselIT (changes requested):
Thanks for taking this on

tests/conftest.py
vytas7: sure

nZac (approved):
:shipit:

PR #1850, opened 2021-01-09
chore: pin tox to work around 3.21 bug

kgriffs:
See also: https://github.com/tox-dev/tox/issues/1777


vytas7 (approved):
The provided pinning instructions will result in `tox` version 3.20.0 used instead of the latest "good" 3.20.1. But OTOH the sole fix in 3.20.0 :arrow_right: 3.20.1 does not really affect us, so :+1: :shipit:

PR #1854, opened 2021-01-20
refactor: Organize imports removing all "import *"

CaselIT:
I've tried to remove all the `import *`, other than the one used for the `status_codes`.
The api should not have changed, and the test pass without any change, but a double check is probably better.

I've also improved the imports to avoid some circular dependencies.

I've semi-used an automated tool, zimports, to update the imports, but it cannot be automated at the moment since in changes the order of the comments.


falcon/asgi/request.py
```diff
@@ -15,23 +15,20 @@
...
 
 from falcon import errors
-from falcon import request_helpers as helpers  # NOQA: Required by fixed up WSGI Request attrs
+from falcon import request
+from falcon import request_helpers as helpers
+from falcon.constants import _UNSET
 from falcon.constants import SINGLETON_HEADERS
-from falcon.forwarded import _parse_forwarded_header  # NOQA: Req. by fixed up WSGI Request attrs
-from falcon.forwarded import Forwarded  # NOQA
-import falcon.media
```
CaselIT, line 22: did this had a particular meaning? it's not used

falcon/asgi/response.py
```diff
@@ -14,18 +14,18 @@
...
 
-from asyncio.coroutines import CoroWrapper  # type: ignore
-from inspect import iscoroutine, iscoroutinefunction
+from asyncio.coroutines import CoroWrapper
+from inspect import iscoroutine
+from inspect import iscoroutinefunction
 
+from falcon import response
 from falcon.constants import _UNSET
-import falcon.media
```
CaselIT, line 21: did this had a particular meaning? it's not used

vytas7: Probably a leftover from different phases of refactoring

falcon/util/uri.py
```diff
@@ -26,10 +26,8 @@
 
 from falcon.constants import PYPY
 try:
-    from falcon.cyutil.uri import (
-        decode as _cy_decode,
-        parse_query_string as _cy_parse_query_string,
-    )
+    from falcon.cyutil.uri import decode as _cy_decode
+    from falcon.cyutil.uri import parse_query_string as _cy_parse_query_string
```
vytas7, line 30: You'll need to either revert, or unroll the `except ImportError` clause for both statements, otherwise one is missed in coverage (as we are not aggregating coverage from tests against the cythonized flavour).

vytas7 (changes requested):
Apart from the missing line in coverage (explained inline), I'm not sure we want to use this pattern for mass-constants or types pertinent to these, i.e., I would leave things as-is for `falcon.constants.*` (for media types), and for `falcon.errors.*` classes corresponding to HTTP status codes.

Other than that, great effort!

CaselIT:
>  I would leave things as-is for falcon.constants.* (for media types), and for falcon.errors.* classes corresponding to HTTP status codes

In those cases I agree, think it makes sense to use the `*` if they are all imports of the same types, like all constants or all errors. I would not use it for a mixed import, like the `utils` one.

I'll check / add an `__all__` to the files so that we export only what's needed and update it.

thanks for the review

CaselIT:
Updated, please take another look

vytas7 (approved):
LGTM (once the merge conflicts are resolved) :+1:

CaselIT:
Resolved conflicts

kgriffs (approved):
LGTM

vytas7 (approved):
:+1:

PR #1866, opened 2021-02-16
perf(asgi): optimize request header decoding

kgriffs:
# Summary of Changes

Avoid decoding of header names when possible, and delay value decoding until
it is necessary.

This should reduce the overhead for constructing Request
objects without significantly increasing overall application performance in
the common case, since we reason that applications do not typically access
all request headers that may be present.

The worst case scenario would be an app that accesses the same header more than
a couple of times, in which case we would have to decode said header multiple
times. If this turns out to be a common use case, we will need to look into
memoization or some other mitigation.

# Related Issues

#1822 

# Pull Request Checklist

### DIFF
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -29,6 +29,9 @@ from .stream import BoundedStream
 __all__ = ['Request']
 
 _UNSET = falcon.request._UNSET
+_SINGLETON_HEADERS_BYTESTR = frozenset([
+    h.encode() for h in SINGLETON_HEADERS
+])
 
 
 class Request(falcon.request.Request):
@@ -347,6 +350,7 @@ class Request(falcon.request.Request):
     __slots__ = [
         '_asgi_headers',
         '_asgi_server_cached',
+        '_cached_headers',
         '_first_event',
         '_receive',
         '_stream',
@@ -364,20 +368,25 @@ class Request(falcon.request.Request):
             # NOTE(kgriffs): According to ASGI 3.0, header names are always
             #   lowercased, and both name and value are byte strings. Although
             #   technically header names and values are restricted to US-ASCII
-            #   we decode using the default 'utf-8' because it is a little
-            #   faster than passing an encoding option.
-            header_name = header_name.decode()
-            header_value = header_value.decode()
+            #   we decode later (just-in-time) using the default 'utf-8' because
+            #   it is a little faster than passing an encoding option (except
+            #   under Cython).
+            #
+            #   The reason we wait to decode is that the typical app will not
+            #   need to decode all request headers, and we usually can just
+            #   leave the header name as a byte string and look it up that way.
+            #
 
             # NOTE(kgriffs): There are no standard request headers that
             #   allow multiple instances to appear in the request while also
             #   disallowing list syntax.
-            if header_name not in req_headers or header_name in SINGLETON_HEADERS:
+            if header_name not in req_headers or header_name in _SINGLETON_HEADERS_BYTESTR:
                 req_headers[header_name] = header_value
             else:
-                req_headers[header_name] += ',' + header_value
+                req_headers[header_name] += b',' + header_value
 
         self._asgi_headers = req_headers
+        self._cached_headers = None
 
         # =====================================================================
         #  Misc.
@@ -422,7 +431,6 @@ class Request(falcon.request.Request):
         self._cached_forwarded = None
         self._cached_forwarded_prefix = None
         self._cached_forwarded_uri = None
-        self._cached_headers = req_headers
         self._cached_prefix = None
         self._cached_relative_uri = None
         self._cached_uri = None
@@ -430,15 +438,15 @@ class Request(falcon.request.Request):
         if self.method == 'GET':
             # PERF(kgriffs): Normally we expect no Content-Type header, so
             #   use this pattern which is a little bit faster than dict.get()
-            if 'content-type' in req_headers:
-                self.content_type = req_headers['content-type']
+            if b'content-type' in req_headers:
+                self.content_type = req_headers[b'content-type'].decode()
             else:
                 self.content_type = None
         else:
             # PERF(kgriffs): This is the most performant pattern when we expect
             #   the key to be present most of the time.
             try:
-                self.content_type = req_headers['content-type']
+                self.content_type = req_headers[b'content-type'].decode()
             except KeyError:
                 self.content_type = None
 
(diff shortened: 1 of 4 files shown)


falcon/asgi/request.py
```diff
@@ -794,26 +802,42 @@ def if_match(self):
...
+
     # ------------------------------------------------------------------------
     # Public Methods
     # ------------------------------------------------------------------------
 
-    def get_header(self, name, required=False, default=None):
+    # PERF(kgriffs): Using kwarg cache, in lieu of @lru_cache on a helper method
+    #   that is then called from get_header(), was benchmarked to be more
+    #   efficient across CPython 3.6/3.8 (regardless of cythonization) and
+    #   PyPy 3.6.
```
kgriffs, line 839: It would be great if someone could check this on Linux/Window to make sure this holds true across platforms.

The comparison in this case that is relevant is between `foo_a()` and `foo_c()` patterns:

```py
def foo_a(name, _cache={}):
    try:
        decoded_name = _cache[name]
    except KeyError:
        decoded_name = name.decode()
        if len(_cache) < 64:
            _cache[name] = decoded_name

    return decoded_name


@lru_cache()
def foo_b(name):
    return name.decode()


def foo_c(name):
    return foo_b(name)
```

CaselIT: The difference seems too small to provide a meaningful difference (meaning that depending in the run one is better than the other:
```
In [13]: %timeit [foo_c(b'foobar%d'%i) for i in range(200)]
124 µs ± 4.44 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [14]: %timeit [foo_a(b'foobar%d'%i) for i in range(200)]
125 µs ± 7.51 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [15]: %timeit [foo_a(b'foobar%d'%i) for i in range(200)]
124 µs ± 5.32 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [17]: %timeit [foo_c(b'foobar%d'%i) for i in range(200)]
125 µs ± 3.6 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

NOTE: I did place a 64 in the lru cache

falcon/asgi/request.py
```diff
@@ -364,20 +368,25 @@ def __init__(self, scope, receive, first_event=None, options=None):
...
 
             # NOTE(kgriffs): There are no standard request headers that
             #   allow multiple instances to appear in the request while also
             #   disallowing list syntax.
-            if header_name not in req_headers or header_name in SINGLETON_HEADERS:
+            if header_name not in req_headers or header_name in _SINGLETON_HEADERS_BYTESTR:
                 req_headers[header_name] = header_value
             else:
-                req_headers[header_name] += ',' + header_value
+                req_headers[header_name] += b',' + header_value
```
CaselIT, line 386: bytes don't have the same optimizations as strings regarding avoiding creating a new object each time, so maybe we could see if append+join is faster? The downside would be that it's a bit ugly tracking which header needs joining and which does not.

falcon/asgi/request.py
```diff
@@ -837,14 +861,19 @@ def get_header(self, name, required=False, default=None):
 
         """
 
-        asgi_name = name.lower()
+        try:
+            asgi_name = _name_cache[name]
+        except KeyError:
+            asgi_name = name.lower().encode()
+            if len(_name_cache) < 64:  # Somewhat arbitrary ceiling to mitigate abuse
```
CaselIT, line 868: this would keep only the first 64 entry forever. not sure if we could use popitem to create an lru-like cache?

I'm not sure if this would make the performance consideration above moot though

CaselIT (approved):
I'm not sure if the two comment I left are actually actionable (or worth acting on)

falcon/asgi/request.py
kgriffs: Good thought. Perhaps we should create an issue to look at this later? I suspect that most request headers for most apps won't have many (if any) duplicate header lines.

falcon/asgi/request.py
kgriffs: Interesting. The kwarg pattern is definitely faster for me both on macOS and from within a pypy docker container.

falcon/asgi/request.py
kgriffs: My reasoning was just that a well-behaved app should actually request a much smaller set of request header names over its lifetime than 64, but just to prevent it from growing unbounded to put a sanity check in there.

CaselIT: Yes I agree on that observation. Was just an observation. As mentioned I'm not sure this comment and the other one are worth acting on

vytas7: I've just verified the below still holds on Py3.8:
```python
>>> previous = 0
>>> for number in range(256):
...     d = {i: i*i for i in range(number)}
...     if sys.getsizeof(d) != previous:
...         previous = sys.getsizeof(d)
...         print(number, previous)
... 
0 64
1 232
6 360
11 640
22 1176
43 2272
86 4696
171 9312
```

Basically, the size of a Python `dict` hash table grows at these numbers of keys in the dict. So it would make sense to cap at 42, 85 etc.

falcon/asgi/request.py
```diff
@@ -794,26 +802,42 @@ def if_match(self):
...
+    def headers(self):
+        # NOTE(kgriffs: First time here will cache the dict so all we
+        # have to do is clone it in the future.
+        if self._cached_headers is None:
+            self._cached_headers = {
+                name.decode(): value.decode()
+                for name, value in self._asgi_headers.items()
+            }
+
+        return self._cached_headers
```
vytas7, line 830: As much as I am a huge fan of comprehensions, a simple loop seems to outperform them more often than not.

On 3.8:
```python
>>> headers = {b'content-type': b'application/yaml', b'x-coolness': b'moderate', b'user-agent': b'Browser', b'via': b'SomeCDN', b'content-length': b'8', b'x-some': b'thing'}
>>> def func1():
...     return {key.decode(): value.decode() for key, value in headers.items()}
... 
>>> func1()
{'content-type': 'application/yaml', 'x-coolness': 'moderate', 'user-agent': 'Browser', 'via': 'SomeCDN', 'content-length': '8', 'x-some': 'thing'}
>>> def func2():
...     result = {}
...     for key, value in headers.items():
...         result[key.decode()] = value.decode()
...     return result
... 
>>> func2()
{'content-type': 'application/yaml', 'x-coolness': 'moderate', 'user-agent': 'Browser', 'via': 'SomeCDN', 'content-length': '8', 'x-some': 'thing'}
>>> timeit.timeit(func1)
1.2187647159989865
>>> timeit.timeit(func1)
1.216893122997135
>>> timeit.timeit(func1)
1.21874491300332
>>> timeit.timeit(func2)
1.1081436029999168
>>> timeit.timeit(func2)
1.1094474159981473
>>> timeit.timeit(func2)
1.116244266002468
```

vytas7 (approved):
LGTM :+1:

falcon/asgi/request.py
kgriffs: Cool, definitely something to keep in mind.

CaselIT: My vote would go to 42 😂

falcon/asgi/request.py
CaselIT: I've added an issue with some potential additional points discussed here and on gitter wrt this issue
https://github.com/falconry/falcon/issues/1870

PR #1869, opened 2021-02-21
test(E2E): add basic E2E tests for SSE & WebSocket

### LINKED ISSUE #1576: Testing: End-to-End SSE & WebSocket
Create an end-to-end Server-Sent Events test using the new ASGI interface. The test should involve an HTML page with an embedded script to log received events to the DOM.

vytas7:
Closes #1576

These tests are very simplistic (for now), but I just wanted to get the ball rolling, as well as to write a somewhat realistic application to test our SSE and WebSocket affordances before the release.

### DIFF
--- a/e2e-tests/conftest.py
+++ b/e2e-tests/conftest.py
@@ -58,6 +58,9 @@ def base_url():
 
 
 @pytest.fixture()
+# NOTE(vytas): sb here is seleniumbase's pytest fixture, see also
+#   "The sb pytest fixture (no class)" at
+#   https://seleniumbase.io/help_docs/syntax_formats/.
 def browser(sb, base_url):
     sb.open(base_url + INDEX)
 
--- a/e2e-tests/conftest.py
+++ b/e2e-tests/conftest.py
@@ -9,6 +9,7 @@ import falcon.testing
 
 
 HERE = pathlib.Path(__file__).resolve().parent
+INDEX = '/static/index.html'
 
 
 @pytest.fixture(scope='session')
@@ -36,10 +37,13 @@ def base_url():
     yield base_url
 
     uvicorn.terminate()
-    uvicorn.communicate(timeout=5)
+
+    # NOTE(vytas): give Unicorn a rather generous time period to stop since it
+    #   is waiting for the browser to close open connections such as WS & SSE.
+    uvicorn.communicate(timeout=30)
 
 
 @pytest.fixture()
 def browser(sb, base_url):
-    sb.open(base_url + '/ping')
+    sb.open(base_url + INDEX)
     return sb
--- a/e2e-tests/conftest.py
+++ b/e2e-tests/conftest.py
@@ -1,3 +1,17 @@
+# Copyright 2020-2021 by Vytautas Liuolia.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import pathlib
 import subprocess
 import time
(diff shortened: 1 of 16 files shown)


e2e-tests/server/chat.py
```diff
@@ -0,0 +1,44 @@
+import re
+
+
+class Chat:
+    ALL = re.compile(r'^/all\s+(.+)$')
+    MSG = re.compile(r'^/msg\s+(\w+)\s+(.+)$')
+
+    def __init__(self, hub):
+        self._hub = hub
+        self._banned = set()
```
CaselIT, line 10: this does not seem to be used

e2e-tests/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+
+    uvicorn.terminate()
+
+    # NOTE(vytas): give Unicorn a rather generous time period to stop since it
+    #   is waiting for the browser to close open connections such as WS & SSE.
+    uvicorn.communicate(timeout=30)
+
+
+@pytest.fixture()
+def browser(sb, base_url):
```
CaselIT, line 61: I'm guessing that `sb` is a selenium base fixture.
It may be useful to add a note (and a link to the documentations?) to improve future readability

e2e-tests/server/hub.py
```diff
@@ -0,0 +1,77 @@
...
+
+    async def broadcast(self, message):
+        for emitter in self._update_emitters():
+            await emitter.enqueue(message)
+
+    async def message(self, name, text):
+        ws = self._users.get(name)
+        if ws:
+            # TODO(vytas): What if this overlaps with another ongoing send?
+            await ws.send_text(text)
```
CaselIT, line 63: would it create an issue?

e2e-tests/static/main.js
```diff
@@ -0,0 +1,82 @@
...
+        this.source = new EventSource(endpoint);
+        this.source.onmessage = this.receiveEvent.bind(this);
+        this.messagesId = messagesId;
+    }
+
+    receiveEvent(event) {
+        let newElement = document.createElement('div');
+        newElement.classList.add('message');
+        newElement.classList.add('sse');
+        newElement.textContent = event.data;
```
CaselIT, line 12: shouldn't we guard form an empty message here? (if `yield None` was sent by the emitter)

CaselIT (commented):
Looks nice, just a couple of minor comments

e2e-tests/static/main.js
vytas7: No, these seem to be some sort of keepalive thing, they are not surfaced as events on the front-end side.

See also [`Response.sse`](https://falcon.readthedocs.io/en/latest/api/request_and_response_asgi.html#falcon.asgi.Response.sse):
```python
        if not some_event:
            # Send an event consisting of a single "ping"
            #   comment to keep the connection alive.
            yield SSEvent()

            # Alternatively, one can simply yield None and
            #   a "ping" will also be sent as above.

            # yield

            continue
```

e2e-tests/conftest.py
vytas7: Ah, right, that's a very good comment :+1: 
In fact, their examples are centered around subclassing some sort of `unittest.TestCase` instance, it took me some googling to find out that a pytest fixture is exposed as `sb`. I'll add a note.

e2e-tests/server/chat.py
vytas7: Heh, I this is probably a leftover from the initial prototype where I had ambitions to build a bigger app. :slightly_smiling_face:

e2e-tests/server/hub.py
vytas7: It's not an issue for this simplistic test where we always wait for things to happen sequentially.
But it was a reminder to myself to dig into what the spec says on the matter, and what different implementations do.

e2e-tests/static/main.js
CaselIT: Oh ok. I looked only at mdn and it was not mentioned that messages with empty body are discarded, but looking at the linker speck it does seems so. From [Interpreting an event stream](https://html.spec.whatwg.org/multipage/server-sent-events.html#dispatchMessage):

> If the data buffer is an empty string, set the data buffer and the event type buffer to the empty string and return.

CaselIT (approved):

kgriffs:
Nice work! Looking forward to some recipes in the docs. 😄

kgriffs (approved):

PR #1875, opened 2021-03-03
dry create wheel workflow using a better action to upload wheels to release

CaselIT:
https://github.com/actions/upload-release-asset is discontinued, and I've found a better action to upload the asset to the release.

it also allows us to dry things quite a bit

### DIFF
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -223,6 +223,7 @@ jobs:
           pip install setuptools>=44 wheel>=0.34 cython>=0.29.21
           # cythonize -3 falcon/**/*.pyx
           python setup.py sdist --dist-dir dist
+          python .github/workflows/scripts/verify_tag.py dist
 
       - name: Upload wheels to release
         # upload the generated wheels to the github release
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -9,7 +9,7 @@ env:
   # set this so the falcon test uses the installed version and not the local one
   PYTHONNOUSERSITE: 1
   # comment TWINE_REPOSITORY_URL to use the real pypi. NOTE: change also the secret used in TWINE_PASSWORD
-  TWINE_REPOSITORY_URL: https://test.pypi.org/legacy/
+  # TWINE_REPOSITORY_URL: https://test.pypi.org/legacy/
 
 jobs:
   # four jobs are defined make-wheel-win-osx, make-wheel-linux, make-source-dist and make-emulated-wheels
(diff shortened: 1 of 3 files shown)


.github/workflows/create-wheels.yaml
```diff
@@ -9,7 +9,7 @@ env:
   # set this so the falcon test uses the installed version and not the local one
   PYTHONNOUSERSITE: 1
   # comment TWINE_REPOSITORY_URL to use the real pypi. NOTE: change also the secret used in TWINE_PASSWORD
-  TWINE_REPOSITORY_URL: https://test.pypi.org/legacy/
+  # TWINE_REPOSITORY_URL: https://test.pypi.org/legacy/
```
CaselIT, line 12: do we want to target the real pipy? (if we do we need to also change the secret in the configuration)

vytas7 (approved):
I guess it is better to just test this in-action, but review-wise it looks great :+1:

.github/workflows/create-wheels.yaml
CaselIT: will wait for kurt, but I can revert that without problems

kgriffs (changes requested):
Thanks for staying on top of this. I would like to try pushing to the real PyPI before the final release. Is there a way to add a check to the GH action to make sure the full version of the wheel including any tags (b2, rc1, etc.) matches the name of the the git tag for the release?

kgriffs:
That would be cool if we could do it, but if it is too much work we can circle back.

CaselIT:
I've tried to give this a go:
- this run should be ok https://github.com/CaselIT/falcon/actions/runs/647249288
- this should error instead https://github.com/CaselIT/falcon/actions/runs/647252686

.github/workflows/scripts/verify_tag.py
```diff
@@ -0,0 +1,47 @@
...
+from os import environ
+from pathlib import Path
+
+# See https://docs.github.com/en/actions/reference/environment-variables
+ENV_VARIABLE = 'GITHUB_REF'
+
+
+def go():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('folder', help='Directory where to look for wheels and source dist')
```
vytas7, line 11: Why wouldn't call the argument `path` or `directory`? `folder` smells Windows :slightly_smiling_face:

.github/workflows/scripts/verify_tag.py
```diff
@@ -0,0 +1,47 @@
...
+    if not candidates:
+        raise RuntimeError('No wheel or source dist found in folder ' + args.folder)
+    raw_value = environ[ENV_VARIABLE]
+    tag_value = raw_value.split('/')[-1]
+
+    errors = []
+    for candidate in candidates:
+        name = candidate.stem
+        if name.endswith('.tar'):
+            name = name[:-4]
```
vytas7, line 29: I would prefer to see an approach base on splitting extension w/ `os.path` or `pathlib`, rather than relying on the exact length of the string `.tar`. Maybe it could even make code look cleaner.

vytas7 (approved):
:+1: 

`verify_tag.py` could probably be bikeshed a bit, bit I realize it is more about stylistic preferences. What is more, it is an auxiliary tool, and not a core part of the framework.

.github/workflows/scripts/verify_tag.py
CaselIT: I started with `dir` but it shadows the built, and directory was too long.

.github/workflows/scripts/verify_tag.py
CaselIT: I do split the extension with pathlib: `candidate.stem`. `.tar.gz` has two extensions, `stem` removes just `.gz`. so at this point I think it's safe to use `.endwith`

kgriffs (approved):
> it is an auxiliary tool

Agreed... LGTM

kgriffs:
Thanks!

PR #1709, opened 2020-04-03
[POC] Adopt pep517 + try compile with cython

CaselIT:
# Summary of Changes

This is just a WIP POC to see if something like this could work.

Add support for pep517 using the setuptools backend. This is based on https://github.com/pypa/setuptools/issues/1698 since currently no documentation exists.
Since we now use pep517 we have cython available during a normal install, so `setuy.py` has been updated to try to compile using cython and then falling back to a normal install. This logic is based mostly on [SQLAlchemy `setup.py](https://github.com/sqlalchemy/sqlalchemy/blob/c7d3ca0da477451885158a923aa9ee7e49794541/setup.py)

# Related Issues

#1683

# Pull Request Checklist

This is just a reminder about the most common mistakes.  Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](../CONTRIBUTING.md) at least once; it will save you a few review cycles!

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing to do.

- [ ] Applied changes to both WSGI and ASGI code paths and interfaces (where applicable).
- [ ] Added **tests** for changed code.
- [ ] Prefixed code comments with GitHub nick and an appropriate prefix.
- [ ] Coding style is consistent with the rest of the framework.
- [ ] Updated **documentation** for changed code.
    - [ ] Added docstrings for any new classes, functions, or modules.
    - [ ] Updated docstrings for any modifications to existing code.
    - [ ] Updated both WSGI and ASGI docs (where applicable).
    - [ ] Added references to new classes, functions, or modules to the relevant RST file under `docs/`.
    - [ ] Updated all relevant supporting documentation files under `docs/`.
    - [ ] A copyright notice is included at the top of any new modules (using your own name or the name of your organization).
    - [ ] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/directives.html?highlight=versionadded#directive-versionadded).
- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/index.html) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)

If you have *any* questions to *any* of the points above, just **submit and ask**! This checklist is here to *help* you, not to deter you from contributing!

*PR template inspired by the attrs project.*

PR #1876, opened 2021-03-03
style: adopt black

### LINKED ISSUE #1453: Consider using black to normalize coding style
See: https://github.com/ambv/black

CaselIT:
Just to get the ball rolling.

Tried to run black ~~20.8b1~~ 21.5b1 with the configuration
```toml
[tool.black]
    target-version = ["py35"]
    skip-string-normalization = true
    line-length = 88
```
The `skip-string-normalization` avoids the conversion `'` to `"` if we want to keep the current flake8 rules

TODO:
- [x] re-run with line length 88
- [x] cleanup string concatenation in the same line (like `'foo bar ' 'baz foobar'`)
- [x] review where we need to add / remove tailing commas to improve the styling
- [ ] update the documentation with the changes done to the example folder

Closes #1453


examples/look/look/images.py
```diff
@@ -8,18 +8,11 @@
...
 
     def on_get(self, req, resp):
-        doc = {
-            'images': [
-                {
-                    'href': '/images/1eaf6ef1-7f2d-4ecc-a8d5-6e8adba7cc0e.png'
-                }
-            ]
-        }
+        doc = {'images': [{'href': '/images/1eaf6ef1-7f2d-4ecc-a8d5-6e8adba7cc0e.png'}]}
```
CaselIT, line 15: NOTE: this could be improved a bit: placing a comma `,` at the end of the end of a definition makes black use new lines, so we can force expanded formatting for dict/string/argument lists, so that the formatted version is closer to the existing one. It would become:

```py
        doc = {
            'images': [
                {
                    'href': '/images/1eaf6ef1-7f2d-4ecc-a8d5-6e8adba7cc0e.png',
                },
            ],
        }
```

examples/things_advanced.py
```diff
@@ -45,48 +43,53 @@ def __call__(self, req, resp, engine):
...
     def process_request(self, req, resp):
         token = req.get_header('Authorization')
         account_id = req.get_header('Account-ID')
 
         challenges = ['Token type="Fernet"']
 
         if token is None:
-            description = ('Please provide an auth token '
-                           'as part of the request.')
+            description = 'Please provide an auth token ' 'as part of the request.'
```
vytas7, line 53: These changes look odd. We would probably need to carefully review them manually if we want to keep the same high standard of Falcon's source code.

kgriffs:
I am going to delay this for post 3.0 but I do think we should move in this direction. Let's just make sure we have as part of the process a review to tweak the occasional oddity.

CaselIT:
> 
> 
> I am going to delay this for post 3.0 but I do think we should move in this direction. Let's just make sure we have as part of the process a review to tweak the occasional oddity.

yes sure, I was not targeting this release, it was just to start looking into it

examples/things_advanced.py
CaselIT: reaching for `' '` finds 92 results. I think it would be manageable

CaselIT:
I've updated the change with the latest black.

kgriffs:
The default of 88 is fine, and avoids excessive line wrapping esp. for deeply nested control structures. If we increased the limit much further I would argue for a separate standard for keeping comments shorter (even if we have to write our own checker for it).

Regardless, I'll take a look through the diff soon and see how things look now.

CaselIT:
I'll re-reun with the 88 line.

If you are not familiar with black, note that you can tell it to split things in multiple lines by setting a comma at the end.
Like
```py
def foo(x,):
  pass
# becomes
def foo(
  x,
):
  pass
````
So we may need to add / remove some to get the desired formatting.

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html?highlight=comma#the-magic-trailing-comma

**Comment by vytas7** [comment]
I'm a bit orthodox when it comes to line length, 100+ sounds a bit too long.

But I'm fine with `black`'s 88, it does sound like a good compromise, and... if we're going to use `black`, we don't need to care about manually keeping lines at 79 or 88 anyway, that's what `black` is for.

File: examples/things_advanced.py (hunk 2 above)
Comment (vytas7) [reply examples/things_advanced.py:53]: Wondering if `black` is making any progress in this area? :thinking: 
IIRC there were discussions to handle this out of the box, but I haven't checked what the status is.

Comment (CaselIT) [reply examples/things_advanced.py:53]: Not much. It now remosed leading and trailing spaces in triple comments where it makes sense.

But thsi caae needs to be handled by hand.

**Comment by CaselIT** [comment]
Ok, I initially went with 100 both because we do have some lines that are on the long side in the source and it's what I usually use.

I'll re-run with the default 88 and update this.

File: examples/look/look/images.py (hunk 1 above)
Comment (vytas7) [reply examples/look/look/images.py:15]: IIRC this look stuff used to be copypasted, so we'll need to sync, or use "literal include" like I did in the ASGI version (thanks @CaselIT).

File: examples/things_advanced.py (hunk 2 above)
Comment (vytas7) [reply examples/things_advanced.py:53]: I think the intention is to handle all strings: https://github.com/psf/black/issues/2188

But it seems there are still some issues with it.

File: examples/look/look/images.py (hunk 1 above)
Comment (CaselIT) [reply examples/look/look/images.py:15]: ok let's add some points to do

File: examples/things_advanced.py (hunk 2 above)
Comment (CaselIT) [reply examples/things_advanced.py:53]: ah, sqlalchemy makes black file, ahah :)

**Comment by CaselIT** [comment]
Updated with 88. From the change lines it seems that our average line was a bit longer, since this version adds a 1000 new lines.

For reference the previous with length 100 was more or less a wash regarding line count

File: examples/things_advanced.py (hunk 2 above)
Comment (CaselIT) [reply examples/things_advanced.py:53]: tried running it. not particularly convinced:

see the diff https://gist.github.com/CaselIT/2e509ce7b7425700a87f190669baeb1c

**Review by kgriffs — CHANGES_REQUESTED** [review]
Comments inline. I am still favorable to adopting black modulo a few tweaks. Also, note that f we miss any strange formatting issues before merging this, we can always fix those up over time, as we discover them in the course of maintaining the project.

File: falcon/app.py [hunk 3]
```diff
@@ -1087,7 +1109,8 @@ class API(App):
     removed in a future release.
     """
 
-    @deprecation.deprecated('API class may be removed in a future release, '
-                            'use falcon.App instead.')
+    @deprecation.deprecated(
+        'API class may be removed in a future release, ' 'use falcon.App instead.'
```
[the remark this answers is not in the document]
CaselIT: sure I've yet to make this pass

falcon/media/validators/jsonschema.py
```diff
@@ -106,22 +106,19 @@ def wrapper(self, req, resp, *args, **kwargs):
...
                 jsonschema.validate(
-                    req.media, req_schema,
-                    format_checker=jsonschema.FormatChecker()
+                    req.media, req_schema, format_checker=jsonschema.FormatChecker()
                 )
             except jsonschema.ValidationError as e:
                 raise falcon.HTTPBadRequest(
-                    title='Request data failed validation',
-                    description=e.message
+                    title='Request data failed validation', description=e.message
```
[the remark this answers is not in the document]
CaselIT: well, it depends if it fits or not on the line (without tailing comma that is).

it's usually ok, and for the cases where it's better as single-lines adding a comma it's easy. I would guess that we will adjust the formatting a bit over time, to accommodate for these when we modify a file that's formatted funny :)

CaselIT:
I'll wait for https://github.com/falconry/falcon/pull/1709 before rebasing and polishing this

examples/look/tests/test_app.py
```diff
@@ -48,9 +42,7 @@ def test_post_image(client, mock_store):
     image_content_type = 'image/xyz'
 
     response = client.simulate_post(
-        '/images',
-        body=b'some-fake-bytes',
-        headers={'content-type': image_content_type}
+        '/images', body=b'some-fake-bytes', headers={'content-type': image_content_type}
```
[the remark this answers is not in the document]
kgriffs: > always add magic commas to be safe

...for code samples that will be included in the docs, that is.

docs/user/quickstart.rst
```diff
@@ -17,46 +17,8 @@ started writing an app.
...
-            app.add_route('/things', things)
-
-            if __name__ == '__main__':
-                with make_server('', 8000, app) as httpd:
-                    print('Serving on port 8000...')
-
-                    # Serve until process is killed
-                    httpd.serve_forever()
+        .. literalinclude:: ../../examples/things.py
+            :language: python
```
CaselIT, line 21: `literalinclude` ftw

CaselIT:
@kgriffs @vytas7 I've updated the docs to use the new formatting, bug I'm not sure I updated all of them.

I've also made a quick pass to the files to check that no funny formatting was added. For this I think we can correct eventual leftover in the future

setup.cfg
```diff
@@ -85,7 +85,7 @@ filterwarnings =
 [flake8]
 max-complexity = 15
 exclude = .ecosystem,.eggs,.git,.tox,.venv,build,dist,docs,examples,falcon/bench/nuts
-ignore = F403,W504
+extend-ignore = F403,W504,E203,I202
 max-line-length = 99
```
vytas7, line 89: Shouldn't we now decrease this to 88?

CaselIT: Yes, I guess so

tests/asgi/test_boundedstream_asgi.py
```diff
@@ -154,14 +164,18 @@ async def test_iteration():
     falcon.async_to_sync(test_iteration)
 
 
-@pytest.mark.parametrize('body', [
```
vytas7, line 157: Ah, these test fixtures got a bit ugly, but I guess we'll have to live with that.
Unless there is some magic opening parenthesis out there :laughing:

CaselIT: I actually don't mind them, but it may be that I'm used to them

vytas7: Me neither, it's just a massive refactoring of something that was aesthetically OK/better before.

CaselIT: I guess it's inevitable when introducing a formatter that a large number of pointless changed are done.

setup.cfg
CaselIT: Note that we have quite a bit of comment/documentation that are over 88 chars

kgriffs (dismissed):
💥

falcon/routing/compiled.py
```diff
@@ -224,13 +224,19 @@ def insert(nodes, path_index=0):
                     return
 
                 if node.conflicts_with(segment):
-                    msg = textwrap.dedent("""
+                    msg = (
+                        textwrap.dedent(
+                            """
                         The URI template for this route is inconsistent or conflicts with another
```
vytas7, line 230: Shouldn't the text be moved further left to match the triple quotes?

falcon/util/uri.py
```diff
@@ -25,21 +25,19 @@
...
     _cy_decode = None
     _cy_parse_query_string = None
 
 
 # NOTE(kgriffs): See also RFC 3986
-_UNRESERVED = ('ABCDEFGHIJKLMNOPQRSTUVWXYZ'
-               'abcdefghijklmnopqrstuvwxyz'
-               '0123456789'
-               '-._~')
+_UNRESERVED = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~'
```
vytas7, line 40: Does `black` insist on joining these, or is possible to leave the separation as-is?

falcon/vendor/mimeparse/mimeparse.py
```diff
@@ -31,8 +31,7 @@ def parse_mime_type(mime_type):
 
     type_parts = full_type.split('/') if '/' in full_type else None
     if not type_parts or len(type_parts) > 2:
-        raise MimeTypeParseException(
-            "Can't parse type \"{}\"".format(full_type))
+        raise MimeTypeParseException("Can't parse type \"{}\"".format(full_type))
```
vytas7, line 34: Maybe we should exclude vendored modules from blackening?

It might make it more cumbersome to backport changes/sync with upstream, even though any changes are looking unlikely at this point.

vytas7 (commented):
This is getting almost ready :+1: 

I just had a handful of open questions (commented inline).

And a more general question: should we try updating Cython code for the delimited stream reader etc? When I wrote it, it was loosely based on the Python version, with changes specific to Cython. Right now, that parity would be disturbed a bit. OTOH mangling that code now after the fact is just a (an unnecessary?) risk to introduce regressions.

falcon/util/uri.py
CaselIT: yes, since they fit in the same line it will do `'abcdefghijklmnopqrstuvwxyz' '0123456789' '-._~'` I joined them together.

we could mark the section as format off, or use a tuple and join them. But I think it's fine to have it as one line.

Also maybe we could use something linke `string.ascii_letters + string.digits + '-._~'` ?

falcon/vendor/mimeparse/mimeparse.py
CaselIT: oh good point. let me try to exclude them.

falcon/routing/compiled.py
CaselIT: I'll check if it can be tweaked

falcon/util/uri.py
vytas7: This one is not critical for me, I just thought that the previous indentation was intentional.

CaselIT: > I just thought that the previous indentation was intentional.

that's most likely true, but I think it's still understandable that's ascii + digits + some special chars

CaselIT:
> And a more general question: should we try updating Cython code for the delimited stream reader etc? When I wrote it, it was loosely based on the Python version, with changes specific to Cython.

black does not support cython, so the formatting would be manual.

Personally I think we could leave it as is or update as needed as a follow on change (maybe in the same issue as the one setting flake line length to 88)

falcon/util/uri.py
kgriffs: string.ascii_letters + string.digits + '-._~' sounds like a good way to make it a little more self-documenting, but not critical to change it.

CaselIT: The only thing to consider is that with the concatenation form the resulting string may not be interned by python (iirc the string literal are).

If we want to change this a follow on pr is probably best

kgriffs (approved):
👍

vytas7 (approved):
Here we go then...

CaselIT:
I've opened https://github.com/falconry/falcon/issues/1926 to do the follow up on this change.

Please add there if there were other deferred points from this change. I'll try to get to it this we

Issue #1872, opened 2021-02-28
falcon.uri.encode_value does not encode the percent character


vytas7:
[`falcon.uri.encode_value`](https://falcon.readthedocs.io/en/latest/api/util.html#falcon.uri.encode_value) says:
> An escaped version of *uri*, where all disallowed characters have been percent-encoded.

However, it seems that the percent character itself is not encoded...

Cf
```python
>>> from urllib.parse import quote
>>> from falcon import uri
>>> quote('%26')
'%2526'
>>> uri.encode('%26')
'%26'
>>> uri.encode_value('%26')
'%26'
>>> uri.decode(uri.encode_value('%26'))
'&'
>>> uri.decode(quote('%26'))
'%26'
```

vytas7:
Found this when trying to fix #1871...

MinesJA:
I'd be interested in taking this.

vytas7:
That's awesome @MinesJA !
(If you're quick, this might still make it into 3.0.)

MinesJA:
Will try to be quick then!

MinesJA:
So I think I found the cause of the bug, but had some questions on the functionality we're looking for. In this snippet, we check to see if a string has already been encoded, so if we encounter a string like 'abc%26def' we assume the '%26' was a result of the string already being encoded: 

`tokens = uri.split('%')
  for token in tokens[1:]:
      hex_octet = token[:2]

      if not len(hex_octet) == 2:
          break

      if not (hex_octet[0] in _HEX_DIGITS and
                  hex_octet[1] in _HEX_DIGITS):
                    break
      else:
         # NOTE(kgriffs): All percent-encoded sequences were
         # valid, so assume that the string has already been
         # encoded.
         return uri`

But it looks like 'quote' doesn't do that. I did notice this comment and was wondering why it was that we think there's a good chance the string has already been escaped by this point?

`# NOTE(kgriffs): There's a good chance the string has already
  # been escaped. Do one more check to increase our certainty.`

vytas7:
Heh, we need to check with @kgriffs then.

However, personally, I think this is a bug regardless of the good intentions mentioned above.
There might be cases where one wants to encode the whole URI as a parameter, and pass it along with other query parameters. (see, for instance this exotic use case: https://github.com/falconry/falcon/issues/1856).

Not encoding already escaped characters could then mangle that URI in the same way as, for instance, `to_query_str` was doing in https://github.com/falconry/falcon/issues/1871.

[...]

Issue #68, opened 2013-02-05
Automatically handle the OPTIONS method


kgriffs:
If not defined, add a stub that will respond with Allows and Accept-Patch (the latter will be set if on_patch is specified, and will assume application/json-patch

PR #1879, opened 2021-03-07
fix(uri): change encode to escape percent

MinesJA:
# Summary of Changes

Changed the default behavior of `uri.encode` and `uri.envode_value` to escape all escapable characters even if they look like they've already been escaped. For example, '%26' would be escaped as '%2526' to conform with behavior of `urllib.parse.quote`. This ensures that decoding an encoded value returns the same original value. Previously, encoding '%26' returned '%26' which, decoded, become '&'.

A `check_is_escaped` flag has been added to the encoder method to allow the ability to retain the original behavior of ignoring previously escaped values (currently being used in `location`, `content_location`, and `append_link()`).

# Related Issues

Address bug in #1872 while retaining fix for #68.

# Pull Request Checklist

- [X] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/index.html) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -16,19 +16,20 @@
 
 import mimetypes
 
-from falcon import DEFAULT_MEDIA_TYPE
 from falcon.constants import _UNSET
+from falcon.constants import DEFAULT_MEDIA_TYPE
 from falcon.errors import HeaderNotSupported
 from falcon.media import Handlers
-from falcon.response_helpers import (
-    format_content_disposition,
-    format_etag_header,
-    format_header_value_list,
-    format_range,
-    header_property,
-    is_ascii_encodable,
-)
-from falcon.util import dt_to_http, http_cookies, structures, TimezoneGMT
+from falcon.response_helpers import format_content_disposition
+from falcon.response_helpers import format_etag_header
+from falcon.response_helpers import format_header_value_list
+from falcon.response_helpers import format_range
+from falcon.response_helpers import header_property
+from falcon.response_helpers import is_ascii_encodable
+from falcon.util import dt_to_http
+from falcon.util import http_cookies
+from falcon.util import structures
+from falcon.util import TimezoneGMT
 from falcon.util.deprecation import deprecated
 from falcon.util.uri import encode_check_escaped as uri_encode
 from falcon.util.uri import encode_value_check_escaped as uri_encode_value
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -14,7 +14,6 @@
 
 """Response class."""
 
-from functools import partial
 import mimetypes
 
 from falcon import DEFAULT_MEDIA_TYPE
@@ -31,12 +30,9 @@ from falcon.response_helpers import (
 )
 from falcon.util import dt_to_http, http_cookies, structures, TimezoneGMT
 from falcon.util.deprecation import deprecated
-from falcon.util.uri import encode
-from falcon.util.uri import encode_value
+from falcon.util.uri import encode_check_escaped as uri_encode
+from falcon.util.uri import encode_value_check_escaped as uri_encode_value
 
-
-uri_encode = partial(encode, check_is_escaped=True)
-uri_encode_value = partial(encode_value, check_is_escaped=True)
 
 GMT_TIMEZONE = TimezoneGMT()
 
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -501,8 +501,7 @@ class Response:
             same_site = same_site.lower()
 
             if same_site not in _RESERVED_SAMESITE_VALUES:
-                raise ValueError(
-                    "same_site must be set to either 'lax', 'strict', or 'none'")
+                raise ValueError("same_site must be set to either 'lax', 'strict', or 'none'")
 
             self._cookies[name]['samesite'] = same_site.capitalize()
 
@@ -607,8 +606,7 @@ class Response:
         name = name.lower()
 
         if name == 'set-cookie':
-            raise HeaderNotSupported(
-                'Getting Set-Cookie is not currently supported.')
+            raise HeaderNotSupported('Getting Set-Cookie is not currently supported.')
 
         return self._headers.get(name, default)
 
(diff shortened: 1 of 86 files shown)


falcon/response.py
```diff
@@ -31,10 +32,13 @@
 from falcon.util import structures
 from falcon.util import TimezoneGMT
 from falcon.util.deprecation import deprecated
-from falcon.util.uri import encode as uri_encode
-from falcon.util.uri import encode_value as uri_encode_value
+from falcon.util.uri import encode
+from falcon.util.uri import encode_value
 
 
+uri_encode = partial(encode, check_is_escaped=True)
```
vytas7, line 39: We're a bit hardcore when it comes to performance of the key parts of the framework, and IIRC it is more performant to simply call the function with the desired _kwargs_ than to incur function call overhead via `functools.partial`.

vytas7 (changes requested):
LGTM bar the performance nitpick.

falcon/response.py
MinesJA: Ah got it, good to know. Will make the change

vytas7: Just to illustrate I'm not making this up (CPython 3.8) :slightly_smiling_face: 

```python
>>> def my_func(param, futile_flag=False):
...     return param.upper()
... 
>>> my_func('hello!')
'HELLO!'
>>> another_func = functools.partial(my_func, futile_flag=True)
>>> timeit.timeit(lambda: my_func('hello', futile_flag=True))
0.13354026100000738
>>> timeit.timeit(lambda: my_func('hello', futile_flag=True))
0.13823660299999574
>>> timeit.timeit(lambda: my_func('hello', futile_flag=True))
0.13869832399996085
>>> 
>>> timeit.timeit(lambda: another_func('hello'))
0.2074820359999876
>>> timeit.timeit(lambda: another_func('hello'))
0.2078785850000031
>>> timeit.timeit(lambda: another_func('hello'))
0.20846412799994596
```

falcon/util/uri.py
```diff
@@ -161,6 +168,8 @@ def encoder(uri):
     uri (str): URI fragment to encode. It is assumed not to cross delimiter
         boundaries, and so any reserved URI delimiter characters
         included in it will be percent-encoded.
+    check_is_escaped (boolean): Indicates a uri may have already been escaped
+        and should be treated as such
```
vytas7, line 172: Full stop (period) missing at the end of the sentence.

falcon/response.py
MinesJA: No way.....is that a 50% difference?? That's wild. I had no idea. Will try to fix this today.

MinesJA: Was just about to push something but noticed that header_property (for content_location in response.py::913) takes a "transformer" method as an argument and then applies a value to that method. I was wondering if in this case it might be ok to use the partial?
    
Alternatively, I was originally thinking of exposing 2 more methods on the URI as encode_check_escaped and encode_check_escaped_value and then just using those in response.

falcon/util/uri.py
```diff
@@ -132,6 +137,8 @@ def encoder(uri):
 
 Args:
     uri (str): URI or part of a URI to encode.
+    check_is_escaped (boolean): Indicates a uri may have already been escaped
+        and should be treated as such
```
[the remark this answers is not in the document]
kgriffs: Don't forget to mention the default value.

falcon/util/uri.py
kgriffs: Don't forget to mention the default value.

falcon/response.py
kgriffs: I'd like to try moving `check_is_escaped` to `_create_str_encoder()` and letting the closure do the work more efficiently as compared to `partial()`.  Having two additional methods would be fine... we can also let people know (in the news fragment) that if they were using `encode()` and `encode_value()` directly, they can work around the breaking change by using the new additional methods.

That being said, I would prefer `encode_value_check_escaped()` in lieu of `encode_check_escaped_value()`. There may also be a more elegant way to say it than "check escaped" if we give it a little more thought.

Thanks for tackling this!

kgriffs (changes requested):

falcon/response.py
MinesJA: Ok, great, I'll make that change today.

kgriffs:
@MinesJA Thanks again for your help on this. We would like to include this fix in 3.0.0rc1 if possible. Do you have time to address the feedback items today? If not, no problem, just let me know and I can put up some patches on your feature branch.

MinesJA:
> n for your help on this. We would like to include this fix in 3.0.0rc1 if possible. Do you have time to address the feedback it

For sure! I'll push something up in the next couple hours if that's ok.

MinesJA:
Ready for review @kgriffs @vytas7

falcon/util/uri.py
```diff
@@ -168,6 +172,59 @@ def encoder(uri):
...
+"""
+
+encode_value_check_escaped = _create_str_encoder(True, True)
+encode_value_check_escaped.name = 'encode_value_check_escaped'
+encode_value_check_escaped.__doc__ = """Encodes a value string according to RFC 3986.
+
+RFC 3986 defines a set of "unreserved" characters as well as a
+set of "reserved" characters used as delimiters. Disallowed characters
+are percent-encoded in a way that models ``urllib.parse.quote(safe="~")``
+unless they appear to have been previously encoded. For example, '%26'
```
kgriffs, line 206: 👍

falcon/util/uri.py
```diff
@@ -168,6 +172,59 @@ def encoder(uri):
...
 """
 
+encode_check_escaped = _create_str_encoder(False, True)
+encode_check_escaped.name = 'encode_check_escaped'
+encode_check_escaped.__doc__ = """Encodes a full or relative URI according to RFC 3986.
+
+RFC 3986 defines a set of "unreserved" characters as well as a
+set of "reserved" characters used as delimiters. This function escapes
+all other "disallowed" characters by percent-encoding them unless they
+appear to have been previously encoded. For example, '%26' will not be
```
kgriffs, line 182: 👍

kgriffs (approved):
LGTM

vytas7 (approved):
Good stuff :100:

PR #1883, opened 2021-03-13
fix(deprecation): do not rely on interpret-specific stack frame introspection

### LINKED ISSUE #1882: deprecated() utility raises AttributeError under Meinheld
The inner function of our [`deprecated()`](https://falcon.readthedocs.io/en/latest/api/util.html#falcon.deprecated) utility generator grabs the current stack frame object via [`inspect.getcurrentframe()`](https://docs.python.org/3/library/inspect.html#inspect.currentframe), and then uses its attributes to provide a more informative deprecation warning.

However, as warned in the latter's docs, this function is not guaranteed to return a valid stack frame object on all Python implementations; it may also return `None`. It seems that running Gunicorn+Meinheld workers can trigger this situation even under CPython.

Discovered using the following command line under CPython 3.7 and 3.8:
```
gunicorn --workers=8 --worker-class="egg:meinheld#gunicorn_worker" test:app
```

For instance, assigning a value to the deprecated [`Response.body`](https://falcon.readthedocs.io/en/latest/api/request_and_response_wsgi.html#falcon.Response.body) yields
```
2021-03-11 23:31:42 [FALCON] [ERROR] GET /things => Traceback (most recent call last):
  File "falcon/app.py", line 361, in falcon.app.App.__call__
  File "/tmp/benchmark/test3.py", line 13, in on_get
    resp.body = ('\nTwo things awe me most, the starry sky '
  File "falcon/util/deprecation.py", line 67, in falcon.util.deprecation.deprecated.decorator.wrapper
AttributeError: 'NoneType' object has no attribute 'f_code'
```

vytas7:
Closes #1882

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -42,6 +42,7 @@ jobs:
           - "e2e_chrome"
           - "e2e_firefox"
           - "no_optional_packages"
+          - "wsgi_servers"
           # TODO(kgriffs): Re-enable once hug has a chance to address
           # breaking changes in Falcon 3.0
           # - "hug"
--- a/docs/_newsfragments/1882.bugfix.rst
+++ b/docs/_newsfragments/1882.bugfix.rst
@@ -1,0 +1,5 @@
+Previously, methods marked with the :func:`~falcon.deprecated` utility wrapper
+could raise an unexpected ``AttributeError`` when running under certain
+applications servers such as Meinheld. This has been fixed so that
+:func:`~falcon.deprecated` no longer relies on the availability of
+interpreter-specific stack frame instrospection capabilites.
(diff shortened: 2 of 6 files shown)


vytas7:
I added some WSGI server tests, although they don't always serve as a perfect regression tests against the bug in question (it seems to be a combination of various factors, but it can sometimes be reproduced on C-based uWSGI and Meinheld).

There's an open question whether the tooling here is good enough, or we should add better checks whether the `uWSGI` executable is coming from the user's environment, and not from system-wide `PATH`. This doesn't affect _Actions_, but it might cause unexpected failures when running `tools/mintest.sh` locally.

CaselIT (approved):
nice test addition

kgriffs (approved):

PR #1885, opened 2021-03-15
tests: make WSGI server tests more resilient

vytas7:
# Summary of Changes

The server startup and teardown machinery in `tests/test_wsgi_servers.py` had a couple of PEBCAKs, plus I was able to reproduce occasional hanging of Meinheld workers.
This patch attempts to rectify these issues, and use Gunicorn's standard machinery to terminate unresponsive workers.

### DIFF
--- a/tests/test_wsgi_servers.py
+++ b/tests/test_wsgi_servers.py
@@ -33,10 +33,10 @@ def _gunicorn_args(host, port, extra_opts=()):
         '--bind', '{}:{}'.format(host, port),
 
         # NOTE(vytas): Although rare, but Meinheld workers have been noticed to
-        # occasionally hang on shutdown.
+        #   occasionally hang on shutdown.
         '--graceful-timeout', str(_SHUTDOWN_TIMEOUT // 2),
         # NOTE(vytas): In case a worker hangs for an unexpectedly long time
-        # while reading or processing request (the default value is 30).
+        #   while reading or processing request (the default value is 30).
         '--timeout', str(_REQUEST_TIMEOUT),
     )
     return args + extra_opts + ('_wsgi_test_app:app',)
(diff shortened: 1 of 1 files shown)


tests/test_wsgi_servers.py
```diff
@@ -29,6 +31,8 @@ def _gunicorn_args(host, port):
         'gunicorn',
         '--access-logfile', '-',
         '--bind', '{}:{}'.format(host, port),
+        '--graceful-timeout', '10',
+        '--timeout', '10',
```
CaselIT, line 35: shouldn't there also use `_STARTUP_TIMEOUT`?

vytas7: Umm.. no, not really. If anything, `--graceful-timeout` could be somehow derived from `_SHUTDOWN_TIMEOUT` so that `--graceful-timeout` < `_SHUTDOWN_TIMEOUT`.

`--timeout` I added just in case, it should actually constitute a test failure if that happens, but we would a see a log item in Gunicorn's `stderr`, so it would be easier to reason about what has happened.

vytas7: I could maybe set `--graceful-timeout` to `_SHUTDOWN_TIMEOUT // 2`, but not sure if it would make anything clearer?

CaselIT: I think the issue is that they have the same value, 10, so they same related. Maybe changing one of the makes sense?

vytas7: I can add `NOTE`s...

vytas7: I've hopefully clean this up now. Let me know what you think.

CaselIT (dismissed):

tests/test_wsgi_servers.py
CaselIT: ok now. btw above instead of issue I should have said that I made that remark because both had the same value

CaselIT (approved):

PR #1886, opened 2021-03-16
doc: reorganize the features list

kgriffs:
This patch tweaks the features list to bubble up Falcon's more unique value propositions.


vytas7 (approved):
:100:

PR #1893, opened 2021-03-25
doc: fix broken image links

kgriffs:


vytas7 (changes requested):
The changes look good, but we need to do the same for `README.rst` as well.

vytas7 (approved):
LGTM :+1:

PR #1900, opened 2021-04-04
chore(http_error): remove deprecated code

jkklapp:
Part of #1853


vytas7 (changes requested):
Hi @jkklapp!
This looks good, we just need to get the newsfragment for #1853 started.

kgriffs:
@vytas7 I added the fragment

vytas7 (approved):
I've taken the liberty of fixing a misspelling in a removed class's name, looks good otherwise :+1:
Thanks @jkklapp & @kgriffs!

vytas7:
It seems there is an unrelated error in the CPython 3.5 gate. There is an issue to remove 3.5 & 3.6, so either we do that first, or we can try to fix the gate for now.

CaselIT:
> It seems there is an unrelated error in the CPython 3.5.2 gate. There is an issue to remove 3.5 & 3.6, so either we do that first, or we can try to fix the gate for now.

Probably best to remove 3.5 first, since it does not make much sense to fix the gate to then remove it

vytas7:
I've just removed that 3.5.2 gate. We still have the newest 3.5 gate left that ought to be enough until we remove 3.5 & 3.6 support altogether

PR #1904, opened 2021-04-06
fix(api_helpers): re-add falcon.api_helpers

### LINKED ISSUE #1902: Unannounced breaking change in renaming falcon.api_helpers
The `falcon.api_helpers` module used to contain several semi-public (not documented, but not underscored either) utilities, that people may be relying on.

During the 3.0 release cycle, `falcon.api_helpers` was renamed to `falcon.app_helpers` in order to reflect the `API` :arrow_right: `App` renaming.
However, the old name disappeared without any deprecation notice, and without any information in the 3.0 changelog notes. This is a bug.

In the spirit of SemVer, we cannot afford marking the change as deprecation in 3.0.1 either.
Let us make the name available as-is in 3.0.1, and mark as deprecated in 3.1; remove in 4.0.

CaselIT:
# Summary of Changes

Re-add removed api_helpers package

# Related Issues

closes #1902

### DIFF
--- a/docs/_newsfragments/1902.rst
+++ b/docs/_newsfragments/1902.rst
@@ -1,0 +1,2 @@
+Re-add ``api_helpers`` module that was renamed ``app_helpers``.
+This modules is to consider deprecated and will be removed in a future falcon version.
--- a/docs/_newsfragments/1902.rst
+++ b/docs/_newsfragments/1902.rst
@@ -1,2 +1,2 @@
 Re-add ``api_helpers`` module that was renamed ``app_helpers``.
-This modules is to consider deprecated and will be removed in a future falcon version.
+This module is considered deprecated, and will be removed in a future Falcon version.
--- a/falcon/api_helpers.py
+++ b/falcon/api_helpers.py
@@ -1,0 +1,6 @@
+from .app_helpers import * # NOQA
+
+import warnings
+from .util.deprecation import DeprecatedWarning
+
+warnings.warn('The api_helpers was renamed to app_helpers.', DeprecatedWarning)
--- a/falcon/api_helpers.py
+++ b/falcon/api_helpers.py
@@ -4,4 +4,4 @@ from .app_helpers import * # NOQA
 # import warnings
 # from .util.deprecation import DeprecatedWarning
 
-# warnings.warn('The api_helpers was renamed to app_helpers.', DeprecatedWarning)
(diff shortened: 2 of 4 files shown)


docs/_newsfragments/1902.rst
```diff
@@ -0,0 +1,2 @@
+Re-add ``api_helpers`` module that was renamed ``app_helpers``.
+This modules is to consider deprecated and will be removed in a future falcon version.
```
vytas7, line 2: > _This modules (sic) is to consider (sic) deprecated and will be removed in a future falcon version._

This sentence could be cleaned up a bit... maybe something along the lines of :arrow_down: 
> This module is considered deprecated, and will be removed in a future Falcon version.

falcon/api_helpers.py
```diff
@@ -0,0 +1,7 @@
+from .app_helpers import * # NOQA
+
+# TODO deprecate
+# import warnings
+# from .util.deprecation import DeprecatedWarning
+
+# warnings.warn('The api_helpers was renamed to app_helpers.', DeprecatedWarning)
```
vytas7, line 7: The `api_helpers` module or just `api_helpers`, IMHO at least.

vytas7 (changes requested):
Looks great. :+1: 

I just nitpicked on the news fragment, it seems the second sentence therein could be improved a bit.

CaselIT:
thanks, updated

vytas7 (approved):
:+1:

falcon/api_helpers.py
```diff
@@ -0,0 +1,7 @@
+from .app_helpers import * # NOQA
```
kgriffs, line 1: Nit: Should be two spaces before the `#`

kgriffs (approved):

falcon/api_helpers.py
CaselIT: nit; using a formatter would take care of it :P

floqqi:
Is there a chance of getting these changes released? It prevents us from upgrading to falcon 3 as we rely on the sentry sdk.

vytas7:
Hi @floqqi! 
Yes, we are planning to cut 3.0.1 shorty, hopefully today or tomorrow.

We will be addressing https://github.com/falconry/falcon/pull/1911 in the same bugfix release.

PR #1914, opened 2021-05-02
fix(ASGI): assume ISO-8859-1 for ASGI headers 

### LINKED ISSUE #1911: Assume ISO-8859-1 (instead of UTF-8) encoding for ASGI HTTP headers
Currently, headers in Falcon's ASGI package will be decoded using the Python's default UTF-8 decoding. 
For instance:
https://github.com/falconry/falcon/blob/933851183ed9cce2d3411f890dca3e19385480f6/falcon/asgi/request.py#L847
and
https://github.com/falconry/falcon/blob/933851183ed9cce2d3411f890dca3e19385480f6/falcon/asgi/request.py#L897

When headers have an encoding different from UTF-8, this leads to `UnicodeDecodeError`. An option to use different charsets here would be much appreciated.

Even though headers are supposed to be plain ASCII where it doesn't matter, in reality, there are still headers sent which contain characters especially from the ISO-8859-1 charset. Compare also [RFC 7230, section 3.2.4](https://tools.ietf.org/html/rfc7230#section-3.2.4):
> Historically, HTTP has allowed field content with text in the
> ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
> through use of [RFC2047] encoding.  In practice, most HTTP header
> field values use only a subset of the US-ASCII charset [USASCII].

vytas7:
Closes #1911

### DIFF
--- a/falcon/asgi/response.py
+++ b/falcon/asgi/response.py
@@ -20,7 +20,7 @@ from inspect import iscoroutinefunction
 
 from falcon import response
 from falcon.constants import _UNSET
-from falcon.util.misc import is_python_func
+from falcon.util.misc import _encode_items_to_latin1, is_python_func
 
 __all__ = ['Response']
 
@@ -415,8 +415,13 @@ class Response(response.Response):
             headers['content-type'] = media_type
 
         try:
-            items = [(n.encode('ascii'), v.encode('ascii')) for n, v in headers.items()]
+            # NOTE(vytas): Supporting ISO-8859-1 for historical reasons as per
+            #   RFC 7230, Section 3.2.4; and to strive for maximum
+            #   compatibility with WSGI.
+            items = _encode_items_to_latin1(headers)
         except UnicodeEncodeError as ex:
+            # TODO(vytas): In 3.1.0, update this error message to highlight the
+            #   fact that we decided to allow ISO-8859-1?
             raise ValueError(
                 'The modern series of HTTP standards require that header names and values '
                 f'use only ASCII characters: {ex}'
--- a/falcon/asgi/response.py
+++ b/falcon/asgi/response.py
@@ -418,6 +418,11 @@ class Response(response.Response):
             # NOTE(vytas): Supporting ISO-8859-1 for historical reasons as per
             #   RFC 7230, Section 3.2.4; and to strive for maximum
             #   compatibility with WSGI.
+
+            # PERF(vytas): On CPython, _encode_items_to_latin1 is implemented
+            #   in Cython (with a pure Python fallback), where the resulting
+            #   C code speeds up the method substantially by directly invoking
+            #   CPython's C API functions such as PyUnicode_EncodeLatin1.
             items = _encode_items_to_latin1(headers)
         except UnicodeEncodeError as ex:
             # TODO(vytas): In 3.1.0, update this error message to highlight the
--- a/falcon/cyutil/misc.pyx
+++ b/falcon/cyutil/misc.pyx
@@ -1,4 +1,4 @@
-# Copyright 2020 by Vytautas Liuolia.
+# Copyright 2020-2021 by Vytautas Liuolia.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -38,3 +38,14 @@ def isascii(unicode string not None):
             return False
 
     return True
+
+
+def encode_items_to_latin1(dict data not None):
+    cdef list result = []
+    cdef unicode key
+    cdef unicode value
+
+    for key, value in data.items():
+        result.append((key.encode('latin1'), value.encode('latin1')))
+
+    return result
--- a/falcon/testing/helpers.py
+++ b/falcon/testing/helpers.py
@@ -337,6 +337,8 @@ class ASGIResponseEventCollector:
                 if self._BAD_HEADER_VALUE_RE.search(value):
                     raise ValueError('Bad header value: {!r}'.format(value))
 
+                # NOTE(vytas): After the name validation above, the name is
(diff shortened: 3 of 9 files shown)


vytas7:
I realized I probably need to add a couple more regression tests for the reported issue itself. Working on this now.

falcon/testing/helpers.py
```diff
@@ -327,11 +331,17 @@ async def collect(self, event: Dict[str, Any]):
                 if not isinstance(value, bytes):
                     raise TypeError('ASGI header names must be byte strings')
 
+                # NOTE(vytas): Ported basic validation from wsgiref.validate.
+                if not self._HEADER_NAME_RE.match(name):
+                    raise ValueError('Bad header name: {!r}'.format(name))
+                if self._BAD_HEADER_VALUE_RE.search(value):
+                    raise ValueError('Bad header value: {!r}'.format(value))
+
                 name_decoded = name.decode()
```
CaselIT, line 340: missed a decode here

falcon/cyutil/misc.pyx
```diff
@@ -38,3 +38,14 @@ def isascii(unicode string not None):
             return False
 
     return True
+
+
+def encode_items_to_latin1(dict data not None):
```
CaselIT, line 43: using cpdef has no advantage here, even if we cython compile response? Does it only has advantages when called from pyx files?

vytas7: Might be worth looking at this in the future, but currently the ASGI source tree is unfortunately not cythonized (although readied for that when we can; see also https://github.com/falconry/falcon/pull/1860#issue-571338714).

falcon/testing/helpers.py
vytas7: Actually not, but maybe I could add a `NOTE`? The regex above rules out any wonky characters.

falcon/cyutil/misc.pyx
CaselIT: I'm actually not sure if cpdef takes affect from cythonized py files or not

vytas7: It can not from pure Python (including cythonized Python), _methinks_.

It only works from inside a `cdef` -- you basically don't need to go via a full Python call, but can partially do it in C.

vytas7: For reference, I have so far stayed away from `cpdef`, and just wrote separate `cdef` and `def` functions, invoking the former from the latter where needed (in order to share code). I think it's basically what `cpdef` does.

CaselIT (dismissed):

tests/test_headers.py
```diff
@@ -125,13 +125,22 @@ def on_head(self, req, resp):
 
 class UnicodeHeaderResource:
 
+    def on_connect(self, req, resp):
+        # A way to CONNECT with people.
+        resp.set_header('X-Clinking-Beer-Mugs', '🍺')
```
kgriffs, line 130: 💯

falcon/asgi/response.py
```diff
@@ -415,8 +415,13 @@ def _asgi_headers(self, media_type=None):
             headers['content-type'] = media_type
 
         try:
-            items = [(n.encode('ascii'), v.encode('ascii')) for n, v in headers.items()]
+            # NOTE(vytas): Supporting ISO-8859-1 for historical reasons as per
+            #   RFC 7230, Section 3.2.4; and to strive for maximum
+            #   compatibility with WSGI.
+            items = _encode_items_to_latin1(headers)
```
kgriffs, line 421: I assume the overhead of the function call itself is eclipsed by the gain in performance with Cython?

vytas7: Yes, and the gain is more than threefold. I was profiling this carefully when developing, but still, I wasn't referencing the final function from cythonized Falcon.

Adding a targeted benchmark here which does just that:
```python
import timeit

from falcon.util.misc import _encode_items_to_latin1


TEST = {
    'content-length': '4',
    'content-type': 'text/plain',
    'x-another': 'Link_aa_bb_x=1',
}

print([(n.encode('ascii'), v.encode('ascii')) for n, v in TEST.items()])
print(_encode_items_to_latin1(TEST))

print(timeit.repeat(
    lambda: [(n.encode('ascii'), v.encode('ascii')) for n, v in TEST.items()]))
print(timeit.repeat(
    lambda: _encode_items_to_latin1(TEST)))
print(timeit.repeat(
    lambda: [(n.encode('ascii'), v.encode('ascii')) for n, v in TEST.items()]))
print(timeit.repeat(
    lambda: _encode_items_to_latin1(TEST)))
```

```python
[(b'content-length', b'4'), (b'content-type', b'text/plain'), (b'x-another', b'Link_aa_bb_x=1')]
[(b'content-length', b'4'), (b'content-type', b'text/plain'), (b'x-another', b'Link_aa_bb_x=1')]
[0.7468254599999682, 0.7431183099999998, 0.7475773680000657, 0.7449554410000019, 0.7483860550000827]
[0.22064724300003036, 0.2204902519999905, 0.22048896700016485, 0.22051541400014685, 0.2205529090001619]
[0.7458620350000729, 0.7445737529999406, 0.7403353429999697, 0.7462570249999771, 0.7480042119998416]
[0.22065620099988337, 0.22055909699997756, 0.22055948600018382, 0.22050846599995566, 0.22057072500001595]
```

CaselIT: I guess the gap increases the more elements are in the db

CaselIT (approved):

falcon/asgi/response.py
kgriffs: Very nice.

kgriffs (approved):

PR #1922, opened 2021-05-23
chore(CI): add CPython 3.10 beta gates

vytas7:
I haven't added the corresponding Trove classifier yet, waiting for @CaselIT 's PEP 517 changes to land, as well as for 3.10 itself to stabilize.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -68,6 +68,12 @@ jobs:
           - python-version: 3.9
             os: ubuntu-20.04
             toxenv: py39_cython
+          - python-version: 3.10.0-beta.1
+            os: ubuntu-20.04
+            toxenv: py310
+          - python-version: 3.10.0-beta.1
+            os: ubuntu-20.04
+            toxenv: py310_cython
           - python-version: 3.8
             os: ubuntu-20.04
             toxenv: py38_sans_msgpack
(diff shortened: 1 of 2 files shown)


CaselIT:
Why don't we also migrate the bulk of the tests to 3.9 from 3.8?

vytas7:
We should, but that's a separate issue I believe.

vytas7:
Moving to _Draft_; it seems there is an issue with the `websockets` client (or our usage pattern thereof) on 3.10.

The [issue](https://github.com/aaugustin/websockets/issues/916) seems, itself, to be fixed, but I haven't checked whether we just need to wait for a new release, or refactoring of our usage patterns is needed.

.github/workflows/tests.yaml
```diff
@@ -69,6 +69,12 @@ jobs:
           - python-version: 3.9
             os: ubuntu-20.04
             toxenv: py39_cython
+          - python-version: 3.10.0-beta.1
```
CaselIT, line 72: it seems that we can use ranges here: `"3.10.0-beta - 3.10"` automatically choses the most recent version.

I'm using it in another workflow and it works ok

vytas7: Thanks for the tip, I'll take a look at it. This is low prio though, since a bunch of packages (or our use thereof) are still broken on 3.10, namely `websockets`.

PR #1925, opened 2021-06-22
docs(media): make JSONHandler customization docs clearer

### LINKED ISSUE #1907: Make JSONHandler customization docs clearer
As pointed out by @Stargateur in https://github.com/falconry/falcon/issues/1906#issuecomment-817374057, our [`JSONHandler`](https://falcon.readthedocs.io/en/stable/api/media.html#falcon.media.JSONHandler) customization docs could be made clearer by separately illustrating different (albeit closely related) concepts:
* Use a custom JSON library (such as the exemplified `rapidjson`). Customize parameters.
* Use the stdlib's `json` module, just provide custom serialization or deserialization parameters. Also link to the ["Prettifying JSON Responses" recipe](https://falcon.readthedocs.io/en/stable/user/recipes/pretty-json.html), which illustrates customization of `dumps` parameters.
* Add a sentence or two about replacing the default JSON handlers, not just toss in a code snippet as it is at the time of writing this. Also link to [Replacing the Default Handlers](https://falcon.readthedocs.io/en/stable/api/media.html#custom-media-handlers) from that explanation.

dflss:
# Summary of Changes

I have modified the JSONHandler docs by separating the following concepts:
- using the stdlib's json module with custom parameters
- using a custom JSON library and customizing parameters

# Related Issues

Closes #1907


falcon/media/json.py
```diff
@@ -26,8 +26,35 @@ class JSONHandler(BaseHandler):
         library's JSON implementation, since it will be faster in most cases
         as compared to a third-party library.
 
-    Overriding the default JSON implementation is simply a matter of specifying
-    the desired ``dumps`` and ``loads`` functions::
+    If you are using the stdlib's json module, you can override the default
```
CaselIT, line 29: > If you are using the stdlib's json module, 

I think it's best to remove that par since what follows is valid also for external libraries, meaning that I can use partial also with something like `rapidjson` to specify some parameters.

CaselIT (changes requested):
Thanks for the pr!

falcon/media/json.py
dflss: Done!

CaselIT (dismissed):

vytas7 (approved):
Thanks for this, looks great to me! :+1: 

I've just taken the liberty of tweaking and reordering your changes slightly, please check if I haven't distorted your changeset too much.

CaselIT (approved):

PR #1933, opened 2021-07-26
Clarify CORS options in code example (resolves #1930)

### LINKED ISSUE #1930: Clarify mutual exclusivity of `cors_enable=True` and `falcon.CORSMiddleware`
A tired/frustrated reader (such as myself) might see https://falcon.readthedocs.io/en/stable/api/cors.html#usage and incorrectly interpret the example code comments to mean that a "correct" solution requires both setting `cors_enable=True` in the `App` constructor, and passing a `falcon.CORSMiddleware` object to it. This is incorrect, but in a very subtle way - because the `CORSMiddleware` header effectively consumes the `Allow` header from the response (https://github.com/falconry/falcon/blob/7efb46f4068428eb62b422b4b3c2e40129e5a914/falcon/middleware.py#L112-L113), a request including the `Access-Control-Request-Method` header will receive a response with `access-control-allow-methods: None`.

forana:
# Summary of Changes

This PR modified the comments in the CORS code to make it clearer that the two options are alternatives.

# Related Issues

Fixes #1930

### DIFF
--- a/docs/api/cors.rst
+++ b/docs/api/cors.rst
@@ -41,7 +41,8 @@ Usage
             # Enable a simple CORS policy for all responses
             app = falcon.App(cors_enable=True)
 
-            # Enable CORS policy for example.com and allows credentials
+            # Alternatively, enable CORS policy for example.com and allow
+            # credentials
             app = falcon.App(middleware=falcon.CORSMiddleware(
                 allow_origins='example.com', allow_credentials='*'))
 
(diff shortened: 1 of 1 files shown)


CaselIT (approved):
Ok for me! Thanks

YajJackson (approved):

vytas7:
Indeed, I've already fixed that as part of https://github.com/falconry/falcon/pull/1922, but the PR itself has to stay in Draft for a while now, so I carved out that as a separate PR in order to unblock the CI: https://github.com/falconry/falcon/pull/1935.

vytas7 (approved):
LGTM :+1: 
In addition, it might make sense to also add an explicit _Warning_ box explaining that mixing the two ways is unsupported, and might lead to unexpected results.

But we can do that in a follow-up PR.

Issue #1938, opened 2021-08-01
Update asgi look examples to use aioredis 2


CaselIT:
A new major vesion of aioredis was released, and it has some api changes.
This is the changelog https://github.com/aio-libs/aioredis-py/blob/master/CHANGELOG.md

the-bets:
Myself and [Mihai Todor](https://github.com/mihaitodor) can pick this up and work on it, if nobody else is!

vytas7:
Awesome @the-bets & @mihaitodor, go ahead!
(Bonus points if you find and fix any other funky things in that tutorial.)

PR #1939, opened 2021-08-01
test: constrain aioredis to <2 in example requirements.

CaselIT:
aioredis 2 instroduces breaking changes. 

I've just constrained it to <2 to fix the tests
I've opened https://github.com/falconry/falcon/issues/1938 to add support for v2 instead


vytas7 (approved):
LGTM :+1: 

As a stopgap measure, I've also pinned the corresponding installation instruction in the tutorial's text, so that at least our `latest` version of the docs is usable.

Issue #1895, opened 2021-03-26
FAQ: explain the case where URL path contains escaped slash `%2F` 


seamons:
For example, i've set up the following route with field expression:
```
/user/{name}
```
However sometimes the variable will contains slashes `/`, such as:
```
/user/foo/bar
```
I know the slash character is a reserved character of URL, so i encoded my request URL as below:
```
/user/foo%2Fbar
```
I was expecting receiving the argument `name` with value `foo/bar` in responder method, but instead, i got a 404 error.

vytas7:
Hi @seamons !
Yes, this is an unfortunate artefact of the WSGI [`PATH_INFO`mechanics](https://www.python.org/dev/peps/pep-3333/#environ-variables), and unfortunately nothing Falcon can easily work around in the general case.

See also @kgriffs' answer here: https://github.com/falconry/falcon/issues/897#issuecomment-247486014

You could probably work around this in middleware, either Falcon or generic WSGI, by inspecting the WSGI environ extension variables specific to your app server of choice (such as `RAW_URI` in Gunicorn).

I'll leave this issue open in terms of documentation. We ought to add at least a FAQ item or a recipe explaining this.

vytas7:
Also take a look at what the ASGI spec has to say on this.

vytas7:
When working on the related docs, also revise this answer: [How can I handle forward slashes within a route template field?](https://falcon.readthedocs.io/en/3.0.0/user/faq.html#how-can-i-handle-forward-slashes-within-a-route-template-field).

At the time of writing this, the way the answer is worded might give a false impression that it is possible for a client to escape slashes in a way that is easily accessible to the framework.

flying-sheep:
How would an example middleware look that does this?

vytas7:
Hi @flying-sheep!
This is how it could look for WSGI:

As for ASGI, I don't think it's any different. No, we don't handle it out of the box, it seems.
The spec states (emphasis mine):
>  **Some web server implementations** may be **unable** to provide this. Optional; if missing defaults to `None`.

flying-sheep:
That’s almost it! Since the parameters should be decoded, it needs some code in the router, right?

So how about we make Falcon’s default router should look something like this? (remove the second if branch if you don’t want to support nonstandard things)

```py
from urllib.parse import unquote

class CompiledRouter:
    def find(self, uri, req=None):
        if isinstance(req, falcon.asgi.Request):
            raw_path = req.scope.get('raw_path')
[...]

Issue #1567, opened 2019-09-01
Use pattern hints from converters when matching fields


vytas7:
At the time of writing, the `CompiledRouter` is first matching fields, and then subsequently using converters to validate/adapt them to parameters that are sent to a responder.

However, information such as allowed characters, desired length of a field etc is completely ignored when building routing code and/or regexps. In complex URI patterns with multiple fields in a single path segment, a user may expect these are correctly split, if, for instance, the first field was specified to be a 6-digit integer.

Consider the following example:
```
import falcon
import falcon.testing
import pytest


class DebugResource():
    def on_get(self, req, resp, **kwargs):
        resp.body = str(kwargs)
        resp.content_type = falcon.MEDIA_TEXT


@pytest.fixture
def client():
    api = falcon.API()
    api.add_route('/{yymmdd:int(6)}-{item_id:uuid}', DebugResource())
    return falcon.testing.TestClient(api)


def test_routing(client):
    resp = client.simulate_get('/190101-59559ee7-e661-43c8-9da1-0ab83b9e0d8e')
    assert resp.status_code == 200
```

Running `pytest` fails with `assert 404 == 200`, whereas the user expectation was that `yymmdd` was assigned with `190101`, and the UUID parameter with the remainder.

vytas7:
I also added the **documentation** label -- even if we don't properly fix this in the foreseeable future, documentation should clearer reflect the current limitations.

CaselIT:
The also fails `/r/{foo:int}-{bar}` to match `/r/42-foo-bar`
The issue is that the regexp created is `r'^(?P<foo>.+)-(?P<bar>.+)$'` and matches `{'foo': '42-foo', 'bar':'bar'}` instead of the correct `{'foo': '42', bar: 'foo-bar'}`
It may make sense to use instead `r'^(?P<foo>.+?)-(?P<bar>.+?)$'` this regexp, (it would potentially be breaking change, that we may make pass as a bugfix)

vytas7:
IMHO changing to non-greedy matching would not improve anything, it's just a mirror problem.

If your pattern was `/path/{foo}-{bar:int}`. Using the current version, a path like `/path/foo-bar-42` would work correctly (`foo-bar` and `42` would be matched). Conversely, matching in a non-greedy fashion would introduce a regression.

CaselIT:
Indeed, still if we implement this it may be an option to switch che default to nonl-greedy match

Issue #857, opened 2016-08-04
Regular expression route converter


vit1251:
I have question about embedding regular expression in route object like it doing in flask.

Let's I show you example:

```
/user/<name:[A-Za-z0-9]+>
/article/<article_id:[0-9A-F]+>
```

   ...

Could you have this or maybe a good idea porting Flask router to falcon system?

kgriffs:
Hi @vit1251, we should be able to support this inside Falcon's own high-performance router as part of #423. The syntax might look something like this:

/user/{name:re:[A-Za-z0-9]+}

I'd love to get your feedback and help on reviewing the (up and coming) PR for this feature.

kgriffs:
(closing since this is covered by #423)

vytas7:
Reopening in order to reduce the scope of #423 (and close it by adding `path` support).

vytas7:
I think we might want to support this one in conjunction with `path`, either always, or split into `re` and `repath`.

vytas7:
This issue shouldn't be too terribly hard to build.
We need to watch out for #2062 though, and discuss whether we want to give the matched groups any meaning.

infraredCoding:
@vytas7 hey there. is this issue open to contribute?

vytas7:
> [@vytas7](https://github.com/vytas7) hey there. is this issue open to contribute?

Sure @infraredCoding , go ahead! Since this is a "good first issue", try not to use [LLM](https://falcon.readthedocs.io/en/latest/community/contributing.html#use-of-llms-ai)s at all in the first iteration.

PR #1945, opened 2021-08-05
feat(routing): add support for multiple path segment fields

### LINKED ISSUE #423: Add support for filters in route templates
This is a request I've heard from the community, and that was previously mentioned in one of my comments on #161.  It would be nice to if we could support this within the path and for query string params (assuming we implement the latter). We can take inspiration from [bottle](http://bottlepy.org/docs/dev/tutorial.html#dynamic-routes). Default filters might include:
- [x] `int` (already implemented)
- [x] `float` deferred to https://github.com/falconry/falcon/issues/2022
- [ ] `path` prototype PR: https://github.com/falconry/falcon/pull/1945
- [x] `re` deferred to https://github.com/falconry/falcon/issues/857
- [ ] `in` will be deferred to another issue if needed

~~Also, we need to consider how this might be applies to query string parameters as well (which would require defining such parameters in the URI template)~~ :arrow_left: tracked separately in https://github.com/falconry/falcon/issues/838

### LINKED ISSUE #648: Routing all paths to one handler function
Is it possible to route all paths that start with something like “/api” to same handler function?

Such as:

```
/api/foo
/api/bar
/api/foo/bar
/api/bar/baz/this/that 
```

All should be handled with one function and I should be able to get the full path after /api.

This feature is very handy and I used it often at Node.js Express framework. Now I am looking for ways to accomplish the same thing with Falcon.

More info can be found here; It defines the feature as "white-listed “global” functionality."
http://expressjs.com/api.html#app.all

CaselIT:
# Summary of Changes

This adds support for a converter to be declared as consuming all the remaining path.

The current implementation does not allow nesting these converters and is quite conservative (I think), but I believe we can improve on this later if we need to.

# Related Issues

Closes #423
Fixes #648
Relates to #1895

# Pull Request Checklist

- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/index.html) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)


tests/asgi/test_ws.py
```diff
@@ -341,7 +341,9 @@ async def on_websocket(self, req, ws):
                             await asyncio.sleep(
                                 0
                             )  # Ensure recv_task() has a chance to get ahead
-                            await asyncio.wait([recv_task, ws.close(4099)])
+                            await asyncio.wait(
+                                [recv_task, asyncio.create_task(ws.close(4099))]
```
CaselIT, line 345: this fixes the warning 
`The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.`

CaselIT: this hangs in py36 -.-'

CaselIT: fixed now

falcon/routing/converters.py
```diff
@@ -31,18 +31,38 @@
 class BaseConverter(metaclass=abc.ABCMeta):
     """Abstract base class for URI template field converters."""
 
+    CONSUME_PATH = False
```
CaselIT, line 34: this should probably renamed in `consume_multiple_segments` or similar so that it keeps making sense if in the future we enable support for `prefix/{rest:path}/suffix`

CaselIT:
After https://github.com/falconry/falcon/pull/1990 is merged we should update the docs there to mention that for simple cases `:path` may be used

CaselIT:
I'm not really sold on the idea, since it's another thing that should be documented, explained etc.

I think your example would work better with `/route/{itemid}/repository/{url:relativepath}`. `path` is not special, meaning that the way it indicates that it can captures more than one segment is available to other converters

vytas7:
Agreed that my idea ain't very pretty either.
But functionality wise, it depends on the way you view it. Say that you have a useful converter that _might_ accept path, but in another case you would want to restrict it to a single field, or just use it in another field than the very last one (suffix). I guess you could subclass it to override that flag, and then register the new type as well, but it won't be particularly pretty to document either.

CaselIT:
Since the interface of the converters changes, I think it makes sense to have two implementation in that case. (path like converters get passed a list of path segments, not a single segment)

vytas7:
Another issue in that useful to have in mind when discussing the big picture: https://github.com/falconry/falcon/issues/1567 (which would actually favour your current design, but maybe we could devise a more generic way to provide hints?)

CaselIT:
From a quick look I don't think that the current implementation is incompatible with that issue

vytas7:
I know that although Falcon has moved away from [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570), the said paper has all kinds of funky expansions, so maybe something familiar could be adapted from [Section 3.2.6](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.6), like `{list*}` :arrow_right: `/red/green/blue`.

Or maybe we could have a discussion [meeting?] about this on chat at some point? I don't have any elegant suggestions, but maybe someone else does.

CaselIT:
That may be an option, but we right having two ways of doing something, but having both the converters and this special syntax.
Also what would happen if both a converter and `*` are used together? Like `{list*:int}`

vytas7:
Dunno, I was just thinking what the options were before we move forward with this. But we should move forward in one or another way 🙂

vytas7:
It would be also good to reason about this change in the perspective of adding `re` converter (https://github.com/falconry/falcon/issues/857). Requiring a `CONSUME_PATH` attribute on the converter class would force a decision to make `re` of one or another type, however, I can imagine both could be useful :thinking:

falcon/routing/compiled.py
```diff
@@ -236,12 +256,24 @@ def insert(nodes, path_index=0):
...
+                nodes.append(new_node)
                 new_node.method_map = method_map
                 new_node.resource = resource
                 new_node.uri_template = uri_template
             else:
+                cpc = consume_path_converter(new_node)
+                if cpc:
+                    print('b')
+                    raise ValueError(no_children_err.format(uri_template, *cpc))
+                nodes.append(new_node)
```
kgriffs, line 276: Why do we `nodes.append(new_node)` in both branches of the `if...else` as opposed to doing it once outside the statement?

falcon/routing/compiled.py
```diff
@@ -513,6 +550,8 @@ def _generate_ast(
                 resource_idx = len(return_values)
                 return_values.append(node)
 
+            assert not (consume_path and node.children)
```
kgriffs, line 553: nice sanity check

falcon/routing/converters.py
```diff
@@ -132,8 +152,22 @@ def convert(self, value):
             return None
 
 
+class PathConverter(BaseConverter):
+    """Field converted used to match the rest of the path.
+
+    This field converter matches the all the remaining url path,
```
kgriffs, line 158: `matches the all the remaining url path` -->  `matches the remainder of the URL path`

falcon/routing/converters.py
```diff
@@ -132,8 +152,22 @@ def convert(self, value):
...
+    returning it as a string.
```
kgriffs, line 159: We should also note what the reader should expect when the path converter comes at the end of a path that is already consumed. For example, given:

Template: `'/foo/bar/{foo:path}'`
Path Requested: `'/foo/bar'`

The path will not match. Also describe what happens when there's a trailing forward-slash in the requested path?

`'/foo/bar/'`

tests/test_default_router.py
```diff
@@ -727,3 +727,78 @@ def test_params_in_non_taken_branches(param_router, route, expected, num):
...
+def capture_path_router():
+    router = DefaultRouter()
+
+    router.add_route('/foo/bar/baz', ResourceWithId(1))
+    router.add_route('/foo/{bar:path}', ResourceWithId(2))
+    router.add_route('/foo/bar/{foo:path}', ResourceWithId(3))
+    router.add_route('/{baz:path}', ResourceWithId(4))
+    router.add_route('/x/{v1:int}/{v2}/{other:path}', ResourceWithId(5))
+    return router
+
```
kgriffs, line 781: We should also test that the following request paths result in 404s when `'/{baz:path}'` is *not* added as a route:

`'/foo'`
`'/foo/'`
`'/foo/bar'`
`'/foo/bar/'`

kgriffs (changes requested):

falcon/routing/compiled.py
CaselIT: I guess the idea was to add it only once the checks are passed since raising an error after may lead to an inconsistent router state?

falcon/routing/compiled.py
```diff
@@ -218,6 +235,9 @@ def insert(nodes, path_index=0):
                         node.resource = resource
                         node.uri_template = uri_template
                     else:
+                        cpc = consume_path_converter(node)
+                        if cpc:
+                            raise ValueError(no_children_err.format(uri_template, *cpc))
```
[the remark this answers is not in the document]
CaselIT: The idea is that this limitation is temporary, removing the limitation once that's done. Also in other places the router already raises a ValueError

CaselIT:
thanks for the review @kgriffs 

sorry for the delay in getting to this

CaselIT:
Btw I've taken another look at supporting additional segments after a "consume multiple segments" converter it's not trivial.
Also there is a decision to take here: should a "consume multiple segments" converter be allowed to match 0 segments?

Assuming the reply is yes, the finder would need to be something like the following
```py
from dataclasses import dataclass, field
from itertools import count
from falcon.routing import CompiledRouter
from falcon.routing.converters import PathConverter

seq = count(1)

@dataclass
class Tmp:
    id: int = field(default_factory=seq.__next__)

    def on_get(self, req, resp, **kw):
        resp.media = {'id': self.id, 'kw': kw}

r = CompiledRouter()

def add(url, res):
    print(res, url.replace('{x}', '{x:path}'))
    r.add_route(url, res)

add('/foo/{x}', Tmp())
add('/foo/{x}/foo', Tmp())
add('/foo/{x}/{baz:int}', Tmp())
add('/foo/{x}/foo/bar', Tmp())
add('/foo/{x}/{baz:int}/baz', Tmp())

r._compile()
# print(r.finder_src)

def find(path, return_values, patterns, converters, params):
    converters = [PathConverter(), *converters]
    rv = {r.resource.id: r for r in return_values}
    path_len = len(path)
    if path_len > 0:
        if path[0] == 'foo':
            if path_len > 1:
                remaining_path = path[1:]
                remaining_path_len = len(remaining_path)
                if remaining_path_len > 1:
                    if remaining_path[-1] == 'bar':
                        if remaining_path[-2] == 'foo':
                            fragment = remaining_path[:-2]
                            field_value_1 = converters[0].convert(fragment)
                            if field_value_1 is not None:
                                params['x'] = field_value_1
                                return rv[4]
                    if remaining_path[-1] == 'baz':
                        fragment = remaining_path[-2]
                        field_value_2 = converters[1].convert(fragment)
                        if field_value_2 is not None:
                            fragment = remaining_path[:-2]
                            field_value_1 = converters[0].convert(fragment)
                            if field_value_1 is not None:
                                params['baz'] = field_value_2
                                params['x'] = field_value_1
                                return rv[5]
                if remaining_path_len > 0:
                    if remaining_path[-1] == 'foo':
                        fragment = remaining_path[:-1]
                        field_value_1 = converters[0].convert(fragment)
                        if field_value_1 is not None:
                            params['x'] = field_value_1
                            return rv[2]
                    fragment = remaining_path[-1]
                    field_value_2 = converters[1].convert(fragment)
                    if field_value_2 is not None:
                        fragment = remaining_path[:-1]
                        field_value_1 = converters[0].convert(fragment)
                        if field_value_1 is not None:
                            params['baz'] = field_value_2
                            params['x'] = field_value_1
                            return rv[3]
                field_value_1 = converters[0].convert(remaining_path)
                if field_value_1 is not None:
                    params['x'] = field_value_1
                    return rv[1]
                return None
            return None
        return None
    return None

r._find = find

def go(url):
    res = r.find(url)
    if res is None:
        print(url, res)
    resource, _, param, _ = res
    print(resource, param, url)

print('-' * 20)
go('/foo/foo')
go('/foo/bar')
go('/foo/bar/123')
go('/foo/bar/123/x')
go('/foo/bar/123/x/foo')
go('/foo/foo/bar/123/x/foo')
go('/foo/bar/123/x/foo/bar')
go('/foo/123/baz')
go('/foo/bar/123/baz')
go('/foo/bar/123/x/foo/bar/123/baz')
```

running the above yields
```
Tmp(id=1) /foo/{x:path}
Tmp(id=2) /foo/{x:path}/foo
Tmp(id=3) /foo/{x:path}/{baz:int}
Tmp(id=4) /foo/{x:path}/foo/bar
Tmp(id=5) /foo/{x:path}/{baz:int}/baz
--------------------
Tmp(id=2) {'x': ''} /foo/foo
Tmp(id=1) {'x': 'bar'} /foo/bar
Tmp(id=3) {'baz': 123, 'x': 'bar'} /foo/bar/123
Tmp(id=1) {'x': 'bar/123/x'} /foo/bar/123/x
Tmp(id=2) {'x': 'bar/123/x'} /foo/bar/123/x/foo
Tmp(id=2) {'x': 'foo/bar/123/x'} /foo/foo/bar/123/x/foo
Tmp(id=4) {'x': 'bar/123/x'} /foo/bar/123/x/foo/bar
Tmp(id=5) {'baz': 123, 'x': ''} /foo/123/baz
Tmp(id=5) {'baz': 123, 'x': 'bar'} /foo/bar/123/baz
Tmp(id=5) {'baz': 123, 'x': 'bar/123/x/foo/bar'} /foo/bar/123/x/foo/bar/123/baz
```

In the above example `/foo/{x:path}/foo` matches `/foo/foo` with `x=''`

I'll copy the above in a new issue

kgriffs:
> There is a decision to take here: should a "consume multiple segments" converter be allowed to match 0 segments? the current implementation allows it.

Personally I would expect it to never match if there is no path to consume, because otherwise what's the point of using the `path`	converter? Or at least, if there is another route registered for the exact path prefix, that one would take precedence over the route containing the `path` converter.

One could argue it my be useful to be able to handle `/foo/bar/bang` and /foo/bang` within the same resource controller, but that doesn't feel like a good API design to me (though it may be useful for re-implementing a legacy API?). So perhaps we could introduce an option to allow 0 segments but it is disabled by default? Or we introduce that along with #2061 ?

kgriffs:
Code looks good, we just need a consensus on what to do about allowing 0-length path matches.

vytas7:
If I understand the problem right, IMHO, for the sake of consistency, the new converter should behave like fields behave now.
They do match an empty segment after a trailing slash, and they do not otherwise.

Given the following test app,
```python
import falcon

class Resource:
    def on_get(self, req, resp, resourceid):
        resp.media = {'path': req.path, 'resourceid': resourceid}

(app := falcon.App()).add_route('/api/{resourceid}', Resource())
```

Cf
```
$ xh http://localhost:8000/api/
HTTP/1.1 200 OK
Content-Length: 35
Content-Type: application/json

{
    "path": "/api/",
    "resourceid": ""
}
```
vs
```
$ xh http://localhost:8000/api
HTTP/1.1 404 Not Found
Content-Length: 26
Content-Type: application/json
Vary: Accept

{
    "title": "404 Not Found"
}
```

CaselIT:
That's the current behavior, unless `strip_url_path_trailing_slash=True`, in which case both yield 404

Also note that my question was only once we add support for additional parts after the path. Basically the question what if
`/foo/{bar:path/baz` should match `/foo/baz` or not.

falcon/routing/compiled.py
CaselIT: good idea, will add a comment

vytas7:
Aha, right, I didn't realize question was only about when we lift the requirement for a `path` segment to be last.
Then maybe we don't need to make a decision _right now_?

In any case, I would perceive routing as strange and confusing if `/foo/{bar:path}/baz` matched `/foo/baz`, it wouldn't be consistent with the current behaviour where a field essentially needs a preceding `/` to match. So my vote is it shouldn't match.

falcon/routing/compiled.py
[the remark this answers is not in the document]
CaselIT: I've created a subclass and use it for each error in add_route

falcon/routing/compiled.py
CaselIT: In the end it's probably better to just add the node and remove if before raising an error if needed

CaselIT:
> Aha, right, I didn't realize the question was only about when we lift the requirement for a `path` segment to be last. Then maybe we don't need to make a decision _right now_?
> 
> In any case, I would perceive routing as strange and confusing if `/foo/{bar:path}/baz` matched `/foo/baz`, it wouldn't be consistent with the current behaviour where a field essentially needs a preceding `/` to match, and in this case the `path` field would rob `/baz` of that slash. So my vote is on "it shouldn't match".

We really ought to move the conversation here https://github.com/falconry/falcon/issues/2061

vytas7:
Sure -- I was just pinged here on this issue :slightly_smiling_face:

falcon/routing/converters.py
```diff
@@ -132,8 +155,32 @@ def convert(self, value):
...
+class PathConverter(BaseConverter):
+    """Field converted used to match the rest of the path.
+
+    This field converter matches the remainder of the URL path,
+    returning it as a string.
+
+    This converter is currently supported only when used at the
+    end of the url template.
+
+    The classic routing rules of falcon apply also to this converter:
```
kgriffs, line 167: Given this is documented for the developer, I'm OK with leaving the behavior as-is.

falcon/routing/compiled.py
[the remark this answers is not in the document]
CaselIT: I've updated, let me know what you think

kgriffs:
> If I understand the problem right, IMHO, for the sake of consistency, the new converter should behave like fields behave now. They do match an empty segment after a trailing slash, and they do not otherwise.

Good point, I'm OK with moving forward w/ this as long as the behavior is explicitly documented for the developer.

> the question was only about when we lift the requirement for a path segment to be last.

Actually, I wanted your input on both questions. 😄 Second one is not urgent. Thanks!

falcon/routing/converters.py
```diff
@@ -31,18 +31,41 @@
 class BaseConverter(metaclass=abc.ABCMeta):
     """Abstract base class for URI template field converters."""
 
+    CONSUME_MULTIPLE_SEGMENTS = False
+    """When set to ``True`` it indicates that this converted will consume
```
kgriffs, line 35: `converted` --> `converter`

falcon/routing/converters.py
```diff
@@ -31,18 +31,41 @@
...
+    multiple url path segments. Currently a converter with
+    ``CONSUME_MULTIPLE_SEGMENTS=True`` must be at the env of the url template
+    effectively meaning that it will consume all the remaining url path.
```
kgriffs, line 38: `all the remaining url path` --> `all of the remaining URL path segments`

falcon/routing/converters.py
```diff
@@ -31,18 +31,41 @@
 class BaseConverter(metaclass=abc.ABCMeta):
     """Abstract base class for URI template field converters."""
 
+    CONSUME_MULTIPLE_SEGMENTS = False
+    """When set to ``True`` it indicates that this converted will consume
+    multiple url path segments. Currently a converter with
```
kgriffs, line 36: capitalize `URL` here and in the next line since it is an acronym.

falcon/routing/converters.py
```diff
@@ -132,8 +155,32 @@ def convert(self, value):
...
 
 
+class PathConverter(BaseConverter):
+    """Field converted used to match the rest of the path.
+
+    This field converter matches the remainder of the URL path,
+    returning it as a string.
+
+    This converter is currently supported only when used at the
+    end of the url template.
```
kgriffs, line 165: `url` --> `URL`

falcon/routing/converters.py
```diff
@@ -132,8 +155,32 @@ def convert(self, value):
...
+
+    The classic routing rules of falcon apply also to this converter:
+    considering the template ``'/foo/bar/{matched_path:path}'``, the path
+    ``'/foo/bar'`` will *not* match the route; ``'/foo/bar/'`` will
+    match, producing ``matched_path=''``, when
+    :attr:`~falcon.RequestOptions.strip_url_path_trailing_slash` is ``False``,
```
kgriffs, line 171: Perhaps also mention that `False` is the default for this option.

kgriffs (approved):

vytas7:
Yes so, I think no converter should match 0 fields, because that would be a substantial change in comparison to how the framework behaves now. Matching an empty string after a trailing slash is a bit special edge case (or oversight) which already exists now, so path converter can also do the same, but `/foo/{bar}/{baz:path}` ought not to match `/foo/bar`.

docs/api/routing.rst
```diff
@@ -295,6 +295,7 @@ Built-in Converters
  ``int``       :class:`~.IntConverter`            ``/teams/{tid:int(8)}``
  ``uuid``      :class:`~.UUIDConverter`           ``/diff/{left:uuid}...{right:uuid}``
  ``dt``        :class:`~.DateTimeConverter`       ``/logs/{day:dt("%Y-%m-%d")}``
+ ``path``      :class:`~.PathConverter`               ``/prefix/{other:path}``
```
vytas7, line 298: It seems that ``prefix/{other:path}`` is misaligned

docs/_newsfragments/648.rst
```diff
@@ -0,0 +1 @@
+Add ``path`` field converted to match all paths that start with a prefix.
```
vytas7, line 1: In newsfragments, IMHO we should tell a story, not phrase in the imperative as docstrings.
Also, _converted_ (sic) :arrow_right: _converter_.

falcon/routing/converters.py
```diff
@@ -31,18 +31,42 @@
...
         Returns:
             object: Converted field value, or ``None`` if the field
                 can not be converted.
         """
 
+    @classmethod
+    def patch_converter_class(cls, converter):
+        """Patches the input converter class.
+
+        Ensures that ``converter`` is compatible with the ``BaseConverter``
```
vytas7, line 61: Cosmetic: `BaseConverter` would look nicer as a class reference.

falcon/routing/compiled.py
```diff
@@ -199,14 +199,25 @@ class can use suffixed responders to distinguish requests
...
 
         used_names = set()
         for segment in path:
             self._validate_template_segment(segment, used_names)
 
+        def find_cp_converter(node):
+            value = [
+                (field, converter)
+                for field, converter, _ in node.var_converter_map
+                if self._converter_map[converter].CONSUME_MULTIPLE_SEGMENTS
```
vytas7, line 214: Do we really need to do that patching? We could just do a `getattr` here instead.

falcon/routing/converters.py
```diff
@@ -132,8 +156,32 @@ def convert(self, value):
...
+
+    This converter is currently supported only when used at the
+    end of the URL template.
+
+    The classic routing rules of falcon apply also to this converter:
+    considering the template ``'/foo/bar/{matched_path:path}'``, the path
+    ``'/foo/bar'`` will *not* match the route; ``'/foo/bar/'`` will
+    match, producing ``matched_path=''``, when
+    :attr:`~falcon.RequestOptions.strip_url_path_trailing_slash` is ``False``
+    (the default), while it will *not* match when that option is ``True``.
```
vytas7, line 173: It might be useful to link to ["How does Falcon handle a trailing slash in the request path?"](https://falcon.readthedocs.io/en/stable/user/faq.html#how-does-falcon-handle-a-trailing-slash-in-the-request-path) from here.

vytas7 (changes requested):
This is shaping up, well done :+1:

A couple of inline nitpicks aside, we also ought to revise this FAQ item: [How can I handle forward slashes within a route template field?](https://falcon.readthedocs.io/en/stable/user/faq.html#how-can-i-handle-forward-slashes-within-a-route-template-field)

CaselIT:
thanks for the review

> but `/foo/{bar}/{baz:path}` ought not to match `/foo/bar`.

it does not at the moment.
without `/` in the last segment the behavior will be the same as if using  `/foo/{bar}/{baz}` as template

kgriffs:
> [vytas7](https://github.com/vytas7) requested changes [20 hours ago](https://github.com/falconry/falcon/pull/1945#pullrequestreview-971407378)
lol, this is why it is good to require >= 2 approvals. You guys inevitably catch things I miss.

falcon/routing/compiled.py
CaselIT: no clue why I imokemented that way, but it seems we could avoid that

I think the idea was to have a place that could be used to normalize the api of the converters, in case in the future we wanted to do some other vhange to them.

but it does indeed seem overkill for this case

docs/_newsfragments/648.rst
CaselIT: tried rewording. I'm afraid I'm not a story teller, so feel free to suggest an alternative

docs/_newsfragments/648.newandimproved.rst
```diff
@@ -0,0 +1,2 @@
+A new ``path`` converter capable of matching segments that include ``/``
```
vytas7, line 1: Would be nice to link to the new converter class from here, although not required to pass review.
For example, the RST code could read:
```rst
A new ``path`` :class:`converter <~falcon.routing.PathConverter>` capable of matching segments that include ``/``...
```

vytas7 (changes requested):
LGTM sans the still pending FAQ update: [How can I handle forward slashes within a route template field?](https://falcon.readthedocs.io/en/stable/user/faq.html#how-can-i-handle-forward-slashes-within-a-route-template-field)

docs/_newsfragments/648.newandimproved.rst
CaselIT: will use that

CaselIT:
> LGTM sans the still pending FAQ update: [How can I handle forward slashes within a route template field?](https://falcon.readthedocs.io/en/stable/user/faq.html#how-can-i-handle-forward-slashes-within-a-route-template-field)

sorry I missed the previous message. Will do that

vytas7 (approved):
LGTM :+1:

PR #1947, opened 2021-08-06
feat: introduce type hints (MVP)

### LINKED ISSUE #1350: feat: add type information - mypy support
As of [PEP 561](https://www.python.org/dev/peps/pep-0561/) it is now possible for packages to provide typing information, with different options of how to do so. I'm wondering what would be the preferred course of action to include these.

It seems we could start with [partial stubs](https://www.python.org/dev/peps/pep-0561/#partial-stub-packages) where we could define things that would be most helpful/commonly used (i.e. request/response)

vytas7:
Closes #1350

This is only to get the ball rolling though, with an emphasis on the most commonly used, publicly documented functionality.
We'll have to expand typing later in follow-up PRs. One of such PRs is already in Draft: #1944.


falcon/app.py
```diff
@@ -226,6 +229,9 @@ def process_response(self, req, resp, resource, req_succeeded)
         'resp_options',
     )
 
+    req_options: RequestOptions
+    resp_options: ResponseOptions
```
CaselIT, line 233: if sphinx like is we could also inline the documentation so that ides automatically pick it up. but it would require checking the docs output

falcon/app.py
```diff
@@ -438,7 +444,7 @@ def __call__(self, env, start_response):  # noqa: C901
     def router_options(self):
         return self._router.options
 
-    def add_middleware(self, middleware):
+    def add_middleware(self, middleware: object) -> None:
```
CaselIT, line 447: we could probably add several protocols for the middleware. sadly is seems that all members of a protocol are required, so I guess we would need to define one for each optional method :(

overall this does not have much impact on the users, so we can likely just do this later on

falcon/typing.py
```diff
@@ -0,0 +1,43 @@
...
+# Error serializers
+ErrorSerializer = Callable[[RequestClass, ResponseClass, ExceptionClass], Any]
+
+
+# Sinks
+# TODO(vytas): Is it possible to specify a Callable or a Protocol that defines
+#   type hints for the two first parameters, but accepts any number of keyword
+#   arguments afterwords?
+# class SinkCallable(Protocol):
+#     def __call__(sef, req: Request, resp: Response, <how to do?>): ...
```
CaselIT, line 42: `**kwargs:  Any` (kwargs can be named whatever).

for any positional arg instead you can do `*args: Any` (again args can be named whatever)

falcon/util/misc.py
```diff
@@ -459,7 +461,7 @@ def code_to_http_status(status):
     try:
         code = int(status)
         if not 100 <= code <= 999:
-            raise ValueError('{} is not a valid status code'.format(status))
+            raise ValueError('{!r} is not a valid status code'.format(status))
```
CaselIT, line 464: let's do f-strings

pyproject.toml
```diff
@@ -6,6 +6,34 @@
         "cython>=0.29.21; python_implementation == 'CPython'", # Skip cython when using pypy
     ]
 
+[tool.mypy]
+    exclude = "falcon/bench/|falcon/cmd/"
+
+    [[tool.mypy.overrides]]
```
CaselIT, line 12: I don't think you are supposed to indent inner prop. at least the toml formatter I've used will unintended this

pyproject.toml
```diff
@@ -6,6 +6,34 @@
...
+            "pyximport",
+            "testtools",
+            "uvicorn"
+        ]
+        ignore_missing_imports = true
+
+    [[tool.mypy.overrides]]
+        # Pure Cython modules
+        module = [
+            "falcon.cyutil.misc",
```
CaselIT, line 31: I think mypy ignores cython files

tox.ini
```diff
@@ -142,16 +143,27 @@ deps = {[with-debug-tools]deps}
...
        types-requests
        types-PyYAML
        types-ujson
        types-waitress
        types-aiofiles
+       types-jsonschema
 commands = python "{toxinidir}/tools/clean.py" "{toxinidir}/falcon"
-           pytest --mypy -m mypy tests []
+           mypy e2e-tests/server
+           mypy tests
```
CaselIT, line 166: I kinda liked that it would appear as a pytest output but I'm fine either way

CaselIT (approved):
looks mostly ok to me. I've left a suggestion and replied to a TODO

falcon/app.py
```diff
@@ -285,7 +291,7 @@ def __init__(
         self.add_error_handler(HTTPError, self._http_error_handler)
         self.add_error_handler(HTTPStatus, self._http_status_handler)
 
-    def __call__(self, env, start_response):  # noqa: C901
+    def __call__(self, env: dict, start_response):  # noqa: C901
```
kgriffs, line 294: Typing this function is probably not super helpful, but for the sake of completeness you could annotate `start_response` as well, something like:

```py
def __call__(self, env: dict, start_response: Callable):  # noqa: C901
    ...
```

Although I had fun researching the following hints, I'm not certain they are correct (e.g., `Callable` doesn't technically let you type individual optional keyword arguments). Regardless, I don't think we need to be this specific. :trollface: 

```py
def __call__(self, env: dict, start_response: Callable[[str, List[Tuple[str, str]], Optional[tuple]], Callable[bytes, None]]):  # noqa: C901
    ...
```



https://peps.python.org/pep-0333/#the-start-response-callable

falcon/app.py
```diff
@@ -882,7 +897,7 @@ def _prepare_middleware(self, middleware=None, independent_middleware=False):
             middleware=middleware, independent_middleware=independent_middleware
         )
 
-    def _get_responder(self, req):
+    def _get_responder(self, req) -> Tuple[Callable, dict, object, Optional[str]]:
```
kgriffs, line 900: Should we fully hint this by adding `req: Request`?

falcon/app_helpers.py
```diff
@@ -30,7 +31,9 @@
 )
 
 
-def prepare_middleware(middleware, independent_middleware=False, asgi=False):
+def prepare_middleware(
+    middleware: Iterable, independent_middleware=False, asgi=False
```
kgriffs, line 35: Should this be:

```py
middleware: Iterable, independent_middleware: bool = False, asgi: bool = False
```

falcon/app_helpers.py
```diff
@@ -196,7 +199,7 @@ def prepare_middleware_ws(middleware):
     return request_mw, resource_mw
 
 
-def default_serialize_error(req, resp, exception):
+def default_serialize_error(req, resp, exception: HTTPError):
```
kgriffs, line 202: `req: Request, resp: Response, exception: HTTPError`

falcon/media/multipart.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019-2022 by Vytautas Liuolia.
+# Copyright 2019-2023 by Vytautas Liuolia.
```
kgriffs, line 1: 🙌

falcon/media/base.py
```diff
@@ -114,7 +116,9 @@ def deserialize(self, stream, content_type, content_length) -> object:
         else:
             raise NotImplementedError()
 
-    async def deserialize_async(self, stream, content_type, content_length) -> object:
+    async def deserialize_async(
+        self, stream: IO, content_type: str, content_length: Optional[int]
```
kgriffs, line 120: We should update the docstring to note that `content_length` may be `None`

kgriffs (commented):
Very nice! This patch is a solid improvement to typing of the framework.

tox.ini
vytas7: I had issues with that plugin, it was concealing errors in some cases, thought it would be much less maintenance burden to stay with vanilla `mypy` which is very widely used compared to running through the said plugin.

CaselIT: Understood

kgriffs (commented):
Just https://github.com/falconry/falcon/pull/1947/files#r1145637935 otherwise LGTM regarding my previous comments.

falcon/app_helpers.py
```diff
@@ -30,7 +32,9 @@
 )
 
 
-def prepare_middleware(middleware, independent_middleware=False, asgi=False):
+def prepare_middleware(
+    middleware: Iterable, independent_middleware: bool = False, asgi: bool = False
```
CaselIT, line 36: middleware: Iterable[Any]

falcon/app_helpers.py
```diff
@@ -52,9 +56,9 @@ def prepare_middleware(middleware, independent_middleware=False, asgi=False):
 
     # PERF(kgriffs): do getattr calls once, in advance, so we don't
     # have to do them every time in the request path.
-    request_mw = []
-    resource_mw = []
-    response_mw = []
+    request_mw: List = []
+    resource_mw: List = []
+    response_mw: List = []
```
CaselIT, line 61: Here too we should use List[Any]. Doing just list is not really telling much to the type checker and it may complain that a list is missing the generic parameter

CaselIT (approved):
I see that the generic were not added like in list, dict, tuple, etc.

I guess we can to a pass on them later

vytas7:
I haven't had the time to finish this @CaselIT @kgriffs... As discussed on Gitter, I'll remove the generics and try to type as much as possible using plain `Request` and `Response`.

CaselIT:
> I haven't had the time to finish this yet @CaselIT @kgriffs... As discussed on Gitter, I'll remove the generics and try to type as much as possible using plain `Request` and `Response`.

I was referring to generic only on builtin containers such as list, dict, iterable, tuple, etc

falcon/app_helpers.py
vytas7: Not sure I understand the reasoning to do so tbh... :thinking: 

According to https://docs.python.org/3/library/typing.html#user-defined-generic-types:
> Using a generic class without specifying type parameters assumes [Any](https://docs.python.org/3/library/typing.html#typing.Any) for each position. In the following example, `MyIterable` is not generic but implicitly inherits from `Iterable[Any]`:
```python
from collections.abc import Iterable

class MyIterable(Iterable): # Same as Iterable[Any]
    ...
```

This is for subclassing, but I would assume the checker interprets signatures in the same way?

vytas7:
@CaselIT See my other comment, I'm not convinced there is any difference in elaborating that way with `[Any]`.

See, for instance:
`test.py`
```python
from typing import Iterable


def print_all(iterable: Iterable, header: str = '== Items ==') -> None:
    print(header)
    for index, item in enumerate(iterable):
        print(f'{index+1}: {item!r}')


def main() -> None:
    print_all('ABC', header='=====')
    print_all(None)
    print_all(header=3)


if __name__ == '__main__':
    main()
```
```
$ mypy test.py 
test.py:12: error: Argument 1 to "print_all" has incompatible type "None"; expected "Iterable[Any]"  [arg-type]
test.py:13: error: Missing positional argument "iterable" in call to "print_all"  [call-arg]
test.py:13: error: Argument "header" to "print_all" has incompatible type "int"; expected "str"  [arg-type]
```

Judging from the error messages, `Iterable` and `Iterable[Any]` are simply synonymous.

falcon/app_helpers.py
CaselIT: Depending on configuration, mypy will complain.

The following snippet
```py
from typing import Iterable

def foo(x: Iterable) -> None:
    print(list(x))
```
reports
```
error: Missing type parameters for generic type "Iterable"  [type-arg]
```

Like I commented we can ignore this for now, and revisit once we have more typing in falcon

vytas7: Aha, I see, this is what [`--disallow-any-generics`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-generics) does... Would this explode for the user even if this was inside Falcon, or are packages used not checked the same way?

falcon/app.py
vytas7: Sphinx should pick it up, I use it in other projects, not sure if it plays well with Napoleon, but I can try in a followup PR.

falcon/app.py
vytas7: Yes, I'll add `Request` where applicable. I was trying to use generics for `Request` and `Response` to highlight the fact that one can pass a custom class, but that still seems to have too many rough edges in `mypy`, cannot achieve all I want.

falcon/app.py
vytas7: Yeah this is totally out of scope for now unfortunately, my experience with protocols so far was: :cold_sweat:.

pyproject.toml
vytas7: Most probably you're right, I'm no expert (nor fan for that matter) of TOML... will check.

pyproject.toml
vytas7: Doesn't hurt to list them here though.

falcon/typing.py
vytas7: Unfortunately, that does not work for me.

Any help is very welcome on this one... until then, I'll have to leave it for better times...

This is what I tried:
```
from typing import Protocol


class FancyCallable(Protocol):
    def __call__(self, name: str, **kwargs):
        ...


def fancy_call(name: str, callback: FancyCallable):
    print(f'Hello, {name}!')
    callback(name, style='fancy', version=2.1)


def log_call(name: str, style: str = 'plain', version: float = 2.0):
    print(f'[log_call] {name=} {style=}')


def main() -> None:
    fancy_call('Federico', log_call)


if __name__ == '__main__':
    main()
```
Works perfectly just to run it, and I don't see any problems in code whatsoever.

However, `mypy` complains:
```
test_prot.py:19: error: Argument 2 to "fancy_call" has incompatible type "Callable[[str, str, float], Any]"; expected "FancyCallable"  [arg-type]
```

vytas7: I even tried forcing kw-only args on `log_call`, but it still complains...
```
test_prot.py:19: error: Argument 2 to "fancy_call" has incompatible type "Callable[[str, DefaultNamedArg(str, 'style'), DefaultNamedArg(float, 'version')], Any]"; expected "FancyCallable"  [arg-type]
```

falcon/app_helpers.py
CaselIT: Not sure actually if it effects libraries.

Yes the fact that you have to use the upper case collections is annoying.

CaselIT: also note that I think we can keep the implicit generics, since what we have here is better than nothing.

We can improve later on!

vytas7: Indeed, I think it is not a problem for now.

I have tested this, and `mypy` seems to accept it:
```python
import falcon


def hello_world(req, resp):
    resp.media = {'msg': 'Hello, World!'}


def create_app() -> falcon.App:
    app = falcon.App()
    app.add_sink(hello_world)

    return app


if __name__ == '__main__':
    app = create_app()
    print(app)
```

```
mypy --disallow-any-generics test.py 
Success: no issues found in 1 source file
```

`add_sink()` has a generic `Callable`, and if you try typing Falcon itself with `--disallow-any-generics`, it complains. But it doesn't complain when just using Falcon. If I tried passing an `int` for sink pattern, for instance, it does correctly detect the error.

CaselIT: That's great, thanks for testing

pyproject.toml
vytas7: I'm going to leave this as-is as the [specification states](https://github.com/toml-lang/toml#example):
> `# Indentation (tabs and/or spaces) is allowed but not required`

kgriffs:
🥳

vytas7:
Hi @dkbarn! We re still working on improved typing, this was just a start. But the main reason you not even seeing any `py.typed` is that the changes haven't been incorporated into any stable release yet. This and other typing improvements will be shipped as part of 4.0.0.

vytas7:
Thanks for understanding, I know our velocity has been less than stellar...

For the time being, you can also give the latest development snapshot a shot, and see if it works better with Mypy:
```
pip install git+https://github.com/falconry/falcon
```

PR #1948, opened 2021-08-06
fix(asgi): Incomplete, add finally to stream close

dimucciojonathan:
# Summary of Changes
@vytas7 

- `finally` clause was added to `falcon/asgi/app.py`, so that the stream is closed even when an exception occurs. 
- There are now two failed assertions in `tests/asgi/test_hello_asgi.py`, which are related to closing of the file. Assistance / guidance from an expert is needed.
- No tests have been drafted yet, since I haven't been able to pass the current tests yet.

# Related Issues

#1943 

# Pull Request Checklist

This is just a reminder about the most common mistakes.  Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](../CONTRIBUTING.md) at least once; it will save you a few review cycles!

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing to do.

- [ ] Applied changes to both WSGI and ASGI code paths and interfaces (where applicable).
- [ ] Added **tests** for changed code.
- [ ] Prefixed code comments with GitHub nick and an appropriate prefix.
- [ x] Coding style is consistent with the rest of the framework.
- [ ] Updated **documentation** for changed code.
    - [ ] Added docstrings for any new classes, functions, or modules.
    - [ ] Updated docstrings for any modifications to existing code.
    - [ ] Updated both WSGI and ASGI docs (where applicable).
    - [ ] Added references to new classes, functions, or modules to the relevant RST file under `docs/`.
    - [ ] Updated all relevant supporting documentation files under `docs/`.
    - [ ] A copyright notice is included at the top of any new modules (using your own name or the name of your organization).
    - [ ] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/directives.html?highlight=versionadded#directive-versionadded).
- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/index.html) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)

PR #1949, opened 2021-08-09
fix(asgi): add `finally` statement for stream.close()

### LINKED ISSUE #1943: [ASGI] close() not called if iterating over resp.stream raises an exception
At the time of writing, the `close()` method of [`resp.stream`](https://falcon.readthedocs.io/en/stable/api/request_and_response_asgi.html#falcon.asgi.Response.stream) is only called if the provided generator is successfully exhausted (or yields `None`).

However, the primary use case of the `close()` method is freeing up resources, so one might expect the method to be called regardless. It is also what [PEP 3333 compliant](https://www.python.org/dev/peps/pep-3333/#specification-details) WSGI servers do.

I would suggest introducing a `finally` clause, and moving the invocation of `close()` there.

Make sure to also write proper regression tests:
>  Kurt Griffiths @kgriffs
> We also need a test that forces stream to raise an error.
> (and ensures close() is called)

dimucciojonathan:
# Summary of Changes

This PR has the same implementation from the closed PR #1948, except this is on a new branch.

# Related Issues

Closes #1943


falcon/asgi/app.py
```diff
@@ -698,9 +698,9 @@ async def watch_disconnect():
                         '__aiter__ method. Error raised while iterating over '
                         'Response.stream: ' + str(ex)
                     )
-
-            if hasattr(stream, 'close'):
-                await stream.close()
+                finally:
+                    if hasattr(stream, 'close'):
```
vytas7, line 702: We shouldn't move `.close()` in here. This way, `.close()` is only attempted for the second branch of the main conditional statement (whether `stream` has `read()` or not). That's why the tests are failing too.

vytas7 (changes requested):
Hi again, that's a good start.

However, as pointed out inline, you should keep the original scope of `.close()`, attempting it for file-like objects with a `.read()` method too (that's probably even the primary use case for `.close()`).

- [ ] Maintain the original `.close()` scope
- [ ] Add unit tests asserting that `.close()` is attempted even upon exception while iterating over `stream`, also when calling its `read()` method.
- [x] Add a news fragment describing the bug fix

falcon/asgi/app.py
dimucciojonathan: Thanks, this makes a lot of sense!

dimucciojonathan:
@vytas7 

I have another beginner question / progress for #1943. When adding tests for `.close()` when an exception occurs, it seems like the best approach is adding assertions to the `test_genfunc_error` and `test_nongenfunc_error` functions in `test_hello_asgi.py` since this is where the exception is tested. 

I see that `.close()` is tested directly under these functions, but a different `resource` variable is used. So if I'm on the right path here, how can I verify the `HelloResource` is closed?

I will also update the branch containing the PR with the change to complete bulletpoint 1 of your previous comment.

prabin-lp (dismissed):

vytas7:
Hi again @dimucciojonathan! Just checking if you are planning to continue working on this issue, or could we possibly join forces with another contributor?

dimucciojonathan:
Hi @vytas7 , thank you for asking. I am all out of ideas for how to properly close this issue, so feel free to assign another contributor and I will follow along!

vytas7:
Thanks for the prototyping so far @dimucciojonathan -- maybe this issue is not that trivial as it looks at first glance.
For instance, in addition to what has been prototyped above, one non-obvious choice is whether to call `await close()` as it is now, or to schedule that later asynchronously.

@wsz you had some ideas how to resolve this, would you like to take over this?

vytas7:
Hi again @dimucciojonathan, I believe this is how one of the ways to fix this could look like.
Still remaining are a Towncrier newsfragment, and proper unit test coverage for the change.

vytas7 (approved):
I've taken the liberty of adding further patches to this PR; hopefully it's ready for review now.

CaselIT (approved):
lgtm

PR #1957, opened 2021-10-01
chore(Actions): add a dedicated CPython 3.5.2 gate

vytas7:


.github/workflows/tests-3.5.2.yaml
```diff
@@ -0,0 +1,40 @@
...
+        uses: docker://ubuntu:16.04
+        env:
+          DEBIAN_FRONTEND: noninteractive
+        with:
+          # NOTE(vytas): We upgrade pip in the venv in two steps, because 8.1.1
+          #   does not properly resolve python_requires.
+          args: |
+            /bin/bash -c "
+            apt-get update &&
+            apt-get install -y python3.5 python3-venv &&
```
CaselIT, line 32: shouldn't we specify 3.5.2 here?

vytas7: Ubuntu 16.04 is not updated anymore, we can IMHO just rely on the knowledge that the version is exactly 3.5.2.

CaselIT: strange that they did not update it though. 16.04 was supported until 6 month ago, did they not update py3.5?

it log seems to indicate it's 3.5.2 even if apt seems a bit confused:
```
Preparing to unpack .../python3_3.5.1-3_amd64.deb ...
Unpacking python3 (3.5.1-3) ...
```

CaselIT (approved):

PR #1958, opened 2021-10-01
feat(Response): Add viewable_as property

tipabu:
# Summary of Changes

This is very similar to `Response.downloadable_as`, but uses an `inline` disposition type to display in-browser while still offering a suggested filename when saving.

# Related Issues

#1951

# Pull Request Checklist

- [X] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/index.html) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)


falcon/response_helpers.py
```diff
@@ -105,7 +105,7 @@ def format_content_disposition(value):
     #   * When a "filename" parameter is included as a fallback (as per above),
     #     "filename" should occur first, due to parsing problems in some
     #     existing implementations.
-    return "attachment; filename=%s; filename*=UTF-8''%s" % (
+    return disposition_type + "; filename=%s; filename*=UTF-8''%s" % (
```
CaselIT, line 108: just a style thing. I would use '%s' also for `disposition_type`. also above

falcon/response.py
```diff
@@ -1014,6 +1015,24 @@ def append_link(
         format_content_disposition,
```
CaselIT, line 1015: I think it would make sense to use `functools.partial(format_content_disposition, disposition_type='attachment'),`
so that's cleared to the reader, what's passed

CaselIT (commented):
Thanks for the pr. Just a could of minor things

falcon/response.py
tipabu: Done

falcon/response_helpers.py
tipabu: Done.

falcon/response_helpers.py
```diff
@@ -85,14 +85,14 @@ def format_range(value):
     return result
 
 
-def format_content_disposition(value):
+def format_content_disposition(value, disposition_type='attachment'):
```
tipabu, line 88: I was torn about whether to add a `disposition_type: str` arg like this, or `inline: bool` and drop a
```
disposition_type = 'inline' if inline else 'attachment'
```
at the start of the function. Thoughts? If I keep it this way, should there be any sort of validation on the value of `disposition_type`? Say, to ensure there are no newlines at least?

CaselIT: I think the current version is ok, better than `inline:bool` personally, since if there is new disposition in the future, say `'tophone'`, we can just reuse the function with deprecation/etc.

Regarding validation, I think it's fine as is, since this is not really a public facing api, and even if called it just returns a string. You can't prevent an user from shooting themself in the foot :)

CaselIT (approved):
Thanks!

PR #1986, opened 2021-11-01
remove deprecated code

sneznaj:
Part of #1853


vytas7 (changes requested):
@sneznaj thanks for this, the changes look good at first glance, however, it seems that you have forgot to unlink some reference to the deprecated module or something like that -- **_all_** CI gates are failing.

Furthermore, we need to describe the removals in [`1853.breakingchange.rst`](https://github.com/falconry/falcon/blob/master/docs/_newsfragments/1853.breakingchange.rst).

PR #1990, opened 2021-11-08
docs(user): add a FAQ item + recipe on decoding raw URL

### LINKED ISSUE #1895: FAQ: explain the case where URL path contains escaped slash `%2F`

vytas7:
Closes #1895


CaselIT:
I think it would make more sense to have this include also https://github.com/falconry/falcon/pull/1945

docs/user/faq.rst
```diff
@@ -379,6 +379,33 @@ order to handle all three routes:
...
+
+Although not standardized, some WSGI servers provide the raw URL as a
+non-standard extension; for instance, Gunicorn exposes it as ``RAW_URI``,
+uWSGI calls it ``REQUEST_URI``, etc. You can implement a WSGI (or ASGI, see the
+discussion below) middleware component to overwrite the request path with the
+path component of the raw URL, see more in the following recipe:
+:ref:`raw_url_path_recipe`.
+
+In contrast to WSGI, the ASGI specification does define a standard connection
+HTTP scope variable name (``raw_path``, of type ``bytes``) for the unmodified
```
CaselIT, line 403:
```suggestion
HTTP scope variable name (the ``bytes`` string ``raw_path``) for the unmodified
```

vytas7:
When #1945 is ready for merging, we could coordinate. But I don't think the problem is the same here, only "How can I handle forward slashes within a route template field?" is.

Having path segment support is not exactly the same as being able to route percent-encoded paths. In the latter case, you can percent-encode any/multiple fields, anywhere where an arbitrary value is required.

docs/user/recipes/raw-url-path.rst
```diff
@@ -0,0 +1,110 @@
...
+``req.path`` with a value extracted from the raw URL:
+
+.. code:: python
+
+    import falcon
+
+
+    class RawPathComponent:
+        def process_request(self, req, resp):
+            raw_uri = req.env.get('RAW_URI', req.env.get('REQUEST_URI'))
```
CaselIT, line 30: maybe we could the uglier 
```
raw_uri = req.env.get('RAW_URI')
if not raw_uri:
   raw_uri  = req.env.get('REQUEST_URI')
```
since it should be faster in when finding `RAW_URI`

docs/user/recipes/raw-url-path.rst
```diff
@@ -0,0 +1,110 @@
...
+Running the above app with a supported server such as Gunicorn or uWSGI, the
+following response is rendered to a ``GET /example/item%2F001?p=value``:
+
+.. code:: json
+
+    {
+        "path": "item/001"
+    }
+
+If we removed ``RawPathComponent()`` from the app's initializer, the request
```
CaselIT, line 57:
```suggestion
If we removed ``RawPathComponent()`` from the app middleware list, the request
```

docs/user/recipes/raw-url-path.rst
```diff
@@ -0,0 +1,110 @@
...
+would be routed as ``/example/item/001``, and no matching resource could be
```
CaselIT, line 58:
```suggestion
would be routed as ``/example/item/001``, and no matching resource would be
```

docs/user/recipes/raw-url-path.rst
```diff
@@ -0,0 +1,110 @@
...
+    import falcon.asgi
+
+
+    class RawPathComponent:
+        async def process_request(self, req, resp):
+            raw_path = req.scope.get('raw_path')
+
+            # NOTE: Decode the tunneled raw path from the raw_path bytestring.
+            if raw_path:
+                req.path = raw_path.decode('latin1')
```
CaselIT, line 88: can't path contain any utf8 char? are they encoded to latin1 using the `%` format?

docs/user/recipes/raw-url-path.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Running the above snippet with ``uvicorn`` (that supports ``raw_path``), the
+percent-encoded ``path`` field is now correctly handled for a
+``GET /example/item%2F001?p=value`` request:
+
+.. code:: json
+
+    {
+        "path": "item/001"
+    }
```
CaselIT, line 110: maybe the not found part could be at the end here since it's valid in both cases

CaselIT (changes requested):
As mentioned I think this should take into consideration the pr that adds support for path

CaselIT:
> 
> 
> When #1945 is ready for merging, we could coordinate. But I don't think the problem is the same here, only "How can I handle forward slashes within a route template field?" is.
> 
> Having path segment support is not exactly the same as being able to route percent-encoded paths. In the latter case, you can percent-encode any/multiple fields, anywhere where an arbitrary value is required.

yes, but.. for example the example presented here would be better solved by the `:path` field converter, so I think a different example would be more appropriate

CaselIT:
> When #1945 is ready for merging

It's currently waiting review

docs/user/recipes/raw-url-path.rst
vytas7: No, it normally shouldn't. `raw_path` is a bytestring, and any utf8 char should be percent-encoded. Maybe we should even use `ascii` instead :thinking: I know Nginx does, for instance.

In any case, that's a good question. Let me take a look at this.

docs/user/recipes/raw-url-path.rst
CaselIT: maybe just a note above is enough. the if form is not pretty

vytas7: Heh, I wasn't trying to optimize recipes _that_ much.

Maybe `or` could be more performant?
```python
            raw_uri = req.env.get('RAW_URI') or req.env.get('REQUEST_URI')
```

CaselIT: seems the best option 👍

docs/user/recipes/raw-url-path.rst
vytas7: Hm, I actually wrote _could_ intentionally here, but on consideration, your suggestion makes sense.

docs/user/recipes/raw-url-path.rst
vytas7: I've changed to `ascii`, and added a short `NOTE` explaining the revised choice. Nginx does that too.

vytas7:
Those were good points @CaselIT, I have now updated the examples to illustrate a case where no routing system could tackle this without having access to the raw path, no matter how advanced the router was.

CaselIT (approved):

CaselIT:
thanks for the updates!

PR #1991, opened 2021-11-13
feat(static): set `Content-Length` for static file responses

vytas7:
Also fix range request support across different WSGI servers by always providing a Python wrapper without a file descriptor, since many `sendfile` implementations are buggy wrt not respecting the current file offset; both Gunicorn (https://github.com/benoitc/gunicorn/issues/2669) and uWSGI are affected according to my testing.

There is still an open question remaining what to do with Meinheld: I've filed this as a separate issue (#1992).

### DIFF
--- a/falcon/routing/static.py
+++ b/falcon/routing/static.py
@@ -15,7 +15,7 @@ def _open_range(file_path, req_range):
         file_path (str): Path to the file to open.
         req_range (Optional[Tuple[int, int]]): Request.range value.
     Returns:
-        tuple: Three-member tuple of (stream, size, content-range).
+        tuple: Three-member tuple of (stream, content-length, content-range).
             If req_range is ``None`` or ignored, content-range will be
             ``None``; otherwise, the stream will be appropriately seeked and
             possibly bounded, and the content-range will be a tuple of
@@ -40,7 +40,7 @@ def _open_range(file_path, req_range):
         # NOTE(vytas): Wrap in order to prevent sendfile from being used, as
         #   its implementation was found to be buggy in many popular WSGI
         #   servers for open files with a non-zero offset.
-        return _BoundedFile(fh, -start), size, (size + start, size - 1, size)
+        return _BoundedFile(fh, -start), -start, (size + start, size - 1, size)
 
     if start >= size:
         fh.close()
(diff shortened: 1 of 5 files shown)


tests/test_wsgi_servers.py
```diff
@@ -232,3 +239,43 @@ def test_static_file(self, server_url):
...
+        file_size = os.path.getsize(__file__)
+        assert len(resp.content) == content_length == file_size
+
+    @pytest.mark.parametrize(
+        'byte_range,expected_head',
+        [
+            ('7-', b'hashlib'),
+            ('2-6', b'port'),
+            ('32-38', b'random'),
+            ('-47', b'The content of this comment is part of a test.\n'),
```
CaselIT, line 253: I'm -1 on using this file to test stuff, since each time we change something this test will break for no particular reason.

Let's create a sample file we can use so that we avoid that issue

CaselIT (commented):
looks ok. Just a test issue that I think would make sense changing before merging

Also don't we also miss the asgi tests?

tests/test_wsgi_servers.py
vytas7: Right... hopefully you could accept it this time, because apart from that last comment, this was already in place for testing static files (see the above test), which you accepted on review before.

vytas7:
We already have unit tests covering ASGI too, but I will check if all the functionality is verified in unit tests as well.

Otherwise, ASGI has no concept of `sendfile`, and the response is rendered purely by the app, so adequate unit testing should be enough.

tests/test_wsgi_servers.py
CaselIT: I guess we could leave a `# TODO` comment for this?

vytas7: Will add.

CaselIT (dismissed):

tests/test_wsgi_servers.py
CaselIT: thanks

falcon/routing/static.py
```diff
@@ -15,37 +15,48 @@ def _open_range(file_path, req_range):
...
-        return fh, None
+        return fh, size, None
 
     start, end = req_range
-    size = os.fstat(fh.fileno()).st_size
     if size == 0:
-        # Ignore Range headers for zero-byte files; just serve the empty body
-        # since Content-Range can't be used to express a zero-byte body
-        return fh, None
+        # NOTE(tipabu): Ignore Range headers for zero-byte files; just serve
```
vytas7, line 31: For reference: @tipabu I prefixed your comments with your GH nick, and refactored the code slightly to work around issues with buggy `sendfile` implementations in many popular WSGI servers such as Gunicorn, uWSGI, etc.

CaselIT (approved):

PR #2001, opened 2022-01-07
feat(validators): add a specialized exception class for media validation

### LINKED ISSUE #1320: JSON validator - change exception class
I found a little bit harder to override the exception thrown by falcon.media.validators.jsonschema.validate(). When I raise an instance of HTTPBadRequest is really easy to override it to get my custom error message.

Looking at the code it seems like the exception class is hardcoded, so it will always be the pristine HTTPBadRequest that comes with falcon, not mine, so the error is not myapi-compilant.

¿Is there another way of doing this by not duplicating the validator code to raise my own exception?

vytas7:
Closes #1320

### DIFF
--- a/falcon/errors.py
+++ b/falcon/errors.py
@@ -2472,9 +2472,7 @@ class MediaValidationError(HTTPBadRequest):
     attribute using the "raise ... from" form when raising.
 
     Note:
-        All the arguments should be passed as keyword only. Using them as positional
-        arguments will raise a deprecation warning and will result in an
-        error in a future version of falcon.
+        All the arguments must be passed as keyword only.
 
     Keyword Args:
         title (str): Error title (default '400 Bad Request').
@@ -2506,8 +2504,7 @@ class MediaValidationError(HTTPBadRequest):
             base articles related to this error (default ``None``).
     """
 
-    @deprecated_args(allowed_positional=0)
-    def __init__(self, title=None, description=None, headers=None, **kwargs):
+    def __init__(self, *, title=None, description=None, headers=None, **kwargs):
         super().__init__(
             title=title,
             description=description,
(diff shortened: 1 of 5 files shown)


falcon/errors.py
```diff
@@ -2458,6 +2459,63 @@ def description(self, value):
...
+    when ``req.media`` is successfully deserialized, but fails to validate
+    against the configured schema.
+
+    The cause of this exception, if any, is stored in the ``__cause__``
+    attribute using the "raise ... from" form when raising.
+
+    Note:
+        All the arguments should be passed as keyword only. Using them as positional
+        arguments will raise a deprecation warning and will result in an
+        error in a future version of falcon.
```
CaselIT, line 2477: since this is new why we don't just place a `*` in the init?

CaselIT (approved):
just a suggestion, but it's fine either way

falcon/errors.py
vytas7: Hm, good point. I've just copypasted from other exceptions :slightly_smiling_face:

falcon/errors.py
```diff
@@ -2472,9 +2472,7 @@ class MediaValidationError(HTTPBadRequest):
     attribute using the "raise ... from" form when raising.
 
     Note:
-        All the arguments should be passed as keyword only. Using them as positional
-        arguments will raise a deprecation warning and will result in an
-        error in a future version of falcon.
+        All the arguments must be passed as keyword only.
```
CaselIT, line 2475: not sure if we need this note. probably just for the 3.x series, since it behaves differently from the others.
then on v4 we can cleanup or remove

CaselIT (approved):

falcon/errors.py
vytas7: Yeah, I thought it wouldn't hurt to stress the difference. Yes, I was thinking we could bulk-remove all these notes when we drop non-kw-args in 4.0.

PR #2002, opened 2022-01-09
feat(testing): add shorthand aliases to `simulate_*` methods

### LINKED ISSUE #1806: Add non-prefixed aliases to test client methods, e.g., simulate_get() -> get()
The original intention of including the `simulate_*` prefix was to make tests self-documenting in terms of which requests are real, and which are simulated. However, it does make test cases more verbose and requires a bit more typing. Is it worth it?

Also, consider that httpx uses the same interface regardless of whether requests are being sent directly to an app vs. over a socket.

Thoughts?

vytas7:
Closes #1806


CaselIT (approved):
looks ok!

PR #2013, opened 2022-01-31
chore: the new `black` is the new black

vytas7:
It seems that `black` has finally turned stable, which also brings the usual assortment of formatting changes.


CaselIT:
They seem reasonable enough!

Issue #473, opened 2015-03-24
Document the release process


kgriffs:
Proposed steps to include:
- [ ] Final performance testing/profiling and tuning (mitigate any regressions vs. previous release)
- [ ] Generate all docs, audit, and polish/fix as necessary
  - [ ] Generate Dash documents and submit PR
- [ ] Updating website with perf numbers (would be great to automate perf testing/tracking at some point)
- [ ] Create changelog
    - Update `AUTHORS` and contributors in the `towncrier` template (by running `tools/add_contributors.py`).
    - `towncrier` should generate the changes and put them in the correct _docs_ location.
- [ ] Modifying version.py
- [ ] Tagging/branching
- [ ] Beta(s)
- [ ] Release candidate(s)
- [ ] Packaging steps, including signing and uploading to pypi
- [ ] Announcement email

vytas7:
The document should also mention the last sanity checks verifying that there are no breaking changes in a SemVer-minor release.
Because our users love SemVer (particularly @hynek :smiling_imp:).

PR #2014, opened 2022-02-06
chore(tools): add a tool to aggregate contributors from GitHub commits

### LINKED ISSUE #1533: Automate aggregation of the contributor list for release
Walk GitHub history since the given previous release treeish, and aggregate contributors' GitHub usernames if possible, otherwise pick their name and surname, in a similar fashion to https://falcon.readthedocs.io/en/stable/changes/2.0.0.html#contributors-to-this-release (this was done by @vytas7 by a mix of manual work and scripts in REPL).

The output would be appended or linked to the change list distilled by Towncrier when doing release.

- [x] Output contributors list to changelog
- [x] Update AUTHORS file

vytas7:
Closes #1533

Actually running this tool should be done as part of the release process. I added a mention in #473.
We could probably add a check in CI for pre-release tags... but I left that as a future exercise for now.

Another limitation that we could hopefully live with for a while is that only the primary author with a GitHub login is considered, so commits made by multiple first-time contributors in a single PR might still need a manual review.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -114,6 +114,7 @@ listed below by date of first contribution:
 * Vasilis Gerakaris (vgerak)
 * Eleni E (Contessina)
 * Betsy Behun (the-bets)
+* Mihai Todor (mihaitodor)
 * Abid Ahmad (abidahmadq)
 * KhiemTon (nix010)
 * dimjon (dimucciojonathan)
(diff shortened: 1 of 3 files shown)


CaselIT (approved):
I feel like we could have used git for this, but I guess we would not get the github links

in any case lgtm! thanks

PR #2027, opened 2022-02-13
fix(asgi.stream): no longer hang upon last empty body chunk

### LINKED ISSUE #2024: ASGI: iterating over `req.stream` hangs for chunked requests
When reading a streaming request without `Content-Length` (i.e., using "chunked" `Transfer-Encoding`), iterating over `req.stream` hangs in the case the request payload consists of more than one chunk.

It looks like this is caused by the receive loop implementation in `asgi/stream.py`. It only checks for the number of remaining bytes, and disregards the `more_body: False` hint in the case of an empty body. The logic to account for it is in place, but it is not reached due to the `continue` clause for an empty body chunk in the following event sequence:
```
RECV {'type': 'http.request', 'body': b'123456789abcdef\n', 'more_body': True}

RECV {'type': 'http.request', 'body': b'123456789abcdef\n', 'more_body': True}

RECV {'type': 'http.request', 'body': b'', 'more_body': False}
```

Eventually, the client (I was using `httpx`) times out, and an `http.disconnect` event is received which is handled correctly. But the request has already failed (timeout) from the client's perspective at this point.

vytas7:
Fixes #2024


CaselIT (approved):
devil's in the details! should we backport to 3.0.1 and release a 3.0.2?

tests/asgi/test_boundedstream_asgi.py
```diff
@@ -164,6 +164,30 @@ async def test_iteration():
...
 
+@falcon.runs_sync
+async def test_iterate_streaming_request():
+    events = iter(
+        (
+            {'type': 'http.request', 'body': b'Hello, ', 'more_body': True},
+            {'type': 'http.request', 'body': b'World', 'more_body': True},
+            {'type': 'http.request', 'body': b'!\n', 'more_body': True},
+            {'type': 'http.request', 'body': b'', 'more_body': False},
+            {'type': 'http.disconnect'},
```
kgriffs, line 175: Do we want to not break out of the loop before consuming the http.disconnect event? Hopefully ASGI servers don't care if the app ever receives the disconnect...

kgriffs (approved):

tests/asgi/test_boundedstream_asgi.py
vytas7: We do want, and the code handles it correctly (there are many other tests verifying this behavior). Maybe it's not completely obvious what this means since it's an isolated unit test... But if iteration is used to parse the request body, one cannot wait for a disconnect from the client which is still waiting for a response (the very crux of this bug). I'll add a `NOTE` in a follow up PR.

vytas7:
> devil's in the details! should we backport to 3.0.1 and release a 3.0.2?

That's a good idea @CaselIT; let's discuss it on Gitter.
Another alternative is just pushing out 3.1 since it's well overdue too! :shipit:

PR #2031, opened 2022-02-21
chore(mailman): gate non-breaking releases on GNU Mailman 3

### LINKED ISSUE #259: Gate non-breaking releases on large 3rd-party projects that are known to use Falcon

vytas7:
This pull request adds a GNU Mailman 3 CI gate that is proposed to be run on every `master` merge. Upon every run, Mailman Core is cloned from GitLab, its `tox.ini` patched to add an additional `falcon` env where we simply uninstall Falcon and reinstall it from local source (effectively building a cythonized PEP 517 wheel).

~~A successful run is demoed here: https://github.com/falconry/falcon/runs/5279630719?check_suite_focus=true~~

The following concessions had to be made, because of the way the things are, my incompetence, or a combination thereof:
* ~~I couldn't use @maxking 's Docker image, because GitHub Actions seem to be very hostile to setting a non-root `USER`.~~
* I wished to run tests for a stable release of Mailman instead of the latest `master` commit. However, it seems that some tests inside the 3.3.5 tree are failing, so I abandoned this idea for the time being.

**Edit**: now using the Falcon flavour of @maxking's image. An example of the latest incarnation: https://github.com/falconry/falcon/runs/5351987534?check_suite_focus=true.

It is looking unlikely at this point that we'll be able to resurrect the Hug gate (Hug is still on Falcon 2.0), but I have left the scripts and definitions in question intact for now. (Because [HOPE](https://github.com/hugapi/HOPE) dies last as they say!)

Closes #259

### DIFF
--- a/.github/workflows/tests-mailman.yaml
+++ b/.github/workflows/tests-mailman.yaml
@@ -1,10 +1,9 @@
 name: Run tests (GNU Mailman 3)
 
 on:
+  # Trigger the workflow on master but also allow it to run manually.
+  workflow_dispatch:
   push:
-    branches:
-      - "*"
-  pull_request:
     branches:
       - master
 
--- a/.github/workflows/tests-mailman.yaml
+++ b/.github/workflows/tests-mailman.yaml
@@ -1,15 +1,9 @@
 name: Run tests (GNU Mailman 3)
 
 on:
-  # # Trigger the workflow on master but also allow it to run manually.
-  # workflow_dispatch:
-  # push:
-  #    branches:
-  #      - master
+  # Trigger the workflow on master but also allow it to run manually.
+  workflow_dispatch:
   push:
-    branches:
-      - "*"
-  pull_request:
     branches:
       - master
 
--- a/.github/workflows/tests-mailman.yaml
+++ b/.github/workflows/tests-mailman.yaml
@@ -30,7 +30,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y rustc
-          sudo apt-get install -y libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev libbz2-dev
+          sudo apt-get install -y openssh-server libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev libbz2-dev
 
       - name: Install Python dependencies
         run: |
@@ -45,5 +45,5 @@ jobs:
 
       - name: Run tox
         run: |
-          python -m pip install --upgrade pip tox
-          tox -c .ecosystem/mailman/tox.ini -e py38-nocov
+          cd .ecosystem/mailman
+          tox -e py38-nocov
--- a/.github/workflows/tests-mailman.yaml
+++ b/.github/workflows/tests-mailman.yaml
@@ -10,8 +10,8 @@ on:
 
 jobs:
   run_tox:
-    name: tox -e (GNU Mailman 3)
-    runs-on: ubuntu-latest
+    name: tox -e py38_nocov
+    runs-on: ubuntu-20.04
 
     steps:
       - name: Checkout repo
@@ -19,14 +19,31 @@ jobs:
         with:
           fetch-depth: 2
 
+      - name: Set up Python
+        uses: actions/setup-python@v2.1.4
+        with:
+          python-version: 3.8
+
+      - name: Install dependencies
+        env:
+          DEBIAN_FRONTEND: noninteractive
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y rustc
+          sudo apt-get install -y libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev libbz2-dev
+
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip tox
+          python --version
+          pip --version
+          tox --version
+
       - name: Fetch Mailman Core
         run: |
           tools/testing/fetch_mailman.sh
 
       - name: Run tox
-        uses: docker://maxking/mailman-ci-runner
-        env:
-          DEBIAN_FRONTEND: noninteractive
-        with:
-          args: |
-            /bin/bash -c "tox -c .ecosystem/mailman/tox.ini -e py38_nocov"
+        run: |
+          python -m pip install --upgrade pip tox
+          tox -c .ecosystem/mailman/tox.ini -e py38_nocov
--- a/.github/workflows/tests-mailman.yaml
+++ b/.github/workflows/tests-mailman.yaml
@@ -29,8 +29,7 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           sudo apt-get update
-          sudo apt-get install -y rustc
-          sudo apt-get install -y openssh-server libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev libbz2-dev
+          sudo apt-get install -y libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev libbz2-dev
 
       - name: Install Python dependencies
         run: |
@@ -45,5 +44,4 @@ jobs:
 
       - name: Run tox
         run: |
-          cd .ecosystem/mailman
-          tox -e py38-nocov
+          tox -c .ecosystem/mailman/tox.ini -e falcon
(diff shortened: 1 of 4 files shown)


tools/testing/fetch_mailman.sh
```diff
@@ -0,0 +1,29 @@
...
+# git checkout tags/$MAILMAN_VERSION
+
+# NOTE(vytas): Patch tox.ini to introduce a new Falcon environment.
+cat <<EOT >> tox.ini
+
+[testenv:falcon]
+commands =
+    pip uninstall -y falcon
+    pip install $FALCON_ROOT
+    python -m nose2 -v {posargs}
```
maxking, line 28: Instead of overriding the `commands`, you can add a new `commands_pre` section, which will install falcon. It might be better if in future we decide to change the actual `command` we use in CI.

https://tox.wiki/en/latest/config.html#conf-commands_pre

tools/testing/fetch_mailman.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# TODO(vytas): Detect the latest version from git tags or PyPi JSON.
+# NOTE(vytas): One approach: `curl -Ls https://pypi.org/pypi/mailman/json | jq -r .info.version`
+# MAILMAN_VERSION=3.3.5
+
+# Clean up in case we are running locally and not in CI
+rm -rf $MAILMAN_PATH
+
+mkdir -p .ecosystem
+git clone https://gitlab.com/mailman/mailman.git/ $MAILMAN_PATH
```
maxking, line 14: It would be good to do something like `--depth 100` or something, 10k commits can be expensive to clone on every CI run and isn't really needed.

maxking (commented):
Overall, this looks great!

> I couldn't use @maxking 's Docker image, because GitHub Actions seem to be very hostile to setting a non-root USER.

If that's the only issue, i'd say we can create a 2nd one, which shares most of the dockerfile with the current one, except the `USER` directive, that you can use here.  This will achieve two things IMO - 1) Reduce your burden of maintaining Mailman's testing env as we add/remove support for Python versions (happens yearly) or add new system level dependencies (happens much less often); 2) Reduce the environment differences when debugging issues. Debugging test failures that happen in this CI but not in ours or vice-versa would be a huge time sink.

tools/testing/fetch_mailman.sh
vytas7: Ah, right. Since I'm not checking out any tag as I intended, even a depth of 1 would be enough :)

tools/testing/fetch_mailman.sh
vytas7: Spot on. Done :heavy_check_mark:

vytas7:
@maxking Agreed re benefits of a specialized Docker image. I would be happy to try such an image in the case it isn't too much effort for you to create one.

maxking:
> I would be happy to try such an image in the case it isn't too much effort for you to create one.

https://github.com/maxking/mailman-ci-runner/pkgs/container/mailman-ci-runner-falcon

I added an `USER root` [at the end of the Dockerfile](https://github.com/maxking/mailman-ci-runner/blob/master/.github/workflows/docker-publish-2.yaml#L72) and re-built the same image. Let me know if that helps with GH Actions.

maxking:
@vytas7 I think part of the problem is that due to switching user, the $PATH isn't set correctly, so because of `/usr/local/bin/` not being on $PATH.

You might be able to fix that in the tox env, but I think it should really be fixed in the container image. I should've tried the image before giving it out, so my bad :(
I'll try to add it to the image this evening. You might be able to run quick test by adding export PATH=/usr/local/bin:$PATH in commands.

maxking:
Nevermind, looks like you've got them kicking already with moving the checkout to the final command.

vytas7:
No worries @maxking. Against all the odds, something is running there :crossed_fingers:
https://github.com/falconry/falcon/runs/5309572521?check_suite_focus=true

vytas7:
Hm, same problems as I had on my home workstation, IPv6 doesn't work it seems. And some other failures that I haven't seen before :thinking:

maxking:
One suggestion would be to configure mailman to bind to 127.0.0.1 instead of `localhost` and see if that stops mailman from trying to bind to ipv6.

https://gitlab.com/mailman/mailman/-/blob/master/src/mailman/config/schema.cfg#L418

The default is `localhost`, so it depends on the entries in`/etc/hosts` I suppose.

https://gitlab.com/mailman/mailman/-/blob/master/src/mailman/testing/testing.cfg#L31 is where we can make the change upstream. If this fails locally for you, can add `bind = 127.0.0.1` locally and see if that fixes tests for you? We can get this upstream too, atleast in the testing config.

vytas7:
I can test that.

And how about other two failures, is that just fallout from the former?

vytas7:
This was my diff btw (bar `tox.ini` patches):
```diff
--- a/src/mailman/testing/testing.cfg
+++ b/src/mailman/testing/testing.cfg
@@ -29,6 +29,7 @@ incoming: mailman.testing.mta.FakeMTA
 configuration: python:mailman.testing.passlib
 
 [webservice]
+hostname: 127.0.0.1
 port: 9001
 workers: 1
 
```

maxking:
Ok, so I've been trying to get the tests running inside a Github Codespace, hopefully that is close environment to Actions. The failing tests are failing because we prevent running of Mailman command as root by default (some users accidentally run as root and then cause a whole lot of permission issues), so running the tests as root triggered those code paths.

I seem to have fixed the issue with [this MR](https://gitlab.com/mailman/mailman/-/merge_requests/966) but I started seeing another failure running as root, the fix for it [still running CI](https://gitlab.com/mailman/mailman/-/merge_requests/967), but should be merged soon. 

Finally, the ipv6 issue is difficult to fix since it is coming from aiosmtpd and moving to an IP address (127.0.0.1) breaks a bunch of other SSL related stuff we have for smtps testing. So, what I did was get rid of `::1` from `/etc/hosts` and that worked.

```
sed '/::1/d' /etc/hosts > /tmp/hosts && cat /tmp/hosts > /etc/hosts
```
(For some reason, i can't do in-place edit using `sed` in the container, it fails to write to a temp file, even though running as root should allow me to do that. I didn't dig further and just used a temp file instead)

After this, I was able to run the full test suite (from [this MR](https://gitlab.com/mailman/mailman/-/merge_requests/967))

```
----------------------------------------------------------------------
Ran 2095 tests in 467.698s

OK (skipped=1, expected failures=3)
```

vytas7:
Thanks a lot for digging into this @maxking, the tests pass now. 
The question is of course the maintainability of these hacks compared to simply using Actions' default env, and bumping the Python version when needed. But since you've put lots of good work into this it would be a shame to go back.

Could you publish that CI image onto the Docker hub? Or should I refer to it in the current way? (I had to look to its push actions to get the exact reference to `:master` etc on ghcr.io).

.github/workflows/tests-mailman.yaml
```diff
@@ -0,0 +1,36 @@
+name: Run tests (GNU Mailman 3)
+
+on:
+  # # Trigger the workflow on master but also allow it to run manually.
+  # workflow_dispatch:
+  # push:
+  #    branches:
+  #      - master
```
CaselIT, line 8: do we want to make this run in every branch or only on master + manual as the documentation suggests?

vytas7: No, probably only master+manual. I've now converted this PR back to Draft until it's ready for review again.

maxking:
@vytas7 with the docker hub pull restrictions, I had to move all my images to ghcr.io actually for use in CI, because it will go over the limit pretty fast. The `:master` tag is probably the right one and you have the right pull URL. I updated the readme in the repo on instructions to pull the image, but I don’t know how to actually make https://github.com/maxking/mailman-ci-runner/pkgs/container/mailman-ci-runner-falcon/15817109?tag=master this page show up on the top of the package info for users.

About the hacks, yeah the `sed` hack is really not ideal and it did end up being more work that initially expected to get the container image and tests running. I wonder why do those issues not appear in default actions, I thought all of them run in containers 🤔

maxking (approved):

vytas7:
No problems re the ghcr.io URL as long as it is confirmed to be the intended form.

These containers are indeed not created equal, it seems that if you use a customer `docker://` action, it is run via some sort of secondary containerization from inside a top level container (using GH stock images).

.github/workflows/tests-mailman.yaml
vytas7: Ready for review now!

CaselIT (approved):

PR #2032, opened 2022-03-01
test(test_post_read_bounded_stream_large): validate body digest

kgriffs:
# Summary of Changes

This patch adds a check to `test_post_read_bounded_stream_large()` to validate the digest of the body to ensure that the chunks are read completely with no overlap or repetition.

### DIFF
--- a/tests/asgi/_asgi_test_app.py
+++ b/tests/asgi/_asgi_test_app.py
@@ -108,11 +108,15 @@ class Bucket:
         resp.text = await req.stream.read()
 
     async def on_put_drops(self, req, resp):
+        # NOTE(kgriffs): Integrity check
+        sha1 = hashlib.sha1()
+
         drops = 0
         async for drop in req.stream:
             drops += 1
+            sha1.update(drop)
 
-        resp.media = {'drops': drops}
+        resp.media = {'drops': drops, 'sha1': sha1.hexdigest()}
 
 
 class Feed:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,14 +35,14 @@ $ pip install -U black
 $ black .
 ```
 
-You can check all this by running ``tox`` from within the Falcon project directory (requires your environment to be based on Python 3.8 or 3.10):
+You can check all this by running ``tox`` from within the Falcon project directory. Your environment must be based on CPython 3.8 or 3.10:
 
 ```bash
 $ pip install -U tox
 $ tox --recreate
 ```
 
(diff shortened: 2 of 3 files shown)


vytas7 (approved):
LGTM

tests/asgi/_asgi_test_app.py
```diff
@@ -108,11 +108,15 @@ async def on_post(self, req, resp):
         resp.text = await req.stream.read()
 
     async def on_put_drops(self, req, resp):
+        # NOTE(kgriffs): Integrity check
+        sha1 = hashlib.sha1()
```
CaselIT, line 112: let's use something different, even if not necessary since static analysis tools with their immense wisdom may flag it as "unsecure algorithm used" or something similar.

python 3.5 has sha256, 384 and 512, so we could use one of those

CaselIT (changes requested):
just a suggestion. it's mostly to avoid false positive on static analysis tools

tests/asgi/_asgi_test_app.py
vytas7: For integrity checking even CRC32 is still very prevalent, and you surely use it every day, both for your TCP network packets and your disk's sectors, and it doesn't mean it is "insecure". Furthermore, we already use `sha1` for other tests, so we should make a consistent change throughout the whole codebase if we decide this.
So make we could switch to [`crc32()`](https://docs.python.org/3.10/library/zlib.html#zlib.crc32) to make it obvious that it is merely a checksum? :smiling_imp:

@CaselIT do you know of any such tools that would flag references to `sha1`?

CaselIT: I know that here it's not an issue..

> @CaselIT do you know any such tools that would flag references to `sha1`?

sonarqube will flag it, but I don't seem to find a list of rules. I'm sure it's not the only one.

Note that I think we can merge as is, just mentioned since static analysis tools can be quite dumb at times

vytas7: Yeah, that's a good point regarding static analysis tools. I was genuinely curious, good to know wrt sonarcube.

OTOH I don't think we should care too much about these because is inside the `tests/` directory, which is not installed when installing `falcon`. So it is not an attack vector even if it was a real issue.

I'm not opposed to changing to another hashing function, but if we decide to change, it makes sense to update all the tests to use our new checksum function of choice.

CaselIT: makes sense. let's keep the change in mind, and if we decide to do it we should change them in a single issue.

CaselIT (approved):

Issue #2044, opened 2022-03-22
Add `__repr__` to test Result


vytas7:
At the time of this writing, [`falcon.testing.Result`](https://falcon.readthedocs.io/en/stable/api/testing.html#falcon.testing.Result) has no custom `__repr__`:
```python
>>> import falcon
>>> import falcon.testing
>>> app = falcon.App()
>>> client = falcon.testing.TestClient(app)
>>> result = client.get('/something')
>>> result
<falcon.testing.client.Result object at 0x7f292ce046d0>
```

`__repr__` is also picked up and shown by the popular `pytest`.

It could be nice to add a representation with a concise summary describing the most important attributes of the result (status code etc).

CaselIT:
I think it would make sense to log the status and maybe a substring of the text, something like the first 50 / 100 chars? or maybe the first 25 - last 25?

vytas7:
Yeah, something like that. Maybe the response's `Content-Type` too?

RioAtHome:
Hello, i would like to work on this issue.
From my understatement of this issue i just need to add a `__repr__` function to `falcon.testing.client.Result` class.
Should i just log the status code, a part of the text and the header content-type ?

vytas7:
Hi @RioAtHome!
Sure, go ahead!

Yes, your understanding is correct -- you just need to add a `__repr__` for the class in question (and add a test for it). Although it is a bit open ended -- feel free to tweak that representation to your liking or come up with your own suggestions.

RioAtHome:
Hello!.
How about logging the status code, most important headers(Content-Type, Content-length, Authorization and Cookie) and parts of the text/content(as stated above, just log the beginning and end)?

vytas7:
Hi again!
Cookies are probably not the most common case for APIs, although it of course varies.
`__repr__` should be reasonably short... So probably we cannot afford cramming in too much stuff, although it is tempting.

vytas7:
Implemented in https://github.com/falconry/falcon/pull/2046

PR #2046, opened 2022-03-28
feat(testing): add __repr__ to testing.Result

RioAtHome:
# Summary of Changes
Added `__repr__` to `testing.Result`, where it shows the status code, content-type header and first 20 char.
Also added a simple test.

# Related Issues
Issue #2044 

# Pull Request Checklist


falcon/testing/client.py
```diff
@@ -347,6 +347,13 @@ def json(self) -> Optional[Union[dict, list, str, int, float, bool]]:
 
         return json_module.loads(self.text)
 
+    def __repr__(self):
+        body_text = '{}'.format(self.text[:20])
+
+        return '<Status: {}, Content-Type: {}, Text: {}>'.format(
+            self.status_code, self.headers['Content-Type'], body_text
```
vytas7, line 354: What happens if the response payload is invalid UTF-8?
```python
>>> testing.Result([b'\xc3\x28'], falcon.HTTP_418, {})
  File "/tmp/RioAtHome/falcon/falcon/testing/client.py", line 351, in __repr__
    body_text = '{}'.format(self.text[:20])
  File "/tmp/RioAtHome/falcon/falcon/testing/client.py", line 339, in text
    self._text = self.content.decode(encoding)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 0: invalid continuation byte
```

It is IMHO better to prevent `__repr__` from raising almost at any cost, since it can be very confusing to debug (IIRC the popular `pytest` has some safeguard to handle even this scenario, but it ain't pretty).

tests/test_utils.py
```diff
@@ -764,6 +764,13 @@ def test_status(self, app):
         result = client.simulate_get()
         assert result.status == falcon.HTTP_702
 
+    def test_repr_result(self, app):
```
vytas7, line 767: IMHO it would be nice to add more tests both for WSGI and ASGI flavours of the framework, potentially using `simulate_request()` and friends.

vytas7 (changes requested):
Thanks for getting the ball rolling, this is a great start.

I've provided some feedback inline.

falcon/testing/client.py
RioAtHome: How about making `__repr__` log the content of the payload instead of text?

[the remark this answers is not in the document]
RioAtHome: i made some changes, made `__repr__` start with prefix `Result<...>`, and put an if statement to check whether content-type header is available or not, so nothing can get raised from `__repr__`.

falcon/testing/client.py
```diff
@@ -347,6 +347,13 @@ def json(self) -> Optional[Union[dict, list, str, int, float, bool]]:
 
         return json_module.loads(self.text)
 
+    def __repr__(self):
+        content_type = ''
+        if 'Content-Type' in self.headers.keys():
```
vytas7, line 352: Not that we care too much about the performance of the testing utils, but this is a fairly wasteful test for inclusion...
IMHO it is much easier to simply do `self.headers.get('Content-Type', '')`.

falcon/testing/client.py
```diff
@@ -347,6 +347,13 @@ def json(self) -> Optional[Union[dict, list, str, int, float, bool]]:
...
+            content_type = self.headers['Content-Type']
+
+        return 'Result<{} {} {}>'.format(self.status, content_type, self.content[:20])
```
vytas7, line 355: So if `content_type == ''`, would you get two spaces between status and content?

vytas7 (changes requested):
This is shaping up!
Should we also add a towncrier newsfragment for this improvement?

falcon/testing/client.py
RioAtHome: Ahh, sorry about that.
Will correct in the next commit.

RioAtHome:
> This is shaping up! Should we also add a towncrier newsfragment for this improvement?

towncrier newsfragment?
Sorry to be a bother, but any pointers on what this is?, and where to start if i want to add it?

CaselIT:
> towncrier newsfragment?

It's just a few words that should appear on the changelog. See other newsfragments here https://github.com/falconry/falcon/tree/master/docs/_newsfragments and this section of the contributing file https://github.com/falconry/falcon/blob/master/CONTRIBUTING.md#changelog

falcon/testing/client.py
RioAtHome: First way that came to my mind to handle extra spaces is this:
```
    def __repr__(self):
        content_type = self.headers.get('Content-Type', '')
        repr_result = 'Result<{} {} {}>'.format(
            self.status, content_type, self.content[:20]
        )

        return " ".join(repr_result.split())
```
And i have a feeling this isn't the best way to go about it.

Or i can change the separator instead of spaces, to colon or maybe a comma.
Like this `Result<200 OK::b'hello'>`
What do you think?

RioAtHome:
@CaselIT Thanks for the replying.
I like to point out that the link to towncrier in the changelog section of the contributing file does not work.

vytas7:
Unfortunately towncrier isn't particularly well maintained; we're looking for alternatives, but I don't know when the change could actually happen.
Anyway, I'll update the link to https://towncrier.readthedocs.io/, that should be fixed as per https://github.com/twisted/towncrier/issues/367 (although it still says `17.08` in the title? :thinking:)

falcon/testing/client.py
RioAtHome: Changed the syntax to `Result<Status, Content-Type, Content>`.
Where now if no content type header was included, `__repr__` will give a prettier result imo.

falcon/testing/client.py
RioAtHome: i have edited it to `self.headers.get('Content-Type', '')`.

falcon/testing/client.py
```diff
@@ -347,6 +347,11 @@ def json(self) -> Optional[Union[dict, list, str, int, float, bool]]:
 
         return json_module.loads(self.text)
 
+    def __repr__(self):
+        content_type = self.headers.get('Content-Type', '')
+
+        return 'Result<{}, {}, {}>'.format(self.status, content_type, self.content[:20])
```
myusko, line 353: Since we are supporting CPython3.5+, I'd suggest using the `f-strings` instead of `str.format` method, does it make sense?

vytas7: Probably not yet @myusko, there is a separate issue for dropping that support -- let's take one step at a time.

vytas7: @myusko to make it clear, we do support CPython3.5+, but 3.5 does not support f-strings.
We will drop 3.5 & 3.6 support in https://github.com/falconry/falcon/issues/2049.

myusko: @vytas7 Got it, tbh I've thought that `f-strings` was introduced in 3.5:) sorry, my bad!

RioAtHome:
Hello, sorry to be a bother, but can i get a review on this pull request?

tests/test_utils.py
```diff
@@ -764,6 +764,52 @@ def test_status(self, app):
...
+        )
+
+        assert str(result) == expected_result
+
+    def test_repr_without_content_type_header(self, asgi):
+        value = b'huh'
+        header = [('Not-content-type', '!!!')]
+        result = falcon.testing.Result([value], falcon.HTTP_200, header)
+
+        expected_result = 'Result<200 OK, , {}>'.format(value)
```
vytas7, line 810: I know I'm nitpicking, but OTOH a representation is about the esthetics :smiling_imp:
What if `Content-Type` or content are empty, the `__repr__` would look somewhat awkward -- as illustrated here.

Maybe we could skip the `Content-Type` if it is empty? Those commas are probably not needed either, although I'm open for suggestions there.

My suggestions could be
`Result<200 OK application/json b'{}'>`
`Result<201 Created>`
`Result<503 Service Unavailable b'Database error, dunno\n'>`

One can also ellipsize the text that was truncated, use lenient decoding and present data as string, the possibilities are endless :slightly_smiling_face:

docs/_newsfragments/2044.newandimporved.rst
```diff
@@ -0,0 +1,3 @@
+Added ``__repr__`` method to :class:`falcon.testing.client.Result`.
+Syntax of ``__repr__`` as follows ``Result<{status_code}, {Content-Type Header}, {Content}>``.
```
vytas7, line 2: Suggested edit:
> The form of `__repr__` is as follows: `...`, where the _content_ part will reflect up to 20 bytes of the result's content.

vytas7 (changes requested):
Yes, sorry for taking time @RioAtHome...
This PR is looking good, just a couple of tweaks and it's IMHO ready for merging. :shipit:

tests/test_utils.py
CaselIT: this would be nice to have, but can probably be deferred until some later time

CaselIT (approved):

tests/test_utils.py
RioAtHome: > this would be nice to have, but can probably be deferred until some later time

So sorry, i missed this message before implementing the suggested syntax, my latest commit implement the suggestions above... my apologies.

Concerning my latest commit that changes the syntax, i couldn't think of a way to implement the representation of `Result.content` without an if statement, any feedback will be much appreciated!

RioAtHome:
my latest commit changes:
- `Result.__repr__` will not show any `None` values.
- `Result.__repr__` shows a total of 40 bytes of Result.content where it takes the first 20 and the last 20 and ellipsize the rest with `...`.
- made some few adjustments to `Result.__repr__` tests, mainly removed duplicate code (the get and post methods).

docs/_newsfragments/2043.bugfix.rst
```diff
@@ -1,3 +0,0 @@
-Request attributes :attr:`~falcon.Request.forwarded_scheme` and
-:attr:`~falcon.Request.forwarded_host` now no longer raise an ``IndexError``
-while processing an invalid or empty ``Forwarded`` header.
```
CaselIT, line 3: this seems unrelated

CaselIT (commented):
it seems that an unrelated file was removed. looks ok for me otherwise

CaselIT (dismissed):

vytas7 (dismissed):
This LGTM now :+1: 
(I've taken the liberty of polishing the newsfragment somewhat.)

PR #2047, opened 2022-03-29
fix(tests): spawn servers via sys.executable

mgorny:
# Summary of Changes

Spawn the servers (gunicorn, uvicorn, etc.) during tests through
sys.executable rather than directly, in order to ensure that the test
server is run using the same Python executable as the test suite.
Otherwise, the test servers are started using the Python executable
specified in their shebang and can therefore escape the virtualenv they
are being run in.

This is particularly problem for tests being run from Gentoo ebuilds.
We create a --system-site-packages venv to install falcon in, in order
to test whether it works correctly against the dependencies installed
on the system prior to installing it.  However, the installed packages
such as gunicorn specify /usr/bin/python* shebangs that override
the venv and therefore make these servers unable to find venv-installed
falcon.  Running them via sys.executable ensures that they are started
through Python installed in the venv.

# Related Issues

n/a

# Pull Request Checklist

### DIFF
--- a/docs/_newsfragments/2047.bugfix.rst
+++ b/docs/_newsfragments/2047.bugfix.rst
@@ -1,0 +1,3 @@
+The web servers used for tests are now run through `sys.executable`
+in order to ensure that they respect the virtualenv in which tests
+are being run.
--- a/docs/_newsfragments/2047.bugfix.rst
+++ b/docs/_newsfragments/2047.bugfix.rst
@@ -1,3 +1,2 @@
-The web servers used for tests are now run through `sys.executable`
-in order to ensure that they respect the virtualenv in which tests
-are being run.
+The web servers used for tests are now run through :any:`sys.executable` in
+order to ensure that they respect the virtualenv in which tests are being run.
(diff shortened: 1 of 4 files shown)


CaselIT (dismissed):
Thanks!

mgorny:
> Heh I actually had this problem myself for `uWSGI` because I had that globally on my system, but it is unfortunately a harder task.

Yeah, I couldn't figure out how to make this work for uwsgi.

> I'm thinking to myself though whether we shouldn't apply the same treatment for `uvicorn` for consistency too then?

Hmm, I did uvicorn in `test_wsgi_servers` but I see it's also in `test_asgi_servers`. I'll update the PR in a minute.

mgorny:
Covered asgi tests now too.

vytas7 (approved):
Thanks for this tweak @mgorny :+1:

It is somewhat interesting that one has issues with `gunicorn`, `waitress` and `uwsgi` though since none of them is specified directly as test requirements in the latest ebuild you uploaded, but maybe they get dragged in via indirect test dependencies :thinking: 

In any case, we strive after helping distribution maintainers, and not fighting against them. So we're fine with this change.

mgorny:
Well, I'm running a pretty big test environment here and gunicorn etc. just happen to be installed on the system. The goal's pretty much to try and see what could break on production systems — and I suppose people tend to run webservers there ;-).

PR #2056, opened 2022-04-21
feat(websocket): add support for reason in WebSocket.close()

### LINKED ISSUE #2025: Add support for reason in WebSocket.close()
Support for closing with a `reason` was added in version 2.3 of the [HTTP & WebSocket ASGI Message Format](https://asgi.readthedocs.io/en/latest/specs/www.html#http-websocket-asgi-message-format) (2021-02-02).

Also decide whether we want a way to derive the reason from `HTTPError`s like we do with the close `code`. Alternatives include just using the error's `title`, `description`, and adding a new reason parameter to avoid confusion.

wendy5667:
# Summary of Changes

This PR adds `reason` to Websocket.close() as specified in version 2.3 of the [HTTP & WebSocket ASGI Message Format](https://asgi.readthedocs.io/en/latest/specs/www.html#http-websocket-asgi-message-format). Closure due to [`HTTPError`](https://falcon.readthedocs.io/en/stable/api/errors.html#falcon.HTTPError)s are rendered automatically from HTTP status code, while others are rendered by a default mapping, which could be changed by users.

# Related Issues

Closes #2025

# Pull Request Checklist
 

vytas7 (changes requested):
Thanks for the PR @wendy5667, and sorry we've been slow to review this. :imp:

It looks great at first glance, particularly additions to the test suite. :100:

I'm yet to read this thoroughly... just posting a couple of potential mprovements off the top of my head:
* We need to write a newsfragment for the new feature you've implemented.
* I'm not sure if we can always `send()` a `reason` indiscriminately inside an ASGI WebSocket protocol message. Maybe it's safer to only include a `reason` iff we've verified we're dealing with [ASGI HTTP+WebSocket protocol](https://asgi.readthedocs.io/en/latest/specs/www.html#spec-versions) spec version 2.3+?

CaselIT:
> * I'm not sure if we can always `send()` a `reason` indiscriminately inside an ASGI WebSocket protocol message. Maybe it's safer to only include a `reason` iff we've verified we're dealing with [ASGI HTTP+WebSocket protocol](https://asgi.readthedocs.io/en/latest/specs/www.html#spec-versions) spec version 2.3+?

I agree, this seems the safest option

falcon/asgi/app.py
```diff
@@ -976,7 +976,13 @@ async def _handle_websocket(self, ver, scope, receive, send):
             #   we don't support, so bail out. This also fulfills the ASGI
             #   spec requirement to only process the request after
             #   receiving and verifying the first event.
-            await send({'type': EventType.WS_CLOSE, 'code': WSCloseCode.SERVER_ERROR})
+            await send(
+                {
+                    'type': EventType.WS_CLOSE,
+                    'code': WSCloseCode.SERVER_ERROR,
+                    'reason': 'Internal Server Error',
```
CaselIT, line 983: here we should probably send reason only if the `ver` is 2.3 or greater.

This here poses a problem since doing a simple `ver >= "2.3"` does not work assuming future version, since that check would fail if ver ever goes over `"2.10"`.

we probably need to convert `ver` to a tuple of int, but I don't know if we can assume that it will always be in the form `xx.yy` and not have values like `2.3.1` or something. @vytas7 do you have suggestions here?

falcon/asgi/ws.py
```diff
@@ -257,10 +260,18 @@ async def close(self, code: Optional[int] = None) -> None:
         if self.closed:
             return
 
+        if code in self._close_reasons:
+            reason = self._close_reasons[code]
+        elif 3100 <= code <= 3999:
+            reason = falcon.util.code_to_http_status(code - 3000)
+        else:
+            reason = ''
```
CaselIT, line 268: I think in this case we could not set the reason in the event.
we can probably create the response dict before the if and assign `reason` on it if matched.

as above we would need to account for the version

falcon/asgi/ws.py
```diff
@@ -512,6 +523,9 @@ class WebSocketOptions:
             unhandled error is raised while handling a WebSocket connection
             (default ``1011``). For a list of valid close codes and ranges,
             see also: https://tools.ietf.org/html/rfc6455#section-7.4
+        default_close_reasons (dict): A default mapping between the Websocket
+            close code and the reason why the connection is close. HTTPerrors
+            are not included and will be rendered automatically with HTTP status.
```
CaselIT, line 528:
```suggestion
            are not included and will be rendered automatically as HTTP status.
```

not sure what's the proper pronoun here. as sounded better, but I'm not a native English speaker so I'm probably wrong

falcon/testing/helpers.py
```diff
@@ -427,6 +428,10 @@ def closed(self) -> bool:
     def close_code(self) -> int:
         return self._close_code
 
+    @property
+    def close_reason(self) -> str:
+        return self._close_reason
```
CaselIT, line 433: not sure if we could return str | None and not convert `None` to `''`

falcon/testing/helpers.py
```diff
@@ -727,6 +738,7 @@ async def _collect(self, event: Dict[str, Any]):
                 self._state = _WebSocketState.DENIED
 
                 desired_code = event.get('code', WSCloseCode.NORMAL)
+                reason = event.get('reason', '')
```
CaselIT, line 741: here should be consistent with whatever we decide wrt `None` /` ''`

falcon/testing/helpers.py
```diff
@@ -755,6 +771,7 @@ async def _collect(self, event: Dict[str, Any]):
             if event_type == EventType.WS_CLOSE:
                 self._state = _WebSocketState.CLOSED
                 self._close_code = event.get('code', WSCloseCode.NORMAL)
+                self._close_reason = event.get('reason', '')
```
CaselIT, line 774: here too

falcon/testing/helpers.py
```diff
@@ -780,7 +797,11 @@ def _create_checked_disconnect(self) -> Dict[str, Any]:
             )
 
         self._disconnect_emitted = True
-        return {'type': EventType.WS_DISCONNECT, 'code': self._close_code}
+        return {
+            'type': EventType.WS_DISCONNECT,
+            'code': self._close_code,
+            'reason': self._close_reason,
```
CaselIT, line 803: I guess if we decide to keep `None` reason could be omitted in case of `None`

CaselIT (commented):
great work, I've marked some minor things

falcon/asgi/app.py
wendy5667: Maybe we could split `ver` by `.` and check the numbers one by one?

falcon/testing/helpers.py
wendy5667: I followed the way how `close_code()` was written. Should we assume that `close_reason()` will get called before the connection is closed?

falcon/asgi/app.py
wendy5667: I added a new member function in asgi/ws.py to check if the asgi version support `reason`. Is it more reasonable to put it in util/misc.py so both asgi/ws.py and asgi/app.py can utilize this function?

falcon/asgi/ws.py
wendy5667: I guess my words are kind of confusing. Does the following expression makes better sense?

`Close codes corresponding to HTTPerrors are not included as they will be rendered automatically using HTTP status.`

falcon/asgi/app.py
CaselIT: I think we could keep it in ws.py, but we could make `_check_support_reason` a function, so that app could call it too?

falcon/asgi/ws.py
CaselIT: yes, that's better I think!

falcon/testing/helpers.py
CaselIT: Ok, reading also close_code it may be fine to have `''` as the default

falcon/testing/helpers.py
CaselIT: Here I think we could omit reason if empty string?

CaselIT (commented):
Thanks for the update and sorry for the delay.

looks better, I've replied to a few comments

falcon/asgi/app.py
vytas7: Personally, I can accept `>= '2.3'` as good enough until proven otherwise; the ASGI WebSocket spec is not evolving _that_ fast and I'm not sure if we live to see 2.10 :smiling_imp: Moreover, they might cut straight to `3.x` which would still work with this simple comparison.

CaselIT: since that's not too performance critical (happens once per web socket), I think being correct here is worth it since it's not that much more code

wendy5667: The function `check_support_reason()` has been added in `ws.py`. Hopefully, it could deal with any form of `ver` string, whether it is `2.3` or `2.10.1`.

wendy5667:
> * We need to write a newsfragment for the new feature you've implemented.

Could you kindly share the style guide for newsfragments? I couldn't find it in the Contributor's Guide. Thank you!

CaselIT:
sorry @wendy5667 we kinda forgot this PR. I'll try to bringing it over the finishing line.

falcon/asgi/ws.py
```diff
@@ -512,6 +520,10 @@ class WebSocketOptions:
             unhandled error is raised while handling a WebSocket connection
             (default ``1011``). For a list of valid close codes and ranges,
             see also: https://tools.ietf.org/html/rfc6455#section-7.4
+        default_close_reasons (dict): A default mapping between the Websocket
+            close code and the reason why the connection is close. Close codes
+            corresponding to HTTPerrors are not included as they will be rendered
```
CaselIT, line 525:
```suggestion
            corresponding to HTTPErrors are not included as they will be rendered
```

CaselIT (dismissed):
@vytas7 can you take a look when you have a moment?

CaselIT (dismissed):

falcon/testing/helpers.py
```diff
@@ -734,12 +748,16 @@ async def _collect(self, event: Dict[str, Any]):
...
                     #   raised HTTPError status code.
                     self._close_code = desired_code
+                    self._close_reason = reason
                 else:
                     # NOTE(kgriffs): Force the close code to this since it is
                     #   similar to what happens with a real web server (the HTTP
                     #   connection is closed with a 403 and there is no websocket
                     #   close code).
                     self._close_code = WSCloseCode.FORBIDDEN
+                    self._close_reason = falcon.util.code_to_http_status(
```
vytas7, line 758: As I understand, this would include the preceding integer status code as part of reason.
I don't think we want this, at least it is not in line with the definition of the default reasons.

falcon/asgi/ws.py
```diff
@@ -69,8 +71,10 @@ def __init__(
             Union[media.BinaryBaseHandlerWS, media.TextBaseHandlerWS],
         ],
         max_receive_queue: int,
+        default_close_reasons: Dict[Optional[int], str],
     ):
         self._supports_accept_headers = ver != '2.0'
+        self._supports_reason = check_support_reason(ver)
```
vytas7, line 77: I think we could use more lazy evaluation when it comes to these checks, and actually check only when needed unless it is a simple check as the `ver != 2.0` above.

vytas7 (changes requested):
Thanks for this improvement so far @wendy5667 & @CaselIT!

I propose the following tweaks before merging this PR:
* We should either allow setting an arbitrary reason when closing

or
* Rename `default_close_reasons` to just `close_reasons`, and add more documentation how one sets the close reason.

Also, I think we should inline all standard HTTP code-based close reasons in advance, so that getting a default reason for the given code is as simple as performing a lookup on that dict. It would also be more flexible, allowing to use non-standard reasons or also allowing to remove reasons for 300+4xx errors if desired.

vytas7:
Furthermore, unfortunately it seems that the code got affected by a botched master merge. I can try to fix these things.

vytas7 (approved):
I have now tweaked this a bit more to my liking.

Great work @wendy5667! :100:

vytas7 (approved):

falcon/asgi/ws.py
```diff
@@ -521,12 +539,13 @@ class WebSocketOptions:
...
-        media_handlers (dict): A dict-like object for configuring media handlers
-            according to the WebSocket payload type (TEXT vs. BINARY) of a
-            given message. See also: :ref:`ws_media_handlers`.
+            close code and the reason why the connection is close.
+            Close codes corresponding to HTTP errors are also included in this
+            mapping.
+        media_handlers (dict): A dict-like object for configuring media
+            handlers according to the WebSocket payload type
+            (``TEXT`` vs. ``BINARY``) of a given message.
+            See also: :ref:`ws_media_handlers`.
```
CaselIT, line 548: I don't see a reason to change these descriptions here. I've modified them here https://github.com/falconry/falcon/pull/2290

falcon/asgi/ws.py
```diff
@@ -581,6 +600,13 @@ def __init__(self) -> None:
             1011: 'Internal Server Error',
             3011: 'Internal Server Error',
         }
+        for status_constant in dir(status_codes):
```
CaselIT, line 603: I think it may make sense to move this to a method of the class

falcon/asgi/ws.py
```diff
@@ -728,16 +754,12 @@ async def _pump(self):
                 self._pop_message_waiter = None
 
 
-def check_support_reason(asgi_ver):
+@misc._lru_cache_for_simple_logic(maxsize=16)
+def _supports_reason(asgi_ver: str):
```
CaselIT, line 758: any reason for making this private?

falcon/asgi/ws.py
vytas7: Aha, missed that. I'll roll back this part to minimize merge conflicts.

falcon/asgi/ws.py
vytas7: It was not documented publicly. We have an issue in the 4.0 milestone to privatize as much as possible. Do we want this function in Falcon's public interface with SemVer guarantees? I don't think so.

falcon/asgi/ws.py
vytas7: Hm, just to make the code cleaner? Like `_create_default_close_reasons()`?

falcon/asgi/ws.py
CaselIT: Not sure how to best achieve that. Maybe indicating the public modules/packages is better? 

but we cab decide that at some other time

falcon/asgi/ws.py
CaselIT: that works yes.

falcon/asgi/ws.py
vytas7: Rolled back. But I still need to incrementally add `default_close_reasons` here.

vytas7 (approved):
FWIW, I gave up with coverage on 3.11 and added a `pragma: nocover`.

tests/asgi/test_ws.py
```diff
@@ -1113,6 +1117,137 @@ def test_msgpack_missing():
...
+
+    assert ws.close_code == 4099
+    if reason:
+        assert ws.close_reason == reason
+    else:
+        assert ws.close_reason == ''
+
+
+@pytest.mark.parametrize('reason', ['PEBCAK', 'wow such reason', '', None])
+async def test_close_with_reason_no_cm(conductor, reason):
```
CaselIT, line 1150: this is a bit of a duplication, but it should not cause too much problems

CaselIT (approved):

PR #2069, opened 2022-05-21
chore(CI): make sure Mailman tests run under CPython 3.8

vytas7:


tools/testing/fetch_mailman.sh
```diff
@@ -26,3 +26,6 @@ commands_pre =
     pip uninstall -y falcon
     pip install $FALCON_ROOT
 EOT
+
+# NOTE(vytas): Patch a test that started failing around May, 2022.
+sed -i 's/def test_interact_default_banner/def skip_test_interact_default_banner/' src/mailman/utilities/tests/test_interact.py
```
CaselIT, line 31: this is fine, but how about applying a patch?

CaselIT (dismissed):

vytas7:
I modified one file to skip the only (or so I thought) failing test, but now something else is failing... So this will have to wait for @maxking I guess :imp:

CaselIT:
> I modified one file to skip the only (or so I thought) failing test, but now something else is failing... So this will have to wait for @maxking I guess 👿

alternatively you can patch away tests until there are no more failures :)

vytas7:
I can recursively remove all `test_*.py`, then it should almost definitely pass 😁

tools/testing/fetch_mailman.sh
vytas7: I'll reword to "Skip a test".

vytas7:
Okay, so I was digging into this a bit more in depth, and the problem was more mundane than that. The failures were caused by `tox` picking `python3.10`, so I simply specified `basepython = 3.8`, and voila.

PR #2074, opened 2022-05-27
chore: drop Python 3.5 & 3.6 support

vytas7:
~~This is work in progress. I created a Draft PR in order to be able to verify progress with CI.~~

I had to stop short of some planned improvements in favour of getting the bulk of 3.5 & 3.6 removals out, I'll address these things in follow-up PRs.

### DIFF
--- a/falcon/constants.py
+++ b/falcon/constants.py
@@ -9,8 +9,21 @@ PYPY = sys.implementation.name == 'pypy'
 PYTHON_VERSION = tuple(sys.version_info[:3])
 """Python version information triplet: (major, minor, micro)."""
 
-ASGI_SUPPORTED = True
-"""Evaluates to ``True`` when ASGI is supported for the current Python version."""
+FALCON_SUPPORTED = PYTHON_VERSION >= (3, 7, 0)
+"""Whether this version of Falcon supports the current Python version."""
+
+if not FALCON_SUPPORTED:  # pragma: nocover
+    raise ImportError(
+        'Falcon requires Python 3.7+. '
+        '(Recent Pip should automatically pick a suitable Falcon version.)'
+    )
+
+ASGI_SUPPORTED = FALCON_SUPPORTED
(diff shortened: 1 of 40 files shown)


falcon/constants.py
```diff
@@ -8,18 +8,8 @@
...
-
-# TODO(vytas): Remove these hacks in 4.0 once we have dropped 3.5/3.6 support.
-# NOTE(vytas): Do not alter the microversion if already on 3.5
-#   (read: a hack to hit coverage on 3.5 as well).
-if _mock_py35 or _is_py35:
-    PYTHON_VERSION = (3, 5, PYTHON_VERSION[-1] if _is_py35 else 0)
-
-ASGI_SUPPORTED = PYTHON_VERSION >= (3, 6)
+
+ASGI_SUPPORTED = True
```
CaselIT, line 12: I guess removing this would be a backward incompatible change. I guess it will not cause problems to keep it here

CaselIT (approved):

falcon/constants.py
vytas7: Yeah it wasn't underscored so I thought to play it safe.

vytas7:
I'm still polishing various things... Need to clean up the tests and undo some foul import hacks.

falcon/util/__init__.py
```diff
@@ -76,24 +76,18 @@
...
-
-else:
-    import json  # NOQA
+
+def __getattr__(name):
+    if name == 'json':
+        import warnings
+        import json  # NOQA
+
+        warnings.warn(
```
CaselIT, line 85: Maybe we could just remove this?

CaselIT (approved):

falcon/util/__init__.py
vytas7: Good idea, but we need to add a breaking change newsfragment then telling what was removed. I can do this in a follow-up PR.

CaselIT: Fine by me

PR #2105, opened 2022-08-30
pipelines with py3.11 on 3.1.1 branch

CaselIT:
update the pipelines and various docs on the 3.11 branch to add support for python 3.11


AUTHORS
```diff
@@ -119,6 +119,11 @@ listed below by date of first contribution:
 * signalw
 * Laurent Chriqui (laurent-chriqui)
 * Andrii Oriekhov (andriyor)
+* Tom Boshoven (TBoshoven)
+* Jaakko Lappalainen (jkklapp)
+* Oday Alhayek (RioAtHome)
+* Christian Clauss (cclauss)
+* meetshah133
```
CaselIT, line 126: this may be reverted, since it's here due to a cherry pick

vytas7: `TBoshoven` we might want to leave, if we are to include that `Forwarded` bugfix.

CaselIT: re-reverting!

CaselIT:
I think there is something that prevents the actions from starting since the destination branch is not master but 3.1.1

not sure if there is something in the setting to change this (I don't have access to them)

on my fork the pipeline now passes https://github.com/CaselIT/falcon/actions/runs/2959520819

vytas7 (approved):
Look great :100: 
As to _Actions_ not running on such PR, I don't think it is a big deal worth bothering us with, I'm monitoring CI outcome when cherry-picking into the hotfix branch as I go anyway.

PR #2107, opened 2022-08-30
update action versions

CaselIT:


kgriffs (approved):

CaselIT:
there is actually a dependabot thingy that can do this. Does it make sense activating it?

vytas7 (approved):
:100:

PR #2108, opened 2022-09-01
feat: status_code attribute for Response, HTTPError, HTTPStatus

kgriffs:
This patch adds a status_code attribute (property) to the Response, HTTPError,
and HTTPStatus classes to afford introspection by middleware and error
handlers.

### DIFF
--- a/falcon/http_error.py
+++ b/falcon/http_error.py
@@ -18,7 +18,7 @@ from collections import OrderedDict
 import xml.etree.ElementTree as et
 
 from falcon.constants import MEDIA_JSON
-from falcon.util import uri
+from falcon.util import http_status_to_code, uri
 from falcon.util.deprecation import deprecated_args
 
 
@@ -46,7 +46,10 @@ class HTTPError(Exception):
         error in a future version of falcon.
 
     Args:
-        status (str): HTTP status code and text, such as "400 Bad Request"
+        status (Union[str,int]): HTTP status code or line (e.g.,
+            ``'400 Bad Request'``). This may be set to a member of
+            :class:`http.HTTPStatus`, an HTTP status line string or byte
+            string (e.g., ``'200 OK'``), or an ``int``.
 
     Keyword Args:
         title (str): Human-friendly error title. If not provided, defaults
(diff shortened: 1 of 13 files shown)


falcon/http_error.py
```diff
@@ -46,7 +46,10 @@ class HTTPError(Exception):
         error in a future version of falcon.
 
     Args:
-        status (str): HTTP status code and text, such as "400 Bad Request"
+        status (Union[str,int]): HTTP status code or line (e.g.,
```
CaselIT, line 49: we don't seem to handle the case of int/httpstatus here, like in `self.title = title or status`

falcon/http_status.py
```diff
@@ -25,14 +26,18 @@ class HTTPStatus(Exception):
...
+            ``'400 Bad Request'``). This may be set to a member of
+            :class:`http.HTTPStatus`, an HTTP status line string or byte
+            string (e.g., ``'200 OK'``), or an ``int``.
         headers (dict): Extra headers to add to the response.
         text (str): String representing response content. Falcon will encode
             this value as UTF-8 in the response.
 
     Attributes:
-        status (str): HTTP status line, e.g. '748 Confounded by Ponies'.
+        status (Union[str, int]): The HTTP status line or integer code for
```
CaselIT, line 38:
```suggestion
        status (Union[str,int]): The HTTP status line or integer code for
```

CaselIT (commented):
considering typing, I think it would be nicer if `status` had a narrower type, but there has always been the possibility of setting multiple types so nothing new here.

for the rest it seems ok

falcon/http_error.py
kgriffs: Added tests and a fix for this.

CaselIT (approved):

kgriffs:
> considering typing, I think it would be nicer if `status` had a narrower type, but there has always been the possibility of setting multiple types so nothing new here.
> 
> for the rest it seems ok

Ideally we would have added `status_code` before this and left `status` alone, but I'd like to just keep the consistency for now and avoid breaking changes; we can consider changing things up in the next major release.

vytas7 (approved):
LGTM :+1:

PR #2111, opened 2022-09-28
feat: replace obsolete JS media type

### LINKED ISSUE #2110: Replace the obsolete JS media type
[RFC-4329](https://www.rfc-editor.org/rfc/rfc4329) is obsolete.

According to [RFC-9239](https://www.rfc-editor.org/rfc/rfc9239),

- Changed the intended usage of the media type "text/javascript" from OBSOLETE to COMMON.
- Changed the intended usage for all other script media types to obsolete.

euj1n0ng:
### Summary of Changes

A minor fix according to [RFC 9239 - Updates to ECMAScript Media Types](https://www.rfc-editor.org/rfc/rfc9239)

### Related Issues

Closes #2110


CaselIT (approved):

CaselIT (dismissed):

vytas7 (changes requested):
I think this change warrants a `newandimproved` news fragment.

kgriffs:
@vytas7  I can add the news fragment after this is merged if we want to just move forward with this.

vytas7:
@kgriffs another problem is that this PR, while good-to-go from our side, would probably not address the issue the author had at hand. I wrote a comment on the linked issue. Would you think it would be a good idea to override certain media types with our constants when serving static files to make sure the interpreter's defaults (which may even vary with OS) don't take over?

vytas7:
@kgriffs another problem is that this PR, while good-to-go from our side, would probably not address the issue the author had at hand. I wrote a comment on the linked issue. Would you think it would be a good idea to override certain media types with our constants when serving static files to make sure the interpreter's defaults (which may even vary with OS) don't take over?

vytas7 (dismissed):
Hi again @euj1n0ng, and thanks for this PR! :+1:

I've now added a newsfragment, and updated our machinery somewhat to hopefully address https://github.com/WebOfTrust/keripy/pull/316.

falcon/constants.py
```diff
@@ -141,6 +137,28 @@
...
+_DEFAULT_STATIC_MEDIA_TYPES = tuple(
+    (ext, media_type.split(';', 1)[0])
+    for ext, media_type in (
+        ('.bmp', MEDIA_BMP),
+        ('.gif', MEDIA_GIF),
+        ('.htm', MEDIA_HTML),
+        ('.html', MEDIA_HTML),
+        ('.jpeg', MEDIA_JPEG),
+        ('.jpg', MEDIA_JPEG),
+        ('.js', MEDIA_JS),
```
CaselIT, line 153: do we want to add also mjs/cjs?
not that popular to have them served though

CaselIT (dismissed):

falcon/constants.py
vytas7: Hm, not sure. Maybe we can start small with the most popular ones :grin:
`.cjs` is not even there: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types (`.mjs`, OTOH, is there).

CaselIT: I think cjs is node only, so unknown by the browsers. mjs is instead support also by the browser.

It seems that a limited number of server use the js mime for that type already: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js

I think we can go either way for this, so as you prefer

vytas7: We can add `.mjs`

vytas7 (approved):

CaselIT (approved):

Issue #1010, opened 2017-04-04
Simulate multipart/form-data file upload with Falcon's Testing module


dirkschneemann:
This simple Falcon API will take a `HTTP POST` with `enctype=multipart/form-data` and a file upload in the `file` parameter, returning the file's content to the caller:

```
# simple_api.py
import cgi
import falcon
import json

class MultipartFileUpload(object):
    def on_post(self, req, resp):
        upload = cgi.FieldStorage(fp=req.stream, environ=req.env)
        data = upload['file'].file.read().decode('utf-8')
        resp.body = json.dumps(dict(data=data))

app = falcon.API()
app.add_route('/', MultipartFileUpload())
```

To try this out, create a textfile with some content and send to the API via `curl -F "file=@sample.txt" localhost:8000`.

I would like to test this API now with Falcon's [testing utilities](http://falcon.readthedocs.io/en/stable/api/testing.html), simulating such a file upload. The [simulate_request method](http://falcon.readthedocs.io/en/stable/api/testing.html#falcon.testing.simulate_request) has a `file_wrapper` parameter which might be useful but from the documentation I do not understand how it is supposed to be used or whether this is possible at all.

As a workaround, I currently use [webtest](https://webtest.readthedocs.io) where this is supported, another possibility would be [werkzeug](http://werkzeug.pocoo.org/docs/0.11/test/). Anyway, I'm happy to drop dependencies if this should be possible with Falcon itself.

JamesRamm:
@dirkschneemann 

Here is how I test posting a zip archive to one of our endpoints.
The test function:

```python
def test_posted_model_gets_saved(self, client):
        '''The API can accept a data archive, create and
            return a ``Model`` document pointing to the
        location of the extracted folder
        '''
        path = create_zip()
        with open(path, 'rb') as zfile:
            fake_zip_bytes = zfile.read()
        response = client.simulate_post(
            '/models',
            body=fake_zip_bytes,
            headers={'content-type': 'application/zip'}
        )
        assert response.status == falcon.HTTP_CREATED
        assert 'Location' in response.headers
        assert '_id' in response.json
```

the POST handler:

```python
def on_post(self, req, resp):
         # Save and extract the archive
        directory, name = self._archive_store.save(req.stream, req.content_type)

        # Create the mongodb metadata document
        model = self._document(
            name=name,
            folder=directory
        ).save()

        # build response
        resp.status = falcon.HTTP_CREATED
[...]

PR #2141, opened 2023-02-10
feat(testing): simulate multipart file upload

TigreModerata:
# Summary of Changes

Draft of tests section for post-ing files as multipart/form-data. Parameter 'files' added to simulate_requests in the testing client, as well as a file test_multipart_formdata_request of tests.

# Related Issues

#1010 

# Pull Request Checklist

- [x ] Applied changes to both WSGI and ASGI code paths and interfaces (where applicable).
- [x ] Added **tests** for changed code.
- [ x] Prefixed code comments with GitHub nick and an appropriate prefix.
- [ x] Coding style is consistent with the rest of the framework.
- [ x] Updated **documentation** for changed code.
    - [x ] Added docstrings for any new classes, functions, or modules.
    - [ x] Updated docstrings for any modifications to existing code.
    - [ x] Updated both WSGI and ASGI docs (where applicable).
    - [ x] Added references to new classes, functions, or modules to the relevant RST file under `docs/`.
    - [ x] Updated all relevant supporting documentation files under `docs/`.
    - [ x] A copyright notice is included at the top of any new modules (using your own name or the name of your organization).
    - [ x] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/directives.html?highlight=versionadded#directive-versionadded).
- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/latest/quickstart.html#creating-news-fragments) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)


TigreModerata:
I used urllib3 and I added it to requirements, but it isn't found in the tests - where does it need to be added to be seen?

vytas7:
Hi again @TigreModerata, and thanks for this improvement :+1:
However, when it comes to `urllib3`, Falcon is designed to have no hard dependencies on any third part package, so we should ideally rewrite this functionality not to use `urllib3` at all...
Alternatively, we could possibly catch an `ImportError`, and document that this optional functionality requires `urllib3`.

TigreModerata:
I'll try to write it out, it's probably good exercise :)

vytas7:
Yeah, it should be easy at least when it comes to basic functionality, check out [`tests/test_media_multipart.py`](https://github.com/falconry/falcon/blob/master/tests/test_media_multipart.py) where I mostly construct these forms by hand.

falcon/testing/client.py
```diff
@@ -528,6 +534,15 @@ def simulate_request(
             overrides `body` and sets the Content-Type header to
             ``'application/json'``, overriding any value specified by either
             the `content_type` or `headers` arguments.
+        files(dict): same as the  files parameter in requests,
+             dictionary of ``'name': file-like-objects`` (or ``{'name': file-tuple}``)
+             for multipart encoding upload.
+            ``file-tuple``: can be a 2-tuple ``('filename', fileobj)``,
+                3-tuple ``('filename', fileobj, 'content_type')``
+                or a 4-tuple ``('filename', fileobj, 'content_type', custom_headers)``,
```
vytas7, line 542: Not sure if custom headers are really a must to support in the first iteration :thinking: Falcon doesn't really support any of these since they are forbidden by the latest RFCs.

tests/files/loremipsum.txt
```diff
@@ -0,0 +1,11 @@
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
```
vytas7, line 1: Do we really need to add separate files? Maybe we could just use inline strings, or use existing source or image files?
Otherwise we need to surface these new files in `MANIFEST.in` to get them included in the sdist.

docs/changes/4.0.0.rst
```diff
@@ -15,7 +15,6 @@ Changes to Supported Platforms
 - CPython 3.11 is now fully supported. (`#2072 <https://github.com/falconry/falcon/issues/2072>`__)
 - End-of-life Python 3.5 & 3.6 are no longer supported. (`#2074 <https://github.com/falconry/falcon/pull/2074>`__)
 
-
```
vytas7, line 18: Is there any particular reason behind this newline removal?

vytas7 (changes requested):
Looks great so far :100: 

As discussed, we just need to get rid of `urllib3`, or if the functionality in question is non-trivial to implement, we need to make it optional, like msgpack, jsonschema, etc.
(But of course I'd prefer if this was implemented directly in the framework.)

tests/files/loremipsum.txt
TigreModerata: I didn't really know... I wanted to make sure all goes well mainly with bigger stuff (the image), but now that I'm sure I'll remove them.

docs/changes/4.0.0.rst
TigreModerata: I don't know what happened... was that me?

TigreModerata:
> Yeah, it should be easy at least when it comes to basic functionality, check out [`tests/test_media_multipart.py`](https://github.com/falconry/falcon/blob/master/tests/test_media_multipart.py) where I mostly construct these forms by hand.

Now I  just string a body together with the info coming from files, but it's probably pretty ugly (but it works, at least).

falcon/testing/client.py
TigreModerata: I used Content-Disposition for the nested files, just because you had it in test_media_multipart, but I'll remove them.

falcon/testing/client.py
```diff
@@ -2133,8 +2158,161 @@ async def __aexit__(self, exc_type, exc, tb):
...
+    boundary = _make_boundary()
+    body_string = b'--' + boundary.encode() + b'\r\n'
+    header = {'Content-Type': 'multipart/form-data; boundary=' + boundary}
+    # Handle whatever json data gets passed along with files
+    if data:
+        data_fields = _prepare_data_fields(data)
+        for f, val in data_fields:
+            body_string += (
+                f'Content-Disposition: '
+                f'{content_disposition or "fom-data"}; name={f}; \r\n\r\n'.encode()
```
vytas7, line 2270: Probably a misspelling here _fom-data_ (sic) :arrow_right: _form-data_?

TigreModerata: fixed

falcon/testing/client.py
```diff
@@ -2133,8 +2158,166 @@ async def __aexit__(self, exc_type, exc, tb):
...
+            file_name = k
+        file_data = v
+    if hasattr(file_data, 'read'):
+        file_data = file_data.read()
+    return file_name, file_data, file_content_type
+
+
+def _make_boundary():
+    """
+    Create random boundary to be used in multipar/form-data with files.
```
vytas7, line 2245: `multipar/form-data` (sic) :arrow_right: `multipart/form-data`

falcon/testing/client.py
```diff
@@ -736,6 +751,15 @@ async def _simulate_request_asgi(
             overrides `body` and sets the Content-Type header to
             ``'application/json'``, overriding any value specified by either
             the `content_type` or `headers` arguments.
+        files(dict): same as the  files parameter in requests,
```
vytas7, line 754: I find interesting but probably somewhat unintuitive that the `json` parameter is used to pass non-file parts, this isn't really obvious from the proposed documentation either. I know it is partially my fault when I think I suggested to follow `requests` somewhere in the issue, there apparently you need to pass non-file parts as `data`. I think `httpx` also follows this interface.

So I think here we have two viable options. Either try to follow the same interface, and implement `data` too (that would result in an URL-encoded form in the absence of `files`), or break with the tradition, and try to merge both into one parameter, e.g., `form`. (Or just implement `files` to get something started, and create a follow-up issue for non-file fields and URL-encoded forms? :thinking:)

Thoughts @TigreModerata @CaselIT @kgriffs?

vytas7 (changes requested):

falcon/testing/client.py
TigreModerata: ahiiii will I ever learn :|

falcon/testing/client.py
```diff
@@ -2133,8 +2158,166 @@ async def __aexit__(self, exc_type, exc, tb):
...
+    if hasattr(file_data, 'read'):
+        file_data = file_data.read()
+    return file_name, file_data, file_content_type
+
+
+def _make_boundary():
+    """
+    Create random boundary to be used in multipar/form-data with files.
+    """
+    boundary = binascii.hexlify(os.urandom(16))
```
vytas7, line 2247: Since `binascii` is a part of the stdlib, there's no problem with using it, but, OTOH, maybe we could simply use `os.urandom(16).hex()`?

No need to decode anything then either, e.g.,
```python
>>> boundary = os.urandom(16).hex()
>>> boundary
'8bd51c7e4520992fabf833c7be73a72b'
```

falcon/testing/client.py
vytas7: @CaselIT do you have any input here?

I'm leaning towards remodeling our interface to follow the popular `requests` (and `httpx`) since it is what probably the absolute majority of developers are well familiar with.
Which implies:
* Add support for `data` as in `requests` -- if provided as a `dict` or an iterable of items, then if `files` are provided, we use it as in `requests` (or in lieu of `json` in your current code).
* As in `requests`, if `files` are absent, then use `data` items to URL-encode the `POST` payload.
* As in `requests`, if `data` is a string or a bytestring, then in the same way as `body` currently is.
* Providing string `data` and `files` together either results in a `ValueError`, or we clearly describe the precedence (i.e. what overrides what, as with `json` vs `body`).
* We keep `body` as a deprecated alias for `data`.

What do you think yourself @TigreModerata? I can help, but all this would still increase the amount of work for you here :sweat_smile:

vytas7 (changes requested):

falcon/testing/client.py
```diff
@@ -2133,8 +2158,166 @@ async def __aexit__(self, exc_type, exc, tb):
         await self._task_req
 
 
+def _prepare_data_fields(data):
```
CaselIT, line 2161: ok this is the same as the first part of _encode_files at https://github.com/psf/requests/blob/15585909c3dd3014e4083961c8a404709450151c/requests/models.py#L155-L171

falcon/testing/client.py
```diff
@@ -2133,8 +2158,166 @@ async def __aexit__(self, exc_type, exc, tb):
...
+            if v:
+                # v has to come from json serializable obj, so can't be bytes
+                v = str(v)
+                new_fields.append(
+                    (
+                        field.decode('utf-8') if isinstance(field, bytes) else field,
+                        v.encode('utf-8') if isinstance(v, str) else v,
+                    )
+                )
+            else:
```
CaselIT, line 2193: how about we do like in request, and test above `if v is not None` and ignore the none case?

falcon/testing/client.py
```diff
@@ -2133,8 +2158,166 @@ async def __aexit__(self, exc_type, exc, tb):
...
+        if len(v) == 2:
+            file_name, file_data = v
+        else:
+            file_name, file_data, file_content_type = v
+        if (
+            len(v) == 3
+            and file_content_type
+            and file_content_type.startswith('multipart/mixed')
+        ):
+            file_data, new_header = _encode_files(json_module.loads(file_data.decode()))
```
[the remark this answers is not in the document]
vytas7: I don't think we should support nested multipart forms at all, at least not in this issue. There is a FAQ item on how to parse them that probably already gives too much attention to the topic, it is a super legacy edge case. One can construct the payload in other ways; if they have such legacy code, probably there is already some existing test to create such a payload.

TigreModerata:
Hi, I'm back, thanks for all the feedback! Getting back to work on this today, sorry for the long delay (holiday + new job).

CaselIT:
> Hi, I'm back, thanks for all the feedback! Getting back to work on this today, sorry for the long delay (holiday + new job).

Thanks for the update, I'll take a look this week.
(Good luck on your new job!)

TigreModerata:
I added the data arg to simulate_post, in such a way that if both json and (files or data) are present, a badrequest exception is thrown. If json is absent, then files and data will be processed: if files is also absent, data will be url-encoded. 
I'm not sure I got all the requirements right, in particular the "body as an alias of data" (I didn't touch it because I don't know exactly how and I'm not sure if you wanted to leave that for a future version). 
I also didn't remove the possibility for nested forms, just because the possibility exists in other multipart test files and I wasn't sure if you guys agreed to remove them... but it's a quick fix, if you confirm.

[I am really a beginner, so some of what you said may have gone over my head... I apologize. Since you guys are so responsive, this is a great help to me to learn/practice, I really appreciate your patience!]

falcon/testing/client.py
```diff
@@ -2133,8 +2192,163 @@ async def __aexit__(self, exc_type, exc, tb):
...
+    Args:
+        data: dict or list of tuples with json data from the request
+
+    Returns: list of 2-tuples (field-name(str), value(bytes))
+
+    """
+    urlresult = []
+    body_part = b''
+    if isinstance(data, (str, bytes)) or hasattr(data, 'read'):
+        fields = list(json_module.loads(data).items())
```
CaselIT, line 2207: so we are assuming that if provided as string it's a json.

this would mean that we cannot replace deprecate body with data.

let's see @vytas7 opinion

CaselIT (commented):
thanks for the change. let's wait for @vytas7 review

falcon/testing/client.py
TigreModerata: ok, I hadn't really understood what you guys meant by replacing the body... but ok, that should be an easy fix. I'll try to implement it, then you can use whatever works best.

falcon/testing/client.py
vytas7: Absolutely agree regarding the precedence, and in general, I meant to say we could follow the `requests`' suite where wise and applicable, not blindly. :slightly_smiling_face:

docs/changes/4.0.0.rst
vytas7: I think it must have been you, yes. It is coming from you changeset, but it might be some artefact of merging things back and forth on your side, not necessarily something you actively did.

falcon/testing/client.py
vytas7: Taking a look at this again, maybe we should after all diverge from `requests` here? Splitting `files` and `data` looks really clunky...

How about just adding `form` instead?
* If `form` is only a simple key-value dict, serialize using urlencoded
* If `form` has file tuples, use a multipart form
* It would be possible to specify content-type in conjunction w/ `form` where applicable, i.e., it would be possible to supply a custom `boundary` this way, or force multipart on a dict

CaselIT: not having the ability of passing bytes in `data` will break people code though.

if you meant using `form` in addition to `data` we would need to have a specific rule for how they are handled.
same for `json`

vytas7: I was thinking we wouldn't touch anything existing at all, just add a `form` dict for both types of fields. And yes, document the precedence order, same as for `json`. Since `json` existed before, we can maybe make `form` even lower prio.

CaselIT: so no explicit error if two of them are provided and one would be ignored?

vytas7: We can raise an error, I'm not opinionated about this. OTOH, currently `json` overrides silently, so maybe it makes sense to continue the same pattern?

vytas7:
(I'm in the process of terraforming this somewhat... )

TigreModerata:
Hi, did something happen, can I do anything to help? I just got a notification saying "mentioned" but I'm not sure why. I've not looked at this in ages, but I have time and I'm happy to try to contribute.

vytas7:
Hi @TigreModerata!
I just merged the `master` branch into your branch to start with.

I was thinking to update your PR, and to change `files` and `data` to `form`, similar to how `json` is handled. I can make this change, and you can continue from that point!

TigreModerata:
> I was thinking to update your PR, and to change `files` and `data` to `form`, similar to how `json` is handled. I can make this change, and you can continue from that point!

I'm not totally sure what needs to be done, I need to refresh my memory a bit. But yeah, I'm happy if I can contribute!

vytas7:
> I'm not totally sure what needs to be done, I need to refresh my memory a bit.
That's why I thought it would be a good opportunity to restart this from scratch, it doesn't need to be too complex in the first iteration. I was thinking we could simply add a `form` parameter that can be a dict or a list/tuple of key-value pairs. If the whole form is only simple key-values, we do a URL-encoded form, otherwise encode it as multipart, roughly how browsers do; what do you think?

I have written a simple proof-of-concept implementation, but it needs more validation, tests, documentation, and a newsfragment.

> But yeah, I'm happy if I can contribute!
You are very welcome to!

vytas7 (dismissed):

falcon/testing/client.py
```diff
@@ -2136,8 +2146,70 @@ async def __aexit__(self, exc_type, exc, tb):
...
+            data = data.read()
+
+        headers = f'Content-Disposition: form-data; name="{name}"'
+        if filename:
+            headers += f'; filename="{filename}"'
+        headers += f'\r\nContent-Type: {content_type}\r\n\r\n'
+
+        body.append(headers.encode() + data + b'\r\n')
+
+    checksum = hashlib.sha256()
```
CaselIT, line 2192: clever :)

CaselIT (commented):
overall seems ok. I've not checked if the api is consistent with what was suggested before or what request etc do

PR #2169, opened 2023-08-12
chore: remove pinning of `importlib-resources` for Mailman tests

vytas7:
We had to pin `importlib-resources` for some CI gates because its very aggressive deprecation policy.
Now the dependency in question has been dropped upstream, so we don't need to pin anymore (in fact, pinning breaks the gate):  https://gitlab.com/mailman/mailman/-/merge_requests/1130.

### DIFF
--- a/.github/workflows/tests-mailman.yaml
+++ b/.github/workflows/tests-mailman.yaml
@@ -6,6 +6,9 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    branches:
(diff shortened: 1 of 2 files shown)


vytas7:
Apparently it is still broken, maybe the Docker image needs to be updated :thinking: I'll need to check with @maxking.

vytas7:
Closing, it is not as simple as that unfortunately.

PR #2170, opened 2023-08-16
feat(typing): type util package

copalco:
# Summary of Changes

# Related Issues

# Pull Request Checklist

### DIFF
--- a/falcon/util/deprecation.py
+++ b/falcon/util/deprecation.py
@@ -18,6 +18,9 @@ This module provides decorators to mark functions and classes as deprecated.
 """
 
 import functools
+from typing import Any
+from typing import Callable
+from typing import Optional
 import warnings
 
 
@@ -41,7 +44,9 @@ class DeprecatedWarning(UserWarning):
     pass
 
 
-def deprecated(instructions, is_property=False, method_name=None):
+def deprecated(
+    instructions: str, is_property: bool = False, method_name: Optional[str] = None
+) -> Callable[[Callable[..., Any]], Any]:
     """Flag a method as deprecated.
 
     This function returns a decorator which can be used to mark deprecated
@@ -60,7 +65,7 @@ def deprecated(instructions, is_property=False, method_name=None):
 
     """
 
-    def decorator(func):
+    def decorator(func: Callable[..., Any]) -> Callable[[Callable[..., Any]], Any]:
 
         object_name = 'property' if is_property else 'function'
         post_name = '' if is_property else '(...)'
@@ -69,7 +74,7 @@ def deprecated(instructions, is_property=False, method_name=None):
         )
 
         @functools.wraps(func)
-        def wrapper(*args, **kwargs):
+        def wrapper(*args: Any, **kwargs: Any) -> Callable[..., Any]:
             warnings.warn(message, category=DeprecatedWarning, stacklevel=2)
 
             return func(*args, **kwargs)
@@ -79,7 +84,9 @@ def deprecated(instructions, is_property=False, method_name=None):
     return decorator
 
 
-def deprecated_args(*, allowed_positional, is_method=True):
+def deprecated_args(
+    *, allowed_positional: int, is_method: bool = True
+) -> Callable[..., Callable[..., Any]]:
     """Flag a method call with positional args as deprecated.
 
     Keyword Args:
@@ -98,9 +105,9 @@ def deprecated_args(*, allowed_positional, is_method=True):
     if is_method:
         allowed_positional += 1
 
-    def deprecated_args(fn):
+    def deprecated_args(fn: Callable[..., Any]) -> Callable[..., Callable[..., Any]]:
         @functools.wraps(fn)
-        def wraps(*args, **kwargs):
+        def wraps(*args: Any, **kwargs: Any) -> Callable[..., Any]:
             if len(args) > allowed_positional:
                 warnings.warn(
                     warn_text.format(fn=fn.__qualname__),
--- a/falcon/util/misc.py
+++ b/falcon/util/misc.py
@@ -22,12 +22,17 @@ framework itself. These functions are hoisted into the front-door
 
     now = falcon.http_now()
 """
-
 import datetime
 import functools
 import http
 import inspect
 import re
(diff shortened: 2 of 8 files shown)


vytas7 (dismissed):
Looks great, thanks a ton!

falcon/util/deprecation.py
```diff
@@ -60,7 +65,7 @@ def deprecated(instructions, is_property=False, method_name=None):
 
     """
 
-    def decorator(func):
+    def decorator(func: Callable[..., Any]) -> Callable[[Callable[..., Any]], Any]:
```
CaselIT, line 68: we could use https://docs.python.org/3/library/typing.html#typing.ParamSpec but for now it's probably fine, since it would required adding typing extensions as a dependency

The same feature could be used also in some other places, like in async_to_sync and in the other similar method.

@vytas7 do you have preferences here?

falcon/util/misc.py
```diff
@@ -95,7 +100,7 @@ def decorator(func):
     _lru_cache_for_simple_logic = functools.lru_cache  # type: ignore
 
 
-def is_python_func(func):
+def is_python_func(func: Union[Callable, Any, None]) -> bool:
```
CaselIT, line 103: isn't any enough?

falcon/util/reader.py
```diff
@@ -354,22 +372,22 @@ def pipe_until(
                 raise DelimiterError('expected delimiter missing')
             self._buffer_pos += delimiter_len
 
-    def exhaust(self):
+    def exhaust(self) -> None:
         self.pipe()
 
-    def delimit(self, delimiter):
+    def delimit(self, delimiter: bytes) -> 'BufferedReader':
```
CaselIT, line 378: add future annotation so we can avoid the string?

falcon/util/structures.py
```diff
@@ -61,34 +72,36 @@ class CaseInsensitiveDict(MutableMapping):  # pragma: no cover
 
     """
 
-    def __init__(self, data=None, **kwargs):
-        self._store = dict()
+    def __init__(
+        self, data: Optional[Union[Iterable[Tuple[str, Any]]]] = None, **kwargs: Any
```
CaselIT, line 76: the union seems unnecessary

falcon/util/sync.py
```diff
@@ -17,14 +19,17 @@
     'wrap_sync_to_async_unsafe',
 ]
 
+from typing import Optional
+from typing import TypeVar
+from typing import Union
```
CaselIT, line 24: move these above?

falcon/util/sync.py
```diff
@@ -212,7 +228,7 @@ def async_to_sync(coroutine, *args, **kwargs):
     return loop.run_until_complete(coroutine(*args, **kwargs))
 
 
-def runs_sync(coroutine):
+def runs_sync(coroutine: Callable[..., Awaitable[Any]]) -> Callable[..., Any]:
```
CaselIT, line 231: isn't this awaitable[Result] and Callable[..., Result] ?

falcon/util/uri.py
```diff
@@ -143,7 +149,7 @@ def encoder(uri):
 """
 
 encode_value = _create_str_encoder(True)
-encode_value.name = 'encode_value'
+encode_value.__name__ = 'encode_value'
```
CaselIT, line 152: oh this was a bug. nice catch

CaselIT (changes requested):
Great work, I've left a few comments.

Also sometimes to denote a callable `Callable[..., Any]` is used while in other places `Callable` is used. I'm not sure if they are equivalent, but it may make sense to unify them

CaselIT:
Also another question.
Does it make sense to add future-annotation to all modules that have typing?
Personally I think it makes sense, and it should also limit the performance impact of adding typing on the import of a module.
@vytas7 do you have a preference here?

falcon/util/deprecation.py
vytas7: I'm probably a wrong person to ask, but I take all this Python typing stuff with a grain of salt, it needs to stabilize anyway when it comes to more advanced scenarios. The basics work nicely though now. So IMHO let's keep it stupid simple: a plain `Callable` is fine in the first iteration.

falcon/util/misc.py
vytas7: Probably `Optional[Callable]` is enough. When can this `func` be `None`?

falcon/util/reader.py
vytas7: What do you mean by "future annotation" @CaselIT?

Smth along the below lines?
```python
from __future__ import annotations

from copy import deepcopy


class Widget:
    ...

    def copy(self) -> Widget:
        return deepcopy(self)
```

vytas7:
What do you mean by performance impact @CaselIT? Annotations shouldn't affect code perf once they are parsed... or do you mean import times?

falcon/util/reader.py
CaselIT: Yes

CaselIT:
Yes I meant the import time of a module

falcon/util/deprecation.py
copalco: I'm already using typing_extension in the modules i'm typing currently, so maybe ParamSpec is a way to go. I used it in production environment without any problems.

falcon/util/misc.py
copalco: I removed the None type, Any is needed when method is a function.

falcon/util/reader.py
copalco: Done.

falcon/util/structures.py
copalco: Removed.

falcon/util/sync.py
copalco: Done.

falcon/util/sync.py
copalco: Changed

vytas7 (approved):

CaselIT (approved):
thanks. looks ok for me

PR #2171, opened 2023-08-22
feat: type more falcon modules

copalco:
# Summary of Changes

# Related Issues

# Pull Request Checklist


falcon/hooks.py
```diff
@@ -18,8 +18,19 @@
...
-
+from typing import Any
+from typing import Awaitable
+from typing import Callable
+from typing import Dict
+from typing import List
+from typing import Tuple
+from typing import Union
+
+from falcon.asgi.request import Request as AsynchronousRequest
```
CaselIT, line 29: underscore these `import as` since otherwise they mean that these import are public exports from this module

falcon/hooks.py
CaselIT, line 1: missinig from future annotation

falcon/hooks.py
```diff
@@ -28,8 +39,14 @@
...
 )
 
+SynchronousResource = Callable[..., Any]
+AsynchronousResource = Callable[..., Awaitable[Any]]
+Resource = Union[SynchronousResource, AsynchronousResource]
+
 
-def before(action, *args, is_async=False, **kwargs):
+def before(
+    action: Resource, *args: Any, is_async: bool = False, **kwargs: Any
```
CaselIT, line 48: uh ok, I guess a class is accepted by `Callable[..., Any]`.

should probably work as is

falcon/hooks.py
```diff
@@ -266,7 +327,9 @@ def do_before(self, req, resp, *args, **kwargs):
     return do_before
 
 
-def _merge_responder_args(args, kwargs, argnames):
+def _merge_responder_args(
+    args: Tuple[Any, ...], kwargs: Dict[str, Any], argnames: List[Any]
```
CaselIT, line 331:
```suggestion
    args: Tuple[Any, ...], kwargs: Dict[str, Any], argnames: List[str]
```

falcon/http_error.py
```diff
@@ -147,7 +155,9 @@ def __repr__(self):
     def status_code(self) -> int:
         return http_status_to_code(self.status)
 
-    def to_dict(self, obj_type=dict):
+    def to_dict(
+        self, obj_type: Type[Dict[str, Union[str, int, None, Link]]] = dict
```
CaselIT, line 159: probably better to use `MutableMapping` if we want to be 100% correct

falcon/http_error.py
```diff
@@ -194,9 +204,10 @@ def to_json(self, handler=None):
         obj = self.to_dict(OrderedDict)
         if handler is None:
             handler = _DEFAULT_JSON_HANDLER
```
CaselIT, line 206: we can probably type `_DEFAULT_JSON_HANDLER` like this so there is no need of the assert
```
_DEFAULT_JSON_HANDLER: Serializer
```

falcon/inspect.py
```diff
@@ -753,7 +777,17 @@ def visit_app(self, app: AppInfo) -> str:
 # ------------------------------------------------------------------------
 
 
-def _get_source_info(obj, default='[unknown file]'):
+def _get_source_info(
+    obj: ModuleType
```
CaselIT, line 781: union since iirc 3.7 and lower don't like this

CaselIT: also these are internal function, `Any` there is most likely enough

same in the other functions that are similarly typed

falcon/inspect.py
```diff
@@ -753,7 +777,17 @@ def visit_app(self, app: AppInfo) -> str:
...
+    | Type[Any]
+    | MethodType
+    | FunctionType
+    | TracebackType
+    | FrameType
+    | CodeType
+    | Callable[..., Any],
+    default: Optional[str] = '[unknown file]',
+) -> Optional[str]:
```
CaselIT, line 790: default and the return type are actually `str`, not optional

this would avoid the need to add the assert above

falcon/inspect.py
```diff
@@ -780,18 +823,35 @@ def _get_source_info_and_name(obj):
     name = getattr(obj, '__name__', None)
     if name is None:
         name = getattr(type(obj), '__name__', '[unknown]')
+    assert name
```
CaselIT, line 826: strange that this assert is needed.

falcon/inspect.py
```diff
@@ -767,11 +801,20 @@ def _get_source_info(obj, default='[unknown file]'):
...
+def _get_source_info_and_name(
+    obj: ModuleType
+    | Type[Any]
+    | MethodType
+    | FunctionType
+    | TracebackType
+    | FrameType
+    | CodeType
+    | Callable[..., Any]
+) -> Tuple[Optional[str], str]:
```
CaselIT, line 817:
```suggestion
) -> Tuple[str, str]:
```

falcon/vendor/mimeparse/mimeparse.py
```diff
@@ -56,7 +63,7 @@ def parse_media_range(range):
     :rtype: (str,str,dict)
     """
     (type, subtype, params) = parse_mime_type(range)
-    params.setdefault('q', params.pop('Q', None))  # q is case insensitive
+    params.setdefault('q', params.pop('Q', ''))  # q is case insensitive
```
CaselIT, line 66: not sure if this is equivalent to None

falcon/vendor/mimeparse/mimeparse.py
```diff
@@ -77,8 +86,8 @@ def quality_and_fitness_parsed(mime_type, parsed_ranges):
 
     :rtype: (float,int)
     """
-    best_fitness = -1
-    best_fit_q = 0
+    best_fitness = -1.0
+    best_fit_q = 0.0
```
CaselIT, line 90: i've recently become aware that float is a supertype of int for type checkers, so we can do
```py
    best_fitness: float = -1
    best_fit_q: float = 0

File: falcon/middleware.py [hunk 12]
```diff
@@ -75,7 +82,9 @@ def __init__(
                 )
         self.allow_credentials = allow_credentials
 
-    def process_response(self, req: Request, resp: Response, resource, req_succeeded):
+    def process_response(
+        self, req: Request, resp: Response, resource: Resource, req_succeeded: bool
```
CaselIT, line 86: resource is an object, I think it's ok to just type it as object

falcon/media_serializer.py
```diff
@@ -0,0 +1,11 @@
+from typing import Dict
+from typing import Union
+
+from falcon.link import Link
+
+
+class Serializer:
```
CaselIT, line 7: this is to avoid using protocol correct?

I think we should leave a note here.


also move to typing?

falcon/errors.py
```diff
@@ -551,7 +583,12 @@ class HTTPMethodNotAllowed(HTTPError):
 
     @deprecated_args(allowed_positional=1)
     def __init__(
-        self, allowed_methods, title=None, description=None, headers=None, **kwargs
+        self,
+        allowed_methods: List[str],
```
CaselIT, line 587: `iterable[str]` is probably ok, same as challenges above

falcon/errors.py
```diff
@@ -2442,21 +2647,21 @@ class MediaMalformedError(HTTPBadRequest):
...
-    def __init__(self, media_type, **kwargs):
+    def __init__(self, media_type: str, **kwargs: Union[RawHeaders, Kwargs]):
         super().__init__(
             title='Invalid {0}'.format(media_type), description=None, **kwargs
         )
         self._media_type = media_type
 
     @property
-    def description(self):
+    def description(self) -> Optional[str]:
```
CaselIT, line 2657:
```suggestion
    def description(self) -> str:
```

CaselIT: oh ok, this overloads the arg. well if it errors returning just str, keep as is

CaselIT (changes requested):
Great work.

- "from future annotation" is missing from a few modules
- in general I'm not a fan of the type alias union that include None, like RawHeaders for example, since it's a bit strange to read `headers: RawHeaders = None`. That said it's my preference, as they are it's also ok

falcon/link.py
```diff
@@ -0,0 +1,3 @@
+from typing import Dict
+
+Link = Dict[str, str]
```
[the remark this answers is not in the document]
copalco: There is a problem with falcon.typing module that it's very easy to make a circular import error, that's why I created separate typing packages.

falcon/http_status.py
```diff
@@ -46,7 +49,9 @@ class HTTPStatus(Exception):
 
     __slots__ = ('status', 'headers', 'text')
 
-    def __init__(self, status, headers=None, text=None):
+    def __init__(
+        self, status: Status, headers: RawHeaders = None, text: Optional[str] = None
```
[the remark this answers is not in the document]
copalco: Done.

falcon/typing_http_data.py
```diff
@@ -0,0 +1,9 @@
+import http
+from typing import Dict
+from typing import List
+from typing import Tuple
+from typing import Union
+
+NormalizedHeaders = Dict[str, str]
+RawHeaders = Union[NormalizedHeaders, List[Tuple[str, str]], None]
+Status = Union[http.HTTPStatus, str, int]
```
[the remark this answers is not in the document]
copalco: As above circular import errors.

falcon/vendor/mimeparse/mimeparse.py
copalco: Changed.

falcon/media_serializer.py
copalco: Yes this is to avoid Protocol, I cannot move to typing because circular import :(

.coveragerc
```diff
@@ -13,3 +13,4 @@ exclude_lines =
     pragma: nocover
     pragma: no cover
     pragma: no py39,py310 cover
+    raise NotImplementedError()
```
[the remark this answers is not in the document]
copalco: Removed and tested the code

falcon/errors.py
copalco: Changed.

falcon/typing.py
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 """Shorthand definitions for more complex types."""
-
-from typing import Any, Callable, Pattern, Union
+from typing import Any, Callable, Pattern
+from typing import Union
```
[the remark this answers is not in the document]
copalco: Changed.

falcon/hooks.py
copalco: Done.

falcon/hooks.py
copalco: Done.

falcon/http_error.py
copalco: Done.

falcon/http_error.py
copalco: Done. Have a look, because I don't know if this is what you meant.

falcon/inspect.py
copalco: Any instead of all those types?

falcon/inspect.py
copalco: Yes, but there is a case in the tests where default is None, so it is possible that this function will return it.

falcon/inspect.py
copalco: Used Union instead of `|`

falcon/inspect.py
copalco: I cannot since `source_info = _get_source_info(obj, None)` can result in None being returned.

falcon/inspect.py
copalco: I've replaced it with `cast`

falcon/http_status.py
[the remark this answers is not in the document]
CaselIT: I noticed later that raw header includes none so it was actually fine without optional

falcon/link.py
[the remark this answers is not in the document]
CaselIT: that's unfortunate. I'll try taking a look, since I really don't like having numerous typing packages that are 3-4 line each.

falcon/hooks.py
CaselIT: yes, or use `from falcon import asgi` and use `asgi.Request` to type stuff

falcon/inspect.py
CaselIT: > Any instead of all those types?

yes, we don't really gain anything from typing it as the union of a lot of things. We just pass to the inspect library. Also these are internal methods (as they are underscored), so they aren't used by the users

falcon/inspect.py
CaselIT: interesting. I'll take a look, since I think it's not correct (from a quick look everywhere it's used it assumes that it returns a str not None

falcon/vendor/mimeparse/mimeparse.py
CaselIT: cc @vytas7

vytas7: We should IMHO undo annotations to this module altogether, let alone changes that might affect functionality. It is vendored, so we should keep it verbatim.
(I'm planning to reimplement the subset we are actually using and stop vendoring, but until then we should keep it _as is_.)

falcon/http_status.py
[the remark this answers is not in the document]
CaselIT: oh ok, you removed the none, thanks!

CaselIT (commented):
thanks for the update, I've repried to the open messages.

I'll try taking a better look locally to se if we can avoid some of the circular imports

falcon/middleware.py
copalco: Changed.

falcon/vendor/mimeparse/mimeparse.py
copalco: In this case it is because two lines below it's set to '1' either if it is None or ''.

copalco: Removed typing from vendor and excluded it from typing.

falcon/link.py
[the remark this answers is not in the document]
copalco: Ok I'm moved typing to single module, and avoided circular imports have a look.

CaselIT:
Thanks @copalco for the update, I'll try taking a look by the weekend!

falcon/inspect.py
copalco: Moved to typing

falcon/inspect.py
copalco: I've discovered that when we do `from __future__ import annotations` we can use pipes

falcon/typing_http_data.py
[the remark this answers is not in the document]
copalco: Moved

falcon/inspect.py
copalco: Changed to Any.

CaselIT (commented):
Thanks for the updates. In general this looks mostly ok. 

While it works on definition when using future annotation, I'm not sure if it has unintended consequences at runtime for older python

CaselIT:
I've pushed an update to avoid the imports in the functions, using `if TYPE_CHECKING`

falcon/response.py
```diff
@@ -702,7 +706,7 @@ def delete_header(self, name):
         name = name.lower()
 
         if name == 'set-cookie':
-            raise HeaderNotSupported('This method cannot be used to remove cookies')
+            raise HeaderNotSupported('This method cannot be used to remove ookies')
```
vytas7, line 709: This was an accidental keystroke I believe.

falcon/media/handlers.py
```diff
@@ -34,7 +35,7 @@ def _raise(self, *args, **kwargs):
     serialize = deserialize = _raise
 
 
-class Handlers(UserDict):
+class Handlers(MediaHandlers, UserDict):
```
vytas7, line 38: Maybe it's just me, but I don't see much benefit in adding this diamond inheritance here just to be able to type a private (underscored) method. Maybe we could just mark `self._resolve` as `Callable` or smth like that?

falcon/inspect.py
```diff
@@ -226,13 +239,13 @@ def _traverse(roots, parent):
 
                         source_info = _get_source_info(real_func)
                         internal = _is_internal(real_func)
-
+                        assert source_info
```
vytas7, line 242: Are these `assert`s essential to the inspect module's functionality, or debugging remnants?

vytas7 (changes requested):
Looks great, it's a massive chunk of work (bar a couple of nitpicks)!

I'm just not 100% convinced by these new base classes inside `typing.py` :thinking: Do we really need these? It would be easier to get this review merged if it did not introduce any changes to the codebase's behaviour/performance (which in turn needs more extensive testing), just annotations.
Or maybe we could split these into a separate PR?

falcon/media/handlers.py
CaselIT: IIRC they are stand in for protocols that aren't available on of the python version we target.
But maybe we can just use them in an type checking block, without needing to use the typing extension package

falcon/hooks.py
```diff
@@ -13,23 +13,41 @@
...
+    import falcon as wsgi
+    from falcon import asgi
 
 _DECORABLE_METHOD_NAME = re.compile(
     r'^on_({})(_\w+)?$'.format('|'.join(method.lower() for method in COMBINED_METHODS))
 )
 
+SynchronousResource = Callable[..., Any]
+AsynchronousResource = Callable[..., Awaitable[Any]]
+Resource = Union[SynchronousResource, AsynchronousResource]
```
kgriffs, line 45: I think we need to better distinguish between an action callable, a responder callable, and a resource class:

(a) An action callable is the the thing to execute before or after the designated HTTP method responders. It has a signature that is variadic but defining the type as a `typing.Protocol` might work, e.g.:

```py
class SynchronousAction(Protocol):
    def __call__(self, req: Request, resp: Response, resource: type, params: Mapping[str, str], *args: Any, **kwargs: Any) -> None: ...

Action = Union[SynchronousAction, AsynchronousAction]
```

(b) An HTTP method responder type, i.e. `func(self: type, req: Request, resp: Response, **kwargs: Any)`

(c) Resource class being the thing that contains HTTP method  `on_*()` responders. Since there is not base class for these, we can just use `type` to hint these.

falcon/hooks.py
```diff
@@ -13,23 +13,41 @@
...
 
-def before(action, *args, is_async=False, **kwargs):
+
+def before(
+    action: Resource, *args: Any, is_async: bool = False, **kwargs: Any
```
kgriffs, line 49: Referencing my comment above, `action` here would be `Union[Action, type]`. I realize that putting `type` in there makes this hint very lax, but there's nothing for it since it can be any resource class that follows the `on_*()` protocol.

kgriffs: Perhaps it would be best to actually alias `type` as `Resource` so that it is at least more self-documenting.

falcon/hooks.py
```diff
@@ -79,7 +97,7 @@ def do_something(req, resp, resource, params):
             *action*.
     """
 
-    def _before(responder_or_resource):
+    def _before(responder_or_resource: Resource) -> Resource:
```
kgriffs, line 100: The type here would be a Union of a `Resource` and a `Responder`

falcon/hooks.py
```diff
@@ -89,7 +107,7 @@ def _before(responder_or_resource):
                     # responder in the do_before_all closure; otherwise, they
                     # will capture the same responder variable that is shared
                     # between iterations of the for loop, above.
-                    def let(responder=responder):
+                    def let(responder: Resource = responder) -> None:
```
kgriffs, line 110: This should use a more specific type that is yet undefined, `Responder`.

falcon/hooks.py
```diff
@@ -109,7 +127,9 @@ def let(responder=responder):
     return _before
 
 
-def after(action, *args, is_async=False, **kwargs):
+def after(
+    action: Resource, *args: Any, is_async: bool = False, **kwargs: Any
```
kgriffs, line 131: Similar comments to the above; also review code below and apply more precise type hints as required.

kgriffs (changes requested):
Thanks for your help on this! Retrofitting type hints is a lot of work and we really appreciate it. I still need to finish reviewing several of the modules, but wanted to share my comments so far to get you unblocked.

CaselIT:
@copalco let us know if you plan on continuing this or if we should take this over the finishing lie. In any case thanks a lot for the effort on this!

falcon/app_helpers.py
```diff
@@ -280,22 +281,22 @@ class CloseableStreamIterator:
         block_size (int): Number of bytes to read per iteration.
     """
 
-    def __init__(self, stream: IO, block_size: int):
+    def __init__(self, stream: IO, block_size: int) -> None:
```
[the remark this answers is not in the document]
CaselIT: no, you have to type.every.single.one -.-'

copalco:
> @copalco let us know if you plan on continuing this or if we should take this over the finishing lie. In any case thanks a lot for the effort on this!

Sure I can continue. Also I don't mind if you will just propose the actual change as a comment.

falcon/response.py
copalco: Fixed.

falcon/inspect.py
copalco: "_get_source_info" can return None or str, but in this case it's always a string. In such cases I just use asserts instead of handling never occuring cases.

falcon/asgi_spec.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Constants, etc. defined by the ASGI specification."""
+from __future__ import annotations
```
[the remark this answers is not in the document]
CaselIT: in general `from __future__ import annotations` makes sense to have since
- it's how python will eventually work
- substantially speeds up the import of the module, since no type needs to be resolved, all annotations are just strings
- avoids the need of manually using string every time a forward-ref or type only import is referenced

overall, in my opinion, it makes sense to have in every module that uses type checking in any significant capacity

falcon/errors.py
```diff
@@ -169,6 +177,10 @@ class WebSocketServerError(WebSocketDisconnected):
     pass
 
 
+Kwargs = Union[str, int, None]
```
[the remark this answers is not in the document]
copalco: I'd rather leave it where it is, because that's the only module where Kwargs are actually used, so in my reasoning i thought that we should keep the code as close to the usage as possible. Renamed.

falcon/media/handlers.py
copalco: Yes I meant them as a replacement of protocols.

CaselIT: I think if we define them like this
```py
if TYPE_CHECKING:
  from typing import Protocol
  
  class MediaHandlers(Protocol): ...
```
then we use them like this
```py
if TYPE_CHECKING:
  from falcon.typing import MediaHandlers

  def i_get_an_handler(foo: MediaHandlers): ...
```

everything should work at runtime. It will fail type checking on python 3.7, but anyone who cares about type checking is most likely on a more recent version.

NOTE that we cannot use these as bases, so we cannot do `class Handlers(MediaHandlers, UserDict):`, but that should be ok in general, since the point of protocols is mostly to avoid having them as bases.

@vytas7 would that work for you?

falcon/typing.py
```diff
@@ -34,3 +61,6 @@
 #   arguments afterwords?
 # class SinkCallable(Protocol):
 #     def __call__(sef, req: Request, resp: Response, <how to do?>): ...
+NormalizedHeaders = Dict[str, str]
+RawHeaders = Union[NormalizedHeaders, List[Tuple[str, str]]]
```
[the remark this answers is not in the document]
copalco: Ok I didn't know that I was basing it on 
`def _load_headers(headers: Optional[RawHeaders]) -> NormalizedHeaders:
    """Transform the headers to dict."""
    if headers is None:
        return {}
    if isinstance(headers, dict):
        return headers
    return dict(headers)
`
Where headers that we transform are actually in the correct "Normalized" form
But sure I can change it. What would you suggest as a name of this third type?

falcon/hooks.py
```diff
@@ -13,23 +13,41 @@
 # limitations under the License.
 
 """Hook decorators."""
+from __future__ import annotations
 
 from functools import wraps
 from inspect import getmembers
 from inspect import iscoroutinefunction
 import re
+from typing import Any
```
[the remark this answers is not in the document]
copalco: Changed.

falcon/media/handlers.py
copalco: I've implemented it in accordance to @CaselIT . Please check if it's fine.

falcon/inspect.py
copalco: We could use `cast` or split _get_source_info to two functions with and without default. With default would always return str.

falcon/media/handlers.py
vytas7: We are probably going to drop active support for Python 3.7 in 4.0 anyway, so it doesn't matter if typing breaks there. I was just concerned with extra complexity and impact on performance (although it isn't the hottest path but still).

Just not "an handler" but "a handler" @CaselIT :slightly_smiling_face:

falcon/asgi_spec.py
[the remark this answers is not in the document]
vytas7: Agreed, it reduces the import complexity and has no impact on runtime performance, so I only see the positives as well.

falcon/inspect.py
vytas7: That works for me, I was just wondering. Well, ideally you could maybe add `assert`ion messages explaining the error?

falcon/media/handlers.py
CaselIT: I'm hopeless regarding use of an instead of a. I also routinely use it in fort of user :/

falcon/inspect.py
copalco: Added message.

falcon/asgi_spec.py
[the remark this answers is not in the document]
CaselIT: there is also a flake8 plugin the enforces this. maybe we could look into it at some point?

it's called `flake8-future-annotations`

CaselIT (dismissed):
thanks for updating it

falcon/hooks.py
copalco: Added simple typing, will improve it in accordance to your remarks in PR: https://github.com/falconry/falcon/pull/2183

falcon/hooks.py
copalco: Changed.

falcon/hooks.py
copalco: Changed

falcon/hooks.py
copalco: Done

CaselIT (dismissed):

falcon/request.py
```diff
@@ -2090,7 +2093,7 @@ class RequestOptions:
     auto_parse_qs_csv: bool
     strip_url_path_trailing_slash: bool
     default_media_type: str
-    media_handlers: Handlers
+    media_handlers: UserDict
```
vytas7, line 2096: Is there any reason why this is a blank `UserDict` whereas `Response.media_handlers` is typed as `falcon.typing.MediaHandlers`?

falcon/typing.py
```diff
@@ -11,19 +11,50 @@
...
+        ) -> Tuple[Serializer, Optional[Callable], Optional[Callable]]: ...
+
 
-from falcon.request import Request
-from falcon.response import Response
+Link = Dict[str, str]
 
 # Error handlers
-ErrorHandler = Callable[[Request, Response, BaseException, dict], Any]
+ErrorHandler = Callable[['Request', 'Response', BaseException, dict], Any]
```
vytas7, line 54: Why cannot we import `Request` and `Response` here, is it because of cyclic references? We could import just under `TYPE_CHECKING` if that helps.

falcon/typing.py
```diff
@@ -33,3 +64,6 @@
 #   arguments afterwords?
 # class SinkCallable(Protocol):
 #     def __call__(sef, req: Request, resp: Response, <how to do?>): ...
+NormalizedHeaders = Dict[str, str]
```
vytas7, line 67: Maybe we could simply rename these to:
```python
Headers = Dict[str, str]
```
and
```
HeaderList = Union[Headers, Iterable[Tuple[str, str]]]
```
?

falcon/errors.py
```diff
@@ -214,7 +225,13 @@ class HTTPBadRequest(HTTPError):
     """
 
     @deprecated_args(allowed_positional=0)
-    def __init__(self, title=None, description=None, headers=None, **kwargs):
```
vytas7, line 217: Not strictly related to this PR, but we need to make a decision whether we want to follow through this deprecation and disallow positional `HTTPError` arguments in Falcon 4.0, or actually make the breaking change in 5.0.

falcon/media/handlers.py
```diff
@@ -41,7 +47,7 @@ class Handlers(UserDict):
     def __init__(self, initial=None):
         self._resolve = self._create_resolver()
 
-        handlers = initial or {
+        handlers: Mapping[str, Serializer] = initial or {
```
vytas7, line 50: Why do we need to define `Serializer` (probably not the best name since it also does deserialization, not only serialization), cannot we just type it what it is, `base.Handler`?

falcon/response.py
```diff
@@ -1251,7 +1256,7 @@ class ResponseOptions:
 
     secure_cookies_by_default: bool
     default_media_type: Optional[str]
-    media_handlers: Handlers
```
vytas7, line 1254: What is wrong with typing `Handlers` as `Handlers`?

vytas7 (changes requested):
Thanks, this looks like a really great start toward better typing! :100: 

The only part that looks slightly inconsistent, or at least I don't understand why it needs to be done this way, is typing of media `Handlers` in `request.py`, `response.py`, and `media/` itself.

falcon/typing.py
CaselIT: Importing under type_checking requires strings in an union like this, since it's in the right side and it's evaluated

CaselIT:
Will look

falcon/typing.py
```diff
@@ -33,3 +48,6 @@
 #   arguments afterwords?
 # class SinkCallable(Protocol):
 #     def __call__(sef, req: Request, resp: Response, <how to do?>): ...
+Headers = Dict[str, str]
+HeaderList = Union[Headers, List[Tuple[str, str]]]
+Status = Union[http.HTTPStatus, str, int]
```
vytas7, line 53: Maybe we could call it `ResponseStatus`? Just `Status` sounds a bit vague.

falcon/hooks.py
```diff
@@ -14,21 +14,35 @@
...
 
+if t.TYPE_CHECKING:  # pragma: no cover
+    import falcon as wsgi
+    from falcon import asgi
+
 _DECORABLE_METHOD_NAME = re.compile(
     r'^on_({})(_\w+)?$'.format('|'.join(method.lower() for method in COMBINED_METHODS))
 )
 
+Resource = object
```
vytas7, line 37: I think it would make sense to consolidate these kind of things in `typing.py`, that's why I created that module :slightly_smiling_face: 
But let's do that in follow-up PRs.

vytas7 (approved):
LGTM now :shipit:

PR #2172, opened 2023-09-09
chore(testing): allow test to be run using pytest-xdist

CaselIT:
Trying to run `pytest .\tests\ -n 6 --dist=worksteal` on master it results in an error since different test are collected.
This is mostly because we parametrize the test on random uuid, that result in the test signature to be different.

This fixes that by creating a deterministic id for the parametrized test where needed

### DIFF
--- a/tests/_util.py
+++ b/tests/_util.py
@@ -72,6 +72,7 @@ def disable_asgi_non_coroutine_wrapping():
 def as_params(*values, prefix=None):
     if not prefix:
         prefix = ''
+    # NOTE(caselit): each value must be a tuple/list even when using one single argument
     return [
         pytest.param(*value, id=f'{prefix}_{i}' if prefix else f'{i}')
         for i, value in enumerate(values, 1)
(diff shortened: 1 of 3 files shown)


tests/_util.py
```diff
@@ -65,3 +67,12 @@ def disable_asgi_non_coroutine_wrapping():
 
     if should_wrap:
         os.environ['FALCON_ASGI_WRAP_NON_COROUTINES'] = 'Y'
+
+
+def as_params(*values, prefix=None):
+    if not prefix:
+        prefix = ''
+    return [
+        pytest.param(*value, id=f'{prefix}_{i}' if prefix else f'{i}')
```
CaselIT, line 76: I guess this currently does not work when parametrize over a single param.

Should it be fixed?

vytas7: If the whole suite passes with your fixes and `pytest-xdist`, it can stay _as-is_, just add a `NOTE`.

vytas7 (approved):

tests/_util.py
CaselIT: OK adding

CaselIT:
Added a note there, can you take a second look?

vytas7 (approved):
LGTM

PR #2177, opened 2023-10-14
chore: support CPython 3.12, update CI jobs

### LINKED ISSUE #2176: Basic CPython 3.12 support
Implement basic CPython 3.12 infrastructure:
* CI gates 
* Build binary wheels
* Trove classifiers (some say there don't mean anything anymore, but we still keep them up to date)
* Any new warnings (except `cgi.parse_header` which is tracked separately)

Split issues for full support:
* #2178
* #2179

vytas7:
Closes #2176

### DIFF
--- a/falcon/util/misc.py
+++ b/falcon/util/misc.py
@@ -74,10 +74,20 @@ _DEFAULT_HTTP_REASON = 'Unknown'
 _UNSAFE_CHARS = re.compile(r'[^a-zA-Z0-9.-]')
 
 # PERF(kgriffs): Avoid superfluous namespace lookups
-strptime: Callable[[str, str], datetime.datetime] = datetime.datetime.strptime
-utcnow: Callable[[], datetime.datetime] = functools.partial(
+_strptime: Callable[[str, str], datetime.datetime] = datetime.datetime.strptime
+_utcnow: Callable[[], datetime.datetime] = functools.partial(
     datetime.datetime.now, datetime.timezone.utc
 )
+
+# The above aliases were not underscored prior to Falcon 3.1.2.
+strptime: Callable[[str, str], datetime.datetime] = deprecated(
+    'This was a private alias local to this module; '
+    'please reference datetime.strptime() directly.'
+)(datetime.datetime.strptime)
+utcnow: Callable[[], datetime.datetime] = deprecated(
+    'This was a private alias local to this module; '
+    'please reference datetime.utcnow() directly.'
+)(datetime.datetime.utcnow)
 
 
 # NOTE(kgriffs,vytas): This is tested in the PyPy gate but we do not want devs
@@ -134,7 +144,7 @@ def http_now() -> str:
         e.g., 'Tue, 15 Nov 1994 12:45:26 GMT'.
     """
 
-    return dt_to_http(utcnow())
+    return dt_to_http(_utcnow())
 
 
 def dt_to_http(dt: datetime.datetime) -> str:
@@ -178,7 +188,7 @@ def http_date_to_dt(http_date: str, obs_date: bool = False) -> datetime.datetime
         #   over it, and setting up exception handling blocks each
         #   time around the loop, in the case that we don't actually
         #   need to check for multiple formats.
-        return strptime(http_date, '%a, %d %b %Y %H:%M:%S %Z')
+        return _strptime(http_date, '%a, %d %b %Y %H:%M:%S %Z')
 
     time_formats = (
         '%a, %d %b %Y %H:%M:%S %Z',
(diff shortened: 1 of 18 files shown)


CaselIT (approved):

tests/asgi/test_ws.py
```diff
@@ -1118,6 +1118,9 @@ async def on_websocket(self, req, ws):
     async with conductor as c:
         if accept:
             async with c.simulate_ws() as ws:
+                # Make sure the responder has a chance to reach the raise point
+                for _ in range(3):
+                    await asyncio.sleep(0)
```
copalco, line 1123: Since we have same code in two places maybe we could extract it to some function e.g. "wait"?

falcon/util/misc.py
```diff
@@ -75,7 +75,9 @@
 
 # PERF(kgriffs): Avoid superfluous namespace lookups
 strptime: Callable[[str, str], datetime.datetime] = datetime.datetime.strptime
-utcnow: Callable[[], datetime.datetime] = datetime.datetime.utcnow
+utcnow: Callable[[], datetime.datetime] = functools.partial(
```
copalco, line 78: Maybe instead using partial we could actually create a function `utcnow`?

CaselIT, line 78: note that this is a potentially breaking change, since the datetime now has a timezone.

I think we are fine keeping your change though

tox.ini
```diff
@@ -172,6 +172,11 @@ commands = python "{toxinidir}/tools/clean.py" "{toxinidir}/falcon"
 [with-cython]
 deps = -r{toxinidir}/requirements/tests
        Cython
+       # NOTE(vytas): By using --no-build-isolation, we need to manage build
```
CaselIT, line 175: we need to drop that for v4. is there an issue for it?

CaselIT (approved):
great work, sorry for the delay

falcon/util/misc.py
vytas7: What is the advantage of creating it? Do you think we could write it in a slightly more performant way?

vytas7: Ah, right. This `utcnow` is just a performance alias (not published in our docs), so I don't think this is a breaking change. But it is better to be safe than sorry, I'll simply create a new a alias, `_utcnow`, and plop a `DeprecationWarning` on the old one.

vytas7: Privatized both now. Do you think we need a deprecation newsfragment for these?

copalco:
```
def utcnow() -> datetime.datetime:
      return datetime.datetime.now(datetime.timezone.utc)
```
It does exactly the same thing, does not use functools and is one line shorter :)

vytas7: Aha, I see how you reason then :slightly_smiling_face: 

I will, however, stick to `functools`, as I don't see anything wrong with them, and it depends how you look at it regarding the brevity of the two versions. If you disregard formatting/wrapping, the functools variant is actually a oneliner. :smiling_imp:

What is more, the `functools` version seems to be faster:
```python
>>> import timeit
>>> def original(param=None):
...     return param
... 
>>> import functools
>>> import timeit
>>> def original(param=None):
...     return param
... 
>>> def wrapped1():
...     return original('UTC')
... 
>>> wrapped2 = functools.partial(original, 'UTC')
>>> 
>>> timeit.repeat(wrapped1)
[0.07910766499844613, 0.07443887900080881, 0.07431553800051915, 0.074646307999501, 0.07460912899841787]
>>> 
>>> timeit.repeat(wrapped1)
[0.07943654300106573, 0.0741412629995466, 0.07445096799892781, 0.07412566799939668, 0.07406616599837434]
>>> 
>>> timeit.repeat(wrapped2)
[0.050794496999515104, 0.045646627000678563, 0.04568123100034427, 0.045801513999322196, 0.0458722419989499]
>>> 
>>> timeit.repeat(wrapped2)
[0.05152076400008809, 0.046164414001395926, 0.045919030000732164, 0.046146496999426745, 0.046126109000397264]
```

CaselIT: usually partial is a bit faster. probably not that important here

CaselIT (approved):

falcon/util/misc.py
copalco: partial is faster than defining function in pure python? 😯

CaselIT: partial is a class implemented in C, so I guess they use a fast path to execute a function and/or resolve the function at construction time vs when it's called

tox.ini
vytas7: Now, there is: https://github.com/falconry/falcon/issues/2184.
I tentatively scheduled it for 4.0, but we might need to push down the road in favour of a timely release.

PR #2183, opened 2023-11-05
feat(typing): add type hints to hooks

copalco:
# Summary of Changes

# Related Issues

# Pull Request Checklist


CaselIT:
this seems to include https://github.com/falconry/falcon/pull/2171. I'll get to this once the other one is merged (I hope to merge it soon)

CaselIT (approved):

falcon/hooks.py
```diff
@@ -20,29 +20,94 @@
...
+
+
+class SyncBeforeFn(Protocol):
+    def __call__(
+        self,
+        req: wsgi.Request,
+        resp: wsgi.Response,
+        resource: Resource,
+        params: Dict[str, Any],
+        *args: Any,
```
CaselIT, line 57: since we can't use paramspec due to the is_async kw, we can't use a simple `Callable[[request, etc], None]` here since it does not support args and kwargs

CaselIT: left a note

CaselIT (approved):

setup.cfg
```diff
@@ -54,6 +54,7 @@ include_package_data = True
 packages = find:
 python_requires = >=3.8
 install_requires =
+    typing-extensions; python_version<"3.8"
```
vytas7, line 57: This is no longer needed.

falcon/typing.py
```diff
@@ -62,3 +65,30 @@
...
 ResponseStatus = Union[http.HTTPStatus, str, int]
+
+Resource = object
+
+
+class SyncResponderMethod(Protocol):
+    def __call__(
+        self,
+        resource: Resource,
+        req: wsgi.Request,
```
vytas7, line 76: Since `wsgi.Request` is only imported under `TYPE_CHECKING`, but this one is not, is this module even importable at all?

vytas7 (changes requested):
Looks good, we just need to decide how the users should consume `typing.py`. Maybe it would be nice if it is importable without `TYPE_CHECKING` too? Should we document any aliases?

falcon/typing.py
```diff
@@ -62,3 +65,30 @@
 Headers = Dict[str, str]
 HeaderList = Union[Headers, List[Tuple[str, str]]]
 ResponseStatus = Union[http.HTTPStatus, str, int]
+
```
vytas7, line 68: I swear the above didn't work for me with sinks... But maybe I was testing on an older Mypy version back then, or made some other mistake :thinking:

falcon/typing.py
CaselIT: If you have future-annotations all these are treated as strings, so it's not an issue

CaselIT:
Since we dropped 3.7 we can probably avoid having everything inside a type check block. Will do

CaselIT:
> Maybe it would be nice if it is importable without TYPE_CHECKING too? Should we document any aliases?

For the moment I mentioned in the v4 chance log that any type alias should be treated as private, since it can change between releases.

I think once we are adding typing we can do a bit of cleanup and maybe provide a public set to aliaes

.coveragerc
```diff
@@ -7,9 +7,10 @@ parallel = True
 
 [report]
 show_missing = True
-exclude_lines =
+# https://coverage.readthedocs.io/en/latest/excluding.html#advanced-exclusion
+exclude_also =
```
vytas7, line 11: Maybe we could later move this to `pyproject.toml`? Not required to pass review now.

vytas7 (approved):
LGTM now :+1:
Thanks for the effort @copalco & @CaselIT!

.coveragerc
CaselIT: they seem to support it, so why not

PR #2185, opened 2023-11-12
chore(readthedocs): add `.readthedocs.yaml` as required

vytas7:
It seams that our builds have stopped working completely, Read the Docs now [mandates having a config file](https://blog.readthedocs.com/migrate-configuration-v2/):

```
Error

Problem in your project's configuration. No default configuration file found at repository's root.
```

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,21 @@
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
+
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,21 @@
...
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
```
CaselIT, line 11: we could up to 11 here 🚀

CaselIT (approved):

.readthedocs.yaml
vytas7: I was thinking that most of our CI gates were on 3.10, so kept it consistent, we can bump up to .11 or .12 when we update the majority of other gates.

PR #2200, opened 2023-12-15
feat(parse_query_string): change the default value of `csv` to `False`

### LINKED ISSUE #1999: `parse_query_string()`: change default value of `csv` to False
Some url encoders don't encode comma to %2C

And in such cases, the Falcon Query parser creates an array instead of a string. 

If you have a query string "ABC,ABC" instead of "ABC%2CABC" then if you try to fetch if using 

`msg = falcon.uri.parse_query_string(req.query_string).get("message")`

then msg would be an array instead of string

`msg = ['ABC', 'ABC'] `

which is incorrect

vytas7:
Closes #1999.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -380,18 +380,18 @@ class TestFalconUtils:
         result = uri.parse_query_string(query_string)
         assert result['a'] == decoded_url
         assert result['b'] == decoded_json
-        assert result['c'] == ['1', '2', '3']
+        assert result['c'] == '1,2,3'
         assert result['d'] == 'test'
-        assert result['e'] == ['a', '&=,']
+        assert result['e'] == 'a,,&=,'
         assert result['f'] == ['a', 'a=b']
         assert result['é'] == 'a=b'
 
         result = uri.parse_query_string(query_string, True)
         assert result['a'] == decoded_url
         assert result['b'] == decoded_json
-        assert result['c'] == ['1', '2', '3']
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -380,18 +380,18 @@ def test_parse_query_string(self):
         result = uri.parse_query_string(query_string)
         assert result['a'] == decoded_url
         assert result['b'] == decoded_json
-        assert result['c'] == ['1', '2', '3']
+        assert result['c'] == '1,2,3'
```
CaselIT, line 383: Doing a quick look at this file we are loosing tests for csv=True.
Have I missed them?

CaselIT (commented):
Change looks ok, but maybe we miss some tests now?

tests/test_utils.py
vytas7: We have very extensive parametrized tests for `parse_query_string()`, including both values of `csv`. But I can resurrect this specific one too so that we are not losing anything _per se_.

CaselIT: it seems that `test_to_query_str_encoding` does mostly that, but it's a bit harder to follow

CaselIT (approved):

tests/test_utils.py
vytas7: I have now resurrected that old original non-parametrized test too, as well as added a new combo to it.

CaselIT (approved):
thanks!

PR #2201, opened 2023-12-16
feat(cors): disallow `cors_enable` & additional `CORSMiddleware` combo

### LINKED ISSUE #1977: Better handling of `cors_enable=True` + added `CORSMiddleware`
Although similar to https://github.com/falconry/falcon/issues/1930, this issue proposes to improve not only documentation, but also provide a clearer indication to the developer who has happened to pass this unsupported combination of parameters.

Maybe we could emit a warning, silently disable the implicitly added middleware component if we detect that any instance of `CORSMiddleware` is already present, or even raise an Exception? (Although the latter might be seen as a breaking change -- discuss).

vytas7:
Closes #1977.

### DIFF
--- a/falcon/app.py
+++ b/falcon/app.py
@@ -244,6 +244,7 @@ class App:
         cors_enable=False,
         sink_before_static_route=True,
     ):
+        self._cors_enable = cors_enable
         self._sink_before_static_route = sink_before_static_route
         self._sinks = []
         self._static_routes = []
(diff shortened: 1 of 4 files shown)


falcon/app.py
```diff
@@ -461,10 +462,28 @@ def add_middleware(self, middleware: object) -> None:
         #   the chance that middleware may be None.
         if middleware:
             try:
-                self._unprepared_middleware += middleware
+                middleware = list(middleware)  # type: ignore
```
vytas7, line 465: Bonus points if anyone knows how type this correctly. Since we optionally allow passing in any `object`, `mypy` complains:
```
No overload variant of "list" matches argument type "object"  [call-overload]
```

CaselIT: I don't think you can. Maybe you can use
```py
try:
  if TYPE_CHECKING:
    assert isinstance(middleware, Iterable)
  ...
```

But I'm not sure it's worth it

falcon/app.py
```diff
@@ -461,10 +462,28 @@ def add_middleware(self, middleware: object) -> None:
...
             except TypeError:
                 # middleware is not iterable; assume it is just one bare component
-                self._unprepared_middleware.append(middleware)
+                middleware = [middleware]
+
+            if (
+                self._cors_enable
+                and len(
```
CaselIT, line 472: any reason why not using any? I think it's clearer
```py
any(isinstance(mc, CORSMiddleware) for mc in self._unprepared_middleware + middleware)
```

CaselIT (approved):
left a could of suggestions, but it looks ok also without applying them

falcon/app.py
vytas7: No, a simple `any` probably doesn't work like that because I'm looking for a `len` of more than 1, not nonzero.

CaselIT: right. The cors middleware is not added here but in the constructor.

sorry looked only at this section. I guess there is no better option for it

PR #2203, opened 2024-01-14
docs(FAQ): add a FAQ item on shutting down `wsgiref.simple_server`

### LINKED ISSUE #1559: Is there a way of adding a REST API call that causes falcon to cleanly shut down?
Hi.

I'm working on getting coverage information out of a REST API written using Python+Falcon.

To do this, I'm using coverage.py.  It appears that coverage.py doesn't like to do its analysis on processes that never terminate; I think it wants to write the .coverage file on process exit.   

Also, it seems like falcon doesn't want to shut down.  To deal with that, I'd like to add an API call that cleanly shuts down the API I'm testing.  So I would fire up the API, run the automated tests, issue the terminate call to the API, and collect the coverage results.

Does falcon have some sort of "do a clean shutdown" callable somewhere?  I grep'd the sources a little, and didn't find anything.

If not, could one be added?

Thanks for the cool software.  :)

vytas7:
Fixes #1559

### DIFF
--- a/docs/user/faq.rst
+++ b/docs/user/faq.rst
@@ -1277,6 +1277,11 @@ thread (otherwise it may deadlock):
         print('Serving on port 8000, POST to /shutdown to stop...')
         httpd.serve_forever()
 
+.. warning::
+   While ``wsgiref.simple_server`` is handy for integration testing, it builds
+   upon :mod:`http.server`, which is not recommended for production. (See
+   :ref:`install` on how to install a production-ready WSGI or ASGI server.)
+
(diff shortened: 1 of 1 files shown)


CaselIT (approved):
Do you think it make sense adding a note that wsgiref should not be used in prod?

I think it's documented somewhere else, so maybe not needed

vytas7:
Hm, good idea, it doesn't hurt to reiterate this.

CaselIT (approved):
💯

Issue #1820, opened 2020-12-17
Type Hints: Add tripe hinds to code, tutorial, snippets and examples


CaselIT:
To improve code readability and be friendlier wrt prp484 we should start adding type annotation to the code, probably starting from the snippet, tutorials and examples.

Originally discussed here https://github.com/falconry/falcon/pull/1737#issuecomment-744064360

vytas7:
See also https://github.com/falconry/falcon/issues/1350, which strives after full typing support for the whole codebase.

bibekjoshi54:
@vytas7 I can start working on this

vytas7:
@bibekjoshi54 Awesome sauce! Go ahead.

CaselIT:
That would be great!

BrianKlemm-GIT:
Hello, it looks like this issues was reopened. I am happy to add type hints for you. 

Let me know, thanks.

davetapley:
Just to throw it out there: I think a [Protocol](https://peps.python.org/pep-0544/) would be a great fit for `resource` in :
https://github.com/falconry/falcon/blob/4910dd73ecd1b9c8cf6cae045b26ad432fa56128/falcon/routing/compiled.py#L150

Something like:

```py
from typing import Protocol

from falcon.request import Request
from falcon.response import Response

class Resource(Protocol):
    def on_get(self, request: Request, resp: Response)) -> None:
        ...

    def on_post(self, request: Request, resp: Response)) -> None:
        ...

    # and so on ...
```

Then the signature would become:
```py
 def add_route(self, uri_template: str, resource: Resource, suffix: str | None = None, compile: bool = False)
```

The only caveat is that implementor would have to provide **all** the methods
(although could easily just `...` those with no implementation)

... at least until [optional protocol members](https://peps.python.org/pep-0544/#support-optional-protocol-members) are supported 🤞🏻

CaselIT:
That's an option, but I'm not sure it would provide much in terms of help for a user, since once the `add_route` call happens the class is already defined.

Maybe it could be used a "TypingResource" that may be used by user as superclass, where each method is defined in a type_checking block (to avoid it being picked up by the router). This could be useful since ide can autocomplete the signature of methods when overloading one.

Something like
```py
class TypeCheckResource:
  """Superclass of a Resource for typing purposes"""
  if TYPE_CHECKING:
    def on_get(self, req: Request, resp: Response, **path_args: Any) -> None:
      """Handles the GET method"""
    ...
```

[...]

PR #2205, opened 2024-01-29
Some more typing for Request/Response classes

davetapley:
# Summary of Changes

Just a few types.

# Related Issues

#1820

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -12,9 +12,14 @@
 
 """Request class."""
 
+
 from datetime import datetime
 from io import BytesIO
+import sys
 from uuid import UUID
+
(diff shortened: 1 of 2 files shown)


falcon/request.py
```diff
@@ -454,6 +458,9 @@ class Request:
     # Child classes may override this
     context_type = structures.Context
 
+    if sys.version_info >= (3, 11):
+        stream: InputStream
```
CaselIT, line 462: `IO[bytest]` imported from typing.

CaselIT (changes requested):
Hi, sorry for the delay in the review. I've left some comment.

I can take it over if you prefer

PR #2216, opened 2024-03-20
chore(sync): use `asyncio.Runner` for `async_to_sync()` on py311+

### LINKED ISSUE #2178: `DefaultEventLoopPolicy.get_event_loop()` is deprecated (in the case of no loop)
As of Python 3.12, if there is no event loop, `DefaultEventLoopPolicy.get_event_loop()` emits a deprecation warning, and threatens to raise an error in future Python versions.

No replacement is suggested by the official docs. When it does start raising an error, I suppose one can catch it, and create a new loop instead.

We have already updated our code to use this method to combat another deprecation... it seems Python wants to make it harder and harder obtaining the current loop from sync code (which makes no sense to me).

vytas7:
Fixes #2178

### DIFF
--- a/falcon/util/sync.py
+++ b/falcon/util/sync.py
@@ -236,11 +236,12 @@ def async_to_sync(
     Keyword Args:
         **kwargs: Additional args are passed through to the coroutine function.
     """
+    global _runner
     # NOTE(vytas): Sometimes our runner's loop can get picked and consumed by
     #   other utilities and test methods. If that happens, recreate the runner.
-    global _runner
     if _runner.get_loop().is_closed():
-        _runner = _runner_cls()
+        # NOTE(vytas): This condition is never hit on _DummyRunner.
+        _runner = _runner_cls()  # pragma: nocover
     return _runner.run(coroutine(*args, **kwargs))
 
 
--- a/falcon/util/sync.py
+++ b/falcon/util/sync.py
@@ -22,7 +22,27 @@ __all__ = [
     'wrap_sync_to_async_unsafe',
 ]
 
+Result = TypeVar('Result')
+
+
+class _DummyRunner:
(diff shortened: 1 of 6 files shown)


falcon/util/sync.py
```diff
@@ -214,17 +236,13 @@ def async_to_sync(
...
-    # NOTE(vytas): Work around get_event_loop deprecation in 3.10 by going via
-    #   get_event_loop_policy(). This should be equivalent for async_to_sync's
-    #   use case as it is currently impossible to invoke run_until_complete()
-    #   from a running loop anyway.
-    loop = asyncio.get_event_loop_policy().get_event_loop()
-    return loop.run_until_complete(coroutine(*args, **kwargs))
+    global _runner
+    # NOTE(vytas): Sometimes our runner's loop can get picked and consumed by
+    #   other utilities and test methods. If that happens, recreate the runner.
+    if _runner.get_loop().is_closed():
```
CaselIT, line 242: would it make sense to hide this away in a method? Something like

> return _active_runner().run(coroutine(*args, **kwargs))

(that could even be a class instance with a call method

CaselIT (commented):
looks ok

falcon/util/sync.py
vytas7: What would be the benefits of this change, avoiding `global`?
I can change, for me it works either way.

CaselIT: mainly to split the logic of getting an active runner from the execution of the coroutine.
regarding use of global or not, should not make any difference either way so a function is probably easier to do

vytas7: Refactored, check if it looks better now.

CaselIT (approved):
seems ok!

vytas7:
OK, merging this since we want to unblock Uvicorn gates

PR #2217, opened 2024-03-21
feat(parse_header): provide our own implementation of `parse_header()`

### LINKED ISSUE #2066: Replace usage of `cgi.parse_header()` (slated for removal in 3.13)
[PEP 594](https://peps.python.org/pep-0594) specifies removal of "dead batteries" in CPython 3.13, with deprecation announced in 3.11.

Apparently, [cgi](https://peps.python.org/pep-0594/#cgi) is also on the chopping block, so we'll have to replace `cgi.parse_header()`. At first glance, the suggested replacement (instantiating an `email.message.Message`) looks clunky and likely to perform worse. Maybe we can reimplement it in Cython instead?

vytas7:
I have another branch which also introduces Cython's "pure Python mode" in order to speed up `parse_header()`, but we can circle back on that optimization in Falcon 4.1.

I'm also planning to reimplement the parts of `mimeparse` that we actually use, and unvendor it completely.

Fixes #2066

### DIFF
--- a/falcon/util/mediatypes.py
+++ b/falcon/util/mediatypes.py
@@ -17,6 +17,19 @@
 import typing
 
 
+def _parse_param_old_stdlib(s):  # type: ignore
+    while s[:1] == ';':
+        s = s[1:]
+        end = s.find(';')
+        while end > 0 and (s.count('"', 0, end) - s.count('\\"', 0, end)) % 2:
+            end = s.find(';', end + 1)
+        if end < 0:
+            end = len(s)
+        f = s[:end]
+        yield f.strip()
+        s = s[end:]
+
+
 def _parse_header_old_stdlib(line):  # type: ignore
     """Parse a Content-type like header.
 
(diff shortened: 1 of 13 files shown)


falcon/util/mediatypes.py
```diff
@@ -0,0 +1,89 @@
...
+    """Parse a Content-type like header.
+
+    Return the main content-type and a dictionary of options.
+
+    Note:
+        This method has been copied (almost) verbatim from CPython 3.8 stdlib.
+        It is slated for removal from the stdlib in 3.13.
+    """
+
+    def _parseparam(s):  # type: ignore
```
CaselIT, line 30: I know we are just coping this, but it makes no sense to define this function inside this outer function, since each time python needs to re-create it.

Can we move it at module level?

CaselIT (dismissed):
have just glanced at the mediatypes module since it's copied from the stdlib.

Did you make changes to it?

falcon/util/mediatypes.py
vytas7: Ah lol, I actually moved it inside to keep things cleaner without thinking too much about it :facepalm:

vytas7:
> have just glanced at the mediatypes module since it's copied from the stdlib.
> 
> Did you make changes to it?

@CaselIT yes, I wrote a much simpler implementation for the common case, and fall back to the old stdlib impl for more complex cases.

CaselIT:
Ok, I'll take a better look later

tests/test_mediatypes.py
```diff
@@ -0,0 +1,33 @@
...
+        ('text/plain; charset=utf-8', ('text/plain', {'charset': 'utf-8'})),
+        ('stuff/strange; missing-value; missing-another', ('stuff/strange', {})),
+        ('stuff/strange; missing-value\\missing-another', ('stuff/strange', {})),
+        (
+            'application/falcon; P1 = "key; value"; P2="\\""',
+            ('application/falcon', {'p1': 'key; value', 'p2': '"'}),
+        ),
+    ],
+)
+def test_parse_header(value, expected):
```
CaselIT, line 32: we could add one more test with more than one key on the new simpler path

CaselIT (approved):
looks ok but could maybe use one more test

tests/test_mediatypes.py
vytas7: Good point! Updated now.

CaselIT (approved):

PR #2226, opened 2024-04-05
chore(typing): add more descriptive typing to `Context`

vytas7:
This change affords arbitrary attribute access for `req.context` & `resp.context` in typed code without having to mask Mypy warnings similar to `error: "Context" has no attribute "..."`.

### DIFF
--- a/falcon/util/structures.py
+++ b/falcon/util/structures.py
@@ -147,19 +147,13 @@ class Context:
     if TYPE_CHECKING:
 
         def __getattr__(self, name: str) -> Any:
-            try:
-                return self.__dict__[name]
-            except KeyError:
-                raise AttributeError(name) from None
+            ...
 
         def __setattr__(self, name: str, value: Any) -> None:
-            self.__dict__[name] = value
+            ...
 
         def __delattr__(self, name: str) -> None:
-            try:
-                del self.__dict__[name]
-            except KeyError:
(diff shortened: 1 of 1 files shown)


falcon/util/structures.py
```diff
@@ -141,6 +142,25 @@ class Context:
     True
     """
 
+    # NOTE(vytas): Define synthetic attr access methods (under TYPE_CHECKING)
+    #   merely to let mypy know this is a namespace object.
+    if TYPE_CHECKING:
+
+        def __getattr__(self, name: str) -> Any:
```
CaselIT, line 149: there is no need to provide an implementation here.

Just define the function:
```py
def __getattr__(self, name: str) -> Any:...
def __setattr__(self, name: str, value: Any) -> None:...
def __delattr__(self, name: str) -> None:...
```

the type checker does not execute the code and `TYPE_CHECKING` is never true so the implementation has no reason to be there

CaselIT (changes requested):
left a suggestion to dry the code

falcon/util/structures.py
vytas7: Ah, right :man_facepalming: I was first prototyping without `TYPE_CHECKING` and then never removed the impls.

CaselIT (approved):
💯

Issue #2156, opened 2023-07-14
Drop `testtools` support?


vytas7:

vytas7:
Another option is to drop `testtools`, but leave an escape hatch via an envvar along the lines of [Custom HTTP Methods](https://falcon.readthedocs.io/en/stable/api/routing.html#custom-http-methods) if anyone is still using that.

CaselIT:
I have no opinion on this, either way works for me

kgriffs:
I don't think we need to keep supporting it. It's a minor convenience left over from early OpenStack days for a very small community relative to the broader python ecosystem.

vytas7:
We won't be able to fully drop it until 5.0, but we should emit a deprecation warning whenever the `TestCase` is rebased on top of `testtools.TestCase`, and add an escape hatch via an envvar to be able to import a custom testcase base.

EricGoulart:
I’m sorry for the mistake with my previous pull request. I realize now that this wasn’t the correct process, and I apologize for any inconvenience it may have caused. I’m currently in the process of fixing this and will submit the pull request the correct way.

vytas7:
No worries @EricGoulart, and thanks for your interest in contributing to this issue!
These mistakes do no real harm except a notification or two in our mailboxes, which is not the end of the world :sweat_smile:

vytas7:
Note that using `testtools` actually creates a real bug, uncovered by recent versions of Mypy:
```
falcon/testing/test_case.py:34: error: Definition of "patch" in base class "TestCase" is incompatible with definition in base class "TestClient"  [misc]
```

Falcon's [`TestClient.patch`](https://falcon.readthedocs.io/en/stable/api/testing.html#falcon.testing.TestClient.patch) is shadowed by `testtools.TestClient.patch` via diamond inheritance... I guess we need to keep the `testtools` flavour for now, since it was introduced prior to the `patch()` alias (of `simulate_patch`).

CaselIT:
I think it's fine to remove for v5

ApparentlyTejas:
I'd like to take this. Rather than a full deprecation (which touches the public dependency surface and seems worth a separate v5 discussion), I'm going with the minimal fix @vytas7 outlined as a fallback: reword the `TestCase` docstring so it doesn't read as recommending `testtools`, and fix the underlying mypy issue.

[...]

PR #2233, opened 2024-05-07
chore(CI): pin `pytest` + ASGI tutorial fixes

vytas7:
It seems that something breaks when using `testtools.TestCase` under `pytest` 8.2.x along the lines of https://github.com/pytest-dev/pytest/issues/12263. This PR temporarily pins `pytest` until the issue is resolved upstream, or we decide on how we want to move forward with this. An alternative is to drop active support for `testtools` altogether: #2156.

Also patched is some typing import issue in ASGI tutorial tests. The issue seems to also stem from third party packages (`fakeredis`) under Python 3.10 that we had hardwired in the tox env by mistake. I'll file a followup issue, because this also means that our tutorial is not fully usable on 3.10 or older.

### DIFF
--- a/docs/user/tutorial-asgi.rst
+++ b/docs/user/tutorial-asgi.rst
@@ -180,13 +180,12 @@ We can now implement a basic async image store. Save the following code as
 
 
     class Image:
-
         def __init__(self, config, image_id, size):
             self._config = config
 
             self.image_id = image_id
             self.size = size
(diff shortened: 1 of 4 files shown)


CaselIT (commented):
I would also be ok in dropping testtools, but for now this is fine

CaselIT (approved):

PR #2234, opened 2024-06-22
Type Annotations for Before Hook

nZac:
# Summary of Changes

Despite the poorly named branch, this PR brings a brings type annotations to the `before` hook. The approach is a proof of concept and I'll welcome suggestions. I've not typed all the functions beneath it, just the top level API though that should be accomplished at some point. See the commit messages for more details.

# Related Issues

The existing type issues are closed already, so N/A for now.

# Pull Request Checklist

- [ ] Coding style is consistent with the rest of the framework.
- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/latest/quickstart.html#creating-news-fragments) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)


setup.cfg
```diff
@@ -54,6 +54,7 @@ include_package_data = True
 packages = find:
 python_requires = >=3.7
 install_requires =
+    typing_extensions
```
CaselIT, line 57: it would be great if we could require this only for a pthon version lower of a certain version. 

Since at the moment we need ParamSpec I guess we could make it install only for < 3.10

falcon/hooks.py
```diff
@@ -79,6 +91,8 @@ def do_something(req, resp, resource, params):
             *action*.
     """
 
+    is_async = kwargs.get('is_async', False)
```
CaselIT, line 94: if we keep `is_async: bool = False,` listed explicitly this line is not needed

setup.cfg
vytas7: Yes, we want to keep Falcon free from third party dependencies.
It is OK to add dependencies for soon EOL 3.8 or 3.9.

falcon/hooks.py
vytas7: I would like to deprecate `is_async` now that Cython 3.0 properly marks coroutine functions if run on Python 3.10+. I will create a separate issue for this.

nZac: If we can drop `is_async` this PR becomes possible. After I pushed this branch I continued my testing and realized this wont work the way I thought it would otherwise I'm not seeing a way to actually type `before` and `after` methods. I could be missing something though 🤷🏼‍♂️

CaselIT:
this was implemented in https://github.com/falconry/falcon/pull/2183

CaselIT:
will leave a todo in the code to improve that part

PR #2236, opened 2024-07-11
refactor: replace `flake8` plugins with `ruff`

### LINKED ISSUE #2231: Use `ruff` linter instead of `flake8`
#2218 aims to replace the `blue` formatter with `ruff`, however, we could potentially replace all our `flake8` plugins with the `ruff` linter as well (needs to be verified).

CaselIT:
Replace flake8 with ruff

Not 100% sure all the options are the same, but no major changes were required

Fixes #2231


vytas7:
Looks great 💯

Are you sure we are checking the import order though? I remember playing with that, and even when I configured `ruff` to follow the Google style we are currently using, it still generated a fair diff, and here I see no changes in that 🤔

CaselIT:
Honestly no, not sure if it activated thwt check

examples/asgilook/asgilook/store.py
```diff
@@ -3,9 +3,10 @@
 import io
 
 import aiofiles
-import falcon
 import PIL.Image
 
+import falcon
```
vytas7, line 8: On the one hand, this makes sense from Falcon's perspective, but from the tutorial's perspective, `falcon` should be grouped just together with other 3rd parties just like `aiofiles`. But maybe it is hard to reconfigure Ruff just for the tutorials? Regardless, we need to make sure that other copypasted incarnations of this file are synced.

examples/look/look/app.py
```diff
@@ -2,7 +2,8 @@
 
 import falcon
 
-from .images import ImageStore, Resource
```
vytas7, line 5: Same as for ASGI, we need to recheck if we don't need to sync the copypasted copies.

vytas7 (changes requested):
It is shaping up for merging, we just need to revise these tutorial snippets.

vytas7 (approved):
Looks good :100:
(Even if I missed anything, it should be mostly stylistic.)

CaselIT:
Here it goes then

Issue #2237, opened 2024-07-12
Modernize setup.py to avoid custom built_ext subclass


CaselIT:
setuptools a while ago introduced an optional flag that makes the build continue if a cython compilation fails.
this removes the need for having a custom build subclass.

Example of change done in sqlalchemy: https://github.com/sqlalchemy/sqlalchemy/commit/282d86d1086b0aaaa0cc1a6eede922ce8214e8d0

PR #2242, opened 2024-07-13
chore: change `setup.py` to use optional build_ext option

CaselIT:
This allows skipping the compilation of the c file if it fails for some reason, like the compiler is missing.
The build will still fail if cython can't generate the c file from the py/pyx file but that should be ok.

The optional keyword is documented here https://setuptools.pypa.io/en/latest/userguide/ext_modules.html#setuptools.Extension
> optional ([bool](https://docs.python.org/3.11/library/functions.html#bool)) – specifies that a build failure in the extension should not abort the build process, but simply not install the failing extension.

Fixes: #2237


vytas7 (changes requested):
Looks good, and I've tested it to work well in different scenarios.
But apparently it is conflicted against `master`, we need to address these conflicts before merging.

CaselIT:
Thanks. Will merge with master soon!

CaselIT:
updated

vytas7 (approved):

PR #2252, opened 2024-07-18
docs(user, util): address broken links

vytas7:
The broken link itself has been discovered by @e-io.
~~I haven't been able to find what inserts the `_asyncio` module in the index. But it isn't easy to find that index anyway, is it?~~
**Edit**: I've found the culprit for that `_asyncio` in the index. We had aliased functions directly from the stdlib, so I went on to soft-deprecate them for now. I'll create an issue to deprecate them (but there is probably no rush to do so.)

### DIFF
--- a/docs/api/util.rst
+++ b/docs/api/util.rst
@@ -56,7 +56,7 @@ However, as of Falcon 4.0+, these aliases are identical to their :mod:`asyncio`
 counterparts on all supported Python versions. (They are only kept for
 compatibility purposes.)
 
-Simply use :func:`asyncio.get_running_loop` & :func:`asyncio.create_task`
+Simply use :func:`asyncio.get_running_loop` or :func:`asyncio.create_task`
 directly in new code.
 
 Adapters
--- a/docs/api/util.rst
+++ b/docs/api/util.rst
@@ -45,11 +45,19 @@ Async
 Aliases
 ~~~~~~~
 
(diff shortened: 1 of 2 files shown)


docs/api/util.rst
```diff
@@ -45,11 +45,19 @@ Async
...
-.. autofunction:: falcon.create_task
+* ``falcon.get_running_loop()`` → :func:`asyncio.get_running_loop`
+
+* ``falcon.create_task(coro, *, name=None)``  → :func:`asyncio.create_task`
+
+However, as of Falcon 4.0+, these aliases are identical to their :mod:`asyncio`
+counterparts on all supported Python versions. (They are only kept for
+compatibility purposes.)
+
+Simply use :func:`asyncio.get_running_loop` & :func:`asyncio.create_task`
```
CaselIT, line 59:
```suggestion
Simply use :func:`asyncio.get_running_loop` or :func:`asyncio.create_task`
```

CaselIT (approved):
left a suggestion

docs/api/util.rst
vytas7: Hm, I meant _and_. But _or_ does the job too if it feels more natural to you.

CaselIT: i did not like the & there. and worked too. overall all 3 are likely fine

Issue #2256, opened 2024-07-25
Add web browser interaction to the WS tutorial


vytas7:
Add a simple web page that is served from the tutorial's app, and make it connect and send/recv things over WS in a more visual way. Because that's what the majority of users use WS for 🙂

Issue #2255, opened 2024-07-25
Misc WS tutorial polishing


vytas7:
* Address @CaselIT's remaining comments on the PR https://github.com/falconry/falcon/pull/2245
* Replace `print(...)` statements with logging
* (Insert other things that could be improved)

(Excluding #2256.)

PR #2257, opened 2024-07-25
docs(user): improve WS tutorial

CaselIT:
Some improvements to the ws tutorial

Fixes: #2255
Fixes: #2256


CaselIT:
@vytas7 I've added an interface only to the simpler echo ws endpoint, since I did not want to overcomplicate the html.

Let me know if you think it's enough

CaselIT:
note to self: we should probably replace or integrate the link at
https://github.com/falconry/falcon/blob/cbca63dc7739720eab856f48b32f9e782438be7a/falcon/asgi/ws.py#L239

with https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code

docs/user/tutorial-websockets.rst
```diff
@@ -460,6 +461,7 @@ Things we've changed:
 - Added a new middleware class `AuthMiddleware` that will check the token on the first message.
 - Opening a websocket connection is now handled by the middleware.
 - The client now sends a token as the first message, if required for that route.
+- Falcon was configured to server a simple html page to use the echo websocket client for a browser.
```
vytas7, line 464: ~~to server~~ (sic) :arrow_right: _to serve_

vytas7, line 464: Maybe capitalize HTML and WebSocket?

examples/wslook/wslook/static/index.html
```diff
@@ -0,0 +1,80 @@
...
+<body>
+    <h1>WebSocket Echo Test</h1>
+    <form id="messageForm">
+        <input type="text" id="message" placeholder="Type your message here">
+        <input type="submit" value="Send">
+    </form>
+    <pre id="history"></pre>
+</body>
+
+</html>
```
vytas7, line 80: No newline at the end of file

vytas7 (changes requested):
Thanks! I wrote down a couple of comments inline.

Another thing we would need to go through is naming, it seems the tutorial directory is called `wslook`, but in the tutorial itself it is called differently.

CaselIT:
> the tutorial directory is called wslook, but in the tutorial itself it is called differently.

Do you habe a suggestion regarding the new name?

vytas7:
> > the tutorial directory is called wslook, but in the tutorial itself it is called differently.
> 
> Do you habe a suggestion regarding the new name?

I don't have a particular preference, I just find it a bit confusing and error-prone that the same directory is called `wslook` in our repository, and [`asyncws`](https://falcon.readthedocs.io/en/latest/user/tutorial-websockets.html#first-steps) in the tutorial itself.

CaselIT:
makes sense. will think what of the two names makes more sense

vytas7 (approved):
LGTM, thanks! :100:

PR #2261, opened 2024-08-09
docs(user): move examples and recipes to Python files

### LINKED ISSUE #44: Add example code & recipes to test suite
Need to do this in a DRY manner.

vytas7:
Fixes #44.


examples/recipes/output_csv_stream_asgi.py
```diff
@@ -0,0 +1,33 @@
...
+        writer.writerow(('n', 'Fibonacci Fn'))
+
+        previous = 1
+        current = 0
+        for i in range(n + 1):
+            writer.writerow((i, current))
+            previous, current = current, current + previous
+
+            for chunk in stream.result:
+                yield chunk
```
CaselIT, line 27: yield from? this seems the only difference from the wsgi one (other than the async bits)

examples/recipes/request_id_context.py
```diff
@@ -0,0 +1,19 @@
+# context.py
+
+import threading
```
CaselIT, line 3: can't you just update this here, instead of having another PR for #2260 ?

CaselIT: I don't think you would even need this wrapper class.

tests/conftest.py
```diff
@@ -17,6 +24,78 @@ def asgi(request):
     return request.param
 
 
+class _SuiteUtils:
```
CaselIT, line 27: you didn't remove the util file

tests/test_examples.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+
+def test_things_advanced(asgi, util):
+    suffix = '_asgi' if asgi else ''
+    advanced = util.load_module(f'examples/things_advanced{suffix}.py')
+
+    # NOTE(vytas): The ASGI example explicitly requires Content-Length
+    #   (its middleware errors out otherwise with 400).
+    #   Should we change this?
```
CaselIT, line 23: makes sense.

tests/test_recipes.py
```diff
@@ -0,0 +1,108 @@
...
+        'recipe,expected_head',
+        [
+            ('output_csv_text', '"fruit","quantity"\r\n"apples",13\r\n'),
+            ('output_csv_stream', '"n","Fibonacci Fn"\r\n0,0\r\n1,1\r\n'),
+        ],
+        ids=['simple', 'stream'],
+    )
+    def test_csv_output(self, asgi, util, recipe, expected_head):
+        suffix = 'asgi' if asgi else 'wsgi'
+        module = util.load_module(f'examples/recipes/{recipe}_{suffix}.py')
```
CaselIT, line 65: maybe it would make sense to have a `util.load_recepie(recepie, asgi)` since a similar thing is used multiple times

CaselIT (commented):
nice work!

examples/recipes/request_id_context.py
vytas7: I didn't want to alter that recipe too much, I'm not even the author. IIRC it was created by various people during PyCon US 2019 Sprints.

examples/recipes/output_csv_stream_asgi.py
vytas7: See [the note on the recipe itself](https://falcon.readthedocs.io/en/latest/user/recipes/output-csv.html#streaming-large-csv-files-on-the-fly); also here: [Asynchronous `yield from`](https://peps.python.org/pep-0525/#asynchronous-yield-from).

tests/conftest.py
vytas7: No, I was thinking to take one step at a time. The comment is misleading and alluding to that the removal has happened, I will update the comment.

examples/recipes/output_csv_stream_asgi.py
CaselIT: interesting. that's not actually async though, but I guess it doesn't matter to the parser.

tests/conftest.py
CaselIT: since there are test, we can maybe to that in this pr too, but as you prefer

vytas7: I'll just update the comment for now. Better keep shipping in small increments :shipit:

tests/test_examples.py
vytas7: (Being discussed on Gitter)

tests/test_recipes.py
vytas7: Done, added more affordances to load module to be able to easier parametrize.

vytas7:
> nice work!

Thanks!

CaselIT (approved):

PR #2263, opened 2024-08-10
chore(platform): declare CPython 3.13 support

### LINKED ISSUE #2258: CPython 3.13 support
Formalize CPython 3.13 support:

- [x] Make the test suite pass on 3.13. May be blocked on dependencies not being readied for 3.13 yet.
- [x] Add a CPython 3.13 CI gate.
- [x] ~~Build CPython 3.13 wheels.~~ :arrow_right: Will be delivered as part of #2259.
- [x] Advertise support using ["trove classifiers"](https://pypi.org/classifiers/).
- [x] Check if anything needs an update in `CONTRIBUTING.md`.
- [x] Update release notes regarding supported platforms.
- [x] Check for any new warnings emitted when running tests, e.g., whether we are relying on any deprecated functionality that will be removed in future Python versions. 
- [x] Check if all the binary test dependencies (such as `orjson`) are ready.

vytas7:
Closes #2258.


.github/workflows/tests.yaml
```diff
@@ -89,6 +89,12 @@ jobs:
           - python-version: "3.12"
             os: windows-latest
             toxenv: py312_nocover
+          - python-version: "3.13.0-rc.1"
```
CaselIT, line 92: It's likely better to use a range here, like `3.13.0-rc.1 - 3.12`

vytas7: Don't think there's much difference since we'll likely want to specify new required gate names when 3.13 is out.
But, OTOH, on the plus side we'll automatically be getting new releases towards 3.13.

CaselIT: yes, that's the main plus

vytas7:
(Still in _Draft_ as I'm continuing to dig what is going on with that `test_hello.NonClosingBytesIO` object.)

.github/workflows/mintest.yaml
```diff
@@ -15,20 +15,27 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.8"
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13"
```
vytas7, line 21: This was demonstrated to pass in the previous commit.

.github/workflows/mintest.yaml
```diff
@@ -15,20 +15,27 @@ jobs:
...
         uses: actions/setup-python@v4
+        if: ${{ matrix.python-version != '3.13' }}
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Set up Python 3.13
+        uses: actions/setup-python@v4
+        if: ${{ matrix.python-version == '3.13' }}
+        with:
+          python-version: "3.13.0-rc.1 - 3.13"
```
CaselIT, line 37: why you went with this mode? to preserve the label?

CaselIT (approved):

.github/workflows/mintest.yaml
vytas7: Here probably it is pure bikeshedding :slightly_smiling_face:. Bit in `tests.yaml` it is sort of important because if you set branch protection based on names, it is annoying that they change every time you go from prerelease to release.

PR #2265, opened 2024-08-12
refactor: add proper top-level `__all__` specifiers

vytas7:


CaselIT:
Did you use a script of some sort?

CaselIT (approved):
We should do as next

vytas7:
I just used Python REPL :slightly_smiling_face: But we should use AI I suppose.

Like this:
```python
>>> import falcon.status_codes
>>> for name in dir(falcon.status_codes):
...     if not name.startswith('_'):
...         print(f'from falcon.status_codes import {name}')
... 
from falcon.status_codes import HTTP_100
from falcon.status_codes import HTTP_101
<...>
```

PR #2266, opened 2024-08-12
refactor(e2e-tests): improve typing of E2E tests server code

vytas7:
Together with `__all__` changes (#2265), `e2e-tests/server/` now passes `mypy` type checking with `--strict`.
However, there are still two `type: ignore`s for `App`'s lack of typing, and improper typing of `Response.status` -- both we ought address before shipping 4.0.


e2e-tests/server/hub.py
```diff
@@ -11,9 +11,9 @@
 class Emitter:
     POLL_TIMEOUT = 3.0
 
-    def __init__(self):
-        self._done = False
-        self._queue = asyncio.Queue()
+    def __init__(self) -> None:
+        self._done: bool = False
+        self._queue: asyncio.Queue[SSEvent] = asyncio.Queue()
```
CaselIT, line 16: Add `from future import annotations ` otherwise this is unrunnable in python<3.9

falcon/typing.py
```diff
@@ -40,6 +40,16 @@
 # Error serializers
 ErrorSerializer = Callable[['Request', 'Response', BaseException], Any]
 
+JSONSerializable = Union[
+    Dict[str, 'JSONSerializable'],
+    List['JSONSerializable'],
```
CaselIT, line 45: It's actually sequence here I think. Tuple serialises to list.

But maybe we can add tuple instead of any sequence

CaselIT (changes requested):
Couple of small things

e2e-tests/server/hub.py
vytas7: But does it really need to be run on <3.9?

CaselIT: Not really required, but that import may make sense in any case

falcon/typing.py
vytas7: I'm not sure `Sequence` is correct here, but you can suggest what the best annotation for JSON serializable data is, I don't think it matters _that_ much for SSE events.

Re `Sequence`, it doesn't seem that `json` supports arbitrary sequences...
```python
import collections.abc
import json

class MyRange(collections.abc.Sequence):
    def __init__(self, n):
        self.n = n
    def __len__(self):
        return self.n
    def __getitem__(self, key):
        key = int(key)
        if not 0 <= key < self.n:
            raise IndexError('Segfault huh?')
        return key

if __name__ == '__main__':
    mr = MyRange(7)
    print(f'{mr[5]=}')
    items = tuple(mr)
    print(items)
    print(json.dumps(mr))
```
:arrow_down: 
`TypeError: Object of type MyRange is not JSON serializable`

CaselIT: Ok. The current type seems fine, just add `Tuple['JSONSerializable',...]` to the union and it should be fine

e2e-tests/server/hub.py
vytas7: Indeed, added now, and the file compiles under 3.8.

falcon/typing.py
```diff
@@ -43,6 +43,7 @@
 JSONSerializable = Union[
     Dict[str, 'JSONSerializable'],
     List['JSONSerializable'],
+    Tuple['JSONSerializable'],
```
CaselIT, line 46: It should be 
```suggestion
    Tuple['JSONSerializable', ...],
```

PR #2267, opened 2024-08-12
chore: build binary wheels using `cibuildwheel`

### LINKED ISSUE #1899: Build wheels for macOS AArch64
Extend our wheel automation to produce wheels for macOS AArch64 (Apple Silicon).

If we want to do this in GitHub Actions, we might need to wait until an appropriate runner is added: https://github.com/actions/virtual-environments/issues/2187.

### LINKED ISSUE #2145: Build wheels for musllinux (PEP 656) platforms
This is a small feature request to build wheels for [PEP 656](https://peps.python.org/pep-0656/) `musllinux` platform tags, if your team aren't opposed to the idea.

We're seeing `pip install` for falcon on AArch64 Alpine Linux platforms take upwards of 20 minutes to build the wheel from source (on Github Actions nodes).

### LINKED ISSUE #2259: Build wheels using `cibuildwheel`
Migrate the current process of building binary wheels to [`cibuildwheel`](https://cibuildwheel.pypa.io/en/stable/).
`cibuildwheel` automates the process, and it also provides ready-made GitHub Actions for building and publishing.

vytas7:
This is a prototype version since PRs do not have access to secrets anyway, I think.

* Closes #1899 
* Closes #2145
* Closes #2259

The plan is to merge this PR once we are happy with the presented code, and test releasing to TestPyPi using the manual dispatch workflow. Then if we are happy we can add/adapt the scripts to perform sanity checks on the produced wheels, and set up publishing to PyPi, retiring the current "Create wheel" workflow,

I got some warnings from various platforms when building on CPython 3.8. So the plan is to build only `manylinux` and `musllinux` 64-bit x86 wheels for CPython 3.8.

Otherwise we are building the following combos:
CPython: 3.9, 3.10, 3.11, 3.12, 3.13.
Platforms: Linux x86 `manylinux` & `musllinux`, Linux ARM `manylinux` & `musllinux`, Linux IBM Z `manylinux`, macOS x86, macOS ARM, Windows x86. Only 64-bit builds.

### TL;DR Proposed CPython wheels
|                     | 3.8                     | 3.9                     | 3.10                    | 3.11                    | 3.12                    | 3.13                    |
| ------------------- |:-----------------------:|:-----------------------:|:-----------------------:|:-----------------------:|:-----------------------:|:-----------------------:|
| `manylinux_x86_64`  | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `manylinux_i686`    |                         |                         |                         |                         |                         |                         |
| `musllinux_x86_64`  | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `musllinux_i686`    |                         |                         |                         |                         |                         |                         |
| `manylinux_aarch64` |                         | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `manylinux_ppc64le` |                         |                         |                         |                         |                         |                         |
| `manylinux_s390x`   |                         | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `musllinux_aarch64` |                         | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `musllinux_ppc64le` |                         |                         |                         |                         |                         |                         |
| `musllinux_s390x`   |                         |                         |                         |                         |                         |                         |
| `macosx_x86_64`     |                         | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `macosx_universal2` |                         |                         |                         |                         |                         |                         |
| `macosx_arm64`      |                         | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `win_amd64`         |                         | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: |
| `win32`             |                         |                         |                         |                         |                         |                         |
| `win_arm64`         |                         |                         |                         |                         |                         |                         |

Example of the latest build before I removed the workflow from running on every PR: https://github.com/falconry/falcon/actions/runs/10416797908


.github/workflows/cibuildwheel.yaml
```diff
@@ -0,0 +1,146 @@
+# Build wheels using cibuildwheel (https://cibuildwheel.pypa.io/)
+name: Build wheels
+
+on:
+  # Run when a release has been created
+  # release:
+  #   types: [created]
```
CaselIT, line 7: I think this should be uncommented

.github/workflows/cibuildwheel.yaml
```diff
@@ -0,0 +1,146 @@
...
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.20.0
+        # Limit scope for testing
+        # if: ${{ matrix.python== 'cp38' }}
+        env:
+          CIBW_ARCHS_LINUX: all
+          CIBW_BUILD: ${{ matrix.python }}-${{ matrix.platform.name }}
+
+      - name: Upload artifacts
+        # NOTE(vytas): Not using v4 yet because it does not support combining
```
CaselIT, line 107: Outdated comment?

pyproject.toml
```diff
@@ -168,6 +168,14 @@ filterwarnings = [
...
+markers = [
+    "slow: mark Falcon tests as slower (potentially taking more than ~500ms).",
+]
 testpaths = [
     "tests"
 ]
+
+[tool.cibuildwheel]
+build-frontend = "build"
+test-requires = "-rrequirements/cibwtest"
```
CaselIT, line 180: Is a space miss here? Between -r and yhr requirement file

tests/asgi/test_example_asgi.py
```diff
@@ -1,223 +0,0 @@
-# examples/things_advanced_asgi.py
```
CaselIT, line 1: How come it was removed?

CaselIT (commented):
Nice work. Left some comments

.github/workflows/cibuildwheel.yaml
vytas7: Yes, will remove.

pyproject.toml
vytas7: No miss, this is an ugly workaround by design. I did have a space in the first iteration, but cibw was detecting it and splitting as two requirements despite double quoting. It works via an envvar but not via TOML. If you know how to add that space, happy to change.

OTOH Pip takes it fine without space too, so maybe it is not a big deal? I can add a `NOTE`.

tests/asgi/test_example_asgi.py
vytas7: It is no longer needed. This is just a copy from the examples folder.
These are already imported via `importlib` and tested in `tests/test_examples.py`.

CaselIT: Ok!

pyproject.toml
CaselIT: A note sounds good. Maybe there is an option to pass a requirement file instead?

vytas7: I haven't found any, I just found some example repositories passing with or without space via `CIBW_*` envvars.

Now that I have narrowed down that list, we could also list them inside a TOML list, but I think it is cleaner to keep it in a req file maintenance wise?

CaselIT: Makes sense to have a file. Probably the same wrt numpy. Let's just add a bote saying "do not add the space here since ..."

tests/asgi/test_example_asgi.py
vytas7: For reference, this has been improved recently in [#2261](https://github.com/falconry/falcon/pull/2261/files#diff-ef88d18dff1f7f3977c40a0649ba44b2c41d89676352a4d5c1fd4f3e7c81dfc7), I just forgot to remove this file in that PR. I did remove the WSGI counterpart.

pyproject.toml
vytas7: I'll try one more TOML approach, and will add a `NOTE` if even that fails.
Moreover, it is not strange to use the `-rrequirements.file` form, it is in fact quite usual that arguments following a single-letter shorthand option follow without a space, and ones following `--full-option-name` do have a space inbetween.

vytas7: OK this worked actually
```toml
test-requires = ["-r requirements/cibwtest"]
```

vytas7:
Coverage failed to accept reports, it has been really lame recently :grimacing: All green otherwise.

CaselIT (approved):
Nice that a list is supported!

Issue #1457, opened 2019-02-22
Privatize helpers, etc. that really should only be used internally by the framework


kgriffs:
Falcon modules import a number of helper functions that do not begin with an underscore, although such functions are not meant to be part of the public API. Remedy this in the next major version of the framework (as it would be a breaking change).

vytas7:
(See also: https://github.com/falconry/falcon/issues/1176 )

Vorashil:
Hey, I would like to take this issue if it's still open.

vytas7:
Hi @Vorashil !
Yeah, you are very welcome to!
Let us know if you are still interested in working on this though so that we do not duplicate efforts.

Vorashil:
Hi @vytas7 . Yes, I am still keen to do it. Please assign it to me. Thanks in advance.

vytas7:
We don't assign in any particular way :slightly_smiling_face: Your confirmation here is enough.
Looking forward to your PR!

bradleydamato:
@vytas7 @kgriffs do the helper functions include those shown here? https://github.com/falconry/falcon/blob/105c56db6a727da7231512c43fd4974a53fc960c/falcon/util/sync.py

If so, I'd like to fix these specifically.

vytas7:
Hi @bradleydamato !
Do you have any specific function from `sync.py` in mind? Most of these functions are meant to be available as utilites in the public API by design: https://falcon.readthedocs.io/en/latest/api/util.html#async . Furthermore, all this sync-async stuff is not a part of any stable release, as it was introduced only as part of the 3.0.0 release cycle, and we are, at the time of writing, at 3.0.0a2; so we can still make adjustments semi-freely.

As I understand this issue is mostly about functions that are already "public" (i.e., not prefixed with an underscore) in Falcon 2.0, but are not documented, and not meant to be a part of the public API.
I would let @kgriffs confirm, but IMHO `falcon.app_helpers.prepare_middleware` could serve as an illustration. Despite having a decent docstring, it does look to be deeply tied to internal interfaces and implementation details (returning a tuple of three tuples of methods :) ).

abid1998:
Hello I can take this up. What exactly needs to be done here?

vytas7:
Typing PRs such as #2271 introduced new constants like `MISSING` that may be used at runtime without underscore in some places.

These also need to be privatized.

vytas7:
I took a stab at this, and proposed a PR privatizing some request and reponse helpers.
[...]

PR #2271, opened 2024-08-14
feat(typing): type request module

CaselIT:
Type sync and async request modules.

Also move the attribute docs next to the actual attribute declaration, to improve usability with editors.

Some additional odd and end typing improvements

Also incorporate #2205:
```
Co-authored-by: Dave Tapley <dave@tapley.com>
```


falcon/asgi/request.py
```diff
@@ -841,41 +635,64 @@ async def get_media(self, default_when_empty=_UNSET):
...
         #   header property generator that DRY's up the memoization
         #   pattern for us.
-        # PERF(kgriffs): It probably isn't worth it to set
-        #   self._cached_if_match to a special type/object to distinguish
-        #   between the variable being unset and the header not being
-        #   present in the request. The reason is that if the app
-        #   gets a None back on the first reference to property, it
-        #   probably isn't going to access the property again (TBD).
-        if self._cached_if_match is None:
+        if self._cached_if_match is MISSING:
```
CaselIT, line 654: this is different

falcon/asgi/request.py
```diff
@@ -841,41 +635,64 @@ async def get_media(self, default_when_empty=_UNSET):
...
+            else:
+                self._cached_if_match = None
 
         return self._cached_if_match
 
     @property
-    def if_none_match(self):
-        if self._cached_if_none_match is None:
+    def if_none_match(self) -> Optional[List[Union[ETag, Literal['*']]]]:
+        if self._cached_if_none_match is MISSING:
```
CaselIT, line 667: this too

falcon/request.py
```diff
@@ -639,53 +446,111 @@ def content_length(self):
...
         #   header property generator that DRY's up the memoization
         #   pattern for us.
-        # PERF(kgriffs): It probably isn't worth it to set
-        #   self._cached_if_match to a special type/object to distinguish
-        #   between the variable being unset and the header not being
-        #   present in the request. The reason is that if the app
-        #   gets a None back on the first reference to property, it
-        #   probably isn't going to access the property again (TBD).
-        if self._cached_if_match is None:
+        if self._cached_if_match is MISSING:
```
CaselIT, line 494: this is different

falcon/request.py
```diff
@@ -639,53 +446,111 @@ def content_length(self):
...
+        list of :class:`falcon.ETag` objects or ``None`` if the header is
+        missing or its value is blank.
+
+        This property provides a list of all ``entity-tags`` in the
+        header, both strong and weak, in the same order as listed in
+        the header.
+
+        (See also: RFC 7232, Section 3.2)
+        """
+        if self._cached_if_none_match is MISSING:
```
CaselIT, line 515: this is different

CaselIT (commented):
@vytas7 I've marked a couple of point where I've done something slightly different. Honestly not sure we gain much by doing that change

falcon/asgi/request.py
copalco: I prefer MISSING than None.

CaselIT:
@vytas7 it's ready for review. We need to decide what to do with python 3.7

I vote to drop it. We can do best effort for 3.8 in 4.0, but let's keep the mention that it may be dropped in following 4.x releases

vytas7:
> @vytas7 it's ready for review. We need to decide what to do with python 3.7
> 
> I vote to drop it. We can do best effort for 3.8 in 4.0, but let's keep the mention that it may be dropped in following 4.x releases

Yes, all that is already described in the upcoming release notes: [Changes to Supported Platforms](https://falcon.readthedocs.io/en/latest/changes/4.0.0.html#changes-to-supported-platforms).

CaselIT:
> Re 3.7, cannot we just do Protocol = object in typing.py for 3.7, and things would just work? We could just write that typing is unsupported on 3.7. And then drop 3.7 and maybe 3.8 in Falcon 4.1.

I'm not really sold on doing this. Since I don't think type checkers will like it. In my opinion we either do something like https://github.com/falconry/falcon/blob/7341f96eab52f98b16080f2f458fa5fd3799f4f5/falcon/_typing_extensions.py or we drop it.
Doing the aliasing seems an hack

falcon/media/multipart.py
```diff
@@ -582,6 +581,11 @@ class MultipartParseOptions:
     ``application/x-www-form-urlencoded`` media types.
     """
 
+    if TYPE_CHECKING:
+        _DEFAULT_HANDLERS: ClassVar[Handlers]
+    else:
+        _DEFAULT_HANDLERS = None
```
vytas7, line 587: Realized just now (not related to this PR): maybe it would be better to assign `self.media_handlers` (see below) to a copy of this classvar in `__init__`? Because now customizing these handlers modifies the default as well.

vytas7: I went on to file this separately as #2293.

CaselIT: Makes sense

CaselIT:
not sure what's up with coverage.

falcon/constants.py
```diff
@@ -186,5 +188,14 @@
 #   None would be ambiguous.
 _UNSET = object()
 
+
+class Missing(Enum):
```
vytas7, line 192: IMHO we should be careful with what we expose as part of the public interface; see also #1457.

If we really want to expose this constant, let's document it, and add to our docs. Otherwise, let's prefix it with an underscore.

falcon/util/uri.py
```diff
@@ -554,7 +564,6 @@ def unquote_string(quoted: str) -> str:
 
 # TODO(vytas): Restructure this in favour of a cleaner way to hoist the pure
 # Cython functions into this module.
-if not TYPE_CHECKING:  # pragma: nocover
```
vytas7, line 557: Does it hurt to keep `TYPE_CHECKING` here? The idea was not even to mess with cythonized replacements, if we are type checking.

vytas7 (changes requested):
Looks great, thanks! :+1:

I would just like to make a decision on that `MISSING` constant before proceeding -- either we need to document it as part of the public interface, or underscore it.

falcon/util/uri.py
CaselIT: no, but I don't really see a need. the type checker is not complaining, so I don't see why we should bother with `if not TYPE_CHECKING`

falcon/constants.py
CaselIT: I'll move to typing and underscore it, but will keep `MissingOp` as is.

Once we are done typing stuff we can do a PR to decide which type alias are public and which are not

CaselIT:
re missing, I've moved it to the typing module, so that we can decide with all the rest when we are done with the typing what to make public and not to

falcon/typing.py
```diff
@@ -15,25 +15,40 @@
...
     from falcon.request import Request
     from falcon.response import Response
 
 
+class _Missing(Enum):
+    MISSING = auto()
+
+
+_T = TypeVar('_T')
+MISSING = _Missing.MISSING
```
vytas7, line 49: Could we underscore this constant too then, if we are using it outside of this module?

falcon/request.py
```diff
@@ -446,33 +111,152 @@ class Request:
...
-
-    _cookies = None
-    _cookies_collapsed = None
-    _cached_if_match = None
-    _cached_if_none_match = None
+    _cookies: Optional[Dict[str, List[str]]] = None
+    _cookies_collapsed: Optional[Dict[str, str]] = None
+    _cached_if_match: MissingOr[Optional[List[Union[ETag, Literal['*']]]]] = MISSING
+    _cached_if_none_match: MissingOr[Optional[List[Union[ETag, Literal['*']]]]] = (
+        MISSING
```
vytas7, line 118: Could we keep `None` for now? This PR isn't the smallest to follow, so ideally we should minimize the introduced runtime changes outside of typing. We probably need to profile things again otherwise (although these cached headers are probably not the most/important sensitive part; we should move the ones that are less frequently used to being cached in some `_state` attribute to increase perf even more).

vytas7 (changes requested):

falcon/constants.py
```diff
@@ -184,7 +184,8 @@
 
 # NOTE(kgriffs): Special singleton to be used internally whenever using
 #   None would be ambiguous.
-_UNSET = object()
+_UNSET = object()  # TODO: remove once replaced with missing
```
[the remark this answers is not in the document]
vytas7: That works for me, but regardless of the name, let's underscore it for any runtime usage (i.e. outside of type annotations), so we can change it at any point as we wish.

falcon/typing.py
CaselIT: my idea was to cleanup all the typing aliases at the same time in a final PR once all things are typed

falcon/request.py
CaselIT: I don't think we need to profile things for this change, since a comparison using `is` has the same time.

missing is used when the final value can include none. in those cases having a default of None is suboptimal like mentioned in the previous TODO.

I can create a new PR for this, but seemed so small a change that made sense to do inline with the typing one

vytas7: I don't think this is true for cythonized code, but since these methods are not very critical for core performance of the framework, I agree it is no big deal.

Just for reference
```python
_OBJECT = object()


def func1(obj):
    return obj is None


def func2(obj):
    return obj is _OBJECT
```
:arrow_down: 
```
.....................
func1 (None is None): Mean +- std dev: 31.5 ns +- 0.4 ns
.....................
func1 (_OBJECT is None): Mean +- std dev: 32.0 ns +- 2.0 ns
.....................
func2 (None is _OBJECT): Mean +- std dev: 37.0 ns +- 0.6 ns
.....................
func2 (_OBJECT is _OBJECT): Mean +- std dev: 36.9 ns +- 1.0 ns
```

This is because `is None` in Cython can be implemented as checking a C pointer vs `NULL`, whereas a global or local constant still needs to be resolved from name first.

But as said, I don't think this difference is enough to bother. But I do want to underscore these constants when used in runtime, so `_Missing`, `_MISSING`, `_UNSET`, etc.

CaselIT: as mentioned my idea was to worry about the public/private in a final cleanup PR. but if you prefer it here I can change it here.

personally I think all the properties could be re-written using a memorizing property since request has dict in any case

vytas7: OK, but let's not forget about that cleanup then!

vytas7 (approved):

falcon/request.py
CaselIT: roger. Should we also try the momoizing prop approach?

I think it would end up faster both in init than in use

vytas7: Dunno, we could maybe focus on getting 4.0 out the door first, and then circle back on rebuilding benchmarks, Codspeed, and optimizations.

I was thinking to add a private `._state` attribute, `object` or `dict` (whatever suits better), and then store less used cached header properties there. Also store stuff like the state of dependent middleware which we would deprecate and provide a shim for.

CaselIT: if we keep `__dict__` in request, memoized properties are all faster, since they avoid the property after the first run

vytas7: Re cleanup/privatizing, I added a note on #1457.

CaselIT: but sure let's just keep it in mind

PR #2281, opened 2024-08-20
refactor(tests): remove `_util` import in favour of fixture

vytas7:
This changeset follows up on the recent initiative to remove non-module imports in tests, using a session-wide fixture interfacing utility methods instead.


tests/_util.py
```diff
@@ -67,13 +65,3 @@ def disable_asgi_non_coroutine_wrapping():
 
     if should_wrap:
         os.environ['FALCON_ASGI_WRAP_NON_COROUTINES'] = 'Y'
-
-
-def as_params(*values, prefix=None):
```
vytas7, line 72: This wasn't as easy to reuse as fixture, but I reduced it to a oneliner, and it is used only in two places.
So I went on to simply inline it where affected.

tests/test_middleware.py
```diff
@@ -1035,6 +1029,9 @@ class FaultyComponentC:
         def process_response_async(self, req, resp, resource, req_succeeded):
             pass
 
+    if util.HAS_CYTHON:
+        pytest.skip(reason='Cythonized coroutine functions cannot be detected')
```
CaselIT, line 1033: is this still true in cython>3?

tests/test_request_attrs.py
```diff
@@ -32,6 +31,13 @@ def _make_etag(value, is_weak=False):
     return etag
 
 
+# NOTE(vytas): create_req is very heavily used in this module in unittest-style
+#   classes, so we simply recreate the function here.
+def create_req(asgi, options=None, **environ_or_scope_kwargs):
+    create_method = testing.create_asgi_req if asgi else testing.create_req
+    return create_method(options=options, **environ_or_scope_kwargs)
+
+
```
CaselIT, line 40: the `@pytest.mark.parametrize('asgi', [True, False])` applied to the class below could also be removed

CaselIT (commented):
looks ok, just a couple of comments

tests/test_middleware.py
vytas7: I was trying to focus on porting from `_util`, we can take a look at this separately. But probably it is easier to skip, because detection requires not only Cython 3.x, but also CPython 3.10+.

tests/test_request_attrs.py
vytas7: Yeah, I had it removed first but then had issues trying to port too much other stuff, and then forgot about it.
Removed now.

tests/test_middleware.py
CaselIT: ok, may make sense to understand if we still need to do this, since it may lead to simplification in the code too

CaselIT (approved):

PR #2286, opened 2024-08-20
feat(typing): type App classes

CaselIT:
This is a follow up of ##2271

### DIFF
--- a/docs/api/middleware.rst
+++ b/docs/api/middleware.rst
@@ -26,8 +26,11 @@ defined below.
 
         .. code:: python
 
+            from typing import Any
+            from falcon import Request, Response
+
             class ExampleMiddleware:
-                def process_request(self, req, resp):
+                def process_request(self, req: Request, resp: Response) -> None:
                     """Process the request before routing it.
 
                     Note:
@@ -42,7 +45,13 @@ defined below.
                             the on_* responder.
                     """
 
-                def process_resource(self, req, resp, resource, params):
+                def process_resource(
+                    self,
+                    req: Request,
+                    resp: Response,
+                    resource: object,
+                    params: dict[str, Any],
+                ) -> None:
                     """Process the request after routing.
 
                     Note:
@@ -62,7 +71,13 @@ defined below.
                             method as keyword arguments.
                     """
 
-                def process_response(self, req, resp, resource, req_succeeded):
+                def process_response(
+                    self,
+                    req: Request,
+                    resp: Response,
+                    resource: object,
+                    req_succeeded: bool
+                ) -> None:
                     """Post-processing of the response (after routing).
 
                     Args:
@@ -90,8 +105,13 @@ defined below.
 
         .. code:: python
 
+            from typing import Any
+            from falcon.asgi import Request, Response, WebSocket
+
             class ExampleMiddleware:
-                async def process_startup(self, scope, event):
+                async def process_startup(
+                    self, scope: dict[str, Any], event: dict[str, Any]
+                ) -> None:
                     """Process the ASGI lifespan startup event.
 
                     Invoked when the server is ready to start up and
@@ -111,7 +131,9 @@ defined below.
                             startup event.
                     """
 
-                async def process_shutdown(self, scope, event):
+                async def process_shutdown(
+                    self, scope: dict[str, Any], event: dict[str, Any]
+                ) -> None:
                     """Process the ASGI lifespan shutdown event.
 
                     Invoked when the server has stopped accepting
@@ -130,7 +152,7 @@ defined below.
                             shutdown event.
                     """
 
-                async def process_request(self, req, resp):
+                async def process_request(self, req: Request, resp: Response) -> None:
                     """Process the request before routing it.
 
                     Note:
@@ -145,7 +167,13 @@ defined below.
                             the on_* responder.
                     """
 
-                async def process_resource(self, req, resp, resource, params):
+                async def process_resource(
+                    self,
+                    req: Request,
+                    resp: Response,
+                    resource: object,
+                    params: dict[str, Any],
+                ) -> None:
                     """Process the request after routing.
 
                     Note:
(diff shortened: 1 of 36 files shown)


docs/api/middleware.rst
```diff
@@ -26,8 +26,11 @@ defined below.
 
         .. code:: python
 
+            from typing import Any
+            from falcon import Request, Response
+
             class ExampleMiddleware:
-                def process_request(self, req, resp):
+                def process_request(self, req: Request, resp: Response) -> None:
```
vytas7, line 33: I'm not 100% convinced we should type middleware definitions... :thinking:
It makes things clearer, but OTOH, the definitions get a bit dense.

CaselIT: I think is makes sense, since these are docs and typing is helpful also as documentation.

docs/api/websocket.rst
```diff
@@ -35,8 +35,12 @@ middleware objects configured for the app:
 
 .. code:: python
 
+    from typing import Any
+    from falcon.asgi import Request, WebSocket
+
+
     class SomeMiddleware:
-        async def process_request_ws(self, req, ws):
+        async def process_request_ws(self, req: Request, ws: WebSocket) -> None:
```
vytas7, line 43: Shouldn't this read `asgi.Request` in line with the above `middleware.rst`?

falcon/app.py
```diff
@@ -107,7 +135,13 @@ def process_request(self, req, resp):
...
 
-                    def process_resource(self, req, resp, resource, params):
+                        def process_resource(
+                            self,
+                            req: Request,
+                            resp: Response,
+                            resource: object,
+                            params: dict[str, Any],
+                        ) -> None:
                         \"\"\"Process the request and resource *after* routing.
```
vytas7, line 145: Do these docstrings need to be escaped that way?

CaselIT: we would need to use `'''` there or as the doc string marker.

maybe we should

docs/api/websocket.rst
vytas7: No, I would say I prefer as it is here. But in `middeware.rst` just above we were using this form: `wsgi.Request`, `asgi.Request`, etc.

CaselIT: since they should be in different tabs we can probably import the types directly instead of using wsgi. / asgi.

will update

falcon/app.py
vytas7: Ah, no need. These are docstrings inside docstring, makes sense. GitHub's diff view tricked me.

docs/api/websocket.rst
vytas7: This is an Rst file, we can write whatever we want :smiling_imp:

falcon/app.py
CaselIT: I feel like it would be nicer to either use `'''` in the outside docs or here.

vytas7: Ruff might decide to change back :slightly_smiling_face: 
I wouldn't like to use `'''` in the user facing docs. We could do on the outside, but as said, Ruff might be unhappy. Let's just keep it as is.

CaselIT: I think it likes it if inside the docs there is `"""`.
Like it will keep `'foo " bar'` as is, not change it to `"foo \" bar"`

falcon/app.py
```diff
@@ -810,28 +899,18 @@ def handle(req, resp, ex, params):
 
         """
 
-        def wrap_old_handler(old_handler):
-            # NOTE(kgriffs): This branch *is* actually tested by
-            #   test_error_handlers.test_handler_signature_shim_asgi() (as
-            #   verified manually via pdb), but for some reason coverage
-            #   tracking isn't picking it up.
-            if iscoroutinefunction(old_handler):  # pragma: no cover
```
vytas7, line 818: Is it safe to remove this code path for coroutine functions?

falcon/app.py
```diff
@@ -841,6 +920,7 @@ def handler(req, resp, ex, params):
                 )
 
         # TODO(vytas): Remove this shimming in a future Falcon version.
+        assert handler is not None
```
vytas7, line 923: Is this validation needed?
IIRC the best practice is not to rely on `assert` in production code (since it is, e.g., possible to silence `assert`s). If we want to validate `handler` here, we could explicitly write, `if handler is None: raise ValueError(...)`. We could also check if it is callable, not just `not None`.

Also, I think it is better to perform this validation just before my `TODO`, because that shimming is starts with `get_argnames`.

falcon/app.py
CaselIT: this code was a leftover. the asgi.App overloads this method and never calls super, so unless we want to support async in wsgi, this can never happen

falcon/app.py
CaselIT: iirc it was to avoid typing errors. I can remove it and see if I have another way of handling the typing issues

vytas7: I mean, that assert doesn't hurt, and we can even leave it inplace if needed, but it would be better to start off with raising a `ValueError`. In fact, I think it would hurt to validate `handler` this way.

falcon/app.py
vytas7: Isn't this subject to automatic wrapping in our own tests? Tests pass on your branch, so I guess we never exercise that, even if we do have a machinery for that.

CaselIT: I don't follow the test part.

I think this was a leftover added while developing asgi, before it was preferred to overload the method

CaselIT:
sorry for the rebase + squash, but it was starting to become a mess

vytas7 (approved):
Thanks, LGTM now bar these triple `'''` docstrings, I regret even starting that discussion :rofl:
But we can revise them later, no need to block the whole PR.

CodeCov is hallucinating as usual, but I have manually verified coverage by running `tox` on your branch.

CaselIT:
> Thanks, LGTM now bar these triple ''' docstrings, I regret even starting that discussion 🤣

I actually think it's a lot nicer for reading the source code. Having to escape the """ is quite ugly

PR #2287, opened 2024-08-20
docs: note that type checking was introduced

CaselIT:


vytas7 (approved):
We'll need to iterate on the release story anyway, but we can merge this as a good starting point. 👍

PR #2289, opened 2024-08-21
docs(sponsors): sync backers with reality & OpenCollective

### LINKED ISSUE #2278: Update backers for 4.0
Update backers in `BACKERS.md`, `README.rst`, docs, and the website.

I think we can remove all existing sponsors, only leaving GovCert since their prior contribution was an order of magnitude larger than anything else. And also add Sentry, since they are close to the definition of our Gold level, just they are donating via https://thanks.dev instead of OpenCollective.

vytas7:
Fixes #2278.

(We also port the non-dynamic `README.rst` from the `pyproject.toml` branch.)


setup.cfg
```diff
@@ -2,6 +2,7 @@
 name = falcon
 version = attr: falcon.__version__
 description = The ultra-reliable, fast ASGI+WSGI framework for building data plane APIs at scale.
+long_description = file: README.rst
```
CaselIT, line 5: I think it makes more sense to make this follow https://github.com/falconry/falcon/pull/2277

CaselIT (commented):
Why not make this follow the pyproject one? so we avoid changing the same files twice?

setup.cfg
vytas7: I have not decided yet whether we make the switch to `pyproject.toml` before 4.0 or post-4.0. So I suggest we merge this as a stopgap measure.

CaselIT: I think from a tool prospective is the same.
pyproject seems to the suggested way.

I think it would make sense before 4.0, but it's probably not that big of a deal to migrate after

CaselIT (approved):

setup.cfg
vytas7: We can migrate before, but we need to go through things like `CONTRIBUTING.md`, because the legacy `tag_build` stuff is gone, etc. I haven't assessed all the potential impact/fallout yet.

PR #2295, opened 2024-08-24
feat(typing): add type annotation to websocket module

CaselIT:
This issue includes #2286


falcon/testing/helpers.py
```diff
@@ -365,7 +366,12 @@ async def collect(self, event: Dict[str, Any]):
     __call__ = collect
 
 
-_WebSocketState = Enum('_WebSocketState', 'CONNECT HANDSHAKE ACCEPTED DENIED CLOSED')
+class _WebSocketState(Enum):
```
vytas7, line 369: Not crucial, but now that we build a proper `num`, it would be nice to assign them symbolic constants instead of `auto()`, like `CONNECT = 'connect'`, etc.

vytas7 (approved):
LGTM :+1:

falcon/testing/helpers.py
CaselIT: Place a note in the cleanup issue, so we don't forget

vytas7: Which one is the cleanup issue though :thinking:
And do you like the idea re these enums? :arrow_up:

CaselIT: I don't mind auto, since the enum value doesn't really have a defined purpose.

Using the strng is also ok

vytas7: OK, this is not a priority at any rate anyway, nor I don't think this is a breaking change to refine it later.

CaselIT: Created #2313 in any case, since we did have a few things to clean up

PR #2298, opened 2024-08-24
feat(typing): type media

CaselIT:
This includes #2297


falcon/media/json.py
```diff
@@ -148,55 +153,73 @@ def default(self, obj):
...
+        content_type: Optional[str],
+        content_length: Optional[int],
+    ) -> Any:
         return self._deserialize(await stream.read())
 
     # NOTE(kgriffs): Make content_type a kwarg to support the
     #   Request.render_body() shortcut optimization.
-    def _serialize_s(self, media, content_type=None) -> bytes:
-        return self._dumps(media).encode()
+    def _serialize_s(self, media: Any, content_type: Optional[str] = None) -> bytes:
```
vytas7, line 206: I suppose `media` could be of type `JSONSerializable` that I introduced to type SSE. But probably it causes more trouble than benefits...

vytas7 (approved):

falcon/media/json.py
CaselIT: yes, that's my impression.
using Any has the advantage that you can do `foo: MyTypedDict = req.get_media()` and the type checker is happy

vytas7: I added a point to the cleanup issue to weigh pros vs cons of this one.

PR #2299, opened 2024-08-25
feat(typing): type multipart forms

CaselIT:
Includes #2298


tests/test_media_multipart.py
```diff
@@ -291,7 +291,7 @@ def test_body_part_properties():
         if part.content_type == 'application/json':
             assert part.name == part.name == 'document'
         elif part.name == 'file1':
-            assert part.filename == part.filename == 'test.txt'
+            assert part.filename == 'test.txt'
```
vytas7, line 294: I think this might not be oversight as it looks like. I don't remember now, but maybe I want to make sure that this property can be referenced multiple times?

falcon/media/multipart.py
```diff
@@ -54,158 +71,57 @@
 # TODO(vytas): Consider supporting -charset- stuff.
 #   Does anyone use that (?)
 class BodyPart:
-    """Represents a body part in a multipart form.
+    """Represents a body part in a multipart form in a ASGI application.
```
vytas7, line 74: _in a ASGI application_ (sic) :arrow_right: _in **an** ASGI application_

vytas7 (changes requested):
Thanks, looks great as well as your other typing improvements :100: 

Just a couple of nitpicks; see inline.

tests/test_media_multipart.py
CaselIT: I can revert but we should likely leave a note mentioning why it's like this

vytas7: Yes, I agree, we need to add a note.

vytas7 (approved):

PR #2300, opened 2024-08-26
docs: refresh docs with the PyData Sphinx theme (MVP)

### LINKED ISSUE #2198: Docs facelift
Look into an alternative docs theme as the current one (Alabaster) is not too appealing and still exhibits various glitches despite extensive tweaking.
Maybe @pradyunsg's [`furo`](https://pypi.org/project/furo/) could work well for us?

### LINKED ISSUE #2332: Enable the dark theme for docs
If we go forward with #2300 , we should enable the dark theme as well (in the PR it was disabled just to cut scope).

vytas7:
Fixes #2198.
Closes #2332.
Logotype is just a placeholder, we need to create a proper one.

Tabs need rework, it would be better to use `sphinx_design`'s tabs, not the current module we're using, since it plays better with the PyData theme: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/web-components.html#tabs.

We also need to rework how table of contents is shown on pages, because PyData in general works a bit differently from Alabaster. I saw other projects hiding the ToC in some cases so that it is only summarized on the side bar on the left side, and then one tunes the landing page separately.


vytas7:
Unfortunately we'll need to cut corners in order to ensure a timely release.
The new theme probably already looks better or at least on par with the old (i.e., current) one, so maybe we can :shipit:
And iterate on it further in the 4.x cycle.

CaselIT (commented):
code looks ok, will try it.

Personally I think instead of disabling the dark them we should instead just keep the pydata default. If we have time to improve it we can do that later, but the builtin dark theme is good enough to have available if people prefer it

CaselIT:
@vytas7 what do you think about removing `.. contents:: :local:` ?

vytas7:
I'll take a look at these tables of contents, IIRC there are many options to achieve what we want, hidden ToC, etc.

CaselIT:
I mentioned removing it since there is now the TOC on the right side already, and having it also at the top seems a bit of an unnecessary duplication

vytas7:
Removed these local ToCs, and updated the Falcon Pygments theme, esp the `falconry-dark` style.

CaselIT:
we have some manual toc in app.rst and other. will push an update.

CaselIT (approved):

PR #2302, opened 2024-08-26
chore: deprecate TimezoneGMT class

### LINKED ISSUE #2301: Deprecate TimezoneGMT class
There is no need for it in py 3.8+.

CaselIT:
Fixes #2301

### DIFF
--- a/docs/_newsfragments/2301.newandimproved.rst
+++ b/docs/_newsfragments/2301.newandimproved.rst
@@ -1,0 +1,1 @@
+Deprecate ``TimezoneGMT`` class. Use :attr:`datetime.timezone.utc` instead.
--- a/docs/_newsfragments/2301.newandimproved.rst
+++ b/docs/_newsfragments/2301.newandimproved.rst
@@ -1,1 +1,0 @@
-Deprecate ``TimezoneGMT`` class. Use :attr:`datetime.timezone.utc` instead.
--- a/falcon/util/time.py
+++ b/falcon/util/time.py
@@ -11,9 +11,8 @@ for convenience::
 
 import datetime
 from typing import Optional
-import warnings
 
-from .deprecation import DeprecatedWarning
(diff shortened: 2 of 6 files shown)


docs/_newsfragments/2301.newandimproved.rst
vytas7, line 1: Release notes should not be worded in imperative as docstrings, just tell a story what has been done.

Moreover, I think deprecations better suit `.misc`, not `.newandimproved`.

falcon/util/time.py
```diff
@@ -20,6 +23,13 @@ class TimezoneGMT(datetime.tzinfo):
 
     GMT_ZERO = datetime.timedelta(hours=0)
 
+    def __init__(self) -> None:
```
vytas7, line 26: Maybe we could just plop the `@deprecation.deprecated(...)` decorator on `__init__`, just like we do in `falcon.app.API`?

tests/test_cookies.py
vytas7, line 1: Not critical, but we have tests for various `util` functions, maybe it testing `TimezoneGMT` would better fit there than inside cookies? It is not really related to cookies.

vytas7 (changes requested):

tests/test_cookies.py
CaselIT: it was there because it's only used there I think. will move them there

vytas7 (approved):
Thanks, it makes sense to deprecate this class indeed as it brings not much value over simply using UTC. :+1:

As discussed on Gitter, technically GMT and UTC are **not** the same, but CPython does not really support, itself, the subtle differences between the two time systems. (See also: https://github.com/python/cpython/issues/67762,)

CaselIT:
I don't actually think we need gmt in the cookies. It's very likely that if the specification ever mantated gmt it has now amended to mandata utc, since that's the one that's standard.

PR #2304, opened 2024-08-26
feat(typing): type response

CaselIT:
follow up of #2299


falcon/middleware.py
```diff
@@ -120,7 +120,11 @@ def process_response(
                 'Access-Control-Request-Headers', default='*'
             )
 
-            resp.set_header('Access-Control-Allow-Methods', allow)
+            if allow is not None:
+                # NOTE: not sure if it's more appropriate to raise an exception here
+                # This can happen only when a responder class defines a custom
+                # on_option responder method and does not set the 'Allow' header.
+                resp.set_header('Access-Control-Allow-Methods', allow)
```
CaselIT, line 127: Ideas on what to do here? cc @vytas7 @kgriffs ?

falcon/response.py
```diff
@@ -967,19 +1007,24 @@ def append_link(
...
         """Set the Cache-Control header.
 
         Used to set a list of cache directives to use as the value of the
         Cache-Control header. The list will be joined with ", " to produce
         the value for the header.
-
         """,
         format_header_value_list,
     )
+    """Set the Cache-Control header.
```
CaselIT, line 1020: I don't like the duplication, but I'm not sure if there is another way of keeping the property doc set and have ide pick up the docs

falcon/middleware.py
vytas7: OK, so what are our `OPTIONS` (pun intended)?

First, I think middleware is for adding _simple_ CORS, so I think we should minimize the chances of it raising an exception.

Realistically, I think there two main alternatives:
* If `Allow` is empty or missing, assume that no methods are allowed, and disable CORS altogether for the response in question (delete all `Access-Control` headers that this middleware sets).
* If `Allow` is empty or missing, assume that all methods are allowed, and set it to `Access-Control-Allow-Methods`: `*`.

At least [`Allow` docs on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow) state:
> An empty Allow header indicates that the resource allows no request methods, which might occur temporarily for a given resource, for example.

Which dovetails into the first option, strip `Access-Control-*` headers from response if `Allow` is empty.

But for me both alternatives are acceptable, as long as we document the chosen behaviour clearly.

vytas7: Please also note that this may affect sinks as well. Not sure about generic sinks, but I think we do want to allow permissive CORS on static routes, if the user has requested it via that parameter or CORS middleware. I think it is a somewhat common usecase in development, to serve some frontend resources from a Falcon app via a static route.

CaselIT: It doesn't seem that static route is currently adding allow headers. are you proposing to make it set the header?

vytas7: Yes, I think CORS should apply to static routes too.

CaselIT: it may make sense to do this change in another route?

may make sense to revert the behaviour change here and fix it there too

CaselIT: sorry I meant PR

vytas7: Yeah, we can create a separate issue for this as well not to pollute this PR

CaselIT: will be handled in #2325

falcon/response.py
CaselIT: well let's keep for now. we can maybe improve on it later on

CaselIT:
@vytas7 updated

CaselIT:
@vytas7 regarding the failure to install rapidjson in pypy, did you already had a fix in a PR?

vytas7:
> @vytas7 regarding the failure to install rapidjson in pypy, did you already had a fix in a PR?

Yes, but that PR still needs work, it is in Draft: https://github.com/falconry/falcon/pull/2324/files#diff-493968368d3cc25459cbc290f4f38ee1444eb298e0c5f840672c3dfce9b9f291R26.

CaselIT:
will cherry pick b4a6a4d6e339e243e7802c6153f8c9aa35a9edde

CaselIT:
seems to be working. pipeline should now be green

tests/test_cors_middleware.py
```diff
@@ -80,6 +86,23 @@ def test_enabled_cors_handles_preflighting(self, cors_client):
             result.headers['Access-Control-Max-Age'] == '86400'
         )  # 24 hours in seconds
 
+    @pytest.mark.xfail(reason='will be fixed in 2325')
```
vytas7, line 89: Heh, this looks like year 2325. But we are going to fix it soon anyway :)

vytas7 (approved):
This looks good for the most of it, thanks!
As discussed, I want to privatize some things related to typing, but we can take a stab at that in a bit more systematic fashion in the cleanup issue.

PR #2311, opened 2024-08-29
chore(cibw): productify the `cibuildwheel` workflow

### LINKED ISSUE #2310: Finalize `cibuildwheel` tooling
The cibuildwheel gate is almost ready to rock 🤘.

However, minor cleanup is still needed before we can get 4.0.0a1 out the door in one click:

- [x] ~~Disable or remove the old wheels workflow~~ :arrow_right: deferred to #2311.
- [x] Publish at least `sdist` to the release page (do we really need to also upload all binaries like now?).
- [x] Make sure `sdist` is uploaded first in a separate step, otherwise `PyPI` and `pip` get confused by a release without `sdist` in the interim period.
- [x] Clean up and adapt the script(s) to check the built wheels.

vytas7:
Examples of a successful run:
* https://github.com/falconry/falcon/actions/runs/10626781301
* https://github.com/falconry/falcon/actions/runs/10629319023 (latest)

I also bumped the tag to `dev2` in order to be able to dispatch a new build to TestPyPI.

Closes #2310.


.github/workflows/cibuildwheel.yaml
```diff
@@ -107,35 +112,94 @@ jobs:
...
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          pattern: cibw-sdist
+          path: dist
+          merge-multiple: true
+
+      - name: Check collected artifacts
+        run: |
+          tools/check_dist.py ${{ github.event_name == 'release' && format('-r {0}', github.ref) || '' }}
```
CaselIT, line 142: You run this twice. I don't think we need to

CaselIT (commented):
Should we also publish a pure python wheel for pypy and other python versions not covered?

.github/workflows/cibuildwheel.yaml
vytas7: I thought it wouldn't hurt, at least it is nice to check it here verifying we have downloaded the correct thing using the action.
I can remove it from the `sdist` step if you think it is redundant?

vytas7:
> Should we also publish a pure python wheel for pypy and other python versions not covered?

This needs discussion I think, I did take this into consideration, but haven't added any `NOTE` yet. Let's discuss this on Gitter/Matrix.

.github/workflows/cibuildwheel.yaml
CaselIT: Maybe it's fine to keep. It's just a fancy ls

vytas7:
> Should we also publish a pure python wheel for pypy and other python versions not covered?

Added to the workflow now.

.github/workflows/cibuildwheel.yaml
```diff
@@ -9,8 +9,14 @@ on:
   # NOTE(vytas): Also allow to release to Test PyPi manually.
   workflow_dispatch:
 
+  # TODO(vytas): Just testing in a PR.
+  pull_request:
+    branches:
+      - master
```
CaselIT, line 15: Remember to remove this

.github/workflows/cibuildwheel.yaml
```diff
@@ -148,14 +164,14 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           files: 'dist/*.tar.gz'
 
-      - name: Publish sdist to TestPyPI
+      - name: Publish sdist and pure-Python wheel to TestPyPI
```
CaselIT, line 167: I think we don't download the pure python wheel.

It's likely fine to just upload that this the other wheels, so let's just revert the name change here and below

CaselIT (commented):
Couple of minor things.

.github/workflows/cibuildwheel.yaml
vytas7: We do actually download it in this step (I think). Let's wait until the Action passes until that point, the emulation is taking a fair bit of time. :)

CaselIT: Oh sorry, I read `files: 'dist/*.tar.gz'` just above and though it was the download artifact action

CaselIT (approved):

PR #2317, opened 2024-09-01
refactor (response): remove deprecated attributes

### LINKED ISSUE #2303: Remove deprecated attributes from Response
Some attribute in response raise attribute errors, and should be removed

aarcex3:
# Summary of Changes

Removed `stream_len` property and setter from `Response` class as both are marked as deprecated.
Also removed the `test_response_removed_stream_len` test and added two new tests, namely `test_response_set_stream` and `test_response_set_stream_with_zero_content_length`.

# Related Issues
Fixes #2303.

# Pull Request Checklist


tests/test_response.py
```diff
@@ -81,3 +74,25 @@ def test_response_option_mimetype_init(monkeypatch):
...
+
+def test_response_set_stream(resp):
+    stream = BytesIO(b'dummy content')
+    content_length = 12
+
+    resp.set_stream(stream, content_length)
+
+    assert resp.stream == stream
+
+    assert resp._headers['content-length'] == str(content_length)
```
vytas7, line 87: It might be better to test via the public property, [`resp.headers`](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Response.headers), instead of relying on the private interface.

tests/test_response.py
```diff
@@ -81,3 +74,25 @@ def test_response_option_mimetype_init(monkeypatch):
     assert ro.static_media_types['.js'] == 'text/javascript'
     assert ro.static_media_types['.json'] == 'application/json'
     assert ro.static_media_types['.mjs'] == 'text/javascript'
+
+
+def test_response_set_stream(resp):
```
vytas7, line 79: Thanks for adding these tests, while I believe we already have tests for `Response.set_stream()`, the more the merrier. 

However, both of these cases look almost identical, would it be possible to DRY a bit using [`pytest.mark.parametrize(...)`](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions)?

E.g.:
```python
@pytest.mark.parametrize('content', [b'', b'dummy content'])
def test_response_set_stream(resp, content): ...
```

falcon/response.py
```diff
@@ -41,10 +41,6 @@
 from falcon.util.uri import encode_check_escaped as uri_encode
 from falcon.util.uri import encode_value_check_escaped as uri_encode_value
 
-_STREAM_LEN_REMOVED_MSG = (
```
vytas7, line 44: It seems that `ruff` is complaining about the import order, probably just a matter of missing or extraneous newline after imports. Could you simply try `ruff check --fix`, and see if it solves the problem?

vytas7 (changes requested):
Thanks for this contribution!
Looks good, just a couple of nitpicks (commented inline) + failing CI that need to be addressed before merging.

tests/test_response.py
```diff
@@ -81,3 +74,14 @@ def test_response_option_mimetype_init(monkeypatch):
...
+
+
+@pytest.mark.parametrize('content', [b'', b'dummy content'])
+def test_response_set_stream(resp, content):
+    stream = BytesIO(content)
+    content_length = len(content)
+
+    resp.set_stream(stream, content_length)
+
+    assert resp.stream == stream
```
vytas7, line 86: Thanks, looks great now bar one small nitpick: it is not very meaningful to compare instances of `BytesIO` this way. Maybe asserting `assert resp.stream is stream` would be more precise?

vytas7 (changes requested):

tests/test_response.py
aarcex3: Done 👍🏻

vytas7 (approved):
LGTM, thanks for contributing! 💯

PR #2324, opened 2024-09-06
fix(WebSocket): handle `OSError` upon `send()` + fix `max_receive_queue == 0`

### LINKED ISSUE #2292: ASGI WebSocket protocol 2.4: `send()` may raise `OSError`
In the ASGI [HTTP & WebSocket protocol](https://asgi.readthedocs.io/en/latest/specs/www.html) version `2.4`, calling `send()` on a closed connection [should raise an `OSError`](https://asgi.readthedocs.io/en/latest/specs/www.html#id2).

See also [Disconnected Client - `send` exception](https://asgi.readthedocs.io/en/latest/specs/www.html#disconnected-client-send-exception):
> If `send()` is called on a closed connection the server should raise a server-specific subclass of `IOError`. This is not guaranteed, however, especially on older ASGI server implementations (it was introduced in spec version 2.4).
>
> Applications may catch this exception and do cleanup work before re-raising it or returning with no exception.
>
> Servers must be prepared to catch this exception if they raised it and should not log it as an error in their server logs.

Do we need to adjust our code? Or do we just bubble this up to the app server?

Note: this issue is aiming to address this only for WebSocket. The same issue in the usual HTTP protocol is tracked as #2323.

See also:
* #1483
* #1808

vytas7:
- [x] Revise and rewrite parts of [Lost Connections](https://falcon.readthedocs.io/en/latest/api/websocket.html#lost-connections).
- [x] Add unit tests (with full coverage) for the `max_receive_queue == 0` case.
- [x] Revise the test client if it could better simulate both behaviours depending on the simulated spec version.
- [x] Add Uvicorn tests in `test_asgi_servers.py` that exercise the `max_receive_queue == 0` case, and catches an error from the webserver's `OSError` upon `send()`.
- [x] Change E2E tests not to use buffering (since we know that use Uvicorn there).

Closes #2292.


vytas7:
Note that I still need to revise the wording of our WebSocket docs, hence **_Draft_** (for now).

CaselIT:
I'm waiting for this to exit draft. let me know if I should review before then

vytas7:
Thanks @CaselIT , I haven't had much time to work on this, there are a couple of items remaining. Code-wise (not tests) I am not planning any more changes, so time permitting you could take a look at that part.
Documentation is ready too I think.

I need to add more tests, and morph newsfragments.

vytas7:
@CaselIT maybe this PR is still not perfect (nothing is), but it is IMHO ready for review!

falcon/asgi/ws.py
```diff
@@ -529,6 +551,25 @@ def _translate_webserver_error(self, ex: Exception) -> Optional[Exception]:
...
+        # NOTE(vytas): Uvicorn 0.30.6 seems to conform to the spec only when
+        #   using the wsproto stack, it then raises an instance of
+        #   uvicorn.protocols.utils.ClientDisconnected.
+        if isinstance(ex, OSError):
+            close_code = None
+
+            # NOTE(vytas): If using the "websockets" backend, Uvicorn raises
+            #   and instance of OSError from a websockets exception like this:
+            #   "received 1001 (going away); then sent 1001 (going away)"
+            if ex.__cause__:
```
CaselIT, line 566: I would try to be less strict here.
Maybe something like:
```
msg = str(ex.__cause__ or ex)
match = _CLIENT_DISCONNECTED_CAUSE.match(str(ex.__cause__))
...
```

we can assume that this may match a wider range of servers

CaselIT (approved):
looks ok, left a suggestion

falcon/asgi/ws.py
vytas7: Hm, good point, but OTOH from what I have been testing other messages looked very different. Also the existing code for Daphne/Autobahn is very specific, so maybe we could start it like this.

Issue #2276, opened 2024-08-17
Update HTTP status code constants wrt RFC 9110


vytas7:
Update our HTTP status code constants (in [`falcon/status_codes.py`](https://github.com/falconry/falcon/blob/master/falcon/status_codes.py)) to conform to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110).

We should be largely up-to-speed, there is at least one known inaccuracy wrt `413 Content Too Large` (see [RFC 9110, Section 15.5.14](https://datatracker.ietf.org/doc/html/rfc9110#name-413-content-too-large)). We still use the older `Payload Too Large`.

VibhinnS:
Hi. I saw the RFC spec sheet and there are a lot of HTTP codes that are removed, one has been reserved and one new code is added. Should I remove their references from `status.rst` file and falcon `__init__` file as well? 

I'm quite new to this so wanted to confirm. Thanks

vytas7:
Hi again @VibhinnS, just checking if you are still working on this issue, and do you need any help?

aarcex3:
Hi, can I take this one if it's still open?

vytas7:
Hi @aarcex3 !
Yes, go ahead! (We haven't heard from @VibhinnS since.)

VibhinnS:
Hi @vytas7! Sorry for being inactive - had interviews and my grandma died 2 weeks ago, so I went off the grid for some time. 
Do you mind if I stay in the community and work on another issue?

CaselIT:
Sorry to hear...
Sure feel free to pick any issue you like, thanks!

vytas7:
Ouch, sad to hear that @VibhinnS :black_heart: 

And no worries at all, you don't owe us anything, come and contribute whenever you like :+1:

aarcex3:
@VibhinnS Im very sorry for your lost.

aarcex3:
@vytas7 question, which link should I use as source? The one from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#information_responses) (it has more info) or the one from [datatracker](https://datatracker.ietf.org/doc/html/rfc9110#name-413-content-too-large), since is the one on the issue description?

vytas7:
@aarcex3 They shouldn't really contradict each other, MDN also includes other RFCs such as WebDAV (which we also include). I don't think we are looking to remove any codes either, let's add the missing ones and sync existing ones according to RFC 9110.

PR #2328, opened 2024-09-12
feat (status_codes): add new status codes

aarcex3:
# Summary of Changes

Added new status codes to conform to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110).

# Related Issues
#2276

# Pull Request Checklist


vytas7 (approved):
Thanks for this improvement @aarcex3!

We cannot close #2276 yet, because you only added new codes, but haven't synced any of the existing ones. However, I can see there already exists https://github.com/falconry/falcon/pull/2330 by @prathik2401 out there that continues on the same topic, so I guess we can merge this one as a standalone increment.

We'll just need to remember to write a newsfragment as part of @2330.

PR #2331, opened 2024-09-17
feat(testing): add `raw_path` to ASGI scope

### LINKED ISSUE #2262: Improve `TestClient` to automatically add `raw_path` to ASGI scope
Thanks to @liborjelinek, the WSGI flavour of the framework does this out of the box as per #2159.
To make things consistent, we ought to add `raw_path` to ASGI scope.

aarcex3:
# Summary of Changes

Added a 'raw_path' key to the asgi scope in `create_scope` to preserve the path and keep consistency as per https://github.com/falconry/falcon/pull/2159

# Related Issues

Closes #2262

# Pull Request Checklist


aarcex3:
I have an issue with mypy, dunno if it's only me. It complains about line 680 in `testing/helpers.py`.

vytas7 (changes requested):
Thanks for this initiative @aarcex3!

However, it seems there is a misunderstanding regarding the actual implementation.
In ASGI, raw path is presented as the [`raw_path`](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) key inside `scope`. The value is a bytestring path, excluding any query string.

Furthermore, I think this improvement warrants a newsfragment, let's add one.

aarcex3:
Hi @vytas7, I made the requested changes. 
Also, what is meant by _newsfragment_? Should I document the changes somewhere?

falcon/testing/helpers.py
```diff
@@ -713,7 +713,10 @@ async def _send(
 
         # NOTE(kgriffs): From the client's perspective, it was a send,
         #   but the server will be expecting websocket.receive
+        # type: Dict[str, Union[bytes, str]]
```
vytas7, line 716: Mypy says that this is invalid syntax, that's the main issue to start with I think.
Moreover, we shouldn't need type annotations in the comments now that we use `from __future__ import annotations` everywhere.

vytas7 (changes requested):
It seems that some other gates are failing too, not just Mypy.

tests/asgi/test_testing_asgi.py
```diff
@@ -153,3 +153,14 @@ def test_immediate_disconnect():
 
     with pytest.raises(ConnectionError):
         client.simulate_get('/', asgi_disconnect_ttl=0)
+
+
+def test_create_scope_preserve_raw_path():
+    path_no_queries = '/cache/http%3A%2F%2Ffalconframework.org/status'
```
vytas7, line 159: It would be nice if we could reduce the amount of copy-paste in this test case by leveraging `pytest.mark.parametrize()`.

tests/asgi/test_testing_asgi.py
vytas7, line 1: See also the relevant test case in `tests/test_recipes.py`. It should be possible to update it to use and test the new functionality from this changeset.

vytas7 (changes requested):
Thanks for this improvement, it seems to function as intended now!

I have a couple of nitpicks wrt testing (commented inline); and we also need a news fragment for this issue. The file should be named `docs/_newsfragments/2262.newandimproved.rst`.

Unfortunately, we have no bandwidth to help shaping up this PR for the impending beta release that we are trying to haste in order to catch up with CPython 3.13, so we'll have to postpone this PR to Falcon 4.1. No specific action is required from you wrt this point, this is just for reference, that we will not be able to merge this PR until 4.0.0 stable is out.

vytas7 (dismissed):
Thanks once again for this improvement @aarcex3!
I have tweaked the proposed docs a bit more to my liking, and removed minor items from `test_recipes.py` that were obsoleted by your change.

vytas7 (approved):

Issue #2207, opened 2024-01-31
TestClient methods are typed to return _ResultBase, but actually return Result


davetapley:
All these methods are typed to return `_ResultBase`:

https://github.com/falconry/falcon/blob/4910dd73ecd1b9c8cf6cae045b26ad432fa56128/falcon/testing/client.py#L1041-L1046

But actually returns `Result` here:
https://github.com/falconry/falcon/blob/4910dd73ecd1b9c8cf6cae045b26ad432fa56128/falcon/testing/client.py#L638

And `Result` defines the methods which are useful in testing:
https://github.com/falconry/falcon/blob/4910dd73ecd1b9c8cf6cae045b26ad432fa56128/falcon/testing/client.py#L328-L348

---

But because it's typed `_ResultBase` I get e.g. `Cannot access member "json" for type "_ResultBase"`, 
even though the methods work 😬 

I _presume_ this is an oversight and not intentional?

CaselIT:
That seems likely yes. PR to fix would be appreciated!

CaselIT:
it seems doable, with a lot more effort

davetapley:
@CaselIT I agree! See https://github.com/falconry/falcon/discussions/2210 😁

PR #2338, opened 2024-09-23
feat(typing): finalize annotations

CaselIT:
Finish typing falcon.
This also fixes https://github.com/falconry/falcon/issues/2207


examples/ws_tutorial/ws_tutorial/app.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
vytas7, line 1: I think we should remove this for now, this tutorial app is not typed at all anyway.

CaselIT: it has some types, for example:
`async def process_request_ws(self, req: Request, ws: WebSocket):`

vytas7: Aha, checked quickly and didn't see many.
Regardless, as a user, why would I need this? I think it is useful for our own codebase, but users can use the standard mode too if they wish so?

CaselIT: I'll add a skip in the file then

vytas7 (dismissed):
Looks good for the most of it, I'll add a note to the cleanup issue to revise `Headers` though. :+1:

vytas7 (approved):

Issue #2313, opened 2024-08-30
Typing cleanup for 4.0


CaselIT:
Once all the typing of the code has been completed, make a final pass to:
- [ ] normalise names
- [x] decide what type alias should be public and which should not
- [x] add future annotattion in any missed package (there should be a ruff code we can use for it)
- [x] should JSON handlers' main argument be typed as `JSONSerializable` or `Any`?
- [x] The argument type of [`secure_filename()`](https://github.com/falconry/falcon/blob/master/falcon/util/misc.py#L375) has recently been changed to `Optional[str]`, whereas before it was just `str`, which I (@vytas7) believe is correct, although it fails some type checks; should we maybe revise these places instead , or alternatively type-ignore them if it is strictly internal use?
- [x] Revise the typing of headers, the latest iteration reads
```python
Headers = Dict[str, str]
HeaderList = List[Tuple[str, str]]
HeaderArg = Union[Headers, HeaderList]
```
If `HeaderArg` is what is used most commonly, maybe it should be just called `Headers`?
Also, we define `HeaderList` as iterable of tuples in the docstrings, but here it is just called `List[...]`. Will Mypy understand that, or should we use `Iterable[...]` instead?

vytas7:
I think we might want to privatize the whole `typing` module, i.e. call it `falcon._typing` instead of `falcon.typing`, to make it obvious that we don't make any guarantees.

@CaselIT told me that SQLAlchemy did something similar, and they re-export selected names via a public module.

PR #2339, opened 2024-09-25
chore: typing cleanup

CaselIT:
Fixes: #2313

Cannot really came up with many type aliases that may be nice to have public.
For not I've found only 3. https://github.com/falconry/falcon/pull/2339/files#diff-59734eebd58dfd39eafe7094e8725d6d5b3847156d7d0bff3cbdf5095fcf84e4

Also have yet to look at naming.


CaselIT:
Not sure if we could provide a `Request` subclass there that's generic on the media?
So people can type their method as `def on_get(self, req: Request[list[int]], resp: Response)`

falcon/_typing.py
```diff
@@ -0,0 +1,198 @@
...
+    TypeVar,
+    Union,
+)
+
+# NOTE(vytas): Mypy still struggles to handle a conditional import in the EAFP
+#   fashion, so we branch on Py version instead (which it does understand).
+if sys.version_info >= (3, 11):
+    from wsgiref.types import StartResponse as StartResponse
+    from wsgiref.types import WSGIEnvironment as WSGIEnvironment
+else:
```
vytas7, line 46: The `else:` branch looks like a genuine coverage miss. Should we add a pragma, or do we need to update `.coveragerc` to omit `falcon/_typing.py` instead of `falcon/typing.py`?

vytas7 (changes requested):
There is an issue with coverage.

I'll try to skim through other changes today too.

CaselIT:
> There is an issue with coverage.

I believe it's its usual issues, but I haven't checked it

falcon/_typing.py
CaselIT: don't we collect coverage for <3.11?

vytas7: No, we don't.

I think you simply need to update the "omit" section in `.coveragerc` to omit `_typing.py`, not `typing.py`.

CaselIT: well we can add a ignore comment in the else. will do

vytas7: If that's all that's needed for the whole file, then remove the file from coveragerc omissions.

CaselIT: oh I think the issue was that the protocol are not liked by coverage. that's why it was added

vytas7: Oh yeah, that was another reason!

vytas7: I have update `.coveragerc` in the web UI. Let's see how it goes now.

falcon/stream.py
```diff
@@ -166,9 +162,21 @@ def exhaust(self, chunk_size: int = 64 * 1024) -> None:
...
 
-    is_exhausted = eof
+    @property
+    # NOTE(caselit): Deprecated long ago. Warns since 4.0.
+    @deprecated(
+        'Use `eof` instead. '
+        '(This compatibility alias will be removed in Falcon 5.0.)',
+        is_property=True,
+    )
+    def is_exhausted(self) -> bool:
```
vytas7, line 177: This change has no coverage.

vytas7 (changes requested):
Thanks a lot for the whole typing effort, this is almost ready to rock.

We just need to address missing coverage, and make a decision on `UNSET` before merging.

We could probably bikeshed some names too, but since they are private under `_typing`, we can revise them later at any time.

falcon/media/multipart.py
```diff
@@ -80,9 +80,9 @@ class BodyPart:
 
     _content_disposition: Optional[Tuple[str, Dict[str, str]]] = None
     _data: Optional[bytes] = None
-    _filename: MissingOr[Optional[str]] = MISSING
-    _media: MissingOr[Any] = MISSING
-    _name: MissingOr[Optional[str]] = MISSING
+    _filename: UnsetOr[Optional[str]] = UNSET
```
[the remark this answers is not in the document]
CaselIT: how strange. I was convinced I renamed this to `_UNSET`. may have inadvertently does a ctrl-z

falcon/media/multipart.py
```diff
@@ -232,6 +232,8 @@ def secure_filename(self) -> str:
 
         See also: :func:`~.secure_filename`
         """  # noqa: D205
+        if self.filename is None:
```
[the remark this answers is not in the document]
vytas7: Yes, but `filename or ''` is nearly a no-op performance-wise if `filename` is a truthy value (the normal/expected case).

I would prefer to keep `secure_filename()` documented as accepting only a string argument. Such typing annotation would help people find errors if they pass `None`.

Also, if we allow `str | None`, wouldn't we need to also change the implementation of `secure_filename()`? Because IIRC Mypy is quite dumb when it comes to such checks `if value:`, it used to think that this could still lead to `None`.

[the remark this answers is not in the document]
CaselIT: Actually `is/is not None` check is usually faster than the `__bool__` check required by the or. Overall i'm fine with either case here 

No `if not value: raise` is enough for mypy to understand that after only str is a valid type.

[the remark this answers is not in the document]
vytas7: In the cythonized code, you are probably right, but if not, I was surprised this was not the case. Also, it is not enough to do `is not None` or `is None`, because you still need to check for the empty string case in one or another place.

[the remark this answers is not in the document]
vytas7: In pure Python 3.12:
```python
>>> a = None
>>> b = True
>>> def func1():
...     return a is None and b is not None
... 
>>> func1()
True
>>> def func2():
...     return not a and b
... 
>>> import timeit
>>> timeit.timeit(func1, number=int(1e8))
3.7819362650625408
>>> timeit.timeit(func1, number=int(1e8))
3.7773449880769476
>>> timeit.timeit(func1, number=int(1e8))
3.8308213039999828
>>> timeit.timeit(func2, number=int(1e8))
3.569090647972189
>>> timeit.timeit(func2, number=int(1e8))
3.556459318031557
>>> timeit.timeit(func2, number=int(1e8))
3.6141516310162842
```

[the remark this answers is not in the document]
CaselIT: Usually. int are fast. for an arbitrary class `is` is faster

vytas7 (approved):
Looks great now :100: 
(Potentially bar naming refinement, but since these are strictly private under `_typing`, we can improve them at any time.)

Issue #2343, opened 2024-09-27
Remove support for `is_async` without a deprecation window


vytas7:
Normally, we always give an adequate deprecation window when removing features, even if it is a major release.

However, the `is_async` parameter (used for [hooks](https://falcon.readthedocs.io/en/stable/api/hooks.html#before-hooks) and [media validation](https://falcon.readthedocs.io/en/stable/api/media.html#falcon.media.validators.jsonschema.validate)) represents an edge case where I very much doubt that anyone from the community has ever needed this flag.

What is more, very easy workarounds exist: wrapping the method in question inside an `async def`, or even easier, simply using [`inspect.markcoroutinefunction()`](https://docs.python.org/3/library/inspect.html#inspect.markcoroutinefunction) (on CPython 3.12+) to mark the method as such. In addition, Cython 3.0 (on CPython 3.10+) no longer needs this flag, as it is able to convey the information to the interpreter via setting the appropriate code flags.

PR #2344, opened 2024-09-27
chore: remove `is_async` support

CaselIT:
Fixes: #2343


vytas7 (approved):
LGTM bar the newsfragment that probably needs minor polishing, but I can take a look in the alpha PR (and we are even not rendering the newsfragments out yet).

PR #2345, opened 2024-09-27
feat: type hooks with `ParamSpec`

CaselIT:
Follow up of #2344 that types hook with paramspec.

Will merge this after 4.0 is out

### DIFF
--- a/docs/_newsfragments/2343.breakingchange.rst
+++ b/docs/_newsfragments/2343.breakingchange.rst
@@ -1,0 +1,5 @@
+Removed ``is_async`` argument from :meth:`~falcon.media.validators.jsonschema.validate`
+and the hooks :meth:`~falcon.before` and :meth:`~falcon.after` since it's
+no longer needed.
+Cython from 3.0 will correctly mark ``asnyc def`` as coroutine, making
+this argument no longer useful.
--- a/falcon/hooks.py
+++ b/falcon/hooks.py
@@ -27,7 +27,6 @@ from typing import (
     cast,
     Dict,
     List,
-    Protocol,
     Tuple,
     TYPE_CHECKING,
     TypeVar,
@@ -38,6 +37,12 @@ from falcon.constants import COMBINED_METHODS
 from falcon.util.misc import get_argnames
 from falcon.util.sync import _wrap_non_coroutine_unsafe
 
+try:
+    from typing import Concatenate, ParamSpec
+except ImportError:
+    from typing_extensions import Concatenate  # type: ignore[assignment]
+    from typing_extensions import ParamSpec  # type: ignore[assignment]
+
 if TYPE_CHECKING:
     import falcon as wsgi
     from falcon import asgi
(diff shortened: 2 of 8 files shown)


vytas7:
It is not decided yet whether we want to do this in 4.1, or even further down the road. But we can keep this PR open for a while.

vytas7:
@CaselIT could you resolve conflicts in this PR in order to focus solely on `ParamSpec`?
AFAICT we have actually released parts of it in 4.0, for instance, dropped `is_async`.

I would like to keep runtime compatibility with 3.8, but it seems we can import stuff we need under `TYPE_CHECKING`, and even use most of the newer syntax under `from __future__ import annotations`, so it doesn't really conflict with 3.8 runtime support.

vytas7:
I am going to close this PR for now since the major part of it has already been merged as #2344.
It is will probably be easier to spell the concatenate part anyway once we drop older typing syntax etc.

CaselIT:
I missed your previous message, sorry.  will fix the merge issues and make it merge-able

vytas7:
Sure, that works too! I was thinking we could merge the generic App & middleware first, then I did some prototyping, hooks should be easy too with `Concatenate` and `ParamSpec`.

docs/_newsfragments/2343.breakingchange.rst
```diff
@@ -0,0 +1,5 @@
+Removed ``is_async`` argument from :meth:`~falcon.media.validators.jsonschema.validate`
```
CaselIT: we can likely remove this changelog, since it's wasn't done here

vytas7: Yes, it belongs to the already merged #2344.

CaselIT:
> Sure, that works too! I was thinking we could merge the generic App & middleware first, then I did some prototyping, hooks should be easy too with `Concatenate` and `ParamSpec`.

they seem to be independent from each other, since before/after don't use the app

pyproject.toml
```diff
@@ -9,7 +9,9 @@ requires = [
 name = "falcon"
 readme = {file = "README.rst", content-type = "text/x-rst"}
 dynamic = ["version"]
-dependencies = []
+dependencies = [
+    "typing-extensions >= 4.2.0; python_version<'3.10'",
```
vytas7: Let's not add this dependency, it is not really needed if we import what we need under `if TYPE_CHECKING:`.
We have already documented that 3.9 support is on a best of effort basis, and only for runtime, type checking does not work. And 3.8 we'll drop I think.

docs/_newsfragments/2343.breakingchange.rst
CaselIT: Ok, removing it.

vytas7:
> > Sure, that works too! I was thinking we could merge the generic App & middleware first, then I did some prototyping, hooks should be easy too with `Concatenate` and `ParamSpec`.
> 
> they seem to be independent from each other, since before/after don't use the app

True. They might be interdependent if I include the `Resource` typing in my PR though. But that can be a separate PR.

pyproject.toml
CaselIT: I'm not sure we can use paramspec without them. unless we move everything under type ignore.

honestly I don't understand why we should complicate our lives like that...

vytas7: I just don't think you should include that dependency, we don't guarantee any 3.9 support outside of runtime.

You can still do conditional import, but I think Mypy understand comparisons against the Python version best.

I.e. I mean if we are under (3, 10), then you can try to import `typing_extensions` anyway. Running Mypy on older Python already installs it anyway.

falcon/hooks.py
```diff
@@ -109,7 +57,20 @@ def __call__(
 )
 
 
-def before(action: BeforeFn, *args: Any, **kwargs: Any) -> Callable[[_R], _R]:
+def before(
+    action: Union[
+        Callable[
+            Concatenate[wsgi.Request, wsgi.Response, Resource, Dict[str, Any], _FN],
```
vytas7, line 63: Since it passes Mypy, we can merge it for now, but beware this `Resource` is not really the correct typing here.

IIRC as I remember from my prototyping, it works here only because `Resource` is defined as `object` which is basically almost equivalent to `Any` (with some exceptions).

falcon/hooks.py
```diff
@@ -39,68 +38,17 @@
...
-        req: asgi.Request,
-        resp: asgi.Response,
-        resource: Resource,
-        *args: Any,
-        **kwargs: Any,
-    ) -> Awaitable[None]: ...
-
-
-AfterFn = Union[SyncAfterFn, AsyncAfterFn]
 _R = TypeVar('_R', bound=Union['Responder', 'Resource'])
```
vytas7, line 52: I meant my comment about `Resource` actually applies here even more so.

Saying that your `TypeVar` is bound to `object` is almost equivalent to not having any bounds.

docs/_newsfragments/2343.breakingchange.rst
vytas7: A slightly tweaked version can already be found above this headline -- https://falcon.readthedocs.io/en/stable/changes/4.0.0.html#new-improved

falcon/hooks.py
CaselIT: sure, but we don't have a base class for resources, so not sure how else we can express this

vytas7: Yes, I meant I just noticed when I replaced `Resource` with a union of protocols, this logic broke, I was forced to restore it back to `object` to make it work in the current form.

falcon/hooks.py
CaselIT: Why not? I don't see what other type we could assign here

vytas7: I had a prototype where `Resource` was a union of signatures similar to middleware, and IIRC it worked just fine with the exception of hooks. But probably with `ParamSpec` and `Concatenate` we can make that work too.

PR #2346, opened 2024-09-28
fix(test_static.py): fix Windows CPython 3.13 build

vytas7:
The codebase now builds and passes tests on Windows and CPython 3.13: https://github.com/falconry/falcon/actions/runs/11082068662/job/30794594018.


CaselIT:
Not sure why the change ls fixes it, but it looks fine

CaselIT (approved):

vytas7:
> Not sure why the change ls fixes it, but it looks fine

Because we use a specialized `client` fixture that does some absolute path mangling for CPython 3.13 on Windows. Same issue as before.

CaselIT:
Oh interesting. Ok

Issue #2318, opened 2024-09-02
Misc documentation cleanup


vytas7:
Documentation improvements and tweaks that are not highest prio, but would be nice to iron out:
- [ ] https://falcon.readthedocs.io/en/latest/community/contributing.html#reviews :arrow_left: checkboxes within use the GitHub flavour of Markdown, which does not render exactly the same in the MyST Markdown flavour that Sphinx uses. Check if there is a suitable plugin, or, in the worst case, just convert `CONTRIBUTING.md` to ordinary bullet lists.

nithin-sudarsan:
@vytas7 since https://falcon.readthedocs.io/en/latest/community/contributing.html#reviews is a list of points that a reviewer will look for while reviewing a PR, it makes more sense to have them as bullet lists rather than checkboxes (similar to https://falcon.readthedocs.io/en/latest/community/contributing.html#code-style-rules)

nithin-sudarsan:
@vytas7 please check the above PR and I would appreciate it if you could take some time to review the changes and provide feedback when convenient.
Thank you!

PR #2351, opened 2024-10-01
fix(contributing): improve checklist

nithin-sudarsan:
# Summary of Changes

Replaced checklist under https://falcon.readthedocs.io/en/latest/community/contributing.html#reviews with bulleted list.

# Related Issues

#2318

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,16 +53,16 @@ Falcon is used in a number of mission-critical applications and is known for its
 
 Project maintainers review each PR for the following:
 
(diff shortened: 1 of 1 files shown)


vytas7 (changes requested):
Thanks for looking into this @nithin-sudarsan!

In a pinch, we could simply change to a list, but in this issue we were first of all looking to render something that looks like checkboxes in Sphinx. Could you explore the alternatives?

nithin-sudarsan:
Thank you for the feedback @vytas7!

I will look further into this to see alternative ways.

PR #2353, opened 2024-10-04
chore: clean up `tox` environments & CI gates

vytas7:
I will change the branch protection rules later if we are OK to proceed with this PR.
That is actually one of the major reasons behind this change, I don't want to keep rewriting all the rules every time we change the default Python version.


.github/workflows/test-other.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+on:
+  push:
+    branches:
+      - "*"
+
+jobs:
+  run-tox:
+    name: tox -e ${{ matrix.toxenv }}
+    runs-on: ubuntu-20.04
```
CaselIT, line 12: can't we use a more modern version? like 22 or 24?

Mainly because I don't know how long will github keep 20 around, since in 6 month id will be unsupported

.github/workflows/tests.yaml
```diff
@@ -11,104 +12,73 @@ on:
...
-            os: ubuntu-latest
-            toxenv: py311_cython
-          - python-version: "3.12"
-            os: ubuntu-latest
-            toxenv: py312
-          - python-version: "3.12"
-            os: ubuntu-latest
-            toxenv: py312_cython
-          - python-version: "3.12"
+        tox:
```
CaselIT, line 25: neat. I did not know you could pass objects here

tox.ini
```diff
@@ -229,12 +196,33 @@ commands = {[with-cython]commands}
...
        gunicorn
-       # NOTE(caselit): meinheld requires greenlet<0.5 that's not compatible with py3.10
-       meinheld
        uwsgi
        waitress
 commands = pytest -v tests/test_wsgi_servers.py
 
+[testenv:wsgi_meinheld]
+# NOTE(caselit): meinheld requires greenlet<0.5 that's not compatible with py3.10
+basepython = python3.8
```
CaselIT, line 205: could think about using 3.9 since 3.8 will be eol soon

CaselIT (approved):
left some minor comments, mainly about versions

tox.ini
vytas7: I'm not sure if Meinheld works with 3.9. I think those old benchmarks unfortunately don't, so I kept everything on 3.8 for now.

.github/workflows/test-other.yaml
vytas7: I can check, but unfortunately I don't think so. We left on 20.04 on purpose when were were upgrading to 22.04.
In a pinch, you can probably compile from source...

.github/workflows/tests.yaml
vytas7: Already using this in CIBW and elsewhere :)

tox.ini
vytas7: (These tests will only be run on master merges, not PR.)

.github/workflows/test-other.yaml
vytas7: Hm, `libunwind-dev` itself seems to work just fine. :thinking: 
I can try a combo of Ubuntu 22.04 and old Python.

vytas7: I'll merge this one, and try to experiment in a new PR.

CaselIT: my suggestion is just to having having to do stuff in 6 month when we don't remember what this is doing

vytas7: That might make sense :smiling_imp: But we can start to demoting stuff to only be run on master merge.

PR #2090, opened 2022-08-04
chore: remove deprecated api_helpers module and the deprecated body attrs

kgriffs:
# Summary of Changes

Removed deprecated api helpers module and body attibutes.

# Related Issues

Supersedes #1986 

# Pull Request Checklist

This is just a reminder about the most common mistakes.  Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](../CONTRIBUTING.md) at least once; it will save you a few review cycles!

If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing to do.

- [x] Applied changes to both WSGI and ASGI code paths and interfaces (where applicable).
- [x] Added **tests** for changed code.
- [x] Prefixed code comments with GitHub nick and an appropriate prefix.
- [x] Coding style is consistent with the rest of the framework.
- [x] Updated **documentation** for changed code.
    - [x] Added docstrings for any new classes, functions, or modules.
    - [x] Updated docstrings for any modifications to existing code.
    - [x] Updated both WSGI and ASGI docs (where applicable).
    - [x] Added references to new classes, functions, or modules to the relevant RST file under `docs/`.
    - [x] Updated all relevant supporting documentation files under `docs/`.
    - [x] A copyright notice is included at the top of any new modules (using your own name or the name of your organization).
    - [x] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/usage/restructuredtext/directives.html?highlight=versionadded#directive-versionadded).
- [x] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/latest/quickstart.html#creating-news-fragments) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)

If you have *any* questions to *any* of the points above, just **submit and ask**! This checklist is here to *help* you, not to deter you from contributing!

*PR template inspired by the attrs project.*

PR #2361, opened 2024-10-09
refactor: restore `AttributeRemovedError` for `resp.body`

vytas7:
The blocked `resp.body` property was introduced in #2090, and then removed later in the 4.0 cycle by oversight.

It is extremely useful to have it, since otherwise setting `resp.body` in legacy code silently does nothing, and it is very hard to understand what is going on.

### DIFF
--- a/falcon/response.py
+++ b/falcon/response.py
@@ -210,6 +210,20 @@ class Response:
         self.status = value
 
     @property
+    def body(self) -> NoReturn:
+        raise AttributeRemovedError(
+            'The body attribute is no longer supported. '
+            'Please use the text attribute instead.'
+        )
+
+    @body.setter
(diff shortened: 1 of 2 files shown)


CaselIT (approved):
So this wasn't in 3.0?

vytas7:
The attribute was already deprecated in 3.x, but the `AttributeRemovedError` has never been part of a stable release.

vytas7:
I am going to disable the Firefox gate for now. I haven't found the reason yet, but I suspect something is not working correctly with `geckodriver` on Linux and the newest Firefox 131.0. I won't have the bandwidth to dig into this.

Locally, it works for me, but I cannot make the browser headless anymore, it always pops up. In the gate, 1/2 tests fail.

PR #2371, opened 2024-10-14
chore(RtD): enable PDF build format

AkshayAwate:
Enabled PDF generation on readthedocs.


CaselIT (approved):

vytas7 (approved):
Thanks @AkshayAwate! :+1:

Unfortunately, it is hard to test this without actually running it on RtD, but let's merge and see how this fares!

vytas7:
It is not showed for unstable releases, but it is possible to download just fine, thanks once again!
https://falcon.readthedocs.io/_/downloads/en/latest/pdf/

PR #2375, opened 2024-10-15
refactor: tweak type annotations to make Pyright pass

jkmnt:
# Summary of Changes

This is not intended to be merged as is, I'm not that familiar with the falcon codebase.

The code is passing the Pyright checks in basic mode. 

In standard mode there are a few more errors. These are mostly related to the unbound variables defined only in some branch of if/else. Pyright is strict here and insist for variables to be always defined. I "fixed" one of these errors in CompiledRouter.

I run the tests. All of them are either green or skipped (?).

The type for headers of falcon.errors was made more strict: `dict[str,str] | list[tuple[str, str]` instead of `Mapping[str, str] | Iterable[tuple[str, str]]`. The catch is that dict is iterable too, so dict keyed by tuple of two strings (i.e. `dict[tuple[str, str], Any]`) is accepted, while should not.

I moved _logger to the separate module. The _logger was used in falcon.asgi/app, but root falcon module was never imported there.

I replaced `hasattr()` applied to some optional methods with `getattr()`. Pyright is strict here too: `hasattr` success means nothing :-)

### DIFF
--- a/falcon/__init__.py
+++ b/falcon/__init__.py
@@ -23,6 +23,8 @@ the framework's classes, functions, and variables::
     app = falcon.App()
 """
 
+import logging as _logging
+
 __all__ = (
     # API interface
     'API',
@@ -407,11 +409,6 @@ from falcon.hooks import after
 from falcon.hooks import before
 from falcon.http_error import HTTPError
 from falcon.http_status import HTTPStatus
-
-# NOTE(jkmnt): Moved logger to leaf module to avoid possible circular imports.
-# the _logging symbol is reexported too - maybe it was used by test or smth.
-from falcon.logger import _logger
-from falcon.logger import logging as _logging
 from falcon.middleware import CORSMiddleware
 from falcon.redirects import HTTPFound
 from falcon.redirects import HTTPMovedPermanently
@@ -644,3 +641,8 @@ from falcon.util import wrap_sync_to_async_unsafe
 
 # Package version
 from falcon.version import __version__  # NOQA: F401
+
+# NOTE(kgriffs): Only to be used internally on the rare occasion that we
+#   need to log something that we can't communicate any other way.
+_logger = _logging.getLogger('falcon')
+_logger.addHandler(_logging.NullHandler())
--- a/falcon/__init__.py
+++ b/falcon/__init__.py
@@ -407,6 +407,11 @@ from falcon.hooks import after
 from falcon.hooks import before
 from falcon.http_error import HTTPError
 from falcon.http_status import HTTPStatus
+
+# NOTE(jkmnt): Moved logger to leaf module to avoid possible circular imports.
+# the _logging symbol is reexported too - maybe it was used by test or smth.
+from falcon.logger import _logger
+from falcon.logger import logging as _logging
 from falcon.middleware import CORSMiddleware
 from falcon.redirects import HTTPFound
 from falcon.redirects import HTTPMovedPermanently
@@ -637,10 +642,5 @@ from falcon.util import uri
 from falcon.util import wrap_sync_to_async
 from falcon.util import wrap_sync_to_async_unsafe
 
-# NOTE(jkmnt): Moved logger to leaf module to avoid possible circular imports.
-# the _logging symbol is reexported too - maybe it was used by test or smth.
-from falcon.logger import _logger, logging as _logging
-
 # Package version
 from falcon.version import __version__  # NOQA: F401
-
--- a/falcon/__init__.py
+++ b/falcon/__init__.py
@@ -23,8 +23,6 @@ the framework's classes, functions, and variables::
     app = falcon.App()
 """
 
-import logging as _logging
-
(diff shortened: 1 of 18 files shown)


falcon/__init__.py
```diff
@@ -23,8 +23,6 @@
     app = falcon.App()
 """
 
-import logging as _logging
```
CaselIT, line 26: I don't understand this change. It's not related to types, so could you revert?

falcon/_typing.py
```diff
@@ -104,6 +104,9 @@ async def __call__(
 HeaderMapping = Mapping[str, str]
 HeaderIter = Iterable[Tuple[str, str]]
 HeaderArg = Union[HeaderMapping, HeaderIter]
+
+NarrowHeaderArg = Union[Dict[str,str], List[Tuple[str, str]]]
```
CaselIT, line 108: I don't understand why we need this. it's more specific that `HeaderArg`

CaselIT (changes requested):
Thanks for the change!

I haven't looked at the complete change yet, but I would avoid doing non-typing related changes, like the logger change.

will have another look later today.

falcon/__init__.py
jkmnt: The reason for this change: the logger is used in asgi.app like that:
https://github.com/falconry/falcon/blob/61c1047db2ef02556da9c28cc42c813b52db3aa9/falcon/asgi/app.py#L1188
but there is no 'import falcon' anythere is this file, so _logger in fact can't be accessed. I think it's the bug actually.
If I add 'import falcon' at top of 'asgi.app', it could lead to circular import if asgi is ever imported in main falcon 
package (now it seems to be not). Should I change it to just `from falcon import _logger` to the top of `asgi.app` ?

falcon/_typing.py
jkmnt: Ok, the reason is this code: 

```python
def _load_headers(headers: Optional[HeaderArg]) -> Headers:
    """Transform the headers to dict."""
    if headers is None:
        return {}
    if isinstance(headers, dict):
        return headers
    return dict(headers)
```

The header args were wide-typed as either Mapping[str, str] or Iterable[tuple[str, str]]. The isinstance(..., dict) is not enough to tell Mapping from Iterable[tuple]. I.e. `for k,v in {('foo', 'bar') : 42, ('a', 'b'): 111}` yields perfect `('foo', 'bar'), ('a', 'b')` pairs matching the Iterable[tuple[str, str]]. That's the pyright complain. 

The docs state the arguments to errors must be list or dict (not list-like or dict-like), I thought it would be the right thing to match the typing to runtime.

CaselIT: this seems overly pedantic. it's likely fine to narrow `HeaderArg` to use `Sequence[Tuple...]` that should solve the complain

jkmnt: I changed the the NarrowHeaderArg to Union[Mapping[], Sequence[]], pyright is ok with it.
Changing the very HeaderArg from Iterable to Sequence is wrong imho, since it's used in `set_headers`.
The docs state the set_headers accepts iterable of tuples but sequence will disallow using generators there.

falcon/__init__.py
CaselIT: I think it's liked by by python and type checker because there is this import in that file:https://github.com/falconry/falcon/blob/61c1047db2ef02556da9c28cc42c813b52db3aa9/falcon/asgi/app.py#L51

CaselIT: but let's import it explicitely

jkmnt: Moved the _logger back to the main __init__.py.
It's explicitly imported in asgi.app now

CaselIT:
I had also made some changes. Will create a new PR

CaselIT:
I've pushed them here https://github.com/falconry/falcon/pull/2379

Everything should pass there. Could you take a look @jkmnt ?

thanks for getting the ball rolling!

jkmnt:
No complains from pyright on my check in both standart and basic modes.
Personally I'm not a big fan of silencing the typechecker, but in internal code it's ok.

I think
```python
meth: Anno = getattr(obj, 'meth_name', None); 
if meth:
    meth()
```

vs

```python
 if hasattr(obj, 'meth_name'):
     ob.meth_name()
```

is still better from both the typecheck and performance standpoints: one attribute lookup vs two.
Though the `try: obj.meth(); except ...` if even faster )

> I've pushed them here #2379
> 
> Everything should pass there. Could you take a look @jkmnt ?
> 
> thanks for getting the ball rolling!

CaselIT:
I agree getatrr. I will cherry pick your chane in the other one

vytas7:
I'm going to close these Pyright prototypes for now, it seems that it is hard to make Pyright pass without many intrusive changes, and/or maintaining a large number of ignores. Maybe we could start testing Pyright on sample apps (i.e , from the user's perspective) as first step.

PR #2379, opened 2024-10-18
refactor: add pyright typecheck

CaselIT:
Follow up of #2375

### DIFF
--- a/requirements/tests
+++ b/requirements/tests
@@ -13,7 +13,7 @@ testtools; python_version < '3.10'
 aiofiles
 httpx
 uvicorn >= 0.17.0
-websockets
+websockets >= 13.1
 
 # Handler Specific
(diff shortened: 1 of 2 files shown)


vytas7:
Does this contain #2375, or builds on top of it?
I'll probably focus on getting the stable version out first, will try to circle back on Pyright once we open 4.1.0 development.

CaselIT:
It contains it but the last couple of commits that were done at the same time. Hence the new PR

vytas7:
I'm going to close these Pyright prototypes for now, it seems that it is hard to make Pyright pass without many intrusive changes, and/or maintaining a large number of ignores. Maybe we could start testing Pyright on sample apps (i.e , from the user's perspective) as first step.

Issue #2372, opened 2024-10-15
Improve typing of custom req/resp type


vytas7:
### Discussed in https://github.com/falconry/falcon/discussions/2370

<div type='discussions-op-text'>

<sup>Originally posted by **jkmnt** October 14, 2024</sup>
Testing f4 beta with the our falcon-powered app. No runtime issues so far, but a few errors from the typechecker.

The `app.add_error_handler` and `app.add_error_handler` expects the callables with signature of  [falcon.Request, falcon.Response, ...etc]. Our app is using the custom Request class:
```python
falcon.App(request_type=MyRequest)

def my_error_handler(req: MyRequest, ...): ...

```

Maybe the request/response types should be typed as generics bound to the base type ?
Something like
```
TReq = TypeVar('TReq', bound=Request)

class App(Generic[TReq, TResp]):
    def __init__( self,  request_type: Optional[type[TReq]] = None, ...): ...

    def add_error_handler(self, exception, handler: Callable[[TReq, TResp, TCtx, ...], None]): ...
```

The same issue is with `before`/`after` hook adders . Yes, it's way harder to type since they know nothing about the app class.</div>

CaselIT:
> BTW, there was some (missed by mypy) errors reported by pyright. I think I'll try to fix some and submit a separate PR.

that would be great. Feel free to also open an issue when you do that.
also if you know tox and github action we could add a gate with it too, so that it stays tested

vytas7:
I have just created an issue to track the Pyright gate: #2373.

jkmnt:
Another thing related to the typing:
I noticed the middleware type is just the `object`. It's the user-facing interface so imho typing should be better.

I suggest of something like
```python
class MwWithReqHandler(Protocol):
    def process_request(self, req: Request, resp: Response) -> None: ...

class MwWithRespHandler(Protocol):
    def process_response(self, req: Request, resp: Response, resource: object, req_succeeded: bool) -> None: ...

class MwWithProcessHandler(Protocol):
    def process_resource(self, req: Request, resp: Response, resource: object, params: dict[str, Any]) -> None: ...

Middleware = Union[MwWithReqHandler, MwWithRespHandler, MwWithProcessHandler]
```

This is not perfect but better than nothing. The bad thing is that it's enough to have one conforming (with others nonconformin) methods to pass the type check. The good thing: the user may inherit it's middleware class from these protocols (kind of abc but w/o runtime overhead) and get full typesafety.

```python
[...]

PR #2390, opened 2024-10-26
feat(typing): add middleware protocols

jkmnt:
# Summary of Changes

Typed the middleware classes with union of protocols.
Middleware class may conform to any protocol of the union to pass the type check

It's not perfect  due the nature of python's static duck typing, but better than the plain `object` ) Class may have correct signature for `process_request()`, but wrong signature for `process_response()`. It would be accepted since `MiddlewareWithProcessRequest`  says nothing about the `process_response`.

But if user explicitly states implemented protocols, then the typecheker will ensure strict conformance.
```python
class MyMiddleware(MiddlewareWithProcessRequest, MiddlewareWithProcessResponse):
    def process_request(...): ... # ok, pass
    def process_responce(...):  # won't pass (typo). process_response is required 
```

The protocols definitions are in the private `falcon._typing` module now. Should they be moved to the public `falcon.typing` to support this usecase ? 

The protocols names are long and noisy. Use some abbreviations ?

# Related Issues
#2372


vytas7:
One thing that we need to investigate (not necessarily related to this change) is how to control annotations in Sphinx. Because if this new union of protocols gets evaluated/expanded, it really clutters the docs... :grimacing:

vytas7:
As mentioned, I'll try to look into whether Sphinx can be tweaked not to unroll the whole union, and how different editors behave.

Because right now it is a bit of a showstopper wrt Sphinx:
image

vytas7:
It might be possible to resolve the Sphinx issue via [`autodoc_type_aliases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases), but haven't tried this myself.

vytas7 (changes requested):
Thanks once again for prototyping this @jkmnt! I agree that this is much better than just `object`, especially given that in many cases middleware implements just a single method. :100: 

However, we need to find a more aesthetic solution than letting Sphinx unwrap these unions. I tried [autodoc_type_aliases](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases), and unfortunately it doesn't really work as advertised... :grimacing: 

Another minor (and easy to fix) issue is that the async union seems to omit the WebSocket variants, probably by oversight.

vytas7:
It seems that Sphinx doesn't traverse into very advanced secondary unions, and probably it doesn't traverse into `if TYPE_CHECKING:` either, so maybe it is possible to exploit these effects to prevent these mammoth unrolled forms from appearing.

**Edit**: No, this still doesn't seem to work for `App.__init__()` parameters...

**Edit2**: See this discussion that seems to touch upon similar issue(s) https://github.com/sphinx-doc/sphinx/issues/10785

vytas7 (changes requested):
OK, I'm thinking to make the following changes:
* Rename `AsgiMiddleware` :arrow_right: `AsyncMiddleware`; I would like to avoid confusion with generic ASGI middleware.
* Fix `AsyncMiddleware` to include the accidentally omitted WebSocket methods.
* Move `Middleware` and `AsyncMiddleware` to `falcon.typing`, and document them as part of the public interface. Note that docstrings probably won't work, we'll need to write `.. class::` docs manually (see #2465 for more info).
* Keep all other new aliases confined to `falcon._typing` for now. Although I understand the value proposition of being able to annotate middleware more precisely, I doubt many users would go the extra mile; moreover, it is quite common that a middleware component implements just one or two methods from the interface, so I think this will practically be a major improvement over just [`object`].

Does that sound like an acceptable plan @CaselIT @jkmnt?

vytas7 (approved):
I have made improvements to this PR, but I cannot commit the changes despite having access to this PR, due to some GH glitch.

So let's merge this as-is, and I'll open a follow-up PR. (**Edit**: :arrow_right: #2472.)

Thanks once again for the effort @jkmnt!

jkmnt:
> I have made improvements to this PR, but I cannot commit the changes despite having access to this PR, due to some GH glitch.
> 
> So let's merge this as-is, and I'll open a follow-up PR. (**Edit**: ➡️ #2472.)
> 
> Thanks once again for the effort @jkmnt!

Looks good to me and it seems you're managed to commit the changes eventually.

PS: Hope to suggest more PRs to the Falcon after returning to the good old Python-land from the JS/TS hell )

PR #2392, opened 2024-10-27
chore: prepare Falcon 4.0.2 (in-tree)

vytas7:
- [x] Aggregate contributors.
- [x] Add a newsfragment for enabling PDF on RtD.
- [x] Decide when we want to ship this.


CaselIT:
> * Decide when we want to ship this.

I think wait a few more days to see if anything more pops up, then release like at the beginning on November.
After that we could start working on 4.1

vytas7:
Hm, maybe we should add #2396 to known typing limitations?

CaselIT:
> Hm, maybe we should add #2396 to known typing limitations?

Makes sense

CaselIT (approved):

PR #2401, opened 2024-11-04
docs(typing): restructure typing docs into a separate chapter

vytas7:


docs/api/typing.rst
```diff
@@ -0,0 +1,75 @@
...
+    app = App(request_type=MyRequest)
+    app.add_error_handler(OSError, handle_os_error)
+
+(Please also see the following GitHub issue:
+`#2372 <https://github.com/falconry/falcon/issues/2372>`__.)
+
+.. important::
+    This is only a typing limitation that has no effect outside of type
+    checking -- the above ``app`` will run just fine!
+
```
CaselIT, line 69: maybe now somewhere in this section that we are working to remove these limitations

CaselIT (approved):
left a suggestion, but it's good also without it

docs/api/typing.rst
vytas7: I'll address this in the 4.0.2 microrelease PR then. That's a good suggestion, but I think it would best fit into the sentence just above this note, where we link to the GitHub issue.

PR #2405, opened 2024-11-10
refactor(testing): deprecate testtools support in TestCase

### LINKED ISSUE #2156: Drop `testtools` support?

EricGoulart:
# Summary of Changes

Remove conditional dependency on testtools in TestCase to simplify maintenance and prepare for Falcon 5.0 where testtools will no longer be supported. Instead, unittest is now the default, and users are encouraged to use pytest for testing Falcon applications. This change emits a deprecation warning if testtools is still enabled via environment variables.

# Related Issues

Closes #2156


vytas7 (changes requested):
Thanks once again for this PR. :+1:

It seems that it needs work before we can merge it, though:

* A newsfragment is missing.
* The environment variable could be called `FALCON_TEST_CASE`, or `FALCON_BASE_TEST_CASE`. The value could be an import path with or without the `.TestCase` or `::TestCase` suffix, e.g., `FALCON_BASE_TEST_CASE=testtools.TestCase`.
* We shouldn't emit a deprecation warning just because `testtools` _can_ be found in the current environment, but only when a `TestCase` is initialized with the base from `testtools` automagically. If provided via the envvar, `testtools` should continue to emit no warning.
* I don't think we should write that we recommend `pytest` over `unittest`; although personally I could maybe sign under that recommendation, `unittest` continues to be very popular too. Moreover, you can use `pytest` to run `unittest.TestCase`s, allowing even for a hybrid approach.

falcon/testing/test_case.py
```diff
@@ -18,36 +18,79 @@
...
-except ImportError:  # pragma: nocover
-    import unittest
+import os
+import unittest
+import warnings
 
 import falcon
 import falcon.request
-
-# TODO hoist for backwards compat. Remove in falcon 4.
```
vytas7, line 29: Please don't remove this note!
We missed it in Falcon 4.0 it seems, so it needs to be updated to say Falcon 5.0, if you want to touch it.

falcon/testing/test_case.py
```diff
@@ -18,36 +18,79 @@
...
+
+        BaseTestCase = testtools.TestCase
+    except ImportError:
+        BaseTestCase = unittest.TestCase
+elif base_case is None:
+    try:
+        import testtools
+
+        warnings.warn(
+            'Support for testtools is deprecated and will be removed in Falcon 5.0.',
```
vytas7, line 44: As requested before, please do not warn just because the package is importable!
Only warn when a `TestCase` is actually initialized.

falcon/testing/test_case.py
```diff
@@ -18,36 +18,79 @@
...
 import falcon
 import falcon.request
-
-# TODO hoist for backwards compat. Remove in falcon 4.
 from falcon.testing.client import Result  # NOQA
 from falcon.testing.client import TestClient
 
+base_case = os.environ.get('FALCON_BASE_TEST_CASE')
+
+if base_case and 'testtools.TestCase' in base_case:
```
vytas7, line 32: We need to actually import the specified module, not just support `testtools.TestCase`.
It can be `something.else.TestCase` too.

docs/_newsfragments/2156.breakingchange.rst
```diff
@@ -0,0 +1,3 @@
+- **Deprecation**: Support for `testtools` in `falcon.testing` is now deprecated and will be removed in Falcon 5.0.
```
vytas7, line 1: Don't name the newsfragment `.breakingchange.`, because we are not making any actual breaking changes until 5.0.

Since we are adding a new way to configure the base test case, I suggest marketing this as new feature.

falcon/testing/test_case.py
```diff
@@ -18,36 +18,79 @@
...
+
+        class TestMyApp(MyTestCase):
+            def test_get_message(self):
+                doc = {'message': 'Hello world!'}
+
+                result = self.simulate_get('/messages/42')
+                self.assertEqual(result.json, doc)
     """
 
-    # NOTE(vytas): Here we have to restore __test__ to allow collecting tests!
```
vytas7, line 50: Why have you removed my note?

vytas7 (changes requested):
Please do not remove important notes that were not added by you!

It seems that the updated changes are not really what was requested, could you take a look again? Thanks!

Moreover, code coverage does not reach 100%. You can run [`tox` locally to check coverage](https://falcon.readthedocs.io/en/stable/community/contributing.html#test-coverage) before pushing changes for a public review; this might be more efficient in terms of the number of unnecessary iterations in the review process.

vytas7:
Hi again @EricGoulart, and thanks for your effort on this and other PRs so far.
Unfortunately, there is still much work needed to move this prototype forward, so I am going to close this one for now. You are welcome to submit this as a new PR if you address all the points, and make the tests/coverage pass.

PR #2411, opened 2024-11-20
fix(typing): use proper type for SinkCallable **kwargs

jap:


CaselIT (changes requested):
Hi,

The proposed typing is wrong. `kwargs` are not typed as dict.

Also when can you have none as the value?

jap:
Oh right, that kwargs thing is one of those typing gotchas.

Anyway, to answer your question, kwargs is constructed using `groupdict()` which can return items with None when a regex matches but individual named groups don't, for example:
```
>>> re.match('(?P<foo>bar)?', 'baz').groupdict()
{'foo': None}
```

I'll give it another thought - what I'd like is being able to have my functions implement SinkCallable with a call signature
```
def my_sink(req: falcon.Request, resp, falcon.Response) -> None:
  ...
```
but mypy and other type checkers cannot infer that **kwargs is going to be empty in those cases so there will be a type mismatch.

I guess for now I'll have to write it as
```
def my_sink(req, resp, **_kwargs: object) -> None:
  assert not _kwargs, 'got unexpected extra arguments in my_sink'
  ...
```
which somewhat nullifies the elegance of passing these things as keyword arguments :(

Long term I can imagine having `add_sink` require an explicit string or regex as prefix (and no more implicit coercion to regex), so that it would be possible to have a typing-overload for add_sink that accepts a SinkCallable without kwargs if a string is passed as prefix. However, this would mean a breaking change for people passing regular expressions as strings to add_sink.

CaselIT:
> kwargs is constructed using groupdict() which can return items with None when a regex matches but individual named groups don't, for example:

understood!

regarding the call signature for sink, maybe we can use `@overload ` so that both types are accepted?

CaselIT (approved):

vytas7 (approved):
Thanks for this improvement @jap, yes, the values may be `None` in the case of optional unmatched regex groups.

CaselIT:
oh I wanted to try overload in this branch, but I guess we can do a follow on change for it

vytas7:
Aha I thought you were planning a follow-up anyway since you approved 😅

PR #2421, opened 2024-12-12
feat(multipart): add max_length to secure_filename()

Krishn1412:
1) adding max_length parameter which truncates the file but includes the ext. 
2) This is presently set to None but can be set 255 characters.

### DIFF
--- a/.github/workflows/test-other.yaml
+++ b/.github/workflows/test-other.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   run-tox:
     name: tox -e ${{ matrix.toxenv }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
--- a/docs/_newsfragments/2243.newandimproved.rst
+++ b/docs/_newsfragments/2243.newandimproved.rst
@@ -1,0 +1,3 @@
+:class:`~falcon.routing.StaticRoute` now renders ``Etag`` headers. It also
+checks ``If-None-Match`` in requests and returns HTTP 304 response if
+appropriate.
--- a/docs/api/media.rst
+++ b/docs/api/media.rst
@@ -80,6 +80,8 @@ do the heavy lifting for you.
     will only process request media the first time it is referenced. Subsequent
     interactions will use a cached object.
 
+.. _media_validation:
+
 Validating Media
 ----------------
 
--- a/docs/user/faq.rst
+++ b/docs/user/faq.rst
@@ -828,6 +828,22 @@ via :meth:`~falcon.Request.get_param`, or use the
 :meth:`~falcon.Request.get_param_as_json` convenience method as
 demonstrated above.
 
+Can I use msgspec with Falcon?
+------------------------------
+
+`msgspec <https://jcristharif.com/msgspec/>`__ is a fast serialization and
+validation library, with built-in support for JSON, MessagePack, YAML, and
+TOML.
+
+You can use ``msgspec`` as :ref:`JSON handler <custom-media-json-library>` out
+of the box. Its awesome performance aside, you will also be able to assign
+instances of ``msgspec.Struct`` to :attr:`resp.media <falcon.Response.media>`
+without any additional configuration.
+
+It is also fairly straightforward to set up
+request media validation and error handling; see more in the following recipe:
+:ref:`msgspec integration <msgspec_recipe>`.
+
 How can I handle forward slashes within a route template field?
 ---------------------------------------------------------------
 
@@ -954,9 +970,10 @@ setting the `default` or `object_hook` params can negatively impact the
 performance of (de)serialization.
 
 If you use an alternative JSON library, you might also look whether it provides
-support for additional data types. For instance, the popular ``orjson`` opts to
-automatically serialize :mod:`dataclasses`, :mod:`enums <enum>`,
-:class:`~datetime.datetime` objects, etc.
+support for additional data types. For instance, the popular ``orjson`` and
+:ref:`msgspec <msgspec_recipe>` libraries opt to automatically serialize
+:mod:`dataclasses`, :mod:`enums <enum>`, :class:`~datetime.datetime` objects,
+etc.
 
 Furthermore, different Internet media types such as YAML,
 :class:`msgpack <falcon.media.MessagePackHandler>`, etc might support more data
--- a/docs/user/recipes/index.rst
+++ b/docs/user/recipes/index.rst
@@ -5,6 +5,7 @@ Recipes
    :maxdepth: 2
 
    header-name-case
+   msgspec-integration
    multipart-mixed
    output-csv
    plain-text-handler
--- a/docs/user/recipes/msgspec-integration.rst
+++ b/docs/user/recipes/msgspec-integration.rst
@@ -1,0 +1,116 @@
+.. _msgspec_recipe:
+
+Msgspec Integration
+===================
+
+This recipe illustrates how the popular
+`msgspec <https://jcristharif.com/msgspec/>`__ data serialization and
+validation library can be used with Falcon.
+
+Media handlers
+--------------
+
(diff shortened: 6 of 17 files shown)


vytas7 (changes requested):
Thanks for this PR @Krishn1412!

In order to speed up the review process though, it is usually more efficient if you make sure that at least [`tox` passes locally for you first](https://falcon.readthedocs.io/en/stable/community/contributing.html#pull-requests). Thanks!

falcon/util/misc.py
```diff
@@ -373,7 +376,13 @@ def secure_filename(filename: str) -> str:
     filename = unicodedata.normalize('NFKD', filename)
     if filename.startswith('.'):
         filename = filename.replace('.', '_', 1)
-    return _UNSAFE_CHARS.sub('_', filename)
+    filename = _UNSAFE_CHARS.sub('_', filename)
+
+    if max_length is not None and len(filename) > max_length:
+        name, ext = os.path.splitext(filename)
+        filename = name[:max_length - len(ext)] + ext
```
CaselIT, line 383: note that `max_length - len(ext)` may be negative

Krishn1412: Can you check the logic I have added right now? Do you want to keep ext upto the max_length or remove the concept of max_length entirely incase the ext is longer than max_length?

falcon/media/multipart.py
```diff
@@ -592,6 +592,15 @@ class MultipartParseOptions:
...
     :class:`.MultipartParseError` will be raised.
     """
+    max_filename_length: Optional[int]
+    """The maximum length of filenames in multipart forms (default ``None``).
+
+    If the filename exceeds this value, it will be truncated while preserving
+    the file extension. Defaults to ``None`` for backward compatibility.
+
+    .. note::
+        Starting from Falcon 5.0, this will default to 255 characters.
```
vytas7, line 602: I suggest a smaller default value. On some systems, e.g., Windows, the max possible path length is around that value, so likely we want to make the file name part of the path shorter.

Maybe 64 could be a sane default?

falcon/util/misc.py
```diff
@@ -373,7 +376,16 @@ def secure_filename(filename: str) -> str:
     filename = unicodedata.normalize('NFKD', filename)
     if filename.startswith('.'):
         filename = filename.replace('.', '_', 1)
-    return _UNSAFE_CHARS.sub('_', filename)
+    filename = _UNSAFE_CHARS.sub('_', filename)
+
+    if max_length is not None and len(filename) > max_length:
+        name, ext = os.path.splitext(filename)
+        if max_length < len(ext):
```
vytas7, line 383: I don't think it is very important how we handle these edge cases, as long as the behavior is not too complicated, and it is clearly documented what to expect.

That said, maybe we could go with the following logic?
* If the filename fits into the requested _max_length_, obviously return it as-is.
* Split the filename into _stem_ & _extension_, just as you proposed.
* If it is possible to get the filename under the limit by shortening the _stem_ so that at remains non-empty (i.e., at least one character), do it.
* Otherwise, disregard the distinction between the stem and extension, and simply return `filename[:max_length]`.

vytas7 (changes requested):
Provided implementation suggestions in-line.

Also, you still need to make all the `tox` tests pass. To that end, you'll also need to add new unit tests verifying the proposed behaviour.

falcon/media/multipart.py
CaselIT: 64 may be a bit small as default, but 96/128 is likely ok.

falcon/util/misc.py
```diff
@@ -373,7 +376,16 @@ def secure_filename(filename: str) -> str:
...
+    filename = _UNSAFE_CHARS.sub('_', filename)
+
+    if max_length is not None and len(filename) > max_length:
+        name, ext = os.path.splitext(filename)
+        if max_length - len(ext) >= 1:
+            filename = name[: max_length - len(ext)] + ext
+        else:
+            filename = filename[:max_length]
+
+    return filename
```
CaselIT, line 388: as a side note I noticed that we don't hold windows hand regarding device file names, like
https://github.com/pallets/werkzeug/blob/7868bef5d978093a8baa0784464ebe5d775ae92a/src/werkzeug/utils.py#L229-L237

this is likely another issue / pr though

CaselIT (approved):
other than the max length for v5 the rest looks fine, thanks!

falcon/util/misc.py
vytas7: Yes, let's create a separate issue if this is desirable. Not really sure this belongs to the framework either. Normally the filename should be appended to a path, is it still an issue on Windows if you write like `C:\Uploads\NUL`?

CaselIT: Yes that's not valid in any folder

vytas7: I created an issue from your comment: #2422.

falcon/media/multipart.py
vytas7: Personally I think 64 ought to be enough for the basename, since `secure_filename()` is mangling the name in any case, so there is no guarantee to get exactly the same string as submitted in the form anyway. But 96 works for me as well.

falcon/media/multipart.py
```diff
@@ -592,6 +592,15 @@ class MultipartParseOptions:
     If the body part headers size exceeds this value, an instance of
     :class:`.MultipartParseError` will be raised.
     """
+    max_filename_length: Optional[int]
```
vytas7, line 595: This new length is not passed to the actual invocation of `secure_filename()` when referencing the corresponding part property.

Also add tests (make sure to parametrize on both WSGI/ASGI) for actually modifying this parsing option, and verifying that it is respected when referencing the said property.

falcon/media/multipart.py
vytas7, line 1: We need to add a [newsfragment](https://falcon.readthedocs.io/en/stable/community/contributing.html#changelog) for this improvement.

falcon/util/misc.py
```diff
@@ -357,6 +358,8 @@ def secure_filename(filename: str) -> str:
     Args:
         filename (str): Arbitrary filename input from the request, such as a
             multipart form filename field.
+        max_length (int, optional): Maximum length of the sanitized filename,
```
vytas7, line 361: This `(int, optional)` looks a bit unconventional. I think you can simply remove this type from here altogether, and let Sphinx pick it up from the type annotation.

vytas7 (changes requested):
Thanks, the algorithm itself looks good now!

It seems that you forgot to actually forgot to wire the new option into the actual implementation of the [`BodyPart.secure_filename`](https://falcon.readthedocs.io/en/stable/api/multipart.html#falcon.media.multipart.BodyPart.secure_filename) property though.

See also other comments inline.

falcon/util/misc.py
Krishn1412: Do you want to pass the max_length as a parameter in the **BodyPart.secure_filename** function or set it as an instance or class variable and reference it from there?

vytas7: I want us to actually use the new option in `BodyPart.secure_filename()`, i.e.:
```python
    @property
    def secure_filename(self) -> str:
        """The sanitized version of `filename` using only the most common ASCII
        <...>
        """  # noqa: D205
        try:
            return misc.secure_filename(self.filename or '')
        except ValueError as ex:
            raise MultipartParseError(description=str(ex)) from ex
```
:arrow_down:
```python
    @property
    def secure_filename(self) -> str:
        """The sanitized version of `filename` using only the most common ASCII
        <...>
        """  # noqa: D205
        try:
            return misc.secure_filename(self.filename or '', max_length=self._parse_options.max_secure_filename_length)
        except ValueError as ex:
            raise MultipartParseError(description=str(ex)) from ex
```
(untested)

vytas7:
Superseded by #2448.

@Krishn1412 your contribution was acknowledged using `Co-authored-by:`. Thanks!

PR #2425, opened 2025-01-02
docs(recipes): clean up recent recipe additions (+misc touchups)

vytas7:


docs/user/recipes/plain-text-handler.rst
vytas7, line 1: This file was originally contributing with CRLF line endings, whereas we use LF. Hence the total diff.

CaselIT (approved):

vytas7:
Codecov is being lame again, doesn't process some files...

PR #2465, opened 2025-06-12
chore(autodoc): add a workaround for Sphinx expanding types too much

vytas7:


falcon/app_helpers.py
```diff
@@ -68,7 +68,7 @@ def prepare_middleware(
 
 @overload
 def prepare_middleware(
-    middleware: Iterable, independent_middleware: bool = ..., *, asgi: Literal[True]
+    middleware: Iterable, independent_middleware: bool = ..., asgi: Literal[True] = ...
```
vytas7, line 71: @CaselIT do you remember why this `*` was here in this overloaded variant?
We don't really support such signature, I simply remove it, and everything seems to pass.

docs/ext/workarounds.py
```diff
@@ -0,0 +1,34 @@
...
+def patch_type_alias_forward_ref(cls: type) -> None:
+    def __hash__(obj) -> int:
+        return hash(obj.name)
+
+    def __repr__(obj) -> str:
+        return obj.name
+
+
+def setup(app):
+    patch_type_alias_forward_ref(sphinx.util.inspect.TypeAliasForwardRef)
```
CaselIT: I don't understand how this could work. isn't `patch_type_alias_forward_ref` effectively a no-op?

falcon/app_helpers.py
CaselIT: may be a typo that was left there

docs/ext/workarounds.py
vytas7: Indeed, it should have been this, plus
```python
    cls.__repr__ = __repr__
```

But now it seems to work correctly even without it?..

vytas7: OK, maybe it was something in my environment...

It seems that, as I've seen on some issue, aliases now work if I define the `.. :class::` manually, but not from `autodoc` aliases, because they are typing constructs, not classes (roughly `autodoc` module attributes/data).

CaselIT: do you want to keep this extension?

vytas7: No, give me a second, I'll push an update. There was something in my environment it seems.

Now I understood I believe what works and what doesn't...

falcon/app_helpers.py
vytas7: OK, I understood why it was there, it was not needed since we type middleware as `Iterable` in both cases.
But if we introduce fine grained protocols along the lines of #2390 , Mypy says overloads overlap since both `asgi=` parameters are optional, so we make it a keyword argument in the second case, but not optional.

docs/ext/workarounds.py
vytas7: We might need this extension after all (the fixed version), I'll open a separate PR or add to @jkmnt's middleware typing PR. I'm getting this dreaded `TypeAliasForwardRef` again for no specific reason there, not sure what it depends on.

It looks like it is a complex combination of this part of Sphinx being totally buggy, which parts are imported under `if TYPE_CHECKING:`, also combo with the Napoleon extension (?).

PR #2467, opened 2025-06-18
fix(typing): fix type definition to match __slots__ value

numberoverzero:
# Summary of Changes

fixed a typo ([PR](https://github.com/falconry/falcon/pull/2290), [diff](https://github.com/falconry/falcon/commit/a452a2d92ea81612b8f8232b3625a9f721dbaf70#diff-17cd1395bb55dc2d7ab7458044c8ded6617eff797edfdbd7dbb7c707753bab91R2007)) in a type annotation.  note that removing the unintended value here is not potentially breaking, because the class uses `__slots__` which prevents the assignment already.  Anyone that tried to assign to this value would hit an `AttributeError` before they accidentally starting using it.  Here's a quick validation of that:

```py
class Demo:
    unknown: int
    __slots__ = ['known']
    def __init__(self, known):
        self.known = known

h = Demo("ok")

#  raises AttributeError: 'Demo' object has no attribute 'unknown' and no __dict__ for setting new attributes
h.unknown = 3
```

here's the non-existent attribute on the stable version of the falcon docs:

https://falcon.readthedocs.io/en/stable/api/app.html#falcon.RequestOptions.keep_black_qs_values

# Related Issues

N/A - didn't file an issue since it's a quick fix.

# Pull Request Checklist

I didn't run any of the checks listed in the PR template, please let me know if they're still necessary for this change.

I'm not sure if there's a way to write a test to guard this - maybe something that checks if a class's type annotations includes a member not in `__slots__`?  But then subclassing could complicate things.

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -2436,7 +2436,7 @@ class RequestOptions:
     respectively.
     """
 
-    keep_black_qs_values: bool
+    keep_blank_qs_values: bool
     """Set to ``False`` to ignore query string params that have missing or blank
     values (default ``True``).
 


numberoverzero:
I updated the commit message to better match the [commit message format](https://falcon.readthedocs.io/en/stable/community/contributing.html#commit-message-format) but I'm not sure this is the right type.  It's somewhere between `fix` and `docs` so I used [this commit](https://github.com/falconry/falcon/commit/ed80f698e987d5709a1360bb1899a603a81bb59b) as a guide.

vytas7:
Hi @numberoverzero, and thanks for the analysis wrt breaking changes!

Normally, we only accept pull requests against the main development branch (`master`). The issue is already fixed there: https://falcon.readthedocs.io/en/latest/api/app.html#falcon.RequestOptions.keep_blank_qs_values.

However, maybe you could contribute a newsfragment with a short explanation why we don't consider it a breaking change, but rather a documentation fix? We would happy to accept such PR.

P.S. I'm on vacation so I might be less responsive than usual.

vytas7:
Re writing tests, I was hoping that Mypy would catch such stuff, but apparently it didn't. We have ASGI E2E tests checked with `--strict`, maybe if we referenced options there, or in another test that was annotated, Mypy could catch it? 🤔

PR #2468, opened 2025-06-26
docs(Context): remove deprecation warnings for the mapping interface

### LINKED ISSUE #1430: Remove deprecation warnings when using context dict interface
Context type has been changed to a bare class in https://github.com/falconry/falcon/issues/1120 , allowing to set attributes on the context object instead of using dict interface. However, the class derives from `dict` for now, allowing applications to use the (now considered deprecated) dict interface.

Emit `DeprecationWarning`s every time `dict` interface is used, or at least when items are set.

This could be achieved two (let us know if you find an even better one!) ways:

1. Switching to a completely bare class, and routing all dict functions to an internal attribute of type dict, i.e. emulating the interface with `__delitem__`, `__getitem__`, `__setitem__` etc. Then it is easy to emit warnings where desired. Is one wants this to work as fast as a true class, we might need to go the foul way of omitting the constructor, and checking for the compatibility dict attribute upon every access.
2. Keeping the current class, and overriding just `__setitem__` in the following fashion:
```python
class Context(dict):
    @deprecated('hey, this is deprecated')
    def __setitem__(self, key, value):
        dict.__setitem__(self, key, value)
```

This approach would be slightly slower than a bare class though (same as now).

vytas7:
See #1430 for discussion/background.

Closes #1430.

### DIFF
--- a/docs/user/faq.rst
+++ b/docs/user/faq.rst
@@ -876,21 +876,21 @@ set attributes on the object:
    # Falcon 2.0
    req.context.cache_backend = MyUltraFastCache.connect()
 
-The new default context type emulates a dict-like mapping interface in a way
-that context attributes are linked to dict items, i.e. setting an object
-attribute also sets the corresponding dict item, and vice versa. As a result,
-existing code will largely work unmodified with Falcon 2.0. Nevertheless, it is
-recommended to migrate to the new interface as outlined above since the
-dict-like mapping interface may be removed from the context type in a future
-release.
+The new :class:`default context type <falcon.Context>` emulates a dict-like
+mapping interface in a way that context attributes are linked to dict items,
+i.e. setting an object attribute also sets the corresponding dict item, and
+vice versa.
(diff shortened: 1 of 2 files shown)


docs/user/faq.rst
```diff
@@ -876,21 +876,21 @@ set attributes on the object:
...
    req.context.cache_backend = MyUltraFastCache.connect()
 
-The new default context type emulates a dict-like mapping interface in a way
-that context attributes are linked to dict items, i.e. setting an object
-attribute also sets the corresponding dict item, and vice versa. As a result,
-existing code will largely work unmodified with Falcon 2.0. Nevertheless, it is
-recommended to migrate to the new interface as outlined above since the
-dict-like mapping interface may be removed from the context type in a future
-release.
+The new :class:`default context type <falcon.Context>` emulates a dict-like
```
CaselIT, line 879: is it still new?

vytas7: Compared to pre-2.0 as per the question, it is "new" as opposed to the "old" dict...

PR #2472, opened 2025-07-02
fix(typing): include protocols for WebSocket middleware + cleanup

vytas7:
Follow-up to https://github.com/falconry/falcon/pull/2390.

### DIFF
--- a/falcon/_typing.py
+++ b/falcon/_typing.py
@@ -200,14 +200,14 @@ DeserializeSync = Callable[[bytes], Any]
 Responder = Union[ResponderMethod, AsgiResponderMethod]
 
 
-# Middleware
-class MiddlewareWithProcessRequest(Protocol):
+# WSGI middleware interface
+class WsgiMiddlewareWithProcessRequest(Protocol):
     """WSGI Middleware with request handler."""
 
     def process_request(self, req: Request, resp: Response) -> None: ...
 
 
-class MiddlewareWithProcessResource(Protocol):
+class WsgiMiddlewareWithProcessResource(Protocol):
     """WSGI Middleware with resource handler."""
 
     def process_resource(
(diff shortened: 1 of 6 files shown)


falcon/_typing.py
```diff
@@ -326,19 +334,34 @@ async def process_response_async(
 # NOTE(jkmnt): This typing is far from perfect due to the Python typing limitations,
 # but better than nothing. Middleware conforming to any protocol of the union
 # will pass the type check. Other protocols violations are not checked.
-Middleware = Union[
-    MiddlewareWithProcessRequest,
-    MiddlewareWithProcessResource,
-    MiddlewareWithProcessResponse,
+SyncMiddleware = Union[
```
CaselIT, line 337: why sync async here instead of wsgi / asgi?

CaselIT: also do we want to expose these protocols in the public typing module?

vytas7: In the commit that I tried to push to the other pull request, I had made them public, hence the names sync/async and docstrings. I wouldn't want to name them WSGI/ASGI in the public interface because that might introduce confusion with generic WSGI/ASGI middleware, which Falcon doesn't support unless you wrap the app yourself.

But then I thought maybe there was not that much value In making them public for now? What do you think?

(If we keep them private, I don't think naming matters that much.)

CaselIT: ok regarding the naming, I wasn't thinking about the general kind of middlewares.

regarding making them public I could see them being used by users while constructing the application, say to keep hold of a list of middlewares while instantiating them, but the usage is likely a stretch. On the other hand I don't see us changing them, so maybe there is little risk in making the public?

vytas7: I will be changing them soon by adding `process_teardown()` :smiling_imp: 

That is not a change per se though as long as we stress (although it is kind of obvious) that these aliases should be treated as opaque, and they only match the middleware interface of the same Falcon version they are defined in.

I'll create a follow-up PR for making them public, or we can make them public after adding teardown? :thinking:

CaselIT: it's ok either way.
in any case I don't think adding new aliases will break existing code, since it would use existing signatures. but it's ok to make them public later

PR #2479, opened 2025-07-11
test(wsgi-servers): add CherryPy's Cheroot to WSGI server tests

vytas7:
CherryPy is maintained by @webknjaz who has been helpful on numerous issues, so it should be a low maintenance cost addition. :sweat_smile: (Everything passed out of the box as soon as I configured the correct options.)

### DIFF
--- a/tests/test_wsgi_servers.py
+++ b/tests/test_wsgi_servers.py
@@ -24,6 +24,27 @@ _STARTUP_TIMEOUT = 10
 _SHUTDOWN_TIMEOUT = 20
 
 
+def _cheroot_args(host, port):
+    """CherryPy's Cheroot"""
+    try:
+        import cheroot  # noqa: F401
+    except ImportError:
+        pytest.skip('cheroot not installed')
+
+    args = (
+        sys.executable,
+        '-m',
+        'cheroot',
+        '--bind',
+        '{}:{}'.format(host, port),
+        # NOTE(vytas): In case a worker hangs for an unexpectedly long time
+        #   while reading or processing request (the default value is 30).
+        '--timeout',
+        str(_REQUEST_TIMEOUT),
+    )
+    return args + ('_wsgi_test_app:app',)
+
+
 def _gunicorn_args(host, port, extra_opts=()):
     """Gunicorn"""
     try:
@@ -119,7 +140,9 @@ def _waitress_args(host, port):
     )
 
 
-@pytest.fixture(params=['gunicorn', 'meinheld', 'uvicorn', 'uwsgi', 'waitress'])
+@pytest.fixture(
+    params=['cheroot', 'gunicorn', 'meinheld', 'uvicorn', 'uwsgi', 'waitress']
+)
 def wsgi_server(request):
     return request.param
 
@@ -127,6 +150,7 @@ def wsgi_server(request):
 @pytest.fixture
 def server_args(wsgi_server):
     servers = {
+        'cheroot': _cheroot_args,
         'gunicorn': _gunicorn_args,
         'meinheld': _meinheld_args,
         'uvicorn': _uvicorn_args,
--- a/tox.ini
+++ b/tox.ini
@@ -195,6 +195,7 @@ commands = {[with-cython]commands}
 [testenv:wsgi_servers]
 setenv = {[with-cython]setenv}
 deps = {[with-cython]deps}
+       cheroot
        gunicorn
        uwsgi
        waitress


tox.ini
webknjaz, line 1: Looks like you need to add the dep to the MyPy env too?

vytas7: Aha, I interpreted an open issue that typing was unsupported. Our type checking of the test code base is pretty much rudimentary anyway, so probably it doesn't make a bit difference, but I can try!

webknjaz: Well, typing is still in the stubs (due to the compatibility requirements in the past) but we run MyPy against 3 Python versions + stubtest: https://github.com/cherrypy/cheroot/blob/fdaa24d/.pre-commit-config.yaml#L201-L274

The typing marker is in place, so MyPy will "see" the types if Cheroot is installed: https://github.com/cherrypy/cheroot/blob/fdaa24d/cheroot/py.typed

Eventually, I want to move the types into the Python modules.

vytas7: The type marker is enough. Since we run these server tests in a subprocess, the actual requirements on the quality of annotations are fairly minimal.

I have now removed the override, it seems to work just fine.

tests/test_wsgi_servers.py
```diff
@@ -24,6 +24,27 @@
 _SHUTDOWN_TIMEOUT = 20
 
 
+def _cheroot_args(host, port):
+    """CherryPy's Cheroot"""
+    try:
+        import cheroot  # noqa: F401
+    except ImportError:
+        pytest.skip('cheroot not installed')
```
CaselIT, line 32: pytest has https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-on-a-missing-import-dependency

vytas7: Hm, maybe we can create a new contributor-friendly issue to apply this treatment throughout the whole test suite? Because right now we are using this pattern throughout.

tests/test_wsgi_servers.py
```diff
@@ -119,14 +140,17 @@ def _waitress_args(host, port):
     )
 
 
-@pytest.fixture(params=['gunicorn', 'meinheld', 'uvicorn', 'uwsgi', 'waitress'])
+@pytest.fixture(
+    params=['cheroot', 'gunicorn', 'meinheld', 'uvicorn', 'uwsgi', 'waitress']
+)
 def wsgi_server(request):
     return request.param
```
webknjaz, line 147:
```suggestion
def wsgi_server(request):
    wsgi_server_importable_name = request.param
    pytest.importorskip(wsgi_server_importable_name)
    return wsgi_server_importable_name
```

tests/test_wsgi_servers.py
webknjaz: @vytas7 +1 to `importorskip`. Though, you could consolidate it in one place: https://github.com/falconry/falcon/pull/2479/files#r2200863676.

webknjaz:
@vytas7 in terms of setting up fixtures, I want to share something I came up with over in aio-libs: https://github.com/aio-libs/multidict/blob/fe4dee0/tests/conftest.py#L24-L156 / https://github.com/aio-libs/propcache/blob/bcb84bf/tests/conftest.py#L12-L64. This strategy is partially inspired by a technique of wrapping params as dataclasses, that one of the pytest maintainers was showing at a EuroPython tutorial + some of my derivative ideas.

webknjaz:
@vytas7 are you coming to EuroPython next week, by the way?

vytas7:
@webknjaz Unfortunately not... Neither me nor @CaselIT (correct me if I'm wrong) participate this year. :imp:

tests/test_wsgi_servers.py
vytas7: Now filed as https://github.com/falconry/falcon/issues/2480.

@webknjaz we are not using this pattern only for WSGI/ASGI servers, but for other libraries too :slightly_smiling_face:

vytas7:
> @vytas7 in terms of setting up fixtures, I want to share something I came up with over in aio-libs: https://github.com/aio-libs/multidict/blob/fe4dee0/tests/conftest.py#L24-L156 / https://github.com/aio-libs/propcache/blob/bcb84bf/tests/conftest.py#L12-L64. This strategy is partially inspired by a technique of wrapping params as dataclasses, that one of the pytest maintainers was showing at a EuroPython tutorial + some of my derivative ideas.

Thanks! I created a new discussion from your suggestion so that it does not get buried in a closed PR: https://github.com/falconry/falcon/discussions/2481.

tests/test_wsgi_servers.py
vytas7: Thanks, I linked to your snippet from #2480!
We wrangle some servers like Gunicorn+Meinheld in a bit more involved way (since the combo needs both dependencies), but it shouldn't be too hard to extend the concept.

PR #2486, opened 2025-07-15
chore(cibuildwheel): enable CPython 3.14 wheels

vytas7:
**Note**: it is important not to publish any stable release wheels compiled against a pre-release CPython.
However, the `rc1` is anticipated in ~1 week, so the risk of us ushering a stable release out the door is highly unlikely anyway.


vytas7:
The failing CI gate is a real issue, on musllinux we get a stack overflow in `uri.pyx`, a variation of https://github.com/cython/cython/issues/4272. I'll refactor that initialization to avoid this issue.

Issue #2337, opened 2024-09-22
Proper `HEAD` support for static routes


vytas7:
Implement proper `HEAD` support for static file serving.

The response should follow the same logic as the default `GET` behaviour (like now), but not open any file streams (because we do not need any as the response to a `HEAD` request cannot have any body anyway).

As a bonus, I think we need to block unsupported methods such as `POST` and `PUT`, and render the correct `Allow` header in response to `OPTIONS`.

akashverma0786:
so, do we need to get the metadata only?

Aanchallllll:
Hello,
would like to contribute.can i be assigned?

vytas7:
Hi @Aanchallllll!
Happy to hear you considering to contribute.
We don't specifically assign per se, a comment here is enough to know that you are working on it!

However, as I understand, @akashverma0786 has also expressed interest, could you two coordinate so that we don't end up duplicating efforts? We have otherwise no shortage of issues :sweat_smile:

akashverma0786:
Hi Anchal, Please go ahead with this issue, I will find myself another one, there are lots here . :)

vytas7:
Cool beans, then it sounds like we have a plan. :sweat_smile:
I've removed the `needs contributor` label, and we are looking forward to your PR @Aanchallllll :+1:

arthurprioli:
Hello, I would like to take this issue, where should I start?

vytas7:
Hi @arthurprioli!
@Aanchallllll is already working on this one, could you coordinate with her?

arthurprioli:
No problem! I'm looking for a refactoring like issue just to get me started. I'm using the "good first issues" label to check them out but I'd appreciate any recommendations for simpler issues to help me get started!

vytas7:
I don't think we have good simple refactoring issues atm.
But since you were looking at this one, this one is similar: https://github.com/falconry/falcon/issues/2244. And it should be fairly straightforward.

vytas7:
I would say #1026 and #2244 are roughly similar in complexity, so it is fine if you pick the one you prefer!

vytas7:
This issue needs new contributors. 🧐

zain-asif-dev:
[...]

PR #2491, opened 2025-07-19
feat(routing): support `HEAD` for static routes

AbduazizZiyodov:
# Summary of Changes

PR is not complete for now, I wanted to have discussion along with development process.

What I did is that in static route call, I'm raising a method not allowed exception if req.method is not GET or HEAD (OPTIONS handled before that line). I also think that in OPTIONS request, we need to change it to that:

```python
...
        if req.method == 'OPTIONS':
            # it's likely a CORS request. Set the allow header to the appropriate value.
            resp.set_header('Allow', 'GET, HEAD') # <- here
            resp.set_header('Content-Length', '0')
            return
...
```

I did not committed it yet, there were some related tests to it, they kept failing. I will change them later (if above is correct).

As described in the issue, HEAD request should behave as GET, but file should not be opened (also doing "stat"), so I added little "switch" to open/stat file if req.method is get. 

I wanted to know whether I'm in right track, previous tests are PASSING after that change. If implementation will become as expected, I will write tests for that.

Thanks.

# Related Issues

#2337

# Pull Request Checklist

- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/latest/tutorial.html#creating-news-fragments) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)


vytas7 (changes requested):
Hi, this is a great start, thanks!

One thing that needs to be changed though, is that `HEAD` should actually perform `stat()` (but we don't need to open the file).

The response to a `HEAD` request should roughly mirror what is rendered in the case of `GET`, bar the actual content.

When in doubt, we can simply check what the popular Nginx (or any other popular all-round HTTP server, like Apache, Caddy, etc) does in one or another case.

We can start Nginx in a Docker container:
```
docker run --rm -it -p 8000:80 nginx
```

Then
```http
HEAD http://localhost:8000/index.html 

HTTP/1.1 200 OK
Accept-Ranges: bytes
Connection: keep-alive
Content-Length: 615
Content-Type: text/html
Date: Sun, 20 Jul 2025 17:51:29 GMT
Etag: "685adee1-267"
Last-Modified: Tue, 24 Jun 2025 17:22:41 GMT
Server: nginx/1.29.0
```
```http
HEAD http://localhost:8000/index.html 'If-None-Match:"685adee1-267"'

HTTP/1.1 304 Not Modified
Connection: keep-alive
Date: Sun, 20 Jul 2025 17:54:01 GMT
Etag: "685adee1-267"
Last-Modified: Tue, 24 Jun 2025 17:22:41 GMT
Server: nginx/1.29.0
```
```http
HEAD http://localhost:8000/index.html If-Modified-Since:"Tue, 24 Jun 2025 17:22:41 GMT"

HTTP/1.1 304 Not Modified
Connection: keep-alive
Date: Sun, 20 Jul 2025 17:53:25 GMT
Etag: "685adee1-267"
Last-Modified: Tue, 24 Jun 2025 17:22:41 GMT
Server: nginx/1.29.0
```
```http
HEAD http://localhost:8000/index.html Range:bytes=0-14

HTTP/1.1 206 Partial Content
Connection: keep-alive
Content-Length: 15
Content-Range: bytes 0-14/615
Content-Type: text/html
Date: Sun, 20 Jul 2025 17:54:49 GMT
Etag: "685adee1-267"
Last-Modified: Tue, 24 Jun 2025 17:22:41 GMT
Server: nginx/1.29.0
```

To summarize, we need to render almost the same responses as to `GET`, but not include the actual content (and ideally not open the file in question, just perform a `stat()`).

I don't know what the best way to structure this is, but I would try to factor out metadata-methods like `_is_not_modified` to primarily accept a `stat` data structure, and then try to reuse as much as possible.

vytas7:
Note that we don't want to follow Nginx blindly here either, though :smiling_imp:

Nginx's `If-Modified-Since` behaviour isn't incorrect _per se_, but it doesn't adhere to the spirit of the [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-if-modified-since). It renders an HTTP 304 only if you pass exactly the same date from `Last-Modified` verbatim instead of performing any date comparison.

So I think our implementation is better in that aspect :slightly_smiling_face:

vytas7:
@AbduazizZiyodov to summarize, in a pinch, conceptually it might be enough to do this for `HEAD`:
```python
current_static_route_get_implementation(req, resp)
resp.stream.close()
resp.stream = None
```

But as discussed, ideally we would like to do `stat(path)` instead of `open(path); fstat(fd)` in the case of `HEAD`. OTOH, the former approach might also make sense since we would directly verify that we can actually open the file.

Regardless of the chosen solution, we should be able to DRY the absolute majority of the implementation with `GET`.

AbduazizZiyodov:
@vytas7 That's also fine, but I guess if we won't "touch" file & keep "DRY" it'll be more satisfiable. I'm thinking about adding flag to existing `_open_file` implementation in that module (e.g. like `should_open: bool = True` with default). `should_open` will become `False` during `HEAD`, and we might return None or empty `BufferedReader` (e.g. `BufferedReader(io.BytesIO(b''))` if I'm not mistaken) & then we have to tune existing parts which handle `GET` request part.

CaselIT:
agree with @vytas7 regarding status that should mirror the get

AbduazizZiyodov:
@vytas7 What should we do regarding to `content-length` ? I've read RFC
* https://datatracker.ietf.org/doc/html/rfc9110#name-head - here it it says it should not be generated.
* https://datatracker.ietf.org/doc/html/rfc9110#section-8.6 - here "MAY" generate

I might skip section related to `_set_range`, which makes content-length 0. Or I just might set it as `st.st_size`.

vytas7:
That's a very good question @AbduazizZiyodov, and I don't have a definite answer yet, I need to reread the newest RFCs on the matter.

Previously, my understanding was that we _SHOULD_ render a `Content-Length` here corresponding to the true length, and a comment by @kgriffs in the source code also conveys the same impression.

AbduazizZiyodov:
@vytas7 As I mentioned, I'm thinking about setting it as `st.st_size`(no `content-range` involved, which makes the length same as size). Then we would reach our goal which is not opening the file(doing only stat) & including `content-length` on response header.

vytas7:
@AbduazizZiyodov I have carefully reread the relevant part of RFC 9110 (9.3.2, HEAD), and I still stand by my original interpretation.
> The server **SHOULD** send the same header fields in response to a `HEAD` request as it **would have sent if the request method had been `GET`**.
> However, a server _MAY_ omit header fields for which a value is determined only while generating the content.  For example, some servers buffer a dynamic response to `GET` until a minimum amount of data is generated so that they can more efficiently delimit small responses or make late decisions with regard to content selection.
> Such a response to `GET` might contain `Content-Length` and `Vary` fields, for example, that are not generated within a `HEAD` response.  These minor inconsistencies are considered preferable to generating and discarding the content for a `HEAD` request, since `HEAD` is usually requested for the sake of efficiency.

(emphasis mine :arrow_up:)

To summarize, we _should_ include `Content-Length` unless we are dealing with dynamically streamed responses where it is hard to calculate the exact length short of actually streaming through all the content and calculating on the fly.
However, in our case, we have an easy way (`stat`) of deducing that `Content-Length` without actually reading any data except `stat`. So we should include the `Content-Length` header.

vytas7:
@AbduazizZiyodov just checking if you're still working on this, or could we find new contributors from Hacktoberfest on this one?

AbduazizZiyodov:
> @AbduazizZiyodov just checking if you're still working on this, or could we find new contributors from Hacktoberfest on this one?

@vytas7 Hi, yes I am! I couldn't make progress a lot prev. month, however I'm sure I'll be able to make it )

AbduazizZiyodov:
@vytas7 Hi, can you verify current implementation ? I tried to achieve the goal with minimal changes/diff (+ DRY).

I'll add test cases.

vytas7:
It seems maintainer edits are not allowed on this one, so I'll go ahead and close this PR for now.

PR #2494, opened 2025-07-21
chore: prepare 4.1.0a1

vytas7:
I have successfully built out more CPython 3.14 wheels, so hopefully we can avoid unpleasant surprises later on: https://github.com/falconry/falcon/actions/runs/16409687738.


CaselIT (approved):
🛳️

PR #2495, opened 2025-07-21
chore(release): grant permissions to upload release content

vytas7:


.github/workflows/cibuildwheel.yaml
```diff
@@ -59,6 +59,8 @@ jobs:
     runs-on: ubuntu-latest
 
     permissions:
+      # NOTE(vytas): The 'contents' permission is needed for writing release assets.
```
webknjaz, line 62: @vytas7 I'd recommend moving things other than upload into a job that runs before this one. Or two jobs. The job that runs validation/linting could even run unconditionally, as it's useful in normal PR runs before releasing. And the job calling the third party action shouldn't have access to OIDC.

FWIW, I'm planning to spend time detecting if `pypi-publish` is called in jobs with steps other than download and issue warnings. It's not something that is generally recommended or supported. https://github.com/pypa/gh-action-pypi-publish/issues/324.

vytas7: Aha, thanks, that's a fair point. I'll try to decouple these into different jobs.

PR #2496, opened 2025-07-21
chore(release): decouple jobs with `id-token` access from other actions

vytas7:
As pointed out by @webknjaz, the best practice recommended by `pypa/gh-action-pypi-publish` is to expose the created _Trusted Publisher_ OIDC tokens to as few third-party actions as possible.

This changeset aims to address the issue (unless I misunderstood it again :cold_sweat:).

### DIFF
--- a/.github/workflows/cibuildwheel.yaml
+++ b/.github/workflows/cibuildwheel.yaml
@@ -52,17 +52,16 @@ jobs:
           name: cibw-sdist
           path: dist/falcon-*
 
-  publish-sdist:
-    name: publish-sdist
+  upload-sdist:
+    name: upload-sdist
     needs:
       - build-sdist
     runs-on: ubuntu-latest
 
     permissions:
       # NOTE(vytas): The 'contents' permission is needed for writing release assets.
+      #   The default permissions seem to suffice, but we specify for clarity.
       contents: write
-      # NOTE(vytas): This permission is mandatory for trusted publishing.
-      id-token: write
 
     steps:
       - name: Checkout repository
@@ -92,6 +91,25 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           files: 'dist/*.tar.gz'
+
+  publish-sdist:
+    name: publish-sdist
+    needs:
+      - build-sdist
+      - upload-sdist
+    runs-on: ubuntu-latest
+
+    permissions:
+      # NOTE(vytas): This permission is mandatory for Trusted Publishing.
+      id-token: write
+
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          pattern: cibw-sdist
+          path: dist
+          merge-multiple: true
 
       - name: Publish sdist and pure-Python wheel to TestPyPI
         uses: pypa/gh-action-pypi-publish@release/v1
@@ -166,16 +184,11 @@ jobs:
           name: cibw-wheel-${{ matrix.python }}-${{ matrix.platform.name }}
           path: wheelhouse/falcon-*.whl
 
-  publish-wheels:
-    name: publish-wheels
+  check-wheels:
+    name: check-wheels
     needs:
       - build-wheels
-      - publish-sdist
     runs-on: ubuntu-latest
-
-    permissions:
-      # NOTE(vytas): This permission is mandatory for trusted publishing.
-      id-token: write
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


webknjaz:
> unless I misunderstood it again

I'd love to hear feedback on how to improve the PyPUG tutorial and the `pypi-publish` README to make things clearer.

.github/workflows/cibuildwheel.yaml
```diff
@@ -52,17 +52,16 @@ jobs:
...
-    name: publish-sdist
+  upload-sdist:
+    name: upload-sdist
     needs:
       - build-sdist
     runs-on: ubuntu-latest
 
     permissions:
       # NOTE(vytas): The 'contents' permission is needed for writing release assets.
+      #   The default permissions seem to suffice, but we specify for clarity.
```
webknjaz, line 63: This would be different for private repos IIRC. But yes, this is correct.

.github/workflows/cibuildwheel.yaml
```diff
@@ -93,6 +92,25 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           files: 'dist/*.tar.gz'
 
+  publish-sdist:
+    name: publish-sdist
+    needs:
+      - build-sdist
```
webknjaz: This is redundant since this job depends on that transitively.
```suggestion
```

.github/workflows/cibuildwheel.yaml
```diff
@@ -93,6 +92,25 @@ jobs:
...
+      # NOTE(vytas): This permission is mandatory for Trusted Publishing.
+      id-token: write
+
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          pattern: cibw-sdist
+          path: dist
+          merge-multiple: true
```
webknjaz, line 114: Since there's just one sdist, pattern-based lookup isn't needed:
```suggestion
          name: cibw-sdist
          path: dist
```

This only makes sense for the platform-specific wheels, where you produce multiple artifacts in parallel.

.github/workflows/cibuildwheel.yaml
```diff
@@ -93,6 +92,25 @@ jobs:
...
+
       - name: Publish sdist and pure-Python wheel to TestPyPI
```
webknjaz, line 119: Normally, we don't support running this action twice in the workflow, and it may cause side effects. However, since they are both conditional on the event name, it should be fine. Just keep this in mind for the future.

.github/workflows/cibuildwheel.yaml
```diff
@@ -199,6 +212,26 @@ jobs:
         run: |
           tools/check_dist.py ${{ github.event_name == 'release' && format('-r {0}', github.ref) || '' }}
 
+  publish-wheels:
+    name: publish-wheels
+    needs:
+      - build-wheels
```
webknjaz: Same as above, redundant.
```suggestion
```

.github/workflows/cibuildwheel.yaml
```diff
@@ -199,6 +212,26 @@ jobs:
...
+      - check-wheels
+      - publish-sdist
+    runs-on: ubuntu-latest
+
+    permissions:
+      # NOTE(vytas): This permission is mandatory for Trusted Publishing.
```
webknjaz, line 228: Maybe, also mention that no other/untrusted actions are allowed to be added into this job (in case people get ideas over time).

.github/workflows/cibuildwheel.yaml
vytas7: No, I know this is a bit misleading, but it is explained at the top of the file. This is one sdist, and one pure-Python wheel.

.github/workflows/cibuildwheel.yaml
vytas7: Good point, I'll add a note.

.github/workflows/cibuildwheel.yaml
vytas7: Is listing all the dependencies that semantically make sense a problem though?

Many dependency systems (including Python Pip's resolver) handle this scenario just fine, you list the packages that you consider dependencies. Of course if you use just Falcon, this is a non-issue :smiling_imp: But if the user, for instance, is building an app with Cheroot, and uses `more-itertools`, normally one just lists both instead of investigating that Cheroot, in fact, also depends on `more-itertools`.

CaselIT (approved):
looks ok

.github/workflows/cibuildwheel.yaml
webknjaz: @vytas7 they are still in a single GHA workflow artifact and not multiple, which means that my point stands. These inputs are redundant here.

vytas7:
> I'd love to hear feedback on how to improve the PyPUG tutorial and the `pypi-publish` README to make things clearer.

The `README` is great and has a great technical depth to it, but maybe one concrete suggestion would be to illustrate with more complete examples (hopefully along the lines of our `cibuildwheel.yaml` now that it is shaping up), so that people would see practically how you create artifacts in one job, and the publishing action does nothing but downloads and publishes them.

Personally, I found Cibuildwheel's documentation very good in this aspect. Although Cibuildwheel is way more complex than the publishing action, sections like [Working examples](https://cibuildwheel.pypa.io/en/stable/working-examples/) help a lot to get things going.

.github/workflows/cibuildwheel.yaml
vytas7: Ah, indeed, you're dead right.
I captured your suggestion in a new issue so that we don't forget it: #2497.

Issue #2480, opened 2025-07-11
Use `pytest.importorskip` to DRY WSGI server tests


vytas7:
Suggested by @CaselIT: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-on-a-missing-import-dependency.

So instead of writing
```python
try:
    import gunicorn  # noqa: F401
except ImportError:
    pytest.skip('gunicorn not installed')
```
...we can use the more succinct form:
```python
gunicorn = pytest.importorskip('gunicorn')  # No need for linter overrides either!
```

Also see a more concrete snippet by @webknjaz how to DRY ASGI/WSGI server fixtures: https://github.com/falconry/falcon/pull/2479#discussion_r2200863676.

webknjaz:
@x612skm wanna give this a try?

x612skm:
@webknjaz Hey! Yeah sure, I can take this! Thanks!

x612skm:
Hi Maintainers! Good Day!
I've created one PR by changing the things at one file only, I went ahead and checked using `findall` that there are many other files as well which are using the approach. For now, it's there in the single file If you allow me I can replace other occurrences as well, maybe in a separate PR where I mention what changed and why changed and if I face any error. 
Well, that's another day approach if you allow me to do that - For now would you mind to review this PR? Thanks! 

Also, just a fyi I was novice for this approach and apologies for that still learning python so I used the LLMs a bit to understand what actually is `importorskip` and what it does, I've included some bullet points in my PR description for the understanding for people maybe who come across with not knowing what it actually is.

Let me know if there's anything I can do! 
Thanks @webknjaz for the help and answering my dumb questions! (Also, would you mind taking a look at the PR too? Thanks!)

x612skm:
Hi @vytas7 So, in this PR https://github.com/falconry/falcon/pull/2490 first I went with the approach `gunicorn = pytest.importorskip('gunicorn')` and it made the pipelines fail so now I went with like `pytest.importorskip('gunicorn')`. Let me know if that works, Thanks!

PR #2509, opened 2025-08-06
use pytest.importorskip for optional dependency imports

### LINKED ISSUE #2492: Use `pytest.importorskip` for remaining tests with optional deps
Split from #2480.

Apply similar treatment to the remaining tests that wrangle optional dependencies by catching `ImportError`.

Some known cases include media handler tests, ASGI server tests. (`grep` tests for `ImportError`.)

x612skm:
This PR is in continuation for the progress made in https://github.com/falconry/falcon/issues/2480 which uses the `pytest.importorskip` to the remaining tests that wrangle optional dependencies by catching ImportError.

Thanks!
Closes #2492 

# Related Issues

#2492 

# Pull Request Checklist

- [ ] Performed automated tests and code quality checks by [running `tox`](https://falcon.readthedocs.io/en/stable/community/contributing.html#pull-requests).
- [ ] Coding style is consistent with the rest of the framework.
- [ ] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/latest/tutorial.html#creating-news-fragments) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)


x612skm:
Hey @vytas7 - I just encountered coverage is low when I run the tox at my local, also when I check the codecov https://app.codecov.io/gh/falconry/falcon/pull/2509/indirect-changes it says it's due to any indirect change. Would you mind help me on this how to get it resolved if that's something to be done via this PR? Also, mind taking a look to this PR? Thanks!

vytas7 (changes requested):
Hi @x612skm, and thanks for the effort so far! :100:

The coverage drop is, IMHO, real, as you seem to have inadvertently changed the behaviour by refactoring.

The `try.. except` clauses before would either succeed to import and by importing set the module variable in question, or fail, and set it `None` instead.

If I understand [`pytest.importorskip`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip) correctly, doing this on a top level would skip the whole test file upon the first failure.
I think `pytest.importorskip` is meant to be used inside a fixture's definition, inside a test function, etc, but not on the top level of the module.

vytas7:
@x612skm if you want to replace global imports, you could turn them into fixtures, e.g.,
```python
import pytest


def import_module(name):
    def importer():
        return pytest.importorskip(name)

    return pytest.fixture()(importer)


numpy = import_module('numpy')
scipy = import_module('scipy')
sympy = import_module('sympy')


def test_one(numpy):
    assert numpy.uint8(1) == 1


def test_integrate(numpy, scipy):
    integral, err = scipy.integrate.quad(numpy.sin, 0, numpy.pi / 2)
    assert integral == pytest.approx(1.0, err)


def test_sympy(sympy):
    # TODO: implement
    pass
```
This now works roughly as you expected:
```
collected 3 items                                                                             

test.py::test_one PASSED                                                                [ 33%]
test.py::test_integrate SKIPPED (could not import 'scipy': No module named 'scipy')     [ 66%]
test.py::test_sympy SKIPPED (could not import 'sympy': No module named 'sympy')         [100%]

================================ 1 passed, 2 skipped in 0.06s =================================
```

But probably it would be even simpler and clearer to just use `pytest.importorskip` where you actually need that module, just like you did in `test_wsgi_servers.py`.

x612skm:
@vytas7 Hey! Thank you very much for taking the time to explain me! Really appreciated! I'd do the following changes and update the PR. Thank you!

x612skm:
Hey @vytas7 👋  - I've reverted the code for the file where it was actually failing, trying to find a way to bypass that but couldn't come up with any. Actually it was happening because of the `msgpack` at the test_ws.py with the reason you noted above. I'd be happy to change if you can suggest anything there or good at this point as well! Thanks for the help and tip above! :)

vytas7 (changes requested):
Just a reminder @x612skm, almost all `pytest.importorskip(...)` invocations in your currently proposed changeset are still defined globally on the module level, causing entire modules to be skipped in case one dependency is missing.

With the exception of very small files where all or almost all tests build upon the same optional dependencies, this behaviour is not desirable as it decreases our overall test coverage and the modularity of our tests, even if in a specific setup we can still attain 100% coverage on paper.

x612skm:
Oh I see! that's why the coverage is saying green when run tox on it. But, one thing I'm confused with the importskip feature how can we actually achieve it inside the test files where it's actually now at the top or we say it's defined globally on the module level, as I see some of the files uses `@pytest.mark.skipif`. Or, we go with the previous of ours what it was using `try...except`?

vytas7:
Maybe it is easier to illustrate with an example, this is roughly what's going on now:
```python
import json

import pytest


yaml = pytest.importorskip('yaml')
msgpack = pytest.importorskip('msgpack')


@pytest.mark.parametrize(
    'loader, payload',
    [
        (json.loads, '{"msg": "Hello"}'),
        (yaml.safe_load, 'msg: Hello'),
        (msgpack.unpackb, b'\x81\xa3msg\xa5Hello'),
    ])
def test_load(loader, payload):
    assert loader(payload) == {'msg': 'Hello'}
```
:arrow_down: 
```
test_example.py::test_load[loads-{"msg": "Hello"}] PASSED                                [ 33%]
test_example.py::test_load[safe_load-msg: Hello] PASSED                                  [ 66%]
test_example.py::test_load[unpackb-\x81\xa3msg\xa5Hello] PASSED                          [100%]

====================================== 3 passed in 0.02s =======================================
```
So far so good, right? However, what happens if, e.g., `msgpack` is missing?
```
collected 0 items / 1 skipped                                                                  

====================================== 1 skipped in 0.01s ======================================
```
The whole file is now skipped. Just because `msgpack` couldn't be imported, we are losing tests for YAML, and even for the stdlib's JSON which requires no third party modules.

vytas7:
@x612skm and as how to resolve this, so it depends on the concrete case, but as said, you need to try to move `pytest.importorskip(...)` inside test fixtures or test cases.

For instance, if a module is needed to perform certain actions, like `requests`, you can convert that into a fixture.
Otherwise, you can use `pytest.importorskip(...)` where certain fixtures are initialized or parameterized.

Refactoring our previous example a bit:
```python
import pytest


@pytest.fixture(scope='session')
def requests():
    return pytest.importorskip('requests')


@pytest.mark.parametrize(
    'module, loader, payload',
    [
        ('json', 'loads', '{"msg": "Hello"}'),
        ('yaml', 'safe_load', 'msg: Hello'),
        ('msgpack', 'unpackb', b'\x81\xa3msg\xa5Hello'),
    ])
def test_load(module, loader, payload):
    loader_func = getattr(pytest.importorskip(module), loader)
    assert loader_func(payload) == {'msg': 'Hello'}


def test_get_localhost(requests):
    resp = requests.get('http://localhost:8000/', timeout=5)
    assert resp.status_code == 200
```
So now, if let's say both `requests` and `msgpack` are missing, we can still run tests for JSON and YAML:
```
test_example.py::test_load[json-loads-{"msg": "Hello"}] PASSED                           [ 25%]
test_example.py::test_load[yaml-safe_load-msg: Hello] PASSED                             [ 50%]
test_example.py::test_load[msgpack-unpackb-\x81\xa3msg\xa5Hello] SKIPPED (could not ...) [ 75%]
test_example.py::test_get_localhost SKIPPED (could not import 'requests': No module ...) [100%]

================================= 2 passed, 2 skipped in 0.02s =================================
```

x612skm:
Oh! that's the case! Thanks for the example, now I understand it perfectly! and appreciate taking the time for the explanation, thank you very much! (I really like how you experienced folks can write it smoothly!, just jealous :)). BTW, I'll try for the second example as you gave and try to run it local and update the PR. 
Thanks, again!

x612skm:
Hi @vytas7 I need some help, so post your example and explanation I tried to implement kinda same thing with the `test_ws.py`, and `test_asgi_servers.py`-
Also, one question I had while changing this portion to the fixture
```
try:
    import websockets
    import websockets.asyncio.client
    import websockets.exceptions
except ImportError:
    websockets = None  # type: ignore
```



I encountered that if I use the structure like below then the tests are failing throwing errors at the `*.asyncio.client` and `*.exceptions`, so I tried to wrap that inside the fixture itself in the new [commit](72bd31aaf6242611326dab8ab789cf0d04c61e39).

```
@pytest.fixture(scope="session")
def websockets_lib():
    ws = pytest.importorskip("websockets")
    return {
        "client": ws.asyncio.client,
        "exceptions": ws.exceptions,
        "module": ws,
    }
 ```


Would you mind to take some time to check if that looks good now, then I'll proceed ahead with the following files. Thanks!

File: tests/asgi/test_asgi_servers.py [hunk 1]
```diff
@@ -11,28 +11,32 @@
...
+@pytest.fixture(scope='session')
+def requests_lib():
+    return pytest.importorskip('requests')
+
+
+@pytest.fixture(scope='session')
+def websockets_lib():
+    websockets = pytest.importorskip('websockets')
+    # Optionally, check for submodules and raise if missing
+    try:
```
vytas7: If we need to import under `try.. except ImportError: ...` anyway, probably let's leave the `websockets` part _as is_ now.

tests/asgi/test_asgi_servers.py
```diff
@@ -11,28 +11,32 @@
...
 
+@pytest.fixture(scope='session')
+def httpx_lib():
+    return pytest.importorskip('httpx')
 
-from falcon import testing
 
-from . import _asgi_test_app
+@pytest.fixture(scope='session')
+def requests_lib():
```
vytas7: I don't think we want to rename it to `requests_lib`. Just call the fixture `requests`, the value returned from it will be the `requests` module.

vytas7 (changes requested):
Maybe we could try to divide and conquer here, change the imports to use the new pattern where it is easy, and leave the complex cases (such as the ones with submodules) _as is_?

tests/asgi/test_asgi_servers.py
x612skm: Ack, I'll leave it as is.

tests/asgi/test_asgi_servers.py
x612skm: Ack

x612skm:
HI @vytas7! 👋 - After reviewing your comments and suggestions finally went near to the finish line, got a great learning in this MR and I hope now the changes looks good I tried best where it's possible to change it to fixture and where it was not or pulling in circular dependency I reverted it to back to what it was. (2 files were like that). Let me know how it looks now! Thanks!

tests/test_wsgi_servers.py
vytas7, line 1: Later in this file we still have 
```python
@pytest.mark.skipif(
    requests is None, reason='requests module is required for this test'
)
```
...which is now redundant.

tests/asgi/test_ws.py
```diff
@@ -1152,16 +1151,18 @@ async def on_websocket(self, req, ws):
                 pass
 
 
-@pytest.mark.skipif(msgpack, reason='test requires msgpack lib to be missing')
 def test_msgpack_missing():
-    options = WebSocketOptions()
-    handler = options.media_handlers[falcon.WebSocketPayloadType.BINARY]
+    try:
+        pytest.importorskip('msgpack')
```
vytas7: Let's revert this to use the normal `ImportError` machinery for now. I think it became more confusing than it was.

tests/asgi/test_ws.py
```diff
@@ -16,22 +16,20 @@
 from falcon.testing.helpers import _WebSocketState as ClientWebSocketState
 from falcon.util.deprecation import DeprecatedWarning
 
-try:
-    import cbor2
-except ImportError:
-    cbor2 = None  # type: ignore[assignment]
 
+@pytest.fixture(scope='session')
+def cbor2_lib():
```
vytas7: Let drop the `_lib` postfix here and from all other "import-or-skip" fixtures.

tests/asgi/test_hello_asgi.py
```diff
@@ -150,9 +151,9 @@ async def on_get(self, req, resp):
 
 
 class AIOFilesHelloResource:
-    def __init__(self):
+    def __init__(self, aiofiles_lib):
```
vytas7: Let drop the `_lib` postfix here and from all other "import-or-skip" fixtures.

I'm wondering also how this test works at all, if the above fixture is called `aiofiles`, but the parameter is called `aiofiles_lib`?

tests/asgi/test_asgi_servers.py
```diff
@@ -11,16 +11,20 @@
...
-    import httpx
-except ImportError:
-    httpx = None  # type: ignore
+from falcon import testing
+
+from . import _asgi_test_app
+
+
+@pytest.fixture(scope='session')
+def httpx_lib():
```
vytas7: Let drop the `_lib` postfix here and from all other "import-or-skip" fixtures.

tests/asgi/test_ws.py
webknjaz: Can't we drop the try/except in here? Wasn't it supposed to just skip the test instead of having conditional hackery?

webknjaz: Ah, I see what you're saying. Looks like this also has a bunch of unrelated changes that multiply the confusion..

vytas7: Aye, this test is meant to be executed when the dependency in question **is missing**, and skipped otherwise!

So quite opposite to what `importorskip` does... Probably it doesn't lend itself well to this scenario, or at least doesn't seem to make it more elegant.

tests/test_wsgi_servers.py
```diff
@@ -7,13 +7,10 @@
...
 import pytest
 
-try:
-    import requests
-except ImportError:
-    requests = None  # type: ignore
-
 from falcon import testing
 
+requests = pytest.importorskip('requests')
```
webknjaz: Wouldn't this skip the entire test module?

x612skm: ah yeah, good catch I think I missed while changing the files, I'll make this a fixture or revert it to its original position.

tests/asgi/test_hello_asgi.py
x612skm: Yeah I actually remember this making like because when I put the aiofiles only replacing the `_lib`, the tests were failing and due to that I had to make it with only go with it. I'll try it again.

x612skm: I tried it again, and fortunately it worked not touching it again :D

tests/asgi/test_hello_asgi.py
```diff
@@ -150,9 +151,9 @@ async def on_get(self, req, resp):
 
 
 class AIOFilesHelloResource:
-    def __init__(self):
+    def __init__(self, aiofiles):
         self.sample_utf8 = testing.rand_string(8 * SIZE_1_KB, 16 * SIZE_1_KB).encode()
-
+        self._aiofiles_lib = aiofiles
```
x612skm: Hey @vytas7 @webknjaz could you folks help me here to understand what's actually happening, I was trying to solve this by not taking the `aiofiles` in the argument and then by not passing this here instead going with the global. Is there any kind of circular dependency coming in? Having a hard time to understand this, actually how to solve this problem or this is only way to solve the problem. Apologies for the dumb questions 🙂 , Thanks for all the support till now!!

x612skm: To add some more context the failure is coming as 

```
    def test_filelike_closing_aiofiles(self, client, aiofiles):
        resource = AIOFilesHelloResource(aiofiles)
        try:
            client.app.add_route('/filelike-closing', resource)
    
            result = client.simulate_get('/filelike-closing')
    
>           assert result.status_code == 200
E           assert 500 == 200
E            +  where 500 = Result<500 Internal Server Error application/json b'{"title": "500 Internal Server Error"}'>.status_code

tests/asgi/test_hello_asgi.py:321: AssertionError

```

vytas7: As suggested, maybe we could divide and conquer here -- i.e., you could undo your changes from any modules where there are ostensibly complex issues like this one, and we merge the improvements to other tests?

x612skm: Hey @vytas7, reverted to it original form. Now I hope the changes looks good to you. Mind taking a look when you get some time? Thanks!

tests/asgi/test_ws.py
```diff
@@ -411,7 +410,9 @@ async def on_websocket(self, req, ws):
 @pytest.mark.parametrize('custom_text', [True, False])
 @pytest.mark.parametrize('custom_data', [True, False])
 @pytest.mark.skipif(msgpack is None, reason='msgpack is required for this test')
-async def test_media(custom_text, custom_data, conductor):  # NOQA: C901
+async def test_media(  # NOQA: C901
+    custom_text, custom_data, conductor, cbor2, rapidjson
```
CaselIT: this is not 100% the same since it now skips all runs if either cbor2 or rapidjson is missing, before the skip was only when neeeded. it may be fine to keep this way

vytas7: Ah indeed... this needs to be addressed.

vytas7 (changes requested):
We still need to address the intersectional imports, as pointed out by @CaselIT.

tests/asgi/test_ws.py
webknjaz: Also, instead of adding unused arguments, it's nicer to go for `@pytest.mark.usefixtures()` instead.

x612skm: I tried using it but again it was going complex and to pass the tests I was writing something which later confused me as well, so went with the divide and conquer and reverted the file to it's original position.

vytas7 (approved):
Thanks for this improvement @x612skm! I think we're good to merge now.

I have unlinked the issue though, because we have only addressed a part of its scope here. I'll update the issue to reflect the remaining work.

PR #2510, opened 2025-08-09
docs(hooks): clarify injecting params in text preceding `falcon.before`

### LINKED ISSUE #2493: Undocumented case of passing an argument in the @before hook to a route
Hello, I want to pass an argument to my endpoint from the `@before` hook, but I don’t see an example of how to do it in the docs. However, the docs do show an example where `project_id` is passed to the route using `@before`, but it’s not explained how.

[Hooks Docs](https://falcon.readthedocs.io/en/stable/api/hooks.html)

```python
@falcon.before(extract_project_id)
class Message:
    def on_post(self, req, resp, project_id):
        pass

    def on_get(self, req, resp, project_id):
        pass
```

vytas7:
Fixes #2493.


CaselIT (commented):
sorry, I must have missed this

vytas7:
No worries. I just rogue merged it because it was rather trivial, and only affecting documentation.

PR #2513, opened 2025-08-09
feat(typing): improve typing of custom req/resp type

### LINKED ISSUE #2372: Improve typing of custom req/resp type

vytas7:
Fixes #2372.

CC @jkmnt @davetapley.

### DIFF
--- a/falcon/_typing.py
+++ b/falcon/_typing.py
@@ -66,17 +66,31 @@ _T = TypeVar('_T')
 _UNSET = _Unset.UNSET
 UnsetOr = Union[Literal[_Unset.UNSET], _T]
 
+_ReqT = TypeVar('_ReqT', bound='Request', contravariant=True)
+_RespT = TypeVar('_RespT', bound='Response', contravariant=True)
+_AReqT = TypeVar('_AReqT', bound='AsgiRequest', contravariant=True)
+_ARespT = TypeVar('_ARespT', bound='AsgiResponse', contravariant=True)
+
 Link = Dict[str, str]
 CookieArg = Mapping[str, Union[str, Cookie]]
+
+
 # Error handlers
-ErrorHandler = Callable[['Request', 'Response', Exception, Dict[str, Any]], None]
+class ErrorHandler(Protocol[_ReqT, _RespT]):
+    def __call__(
+        self,
+        req: _ReqT,
+        resp: _RespT,
+        error: Exception,
+        params: Dict[str, Any],
+    ) -> None: ...
 
 
-class AsgiErrorHandler(Protocol):
+class AsgiErrorHandler(Protocol[_AReqT, _ARespT]):
     async def __call__(
         self,
-        req: AsgiRequest,
-        resp: Optional[AsgiResponse],
+        req: _AReqT,
+        resp: Optional[_ARespT],
         error: Exception,
         params: Dict[str, Any],
         *,
@@ -91,15 +105,13 @@ ErrorSerializer = Callable[['Request', 'Response', 'HTTPError'], None]
 SinkPrefix = Union[str, Pattern[str]]
 
 
-class SinkCallable(Protocol):
-    def __call__(
-        self, req: Request, resp: Response, **kwargs: Optional[str]
-    ) -> None: ...
+class SinkCallable(Protocol[_ReqT, _RespT]):
+    def __call__(self, req: _ReqT, resp: _RespT, **kwargs: Any) -> None: ...
 
 
-class AsgiSinkCallable(Protocol):
+class AsgiSinkCallable(Protocol[_AReqT, _ARespT]):
     async def __call__(
-        self, req: AsgiRequest, resp: Optional[AsgiResponse], **kwargs: Optional[str]
+        self, req: _AReqT, resp: Optional[_ARespT], **kwargs: Any
     ) -> None: ...
 
 
@@ -129,9 +141,11 @@ class ResponderCallable(Protocol):
 
 ProcessRequestMethod = Callable[['Request', 'Response'], None]
 ProcessResourceMethod = Callable[
-    ['Request', 'Response', Resource, Dict[str, Any]], None
+    ['Request', 'Response', Optional[Resource], Dict[str, Any]], None
 ]
-ProcessResponseMethod = Callable[['Request', 'Response', Resource, bool], None]
+ProcessResponseMethod = Callable[
+    ['Request', 'Response', Optional[Resource], bool], None
+]
 
 
 # ASGI
@@ -161,14 +175,14 @@ AsgiReceive = Callable[[], Awaitable['AsgiEvent']]
 AsgiSend = Callable[['AsgiSendMsg'], Awaitable[None]]
 AsgiProcessRequestMethod = Callable[['AsgiRequest', 'AsgiResponse'], Awaitable[None]]
 AsgiProcessResourceMethod = Callable[
-    ['AsgiRequest', 'AsgiResponse', Resource, Dict[str, Any]], Awaitable[None]
+    ['AsgiRequest', 'AsgiResponse', Optional[Resource], Dict[str, Any]], Awaitable[None]
 ]
 AsgiProcessResponseMethod = Callable[
-    ['AsgiRequest', 'AsgiResponse', Resource, bool], Awaitable[None]
+    ['AsgiRequest', 'AsgiResponse', Optional[Resource], bool], Awaitable[None]
 ]
 AsgiProcessRequestWsMethod = Callable[['AsgiRequest', 'WebSocket'], Awaitable[None]]
 AsgiProcessResourceWsMethod = Callable[
-    ['AsgiRequest', 'WebSocket', Resource, Dict[str, Any]], Awaitable[None]
+    ['AsgiRequest', 'WebSocket', Optional[Resource], Dict[str, Any]], Awaitable[None]
 ]
 ResponseCallbacks = Union[
     Tuple[Callable[[], None], Literal[False]],
@@ -201,29 +215,33 @@ Responder = Union[ResponderMethod, AsgiResponderMethod]
 
 
 # WSGI middleware interface
-class WsgiMiddlewareWithProcessRequest(Protocol):
+class WsgiMiddlewareWithProcessRequest(Protocol[_ReqT, _RespT]):
     """WSGI Middleware with request handler."""
 
-    def process_request(self, req: Request, resp: Response) -> None: ...
+    def process_request(self, req: _ReqT, resp: _RespT) -> None: ...
 
 
-class WsgiMiddlewareWithProcessResource(Protocol):
+class WsgiMiddlewareWithProcessResource(Protocol[_ReqT, _RespT]):
     """WSGI Middleware with resource handler."""
 
     def process_resource(
         self,
-        req: Request,
-        resp: Response,
-        resource: object,
+        req: _ReqT,
+        resp: _RespT,
+        resource: Optional[Resource],
         params: Dict[str, Any],
     ) -> None: ...
 
 
-class WsgiMiddlewareWithProcessResponse(Protocol):
+class WsgiMiddlewareWithProcessResponse(Protocol[_ReqT, _RespT]):
     """WSGI Middleware with response handler."""
 
     def process_response(
-        self, req: Request, resp: Response, resource: object, req_succeeded: bool
+        self,
+        req: _ReqT,
+        resp: _RespT,
+        resource: Optional[Resource],
+        req_succeeded: bool,
     ) -> None: ...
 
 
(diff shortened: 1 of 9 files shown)


jkmnt:
Looks good to me. I'm going to test it with our rich app context and report any typing issues.

vytas7:
@jkmnt thanks in advance, your feedback is much appreciated!
I added a simple test that somewhat reflects your user story: [`tests/test_typing.py`](https://github.com/falconry/falcon/pull/2513/files#diff-e59c19a0093884f60bea307ed2969bc936d640ef81a9d8e93f6f99cab392da61).


Note that I was unable to properly annotate the custom `Request.context_type` (i.e. to make it generic). :imp:

So the workaround I made was that I simply marked `context_type` as any customizable type, and `req.context` is still typed as `Context`. However, `Context` is annotated as allowing arbitrary attribute access, so it masks the majority of issues... Check if that could work for you.

On the positive side, it seems to work as expected for me both in Mypy and Pylance/Pyright. VSCode (which I don't normally use myself) seems to unroll the middleware type aliases (that you introduced) making them a bit spammy, but hopefully that's not the end of the world.

I also had a prototype of applying the same treatment to resource typing, i.e. "resource with GET" | "resource with POST" | etc. This won't work if your resource ONLY has suffixed responders, but that shouldn't normally be the case as suffixed responders are primarily used in addition to the main ones. That prototype is not included in this PR yet though.

vytas7:
P.S. the quickest way to install this prototype directly from my branch:
```
pip install git+https://github.com/vytas7/falcon@2372-typing-custom-req-resp-type
```

jkmnt:
Plugged it into the project as is. Pyright reported no extra typing errors, though it was expected. Our Request/Response subclass the falcon ones and replace the (context) type mismatch with type-ignores, that is:
```python
class Req(falcon.Request):
    context_type = RestContext  # NOTE: ignore is not needed here anymore
    context: RestContext  # type: ignore
```

Next I supplied the generics for the App class:
```python
app = falcon.App[Req, Resp]
```

Types correctly propagated to the `falcon.add_error_handler()` callback and middlewares. Quite impressive ! 
A few casts/Anys/ignores may be removed now.

The `falcon.set_error_serializer` triggered the type error, since the `ErrorSerializer` is still typed as `Callable[[Request, Response], ...]`

I'll try a few ideas how to make the `context`/`context_type` generic.

vytas7:
Thanks a lot for testing so far @jkmnt!
Re `falcon.App(...)`, you shouldn't need to supply any custom typing, `falcon.App(request_type=Req, ...)`, etc, should work out of the box.

Re `context_type`, I haven't found any clean solutions yet. If you remove the custom annotation for `context_type`, the most should pass out of the box, but as said, that wouldn't properly detect any issues you might have.
[PEP 526](https://peps.python.org/pep-0526/#class-and-instance-variable-annotations) seems to explicitly state that  a `ClassVar` parameter cannot include any type variables, which makes it hard to work around, if even possible.

jkmnt:
Ok, context idea here (or, maybe, a recipe ?)

```python
class CustomContextMixin:
    def __init__(self):
        self.custom_context = dict()

class Req(falcon.Request, CustomContextMixin):
    pass

app = falcon.App[Req, falcon.Response](..)

def err_handler(req: Req, resp: falcon.Response, *args, **kwargs):
    reveal_type(req.custom_context)
    # Type of "req.custom_context" is "dict[Unknown, Unknown]"Pylance
    pass
```

The custom context is initialized simultaneously with the Request creation, while the base Request don't even care about it )

It's parallel approach, of coarse, but I don't know how to type the existing implementation. I had luck in the past 
with 
```python 
def __init_subclass__(cls, *, custom_type = ):
    cls.custom_type = custom_type
```
Just opened the project with this trick and it no longer typechecks. Apparently is was a bug in the typechecker )

vytas7:
> The `falcon.set_error_serializer` triggered the type error, since the `ErrorSerializer` is still typed as `Callable[[Request, Response], ...]`

P.S. @jkmnt `ErrorSerializer` has now been fixed.

jkmnt:
Just checked the `falcon.App(request_type=Req, ...)` without explicit annotations. 
The req/resp types are correctly inferred from the __init__ arguments, good !

vytas7:
@jkmnt just to circle back on the context problem, as mentioned I created a test file (in this PR) that loosely models your use case, and it passes type checking without any warnings: [`tests/test_typing.py`](https://github.com/vytas7/falcon/blob/2372-typing-custom-req-resp-type/tests/test_typing.py).

Here we use the `RichContext` context type, however, we don't touch the annotation of `Request.context`, so the type checker still thinks it is still `falcon.Context`. However, `falcon.Context` is annotated as allowing arbitrary attribute access, so everything works, sort of. But if I misspelled or mistyped, e.g., `context.userid` as `context.user_id` it wouldn't see any issues. But maybe that could still work for you as a starting point?

An alternative is of course, to correctly annotate the `context: RestContext` as you did. If that's the only place you need an ignore, maybe that's even better?

We'll need to document the ramifications of each of the two tradeoffs.

jkmnt:
Yes, I saw this test case. Typechecker thinks the `context` is `Structures.Context` object, so it's true type is effectively erased.
Single annotation of context in subclass is enough to fix it. Moreover, pyright with the `basic` level of strictness allows it without type-ignore. Maybe because such LSP violations are common in Python projects.

In my another app falcon serves html-over-the-wire.  Request and response contexts are used to pass the whole server-side session, flashes, htmx triggers etc. The untyped context is simply not acceptable here. Single `#type: ignore` is cheap )

vytas7:
Yes, agreed, if a single ignore is what's needed, that's great. We'll just have to document both alternatives.

vytas7:
@kgriffs @CaselIT this is now hopefully ready for review!

falcon/_typing.py
```diff
@@ -66,17 +66,31 @@ class _Unset(Enum):
 _UNSET = _Unset.UNSET
 UnsetOr = Union[Literal[_Unset.UNSET], _T]
 
+_ReqT = TypeVar('_ReqT', bound='Request', contravariant=True)
```
CaselIT, line 69: we type these as contravariant to allow a more generic handler to be provided to an app that is typed to use a request subclass for example?

falcon/asgi/app.py
```diff
@@ -96,10 +97,13 @@
 _TYPELESS_STATUS_CODES = frozenset([204, 304])
 
 _FALLBACK_WS_ERROR_CODE = 3011
-_BE = TypeVar('_BE', bound=Exception)
 
+_ExcT = TypeVar('_ExcT', bound=Exception)
+_ReqT = TypeVar('_ReqT', bound=Request, contravariant=True)
+_RespT = TypeVar('_RespT', bound=Response, contravariant=True)
```
CaselIT, line 103: we could import these from _typing

falcon/asgi/app.py
```diff
@@ -354,34 +358,106 @@ async def process_resource_ws(
...
     ws_options: WebSocketOptions
     """A set of behavioral options related to WebSocket connections.
 
     See also: :class:`~.WebSocketOptions`.
     """
 
+    @overload
+    def __init__(
+        self: 'App[Request, Response]',
+        media_type: str = constants.DEFAULT_MEDIA_TYPE,
```
CaselIT: nit: in the overload you can use `...` for all defaults since they don't have any other meaning. but it's fine either way

falcon/asgi/app.py
```diff
@@ -354,34 +358,106 @@ async def process_resource_ws(
...
+        ] = None,
+        router: Optional[routing.CompiledRouter] = None,
+        independent_middleware: bool = True,
+        cors_enable: bool = False,
+        sink_before_static_route: bool = True,
+    ) -> None: ...
+
+    @overload
+    def __init__(
+        self,
```
CaselIT, line 427: is this implicitly understood as `App[_ReqT, _RespT]`?

falcon/asgi/app.py
```diff
@@ -842,7 +918,23 @@ async def watch_disconnect() -> None:
         if resp._registered_callbacks:
             self._schedule_callbacks(resp)
 
-    def add_route(self, uri_template: str, resource: object, **kwargs: Any) -> None:
+    def add_middleware(
```
CaselIT, line 921: nit: we can also place this under `if TYPE_CHECKING`, but it's likely fine either way since adding a fn call here it's ok

tests/test_typing.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+@dataclass
+class RichContext:
+    userid: Optional[UUID] = None
+    role: str = 'anonymous'
+
+
+class FancyRequest(falcon.Request):
+    context_type = RichContext
```
CaselIT, line 20: don't we want to also type context here?

vytas7: See the proposed documentation changes, I haven't found a way to type this without an ignore comment.

But maybe I could leave, say, one of the req/resp as is, and exercise the type ignore on the other one? I can do that.

falcon/_typing.py
vytas7: IIRC if you want to use a Protocol with generics, you have to use contravariant types for method arguments.

cf
```python
from typing import Protocol, TypeVar


class Shape:
    area: float


class Square(Shape):
    side_length: float

    def __init__(self, length: float) -> None:
        self.side_length = length
        self.area = length ** 2


_ShT = TypeVar('_ShT', bound=Shape)


class ShapeManager(Protocol[_ShT]):
    def print_info(self, shape: _ShT) -> None: ...


class SquareInfo:
    def print_info(self, shape: Square) -> None:
        print(f'{shape}: {shape.area=}')


if __name__ == '__main__':
    mgr: ShapeManager[Square] = SquareInfo()
    mgr.print_info(Square(2.0))
```
:arrow_right: `test.py:19: error: Invariant type variable "_ShT" used in protocol where contravariant one is expected  [misc]`

Using `_ShT = TypeVar('_ShT', bound=Shape, contravariant=True)` fixes the error.

tests/test_typing.py
CaselIT: works for me, but we should then have an impl that does use the typed class

falcon/asgi/app.py
vytas7: It does seem to work _as is_, so maybe let's leave it untouched?

```python
from typing import reveal_type

import falcon

class FancyRequest(falcon.Request): ...

class FancyResponse(falcon.Response): ...

app = falcon.App(request_type=FancyRequest, response_type=FancyResponse)
reveal_type(app)
```
```
$ mypy --strict test.py 
test.py:10: note: Revealed type is "falcon.app.App[test.FancyRequest, test.FancyResponse]"
Success: no issues found in 1 source file
```
```
$ pyright test.py 
/tmp/playground/test.py
  /tmp/playground/test.py:10:13 - information: Type of "app" is "App[FancyRequest, FancyResponse]"
0 errors, 0 warnings, 1 information
```

falcon/asgi/app.py
vytas7: Aha, thanks, I can take a look, that could hopefully reduce the spamminess of these overloads.

**Edit**: addressed.

falcon/asgi/app.py
vytas7: I had them under different names there for the async flavour, and maybe it is even clearer to have them defined here. The type checkers seem to have no issues with it either, so I'm, thinking to leave this _as is_ unless you vehemently object.

tests/test_typing.py
vytas7: Tweaked the tests a little bit, we keep the generic `context` type for `FancyRequest`, but `FancyResponse` now uses a `type: ignore` to override context type to `RichContext`. I have verified that it is type checked.

tests/test_typing.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class FancyResponse(falcon.Response):
+    context_type = RichContext
+
+    # NOTE(vytas): the `type: ignore` exemption is currently required if it is
+    #   desirable to actually check typing of context attributes. See also:
+    #   https://falcon.readthedocs.io/en/latest/api/typing.html#known-limitations
```
vytas7, line 28: This is forward-deployed btw :shipit:. The link will only provide the relevant information after merging this.

falcon/asgi/app.py
vytas7: Let's leave it as is, maybe we want to refine a different docstring for the `async` case in the future too.

falcon/asgi/app.py
CaselIT: Ok

vytas7: I think it makes sense. If you don't annotate `self` inside an object method, it is inferred to be of the same type as the class, `App[_ReqT, _RespT]` in this case.

PR #2514, opened 2025-08-13
chore(test_wsgi): increase startup allowance on macos-15

vytas7:


CaselIT (approved):
Maybe the prints could be conditional on a global variable?

vytas7:
> Maybe the prints could be conditional on a global variable?

Dunno, we already have similar prints in `test_wsgi_servers.py` etc, maybe it's not an issue?
`pytest` captures the `stdout` normally anyway, unless instructed otherwise.

PR #2518, opened 2025-08-15
chore(test_wsgi): perform more debug in Actions on Darwin

vytas7:


vytas7:
Closing, I haven't found any approach that actually works, and probably we don't want to wait 60s for it to come up either.

Interesting enough, Uvicorn works in other tests when started via a separate process (not using `multiprocessing`), so maybe that's the way to go. I tested running in a thread, and it didn't work either.

So just keep skipping for now seems the most reasonable option until this is understood.

PR #2521, opened 2025-08-27
perf(uri): speed up URI encoding by hand-crafted Cython code

### LINKED ISSUE #1169: Consider using percentcoding to speed up encoding speeds under CPython
See also: https://github.com/acg/python-percentcoding

vytas7:
The URI encoding functions [encode()](https://falcon.readthedocs.io/en/stable/api/util.html#falcon.uri.encode), [encode_check_escaped()](https://falcon.readthedocs.io/en/stable/api/util.html#falcon.uri.encode_check_escaped), [encode_value()](https://falcon.readthedocs.io/en/stable/api/util.html#falcon.uri.encode_value), and [encode_value_check_escaped()](https://falcon.readthedocs.io/en/stable/api/util.html#falcon.uri.encode_value_check_escaped) have been reimplemented in Cython, leading to a significant speed bump under CPython (provided Falcon is [installed](file:///home/vytas/progs/falcon/docs/_build/html/user/install.html#install) from a binary wheel, or cythonized from a source distribution).

The performance improvement can reach an order of magnitude, although the actual numbers vary depending on the input data. For instance, on CPython 3.12, calling [encode_value()](https://falcon.readthedocs.io/en/stable/api/util.html#falcon.uri.encode_value) with `'no-reply@falconframework.org'` would be sped up ~9x, and using the already encoded value `'no-reply%40falconframework.org'` with [encode_value_check_escaped()](https://falcon.readthedocs.io/en/stable/api/util.html#falcon.uri.encode_value_check_escaped) could yield a ~12x speed boost. (#1169)

Closes #1169.

### DIFF
--- a/falcon/cyutil/uri.pyx
+++ b/falcon/cyutil/uri.pyx
@@ -17,7 +17,7 @@ from cpython.mem cimport PyMem_Malloc, PyMem_Free
 from libc.string cimport memcpy
 
 
-cdef void _init_hex_table(int* data):
+cdef void _init_hex_chars_table(int* data):
     cdef Py_ssize_t index
 
     for index in range(0x10000):
@@ -31,19 +31,33 @@ cdef void _init_hex_table(int* data):
                 pass
 
 
+cdef void _init_hex_encoded_table(int* data):
+    cdef Py_ssize_t index
+    cdef unsigned char b1
+    cdef unsigned char b2
+
+    for index in range(0x100):
+        b1, b2 = f'{index:02X}'.encode()
+        data[index] = (b1 << 8) + b2
+
+
 # PERF(vytas): Cache hex characters lookup table
-cdef int[0x10000] HEX_CHARS
-_init_hex_table(HEX_CHARS)
+cdef int[0x10000] _HEX_CHARS
+_init_hex_chars_table(_HEX_CHARS)
+
+# PERF(vytas): Encoded hex characters.
+cdef int[0x100] _HEX_ENCODED
+_init_hex_encoded_table(_HEX_ENCODED)
 
 # PERF(vytas): Cache an empty string object.
 cdef EMPTY_STRING = u''
 
 
-cdef inline int cy_decode_hex(unsigned char nibble1, unsigned char nibble2):
-    return HEX_CHARS[(nibble1 << 8) | nibble2]
+cdef inline int _cy_decode_hex(unsigned char nibble1, unsigned char nibble2):
+    return _HEX_CHARS[(nibble1 << 8) | nibble2]
 
 
-cdef unicode cy_decode(unsigned char* data, Py_ssize_t start, Py_ssize_t end,
+cdef unicode _cy_decode(unsigned char* data, Py_ssize_t start, Py_ssize_t end,
                        Py_ssize_t encoded_start, bint unquote_plus):
     # PERF(vytas): encoded_start being -1 signifies that the caller
     #   (cy_parse_query_string) has already verified that no encoding
@@ -81,7 +95,7 @@ cdef unicode cy_decode(unsigned char* data, Py_ssize_t start, Py_ssize_t end,
 
             # NOTE(vytas): Else %
             if pos < end - 2:
-                decoded = cy_decode_hex(data[pos+1], data[pos+2])
+                decoded = _cy_decode_hex(data[pos+1], data[pos+2])
                 if decoded < 0:
                     continue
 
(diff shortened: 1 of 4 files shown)


falcon/cyutil/uri.pyx
```diff
@@ -31,19 +31,33 @@ cdef void _init_hex_table(int* data):
...
 
 
+cdef void _init_hex_encoded_table(int* data):
+    cdef Py_ssize_t index
+    cdef unsigned char b1
+    cdef unsigned char b2
+
+    for index in range(0x100):
+        b1, b2 = f'{index:02X}'.encode()
+        data[index] = (b1 << 8) + b2
```
CaselIT, line 41: don't we need to cast to int at least one of the two? or `b1 << 8` does automatically cast a char (1byte) to short(2 bytes)/int(4bytes)?

falcon/cyutil/uri.pyx
```diff
@@ -252,14 +266,159 @@ cdef cy_parse_query_string(unsigned char* data, Py_ssize_t length,
...
+    cdef bint already_escaped = check_is_escaped
+
+    # PERF(vytas): Here we unroll the code for different values of is_value.
+    #   This could by DRYed at a minor perf cost, but, OTOH, these
+    #   redundant blocks are not too large anyway.
+    if is_value:
+        for pos in range(length):
+            # NOTE(vytas): See _UNRESERVED in uri.py.
+            if data[pos] not in (b'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
+                                 b'0123456789-._~'):
```
CaselIT, line 282: are we sure that a set is not faster? in python it is significantly faster

need to use `ord(data[pos])` in case

CaselIT: same in the other places. it would also simplify drying this

falcon/cyutil/uri.pyx
vytas7: No, this works even in pure Python. Iterating over a bytestring yields bytes (`int` in pure Python), not bytestrings of 1 char:
```python
>>> b1, b2 = b'AB'
>>> b1
65
>>> b2
66
```

falcon/cyutil/uri.pyx
vytas7: No, I'm pretty sure a set is **not** faster. Because in order to test for set inclusion, you would need to compute Python hash etc.

Meanwhile, these statements with a constant get compiled into C `switch` statements, which are usually implemented by a local jump table or other kind of wicked techniques. The whole inclusion test can be often done in just a couple of CPU cycles.

falcon/cyutil/uri.pyx
CaselIT: ok, but we have typed b1 and b2 as unsiched chars, so one byte.

then we do << 8 on it. that would overflow if it's not promoted to an int. all I'm saying is that maybe we could add `data[index] = (<int>b1 << 8) + b2` just to be on the safe side. It's seemingly already doing that since the test work, so it should not have any impact on perf

falcon/cyutil/uri.pyx
CaselIT: oh, have you looked at the c output? that was just a though

vytas7: Yeah, I also read about it before how Cython can optimize this kind of inclusion test.

Cf
```cython
HEX_CHARS = frozenset(b'0123456789ABCDEFabcdef')


def count_chars_in1():
    cdef unsigned char c
    cdef int result = 0

    for c in range(0xFF):
        if c in b'0123456789ABCDEFabcdef':
            result += 1

    return result


def count_chars_in2():
    cdef unsigned char c
    cdef int result = 0

    for c in range(0xFF):
        if c in HEX_CHARS:
            result += 1

    return result
```

You wouldn't get this result in pure Python, but when cythonized, the first function is ~14x faster.

CaselIT: hanks for checking

falcon/cyutil/uri.pyx
vytas7: No, this function has no impact on perf at all, because it is used to prepare a lookup table in advance.

I see what you mean, no, it shouldn't be a problem, these kind of expressions are normally computed at least as 32 bit. But I can change the types just to stay on the safe side, yes!

CaselIT: it's likely just me that's rusty in c semantics. given that everything works it's likely as you say

vytas7: Yeah, but things like that may vary depending on the compiler, OS, and architecture, so I like the idea of fixing it properly.

CaselIT (approved):

PR #2523, opened 2025-08-31
refactor(typing): revert ellipsizing of defaults in `App.__init__` overloads

vytas7:
It seems that Sphinx chooses to only show overloads instead of the main definition, leading to all defaults becoming ellipses: https://falcon.readthedocs.io/en/latest/api/app.html#falcon.App.

Alternatively, we can close this PR if we find a Sphinx option not to show overloads at all.


vytas7:
Superseded by #2524.

PR #2528, opened 2025-09-06
chore(platform): drop Python 3.8 support

### LINKED ISSUE #2527: Drop Python 3.8 support
It seems we now may be forced to drop 3.8 support since `setuptools` does their best to make our life miserable otherwise (see #2451 & #2452).

On that note, we should explore alternatives to `setuptools`, however, it seems that Cython is best suited to work with it, so not sure if there are any elegant choices...

We have announced from the beginning that the 4.x series only guarantees SemVer support for Python 3.10+, so we should be in the clear at least from that point of view.

vytas7:
Closes #2527.

### DIFF
--- a/.github/workflows/test-other.yaml
+++ b/.github/workflows/test-other.yaml
@@ -5,10 +5,6 @@ on:
   push:
     branches:
       - "*"
-  # TODO(vytas): Remove from the default run once verified in a PR.
-  pull_request:
-    branches:
-      - master
 
 jobs:
   run-tox:
--- a/.github/workflows/test-other.yaml
+++ b/.github/workflows/test-other.yaml
@@ -5,20 +5,25 @@ on:
   push:
     branches:
       - "*"
+  # TODO(vytas): Remove from the default run once verified in a PR.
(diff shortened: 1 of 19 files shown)


CaselIT (dismissed):
Lgtm

vytas7:
> Lgtm

Thanks for checking @CaselIT! But I still need to update the Docker containers... We should probably publish them too, it seems we have fallen behind a fair bit there...

CaselIT:
Yes, I meant that it makes sense to me to drop it, since it's a bit of a lost fight to try to keep support for outdated version

vytas7:
I have now updated the Docker containers, and we should publish new benchmarks, but it would be with 3.10 (or maybe 3.11 is doable), because unfortunately `vmprof` and other stuff still doesn't support higher than that.

We will need to overhaul benchmarking to include ASGI anyway.

Meinheld support is on the lifeline too, because it only supports 3.9 from what's left on our side. Meinheld is unfortunately abandoned, it seems.

CaselIT (approved):

PR #2529, opened 2025-09-06
refactor: upgrade typing (and other) syntax

vytas7:
This PR primarily upgrades the typing syntax to largely avoid stuff like `typing.Union`, `typing.Optional`, etc, but also a set of other [`pyupgrade` ("UP")](https://docs.astral.sh/ruff/rules/#pyupgrade-up) rules was applied to the codebase.


examples/things_advanced.py
```diff
@@ -194,7 +194,7 @@ def on_post(self, req, resp, user_id):
         proper_thing = self.db.add_thing(doc)
 
         resp.status = falcon.HTTP_201
-        resp.location = '/%s/things/%s' % (user_id, proper_thing['id'])
+        resp.location = '/{}/things/{}'.format(user_id, proper_thing['id'])
```
CaselIT, line 197: If you run pyupgrade again it should convert format to f-string. It can't go directly to f-string when standing from %s for some reason

CaselIT (approved):
Looks ok, just some formats that could be moved to f-strings

examples/things_advanced.py
vytas7: For some strange reason, it doesn't want to upgrade this one, no.

Maybe because it has a quotation mark in `proper_thing['id']`, or I actually don't know why.

CaselIT: Interesting. They are also some couple of other ones that weren't upgraded. Overall it's fone either watfor me

vytas7: The ones in the main "production" module (`falcon/*`) I excluded from these rules for now, if there are any in the critical code paths, we would need to check performance, and other implications.

Although f-strings should easily outperform `.format()`, I haven't really benchmarked vs plain concatenation with `+`, and `%` formatting. So I don't want to just upgrade blindly for now. They are marked as unsafe in Ruff too IIRC.(Also, ugly as it is, but `%` works on bytestrings too; but this rule wouldn't catch that.)

CaselIT: That was on the example module. We can do a search and replace manually later I guess

vytas7: Yes, so it is covered by the automatic rule. It may be a bug in Ruff, or something related to the use of quotes.

vytas7: @CaselIT it is simply an open issue in Ruff: https://github.com/astral-sh/ruff/issues/2031.

PR #2540, opened 2025-10-02
feat(request): add delimiter support to get_param_as_list

### LINKED ISSUE #2538: Add delimiter support to `req.get_param_as_list()`
As we are planning to add OpenAPI spec affordances down the 4.x road, we'll need to make adjustments to our request utility methods.

Add a new _delimiter_ keyword arg to [`req.get_param_as_list()`](https://falcon.readthedocs.io/en/latest/api/request_and_response_wsgi.html#falcon.Request.get_param_as_list).
If provided (otherwise default `None`), the supported values should be:
* `','`: comma separated
* `' '`: space separated, same as [`spaceDelimited`](https://spec.openapis.org/oas/latest.html#style-values) in the spec
* `'|'`: pipe separated, same as [`pipeDelimited`](https://spec.openapis.org/oas/latest.html#style-values) in the spec

Should we also accept string atoms `'spaceDelimited'` and `'pipeDelimited'` directly as in the spec?

If provided, the splitting on the delimiter will always occur after percent-decoding, document this. The examples in the spec also illustrate that spaces and pipes are percent-encoded. Compare and contrast this with the request option to pre-split on comma.

bricklayer25:
Add a new delimiter keyword argument to 
req.get_param_as_list to support splitting query parameter values on characters other than a comma.

Closes #2538 

# Summary of Changes

This PR introduces a new delimiter keyword argument to the req.get_param_as_list() method.

This enhancement allows for parsing list-style query parameters that use delimiters other than the default comma (e.g., spaces, pipes). The primary motivation is to better support spaceDelimited and pipeDelimited styles as described in the OpenAPI specification.

The splitting logic is only applied when the parameter value is a single string, which preserves the framework's existing behavior for handling multiple instances of the same parameter in a query string.

# Related Issues

Closes #2538 

# Pull Request Checklist

- [x] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/latest/tutorial.html#creating-news-fragments) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)

### DIFF
--- a/docs/_newsfragments/2538.newandimproved.rst
+++ b/docs/_newsfragments/2538.newandimproved.rst
@@ -1,0 +1,1 @@
+req.get_param_as_list() now supports a delimiter argument for splitting values.
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1943,6 +1943,7 @@ class Request:
         *,
         required: Literal[True],
         store: StoreArg = ...,
+        delimiter: str | None = None,
         default: list[str] | None = ...,
     ) -> list[str]: ...
 
@@ -1953,6 +1954,7 @@ class Request:
         transform: Callable[[str], _T],
         required: Literal[True],
         store: StoreArg = ...,
+        delimiter: str | None = None,
         default: list[_T] | None = ...,
     ) -> list[_T]: ...
 
@@ -1963,6 +1965,7 @@ class Request:
         transform: None = ...,
         required: bool = ...,
         store: StoreArg = ...,
+        delimiter: str | None = None,
         *,
         default: list[str],
     ) -> list[str]: ...
@@ -1974,6 +1977,7 @@ class Request:
         transform: Callable[[str], _T],
         required: bool = ...,
         store: StoreArg = ...,
+        delimiter: str | None = None,
         *,
         default: list[_T],
     ) -> list[_T]: ...
@@ -1985,6 +1989,7 @@ class Request:
         transform: None = ...,
         required: bool = ...,
         store: StoreArg = ...,
+        delimiter: str | None = None,
         default: list[str] | None = ...,
     ) -> list[str] | None: ...
 
@@ -1995,6 +2000,7 @@ class Request:
         transform: Callable[[str], _T],
         required: bool = ...,
         store: StoreArg = ...,
+        delimiter: str | None = None,
         default: list[_T] | None = ...,
     ) -> list[_T] | None: ...
 
@@ -2004,6 +2010,7 @@ class Request:
         transform: Callable[[str], _T] | None = None,
         required: bool = False,
         store: StoreArg = None,
+        delimiter: str | None = None,
         default: list[_T] | None = None,
     ) -> list[_T] | list[str] | None:
         """Return the value of a query string parameter as a list.
@@ -2021,6 +2028,10 @@ class Request:
             name (str): Parameter name, case-sensitive (e.g., 'ids').
 
         Keyword Args:
+            delimiter(str): An optional character for splitting a parameter
+                value into a list. Useful for styles like ``spaceDelimited``
+                or ``pipeDelimited``. If not provided, default list parsing
+                applies.
             transform (callable): An optional transform function
                 that takes as input each element in the list as a ``str`` and
                 outputs a transformed element for inclusion in the list that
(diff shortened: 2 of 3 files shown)


docs/_newsfragments/2538.newandimproved.rst
```diff
@@ -0,0 +1 @@
+req.get_param_as_list() now supports a delimiter argument for splitting values.
```
vytas7: It would be nicer to actually link to the request method in question, probably the below could do (untested)
```rst
:meth:`falcon.Request.get_param_as_list`
```

falcon/request.py
```diff
@@ -2021,6 +2028,10 @@ def get_param_as_list(
             name (str): Parameter name, case-sensitive (e.g., 'ids').
 
         Keyword Args:
+            delimiter(str): An optional character for splitting a parameter
+                value into a list. Useful for styles like ``spaceDelimited``
+                or ``pipeDelimited``. If not provided, default list parsing
```
vytas7: It might be useful to link to the OpenAPI spec in the parentheses `(see also: ...)`.

falcon/request.py
```diff
@@ -2004,6 +2010,7 @@ def get_param_as_list(
         transform: Callable[[str], _T] | None = None,
         required: bool = False,
         store: StoreArg = None,
+        delimiter: str | None = None,
```
vytas7: Let's add the _delimiter_ parameter last. We haven't enforced keyword-only arguments in this method yet (maybe we should :thinking:), so this could otherwise theoretically be perceived as a breaking change if one passes everything as positional arguments.

falcon/request.py
```diff
@@ -2066,6 +2077,16 @@ def get_param_as_list(
         if name in params:
             items = params[name]
 
+            # If a delimiter is specified AND the param is a single string, split it.
+            if delimiter and isinstance(items, str):
```
vytas7: Let's only allow the delimiters from the OpenAPI spec in the first iteration, so `','`, `' '`, and `'|'`.
This will make this change easier to reason about in the terms of possible edge cases and side effects.

falcon/request.py
```diff
@@ -2066,6 +2077,16 @@ def get_param_as_list(
...
+                if delimiter == ' ':
+                    items = items.split(' ')
+                elif delimiter == '|':
+                    items = items.split('|')
+                else:
+                    # For commas and others, also strip whitespace for convenience.
+                    items = [v.strip() for v in items.split(delimiter)]
```
vytas7: I don't think we want undocumented side effects like stripping whitespace, moreover, depending on the value of the delimiter.

vytas7 (changes requested):
Thanks for this PR!
This is a great start, but we need to make a couple of tweaks before merging (see the comments inline).

rushevich:
Sounds good. I'm away from my computer for the next 2 days but I'll be able to take care of those adjustments afterwards.

vytas7:
> Sounds good. I'm away from my computer for the next 2 days but I'll be able to take care of those adjustments afterwards.

Thanks, have a nice weekend!

falcon/request.py
CaselIT: it may makes sense to use a dict like similar to the one below to validate delimiter
```py
{' ': ' ', 'spaceDelimited': ' ', ',': ',', 'commaDelimited': ',', '|': '|', 'pipeDelimited': '|'}
```

(we could likely define this at module level)

tests/test_request_attrs.py
```diff
@@ -1012,6 +1012,35 @@ def test_etag_parsing_helper(self, asgi, header_value):
 
         assert _parse_etags(header_value) is None
 
+    @pytest.mark.parametrize('asgi', [False, True])
+    def test_get_param_as_list_space_delimited(self, asgi):
```
CaselIT: I think we already a fixture called asgi that automatically does true/false

CaselIT (changes requested):
thanks, left some suggestions

falcon/request.py
CaselIT: (not sure if we want to also support `comma, space, pipe` as valid keys. I'm not against it, but they are not in the spec)

vytas7: Personally I would vote for keeping just the bare minimum of supported delimiters needed for the OAS in the first iteration (so just like your mapping, maybe even sans `commaDelimited` since it is not in the spec either, but comma is already default in some compact styles there).

We can add more later if there is need/popular demand.

vytas7 (approved):
Thanks once again @bricklayer25!

I took the liberty of tweaking your changeset somewhat, and went on to address the review comments.

PR #2544, opened 2025-10-07
feat(request): add new method `get_param_as_dict`

### LINKED ISSUE #2542: Add new method `req.get_param_as_dict()`
As we are planning to add OpenAPI spec affordances down the 4.x road, we'll need to make adjustments to our request utility methods.

We should add a helper for getting a param as OAS ["object"](https://spec.openapis.org/oas/v3.2.0.html#style-values), the suggested method name is `req.get_param_as_dict()`:
* **required** – Set to True to raise `HTTPBadRequest` instead of returning None when the parameter is not found (default False).
* **store** – A dict-like object in which to place the value of the param, but only if the param is found (default None).
* **deep_object** – Set to True to use the `deepObject` (as in OAS) format. Note that this would deviate from other param methods as it would aggregate the return value from multiple parameters following the deep object name pattern.
* **delimiter** – TBD (?) When not using **deep_object**, to get elements from delimited list items, depends on #2538.
* **default** – If the param is not found returns the given value instead of None.

We could also name the method `req.get_param_as_object()`, but since other `get_param_as_*` methods follow the Python types, `req.get_param_as_dict()` probably sounds more fitting?

StepanUFL:
Add a new method get_param_as_dict in order to handle OAS "objects"

Closes issue #2542

# Summary of Changes

A new method req.get_param_as_dict() has added to the Request class in falcon/util/request.py

# Related Issues

Closes #2542.

# Pull Request Checklist

- [X] Performed automated tests and code quality checks by [running `tox`](https://falcon.readthedocs.io/en/stable/community/contributing.html#pull-requests).
- [X] Changes (and possible deprecations) have [towncrier](https://towncrier.readthedocs.io/en/latest/tutorial.html#creating-news-fragments) news fragments under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `towncrier --draft` to ensure it renders correctly.)

Hope this is worthy enough for you all! If there's improvements to be made, let me know


falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+        name: str,
+        required: bool = False,
+        deep_object: bool = False,
+        store: StoreArg = None,
+        default: Any | None = None,
+    ) -> Any:
+        """Retrieve a query parameter as a dictionary.
+
+        Supports OpenAPI's deep object style (`param[key]=value`)
+        and list-based key/value pairs (e.g., `param=a,b,c,d`).
```
CaselIT:
```suggestion
        and list-based key/value pairs (e.g., `param=k1,v1,k2,v2`).
```

just to make it clearer

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+
+            if values_list is None:
+                if required:
+                    msg = 'Missing query parameter'
+                    raise errors.HTTPMissingParam(msg)
+                output = default
+
+            else:
+                if len(values_list) % 2 != 0:
+                    msg = 'Invalid parameter format'
```
CaselIT:
```suggestion
                    msg = 'Invalid parameter format, list elements must be even'
```

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
         return val
 
+    def get_param_as_dict(
+        self,
+        name: str,
+        required: bool = False,
+        deep_object: bool = False,
+        store: StoreArg = None,
+        default: Any | None = None,
+    ) -> Any:
```
CaselIT: this is likely `dict[str, str | None] | None` even if we don't want to do the overload like the other ones

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+
+        .. versionadded:: 4.1.0
+        """
+
+        output: dict[str, Any] | None = None
+
+        if deep_object:
+            oc: dict[str, Any] = {}
+            for key, value in self._params.items():
+                if not key.startswith(f'{name}[') and key.endswith(']'):
```
CaselIT: I guess we could optimize this a bit by caching the format, but it's likely not that important

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+                    continue
+                inner = key[len(name) + 1 : -1]
+
+                if isinstance(value, (list, tuple)):
+                    oc[inner] = value[0] if value else None
```
CaselIT: not sure if we want to use `''` instead of `None` here. depending on it we may want to update the returned type here https://github.com/falconry/falcon/pull/2544/files#r2411775142

tests/test_request_params.py
```diff
@@ -0,0 +1,118 @@
+from __future__ import annotations
```
CaselIT: we already have a file for these test, called tests/test_query_params.py. could you move the tests there?

also you can probably check there how those methods are tested

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+                output = default
+
+            else:
+                if len(values_list) % 2 != 0:
+                    msg = 'Invalid parameter format'
+                    raise errors.HTTPInvalidParam(msg, name)
+                output = dict(zip(values_list[::2], values_list[1::2]))
+
+        if store is not None and isinstance(output, dict):
+            store.update(output)
```
CaselIT: I think we should do like in the other methods
```py
if store is not None:
  store[name] = output
```

tests/test_request_params.py
```diff
@@ -0,0 +1,118 @@
...
+def test_store_argument_is_updated():
+    req = DummyRequestParams({'pair': ['a', '1', 'b', '2']})
+    store = {}
+    result = req.get_param_as_dict('pair', store=store)
+    assert result == {'a': '1', 'b': '2'}
+    assert store == {'a': '1', 'b': '2'}
+
+
+def test_deep_object_with_list_values():
+    req = DummyRequestParams({'user[name]': ['Bond'], 'user[id]': ['007']})
```
CaselIT: also test an with a list of multiple objects

and if we decide that an empty list can indeed happen, also an empty list

falcon/request.py
CaselIT: Q: does this case ever happen (empty list)?

CaselIT (changes requested):
thanks for the contribution, I've left a few suggestions

falcon/request.py
StepanUFL: probably not in actual requests... left check in to be safe but can take out

falcon/request.py
vytas7: Another thing to try would be to benchmark this flow vs using a regex, but I doubt a regex would be faster.

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+
+        Returns:
+            dict: The value of the param if it is found. Otherwise, returns
+            ``None`` unless required is ``True``.
+
+        Raises:
+            HTTPBadRequest: A required param is missing from the request, or
+                the value could not be parsed from the parameter.
+
+        .. versionadded:: 4.1.0
```
vytas7: This needs to be the current development version, Falcon [4.2.0](https://falcon.readthedocs.io/en/latest/changes/4.2.0.html), not 4.1.0 (which is already released).

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+                if required:
+                    msg = 'Missing deep object parameter'
+                    raise errors.HTTPMissingParam(msg)
+                output = default
+            else:
+                output = oc
+
+        else:
+            try:
+                values_list = self.get_param_as_list(name)
```
vytas7: Could you add a `TODO` comment here to also support **delimiter** from #2538?

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+            except errors.HTTPBadRequest:
+                msg = 'It could not parse the query parameter'
+                raise errors.HTTPInvalidParam(msg, name) from None
```
vytas7: Does raising `from None` have any special purpose here?
Add a code comment explaining the reasoning if there is any clever trick behind this.

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+                    raise errors.HTTPMissingParam(msg)
+                output = default
+            else:
+                output = oc
+
+        else:
+            try:
+                values_list = self.get_param_as_list(name)
+            except errors.HTTPBadRequest:
+                msg = 'It could not parse the query parameter'
```
vytas7: This "It" here sounds slightly odd, could we try to make it more similar to other error messages?

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+                raise errors.HTTPInvalidParam(msg, name) from None
+
+            if values_list is None:
+                if required:
+                    msg = 'Missing query parameter'
+                    raise errors.HTTPMissingParam(msg)
```
vytas7: [`HTTPMissingParam`](https://falcon.readthedocs.io/en/latest/api/errors.html#falcon.HTTPMissingParam) takes the name of the missing parameter as its first argument, not an error message.

falcon/request.py
```diff
@@ -2334,6 +2334,90 @@ def get_param_as_json(
...
+                output = default
+
+            else:
+                if len(values_list) % 2 != 0:
+                    msg = 'Invalid parameter format, list elements must be even'
```
vytas7: To be precise list elements cannot be even _per se_ unless they are integer numbers. Suggested rewording: "...the number of list elements must be even".

tests/test_request_params.py
```diff
@@ -0,0 +1,118 @@
...
+        name: str,
+        required: bool = False,
+        deep_object: bool = False,
+        store: dict[str, Any] | None = None,
+        default: Any | None = None,
+    ) -> Any: ...
+
+
+# NOTE(StepanUFL): If a better way to make this play well with mypy exists...
+DummyRequestParams.get_param_as_dict = Request.get_param_as_dict  # type: ignore[method-assign, assignment]
```
vytas7: Defining a new fake `DummyRequestParams` is a somewhat unorthodox way to approach this :thinking:.

Maybe we could simulate request, or simply construct a normal `req` object via [`create_req()`](https://falcon.readthedocs.io/en/latest/api/testing.html#falcon.testing.create_req) (and [`create_asgi_req()`](https://falcon.readthedocs.io/en/latest/api/testing.html#falcon.testing.create_asgi_req))?

To that end, we also have a `util` fixture (see `tests/conftest.py`), which can already construct both WSGI and ASGI requests, see `util.create_req(asgi, ...)`.

vytas7 (changes requested):
Thanks for this PR!
The base structure looks good, although I had a handful of nitpicks wrt the implementation (see inline).

falcon/request.py
CaselIT: from my point of view we could keep this check just to be paranoid, but I would use `''` in the else case, so that we can just day that the returned values are strings in all cases

vytas7:
Hi again @StepanUFL, just checking if you are still considering to finish this PR?

falcon/request.py
vytas7: 4.3.0 now...

falcon/request.py
vytas7: Optimized

falcon/request.py
vytas7: Prefixed the check with `NOTE(StepanUFL)`.

tests/test_request_params.py
vytas7: moved and refactored

vytas7 (approved):
Thanks for this PR once again @StepanUFL! I addressed the open issues with "AI", then polished the remaining items by hand.

PR #2545, opened 2025-10-07
chore(CI): use stable πthon for 3.14 gates

vytas7:
Waiting for the `versions-manifest` update in the "setup Python" action.


vytas7:
> ok, but github is flagging an hidden character for some reason

Discussed on Gitter, GitHub doesn't seem to like the `π` character in the branch name. :smiling_imp:

PR #2548, opened 2025-10-08
feat(request): add `get_query_string_as_media()` method

### LINKED ISSUE #2546: Add a new method `req.get_query_string_as_media()`
...or `req.get_query_string_as_content()` 🤔 But it I vote for the former.

OpenAPI spec 3.2 introduces a new location for `Parameter`, [`querystring`](https://spec.openapis.org/oas/latest.html#parameter-locations), where the whole query string is deserialized as `content`.

This new request helper method would do just that.

MannXo:
Add new method to deserialize entire query string as media content. This supports OpenAPI 3.2 Parameter Object with content specification, where the query string is treated as serialized data (e.g., JSON).

The method:
- URL-decodes the query string
- Deserializes using configured media handlers
- Caches results for performance
- Supports both WSGI and ASGI (async) variants
- Includes default_when_empty parameter for graceful error handling

Example usage:
`  # Query string: ?%7B%22numbers%22%3A%5B1%2C2%5D%7D data = req.get_query_string_as_media('application/json') # data == {'numbers': [1, 2]}`

Closes #2546.

# Summary of Changes

# Related Issues

# Pull Request Checklist

- [x] Changes (and possible deprecations) have [towncrier news fragments](https://falcon.readthedocs.io/en/stable/community/contributing.html#changelog) under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `tox -e towncrier`, and inspect `docs/_build/html/changes/` in the browser to ensure it renders correctly.)


examples/query_string_as_media_example.py
vytas7, line 1: Although we love good documentation, and this looks like an excellent walkthrough, maybe it devotes too much attention to this rather niche feature. :thinking:

Moreover, is this script included anywhere?

docs/_newsfragments/2546.newandimproved.rst
vytas7, line 1: This file must be named according to the pattern [`{issue_number}.{fragment_type}.rst`](https://falcon.readthedocs.io/en/stable/community/contributing.html#changelog), and this issue's number is not 9999.

:arrow_right: TL;DR rename this file to `2546.newandimproved.rst`.

falcon/asgi/request.py
```diff
@@ -97,6 +97,8 @@ class Request(request.Request):
     _cached_uri: str | None = None
     _media: UnsetOr[Any] = _UNSET
     _media_error: Exception | None = None
+    _query_string_media: UnsetOr[Any] = _UNSET
```
vytas7: This is a rather new niche feature, that is unlikely to be widely used, at least not directly after release.

So let's not allocate new requests attributes for caching it, simply deserialize it every time anew. It is unlikely to be called many times.

falcon/asgi/request.py
```diff
@@ -645,6 +647,113 @@ async def get_media(self, default_when_empty: UnsetOr[Any] = _UNSET) -> Any:
...
+                self._query_string_media = deserialize_sync(
+                    decoded_query_string.encode('utf-8')
+                )
+            else:
+                # For handlers without sync deserializer, use the sync
+                # deserialize method since the query string is already in
+                # memory as a string
+                from io import BytesIO
+
+                query_stream = BytesIO(decoded_query_string.encode('utf-8'))
```
vytas7: Although Falcon is performance-oriented, as pointed out above, this method is going to be a rather niche feature, so maybe it would be possible to DRY the ASGI branch by calling the sync method to do the heavy lifting?

We could possibly just document that the synchronous handler is always used in this case, even for ASGI.

vytas7 (changes requested):
Thanks for this PR, it looks like an already very advanced implementation! :100:

However, as explained inline, we don't want to perform advanced caching of the deserialized media, at least not in the first iteration, given that this is a rather niche feature (it has been introduced in OpenAPI 3.2.0, previously this scenario was impossible to describe in OAS).

docs/_newsfragments/2546.newandimproved.rst
MannXo: Renamed now.

falcon/asgi/request.py
MannXo: Done. ASGI implementation now delegates to parent: `return request.Request.get_query_string_as_media(self, media_type, default_when_empty`

falcon/asgi/request.py
MannXo: Removed `_query_string_media` and `_query_string_media_error` slots. Method now deserializes fresh on each call.

examples/query_string_as_media_example.py
MannXo: Removed  `query_string_as_media_example.py` for now. Can add back later if/when caching is implemented.

MannXo:
> Thanks for this PR, it looks like an already very advanced implementation! 💯
> 
> However, as explained inline, we don't want to perform advanced caching of the deserialized media, at least not in the first iteration, given that this is a rather niche feature (it has been introduced in OpenAPI 3.2.0, previously this scenario was impossible to describe in OAS).

Thank you @vytas7 for the thorough review and your feedback. I have made some changes, to address your comments. Ready for the next round of review.

falcon/request.py
```diff
@@ -246,6 +246,8 @@ def __init__(
         self.uri_template = None
         self._media: UnsetOr[Any] = _UNSET
         self._media_error: Exception | None = None
+        self._query_string_media: UnsetOr[Any] = _UNSET
+        self._query_string_media_error: Exception | None = None
```
CaselIT: I guess we can now remove these

falcon/request.py
```diff
@@ -1154,6 +1156,76 @@ def get_media(self, default_when_empty: UnsetOr[Any] = _UNSET) -> Any:
...
+
+        # URL-decode the query string
+        decoded_query_string = util.uri.decode(self.query_string, unquote_plus=False)
+
+        # Create a BytesIO stream from the decoded query string
+        query_stream = BytesIO(decoded_query_string.encode('utf-8'))
+
+        try:
+            return handler.deserialize(
+                query_stream, media_type, len(decoded_query_string.encode('utf-8'))
```
CaselIT: we could avoid this re-encoding by saving the bytes before creating the bytesio

falcon/asgi/request.py
```diff
@@ -645,6 +645,64 @@ async def get_media(self, default_when_empty: UnsetOr[Any] = _UNSET) -> Any:
         deserialized_media = await req.media
     """
 
+    async def get_query_string_as_media(
```
CaselIT: I wonder if we should leave this async or meke it sync instead?

All other `get_` methods are sync, but making it sync we would preclude any option of doing io in this method.

@vytas7 thoughs

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # JSON with Unicode
+        json_data = '{"emoji": "🚀", "chinese": "你好"}'
+        query_string = quote(json_data, safe='')
+
+        client.simulate_get('/test', query_string=query_string)
+
+        assert resource.captured_media == {'emoji': '🚀', 'chinese': '你好'}
+
+    def test_error_caching(self, asgi, client):
```
CaselIT: this test is likely outdated since caching was removed

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,427 @@
...
+
+        resource = ResourceCachedErrorAsync() if asgi else ResourceCachedError()
+        client.app.add_route('/test', resource)
+
+        # Empty query string will cause MediaNotFoundError
+        client.simulate_get('/test', query_string='')
+
+        assert resource.result == {}
+
+    def test_cached_error_reraises_without_default(self, asgi, client):
```
CaselIT: outdated too? I guess we can keep it, and just remove the comments regarding caches

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,427 @@
...
+        client.app.add_route('/test', resource)
+
+        json_data = '{"uses": "default"}'
+        query_string = quote(json_data, safe='')
+
+        client.simulate_get('/test', query_string=query_string)
+
+        assert resource.captured_media == {'uses': 'default'}
+
+    def test_cached_error_with_default_when_empty(self, asgi, client):
```
CaselIT: here too let's just remove the comments regarding the cache, the test if fine to have even with the current impl

CaselIT (changes requested):

CaselIT:
great work, thanks!

falcon/request.py
MannXo: Done. Removed `_query_string_media` and `_query_string_media_error`

falcon/request.py
MannXo: Good catch. We now store `query_bytes = decoded_query_string.encode('utf-8')` and reuse it for `BytesIO` and `len()` to avoid double-encoding.

falcon/asgi/request.py
MannXo: Kept it `async` to preserve the option to perform async I/O in the future, but simplified it to delegate to the sync implementation (so no duplicate logic). Docstring updated to explain the rationale. Happy to change it to sync in the next round of review if you decide so.

tests/test_request_query_string_media.py
MannXo: Updated the test comment to reflect current behavior (removed references to caching); the test logic remains and still passes.

tests/test_request_query_string_media.py
MannXo: Removed cache-related comments. test retained as-is.

tests/test_request_query_string_media.py
MannXo: Removed outdated cache comments and kept the test. It continues to validate error + default behavior

MannXo:
> great work, thanks!

Thank you for the review and comments. I have pushed a new commit. Ready for the next round of review.

falcon/asgi/request.py
vytas7: I don't think that reading the query string is ever going to be async, normally the HTTP server reads the whole line into the buffer. It is then handed off to an ASGI app as a string value.

Let's remove the async reimplementation.

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,426 @@
...
+
+import pytest
+
+import falcon
+from falcon import errors
+from falcon import testing
+
+
+@pytest.fixture
+def client(asgi):
```
vytas7: I don't think this is needed, the `util` fixture can do this for you.

falcon/request.py
```diff
@@ -1162,6 +1162,76 @@ def get_media(self, default_when_empty: UnsetOr[Any] = _UNSET) -> Any:
...
+            object: The deserialized media representation of the query string.
+
+        .. _Parameter Object with content:
+            https://spec.openapis.org/oas/latest.html#parameter-object-examples
+
+        """
+        if media_type is None:
+            media_type = self.options.default_media_type
+
+        handler, _, _ = self.options.media_handlers._resolve(
```
vytas7, line 1222: What happens in the case no handler could be found for the requested media type?

As I understand, `_resolve(...)` would raise an instance of `HTTPUnsupportedMediaType`, but do we want it to behave this way? I feel like it would be better to raise another type of non-HTTP error that would by default bubble up as an HTTP 500 (unless handled), because the API client has no control over this media type, it is chosen internally by the service.

What do you think @CaselIT?

vytas7 (changes requested):
Hi again @MannXo, and sorry for taking time to respond to this one.

Your PR looks good for most of it, but IMHO we should remove the async version of the method, plus make a couple of other adjustments before merging (see inline).

MannXo:
Thank you for the comments @vytas7 

I pushed a change including the following:
1- Removed the async method from `request.py`. Now inherits from parent WSGI class.
2- Removed custom fixture and async resource variations. All 19 tests now use `util.create_app()` pattern for both ASGI/WSGI.
3- Added try-except in `request.py` to catch `HTTPUnsupportedMediaType` and raise `ValueError` with a suitable message about missing media handler configuration.


P.S, The `mypy_tests` pipeline failure appears to be a pre-existing issue unrelated to my changes. (I get the same errors running the check on master)

vytas7:
> P.S, The `mypy_tests` pipeline failure appears to be a pre-existing issue unrelated to my changes. (I get the same errors running the check on master)

~~Aye, a new Mypy version... I'll try to take a look at that.~~

**Edit**: still not 100% why we started getting this only now, but it is fun that it actually uncovers a real issue. When using `testtools`, Falcon `TestCase`'s `patch` method is shadowed by `testtools.TestCase.patch` via diamond inheritance...
(If anyone is wondering what the heck is `testtools`, see #2156.)

falcon/request.py
```diff
@@ -1162,6 +1162,83 @@ def get_media(self, default_when_empty: UnsetOr[Any] = _UNSET) -> Any:
...
+
+        """
+        if media_type is None:
+            media_type = self.options.default_media_type
+
+        try:
+            handler, _, _ = self.options.media_handlers._resolve(
+                media_type, self.options.default_media_type
+            )
+        except errors.HTTPUnsupportedMediaType:
```
vytas7: Not critical for the first iteration, but it would be nice to optimize this slightly by passing `raise_not_found=False` to the `_resolve` method, and handling the case where `handler is None` instead. It should be much cheaper than raising one extra time unnecessarily.

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,397 @@
...
+
+class TestQueryStringAsMedia:
+    """Test query string deserialization as media."""
+
+    def test_simple_json_query_string(self, asgi, util):
+        """Test deserializing a simple JSON query string."""
+        resource = CaptureQueryStringMedia()
+        app = util.create_app(asgi)
+        app.add_route('/test', resource)
+        client = testing.TestClient(app)
```
vytas7: Could we DRY this block along the below lines (untested)? :arrow_down:
```python
@pytest.fixture()
def client(asgi, util):
    resource = CaptureQueryStringMedia()
    app = util.create_app(asgi)
    app.add_route('/test', resource)
```

vytas7 (changes requested):
Thanks for the prompt updates @MannXo!
A couple more potential improvements (see inline), and I think we can merge this.

falcon/request.py
MannXo: Done

tests/test_request_query_string_media.py
MannXo: Done

falcon/request.py
```diff
@@ -1164,6 +1164,82 @@ def get_media(self, default_when_empty: UnsetOr[Any] = _UNSET) -> Any:
...
+
+        """
+        if media_type is None:
+            media_type = self.options.default_media_type
+
+        handler, _, _ = self.options.media_handlers._resolve(
+            media_type, self.options.default_media_type, raise_not_found=False
+        )
+        if handler is None:
+            raise ValueError(
```
CaselIT, line 1226: let's mention in the doc string that a value error is raised if no handler is found

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # URL-encoded form data
+        query_string = quote('key1=value1&key2=value2', safe='')
+
+        client.simulate_get('/test', query_string=query_string)
+
+        # URLEncodedFormHandler should parse this
+        assert isinstance(resource.captured_media, dict)
```
CaselIT: we could also / instead compare the value

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # JSON with Unicode
+        json_data = '{"emoji": "🚀", "chinese": "你好"}'
+        query_string = quote(json_data, safe='')
+
+        client.simulate_get('/test', query_string=query_string)
+
+        assert resource.captured_media == {'emoji': '🚀', 'chinese': '你好'}
+
+    def test_error_caching(self, asgi, util):
```
CaselIT, line 170: not sure if this test is useful. it's fine to keep though

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,381 @@
...
+        client, resource = standard_client
+
+        json_data = '[1, 2, 3, 4, 5]'
+        query_string = quote(json_data, safe='')
+
+        client.simulate_get('/test', query_string=query_string)
+
+        assert resource.captured_media == [1, 2, 3, 4, 5]
+
+    def test_boolean_and_null_values(self, standard_client):
```
CaselIT: I would remove, we aren't testing a json serializer here

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,381 @@
...
+
+        client.simulate_get('/test', query_string=query_string)
+
+        assert resource.captured_media == {
+            'active': True,
+            'inactive': False,
+            'empty': None,
+        }
+
+    def test_numeric_values(self, standard_client):
```
CaselIT: same, would remove

tests/test_request_query_string_media.py
```diff
@@ -0,0 +1,381 @@
...
+        app = util.create_app(asgi)
+        app.req_options.media_handlers['application/custom'] = handler
+        app.add_route('/test', resource)
+        client = testing.TestClient(app)
+
+        client.simulate_get('/test', query_string='data')
+
+        assert resource.error_message == 'Custom error'
+
+    def test_uses_default_media_type_when_none_specified(self, asgi, util):
```
CaselIT: same as test_simple_json_query_string, would remove

CaselIT (commented):
thanks for the change!

I looks mostly ok to me, I just just document the error if the media is missing and dry up a bit the tests

falcon/request.py
MannXo: Added `Raises` section

tests/test_request_query_string_media.py
MannXo: Done

tests/test_request_query_string_media.py
MannXo: kept it for now

tests/test_request_query_string_media.py
MannXo: removed

tests/test_request_query_string_media.py
MannXo: removed

tests/test_request_query_string_media.py
MannXo: removed

vytas7 (approved):
Thanks once again @MannXo!
I think this is a good start, and we can merge it as is.

I'm thinking to remove the default value of `media_type`, or make it default to URL-encoded, because that's the most common media type for the query string (see also: https://spec.openapis.org/oas/latest.html#parameter-locations).

But we can make that decision later (before the first stable release of this though).

PR #2555, opened 2025-10-18
feat(platform): provide selected `cp314t` (free-threading) wheels

### LINKED ISSUE #2501: Finalize testing on free-threaded CPython, publish `cp314t` wheels
We also need to decide what to do about thread-safety there in general, do we need to make configuration methods like `App.add_route()` etc fully thread-safe too?

This issue is probably not the top prio though, given that where massive concurrency and throughput are needed, Falcon apps are typically run asynchronously (using ASGI or WSGI+gevent) and/or are scaled via multiple worker processes, containers, server nodes.

vytas7:
This is just a draft PR to get the ball rolling.

- [x] Update the installation docs.
- [x] Update the free-threading FAQ item.
- [x] Update the wheel matrix in the docs.
- [x] Stress test the `cp314t` build using a multithreaded WSGI server.

Fixes #2501.

Although everything _should_ work, it is important to perform extensive testing ricky now that we aas thread-safected by the GIL.ty can be tre no longer prote

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -214,6 +214,9 @@ listed below by date of first contribution:
 * Soubhik Kumar Mitra (x612skm)
 * Sviatoslav Sydorenko (webknjaz)
 * Soney (sonephyo)
+* Ayan Ahmed Khan (AyanAhmedKhan)
+* Parman Mohammadalizadeh (MannXo)
+* Tudor Gradinaru (TudorGR)
 
 (et al.)
 
--- a/docs/changes/4.2.0.rst
+++ b/docs/changes/4.2.0.rst
@@ -41,8 +41,11 @@ Contributors to this Release
 
 Many thanks to all of our talented and stylish contributors for this release!
 
+- `AyanAhmedKhan <https://github.com/AyanAhmedKhan>`__
 - `CaselIT <https://github.com/CaselIT>`__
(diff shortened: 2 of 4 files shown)


CaselIT (approved):

.github/workflows/cibuildwheel.yaml
```diff
@@ -153,6 +153,15 @@ jobs:
           - cp312
           - cp313
           - cp314
+        include:
+          - platform:
+              name: manylinux_x86_64
```
CaselIT, line 158: no arm?

vytas7: No, I was thinking we could start small with just x86. Other platforms can still use the pure Python wheel. Or do you think we should add ARM too?

CaselIT: I mean if there aren't other particular blockers, why not?

vytas7: No, there shouldn't be any except more wheels to produce, and more storage used on PyPI.

vytas7: Fair enough, I have now added ARM to that selection.

CaselIT (approved):

PR #2558, opened 2025-10-31
Improve docstring for _encode_items_to_latin1 function in falcon/util/misc.py

swati-londhe:
### Summary
This PR enhances the documentation of the `_encode_items_to_latin1` utility function
in `falcon/util/misc.py`. The previous docstring incorrectly mentioned "decode"
instead of "encode" and lacked clarity on input/output format.

### Changes
- Updated the docstring for `_encode_items_to_latin1` to accurately describe its purpose.
- Added an example showing expected behavior.

### Why
Improved clarity for contributors and maintainers reading Falcon’s internal utility code.
No functional changes were made.

PR #2561, opened 2025-10-31
Add unit test for unquote_string() utility function

swati-londhe:
This pull request adds new unit tests for the unquote_string() function in tests/test_utils.py.

Key updates:
Added test_unquote_string_basic() to verify correct behavior for:
Properly quoted strings (e.g., "hello")
Escaped quotes inside quoted strings
Unquoted strings (should remain unchanged)

Empty strings
Invalid (non-string) input, ensuring a TypeError is raised

Why this change:
Increases test coverage for Falcon’s utility functions.
Improves code reliability and helps prevent regressions.
Safe and minimal update — affects only the test suite.

Testing:
✅ All tests pass locally using pytest.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -17,6 +17,8 @@ import json
 import random
 from urllib.parse import quote
 from urllib.parse import unquote_plus
+import pytest
+from falcon.uri import unquote_string
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


CaselIT:
see https://github.com/falconry/falcon/pull/2558#issuecomment-3471927735

duplicate of #2562

PR #2579, opened 2025-11-11
docs(hooks): demonstrate how to use own decorator instead of hooks

### LINKED ISSUE #2533: Demonstrate how to use own decorator instead of hooks
> Can I just use my own decorator instead of `@before`?

_Originally posted by @0x1618 in https://github.com/falconry/falcon/discussions/2532_

I think we do mention in the docs that you can use an own decorator, but we never provide a concise example how to actually implement it.

Falcon responders are normal Python methods without any magic dust, but we could still show how to preserve the method's name for routing, etc.

0x1618:
Closes #2533.

### DIFF
--- a/docs/api/hooks.rst
+++ b/docs/api/hooks.rst
@@ -99,6 +99,62 @@ requests.
     are intended to hold request and response data specific to your app as it
     passes through the framework.
 
+You can also create your own decorators to wrap responder methods without needing to use hooks.
+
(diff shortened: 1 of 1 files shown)


vytas7 (approved):
Thanks, this LGTM :+1:
Going forward, we probably want to factor out larger snippets into separate files, and use `literalinclude`, but the majority of docs is just inline code atm.

PR #2584, opened 2025-11-22
fix(asgi.Request): handle `scope['client']` being `None`

### LINKED ISSUE #2583: Uvicorn may explicitly set `'client': None`, breaking `req.remote_addr`
Uvicorn may explicitly set `'client': None`, breaking `req.remote_addr`:
```
  File "<venv>/falcon/asgi/request.py", line 496, in access_route
    client, __ = self.scope['client']
    ^^^^^^^^^^
TypeError: cannot unpack non-iterable NoneType object
```

I have also opened a discussion in Uvicorn whether this behaviour is in the spirit of the spec (my interpretation is that it is not, but the spec is very unclear on this). (https://github.com/Kludex/uvicorn/discussions/2761)

vytas7:
Fixes #2583.

### DIFF
--- a/docs/_newsfragments/2583.bugfix.rst
+++ b/docs/_newsfragments/2583.bugfix.rst
@@ -1,0 +1,9 @@
+Due to the differences in interpretation of the ASGI specification, Uvicorn
+could set `client` in the HTTP connection scope in a way that broke
+:meth:`req.access_route <falcon.asgi.Request.access_route>` and
+:meth:`req.remote_addr <falcon.asgi.Request.remote_addr>` when the app server
+was bound to a Unix domain socket.
+
+This discrepancy was addressed, and Falcon should now fall back to the same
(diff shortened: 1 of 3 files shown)


vytas7:
> good catch
> 
> question: can we expect it to be present in most cases? because otherwise a `if client := self.scope.get('client'):` may be a better option

I found it the hard way, by trying to actually build something :slightly_smiling_face:

We do expect it to be present in the absolute majority of cases, yeah. It may be missing only when binding to a Unix domain socket, but even in that case it supports mapping the address from `X-Forwarded-For`.

Moreover, if you are actually interested in `req.remote_addr`, you normally set up something rather than getting the `127.0.0.1` fallback in 100% of cases. So this issue might be well seen as cosmetic for all practical purposes. But still we want to avoid unhandled tracebacks like that.

PR #2591, opened 2026-01-11
feat(typing): parametrize App by default req/resp types on 3.13+

### LINKED ISSUE #2586: typing changes in 4.2 require code changes
After upgrading one of our projects to 4.2, our type checker had a small nervous breakdown.

In our testing suite, we have some fixtures that return an App instance, annotated as bare `falcon.App`. With the generics introduced in #2513 these now need to be annotated as `falcon.App[falcon.Request, falcon.Response]` which is slightly cumbersome.
This also applies to other things like In our case the `SyncMiddleWare` type which appears to have become impossible to use as it lacks a way to inject generics (we're still annotating a list of middlewares as `list[falcon._typing.SyncMiddleWare]` but that is just for documentation I guess.)

It looks like it is possible to make those `_ReqT` etc typevars have a default value of `Request`, which would remove the need to explicitly add those types to the annotation when referring to `falcon.App` with default request/response types, but that default argument is only available with Python 3.13+ so not in all versions supported by falcon.

One solution that does not require waiting for 3.13 to be baseline is to implement a version check and add these defaults only when on 3.13+, so something like
```
if sys.version >= (3,13):
  _ExcT = TypeVar('_ExcT', bound=Exception, default=Exception)
  _ReqT = TypeVar('_ReqT', bound=Request, contravariant=True, default=Request)
  _RespT = TypeVar('_RespT', bound=Response, contravariant=True, default=Response)
else:
  _ExcT = TypeVar('_ExcT', bound=Exception)
  _ReqT = TypeVar('_ReqT', bound=Request, contravariant=True)
  _RespT = TypeVar('_RespT', bound=Response, contravariant=True)
```

I also saw that identical typevars exist in _typing.py ; maybe it is a good idea to import them from there in app.py so they only need to be defined once -- or is there some coding style rule in place that forbids importing types marked as private?

Anyway, I'll whip up one or a bunch of PRs addressing these things.

vytas7:
Thanks for the idea @jap, I took a stab @ implementing this.

Fixes #2586.

### DIFF
--- a/docs/user/tutorial-asgi.rst
+++ b/docs/user/tutorial-asgi.rst
@@ -83,7 +83,7 @@ include:
 For a simple tutorial application like ours, any of the above should do.
 Let's pick the popular ``uvicorn`` for now::
 
-  $ pip install uvicorn
+  $ pip install uvicorn[standard]
 
 See also: :ref:`ASGI Server Installation <install_asgi_server>`.
 
--- a/falcon/inspect.py
+++ b/falcon/inspect.py
@@ -382,7 +382,7 @@ class ErrorHandlerInfo(_Traversable):
     """Describes an error handler.
 
     Args:
-        error (name): The name of the error type.
+        error (str): The name of the error type.
         name (str): The name of the handler.
         source_info (str): The source path where this error handler was defined.
         internal (bool): Whether or not this is a default error handler added by


falcon/app.py
```diff
@@ -1167,7 +1164,7 @@ def _get_responder(
                 if m:
                     if is_sink:
                         params = m.groupdict()  # type: ignore[union-attr]
-                    responder = obj
+                    responder = obj  # type: ignore[assignment,unused-ignore]
```
CaselIT, line 1167: why is this ignore needed?

vytas7: Our typing is inexact, and not parametrized properly either, but this is masked on <=3.12 by earlier ignores/casts. On, 3.13 it applies the default leading to further mismatch, as I understand.

CaselIT: Oh ok, so we have a type warning only in some python version.

vytas7: Yes, only in 3.13+

PR #2595, opened 2026-01-18
feat(asgi): no longer add `NullHandler` to the `falcon` logger

### LINKED ISSUE #2594: falcon.asgi.App() shows 500 error instead of async exception
Trying to convert https://github.com/lucianmarin/subreply to ASGI.

vytas7:
We used to add a `NullHandler()` to `falcon._logger` which *could* be done according to the stdlib's docs,
> "*if* you want to prevent your library's logged events being output to `sys.stderr` in the absence of logging configuration."

However, this has mostly resulted in confusion for people trying the ASGI flavour of the framework as HTTP 500 tracebacks may disappear completely, so the revised choice is NOT to prevent last resort logging to `sys.stderr`.

Fixes #2594.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 recursive-include docs *.rst *.md *.html *.ico *.png *.py *.svg
 recursive-include e2e-tests *.py *.css *.html *.js
-recursive-include examples *.py
+recursive-include examples *.py *.yaml
 recursive-include falcon *.pyx
 recursive-include tests *.py *.pyx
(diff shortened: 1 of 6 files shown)


vytas7:
> I think it's fine to keep as part of a point release.

Aye, don't think it's an issue SemVer wise since it's an implementation detail of the private `falcon._logger`. We only specify that we use the `falcon` logger in rare cases where there is no other option, and that we don't configure anything behind the user's back (not any less true than before).

PR #2613, opened 2026-04-03
chore(E2E): use bounded Queue in Hub

tang-vu:
## Problem

The `Emitter` class in the E2E test server initializes an `asyncio.Queue` without a `maxsize`. In the `Hub.broadcast` method, messages are enqueued for all active emitters. If a client (SSE consumer) stops reading but the connection remains open, the queue will grow indefinitely, potentially leading to a memory leak in the test server.

**Severity**: `medium`
**File**: `e2e-tests/server/hub.py`

## Solution

Set a reasonable `maxsize` for the queue and handle potential `asyncio.QueueFull` errors or use `put_nowait` with a drop-oldest strategy.

## Changes

- `e2e-tests/server/hub.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed

### DIFF
--- a/e2e-tests/server/hub.py
+++ b/e2e-tests/server/hub.py
@@ -15,7 +15,7 @@ class Emitter:
 
     def __init__(self) -> None:
         self._done: bool = False
-        self._queue: asyncio.Queue[SSEvent] = asyncio.Queue()
+        self._queue: asyncio.Queue[SSEvent] = asyncio.Queue(maxsize=100)
 
     async def events(self) -> typing.AsyncGenerator[SSEvent | None, None]:
         try:
@@ -36,7 +36,10 @@ class Emitter:
 
     async def enqueue(self, message: str) -> None:
         event = SSEvent(text=message, event_id=str(uuid.uuid4()))
-        await self._queue.put(event)
+        try:
+            self._queue.put_nowait(event)
+        except asyncio.QueueFull:
+            pass
 
     @property
     def done(self) -> bool:
--- a/e2e-tests/server/hub.py
+++ b/e2e-tests/server/hub.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import typing
 import uuid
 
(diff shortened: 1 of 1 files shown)


CaselIT:
Hi,

Thanks for the contribution, while the change seems fine, I'm not sure how useful it will be, since that's a test harness that's unlikely to ever cause issues

cc @vytas7 what do you think? MAybe we could accept if we were to add a comment to indicate that while not important here it's important in other contexts

vytas7:
I don't know... I do like the spirit of writing safe(r) code despite the context, but, OTOH, this is just a test app... 

Moreover, the description looks AI generated, and our official PR template has been removed. If we do make this change, let's remove error handling when the queue is full, or we need to at least log the exception. Silently swallowing exceptions is usually an antipattern.

vytas7 (changes requested):
As said before, we could consider merging this if you refactor the code not to swallow errors silently.

We have had strange E2E errors against Firefox, so it should be good to see more what's going on if the cause turns out to be unbounded queue (it wasn't in that specific case).

tang-vu:
@vytas7 @CaselIT Thanks for the feedback! I've updated the PR to address both concerns:

e2e-tests/server/hub.py
```diff
@@ -39,7 +46,14 @@ async def enqueue(self, message: str) -> None:
         try:
             self._queue.put_nowait(event)
         except asyncio.QueueFull:
-            pass
+            # NOTE(vytas): Drop message and log the incident.
```
vytas7: Please don't impersonate me here in the notes. Use your own nickname.

e2e-tests/server/hub.py
```diff
@@ -39,7 +46,14 @@ async def enqueue(self, message: str) -> None:
...
+            # In production, consider implementing backpressure or
```
vytas7: This is a test, not a tutorial or documentation item. Not sure if we need to mention production is here.

e2e-tests/server/hub.py
```diff
@@ -9,13 +10,19 @@
...
+logger = logging.getLogger(__name__)
+
 
 class Emitter:
     POLL_TIMEOUT = 3.0
 
     def __init__(self) -> None:
         self._done: bool = False
-        self._queue: asyncio.Queue[SSEvent] = asyncio.Queue(maxsize=100)
+        # NOTE: Bounded queue prevents unbounded memory growth when SSE
```
vytas7: Please prefix with `# NOTE(tang-vu): `.

Indent the following lines to be consistent with other parts of the codebase.

vytas7 (changes requested):

tang-vu:
@vytas7 Thanks for catching those issues! I've fixed:

vytas7 (approved):
Okay, I can merge this with some tweaks, it doesn't really hurt. Thanks @tang-vu!

PR #2618, opened 2026-04-07
Security: Unsafe tarfile extraction allows path traversal (CVE-2007-4559)

barttran2k:
## Problem

The `tarfile.extractall()` method is called without any path validation or filtering. A maliciously crafted tar archive could contain entries with absolute paths or `../` sequences, allowing files to be written outside the intended target directory. This is a well-known vulnerability (CVE-2007-4559).

**Severity**: `high`
**File**: `tools/tox_sdist.py`

## Solution

Use the `filter='data'` parameter (Python 3.12+) or manually validate each member's path before extraction: `targz.extractall(target, filter='data')`. For older Python versions, iterate members and check that resolved paths stay within the target directory.

## Changes

- `tools/tox_sdist.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed

### DIFF
--- a/tools/tox_sdist.py
+++ b/tools/tox_sdist.py
@@ -28,6 +28,11 @@ def build_sdist():
 
 def extract_sdist(archive, target):
     with tarfile.open(archive) as targz:
+        resolved_target = target.resolve()
+        for member in targz.getmembers():
+            member_path = (resolved_target / member.name).resolve()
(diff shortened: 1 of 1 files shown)


CaselIT:
Hi, 

This is an internal script that's called with safe arguments.

Please review if a pr makes sense before blindly spamming ai generated prs, thanks

PR #2648, opened 2026-05-28
chore(typing): align async and sync middleware params type

CaselIT:
just a minor typing fix


vytas7 (approved):
Looks good, as I told you, I was typing stuff at work and was wondering why we needed an abstract `Mapping` there instead of `dict` :)

PR #2673, opened 2026-07-02
fix: omit Allow header in default OPTIONS responder when no explicit methods are mapped

### LINKED ISSUE #2671: The default options responder does not take into consideration on_request, leading to allow=''
When using `on_request` in a resource, the default options header will reply with an empty allow header.
This is wrong since the [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Allow) says:

> An empty Allow value indicates that the resource allows no request methods, which might occur temporarily for a given resource.

sadly there does not seem to be a `*` to indicate all methods, but we can likely use the COMBINED_METHODS list in that case.

MeGaurav4:
When a resource only defines `on_request` (catch-all responder), `map_http_methods` finds no explicit methods, so `set_default_responders` computes an empty `allowed_methods` list. The default OPTIONS responder then sets `Allow:` (empty), which per spec means "no methods allowed."

Fix: skip setting the `Allow` header when there are no explicitly mapped methods. The resource responds to all methods via `on_request`, but since HTTP has no "all methods" token for `Allow`, omitting the header is the correct behavior (clients can try any method).

Closes #2671


vytas7 (changes requested):
Thanks, the fix LGTM.

Since a real bug was fixed, could you add a newsfragment for this change?

chbndrhnns (approved):
LGTM

vytas7 (approved):
Thanks once again @MeGaurav4, and thanks @chbndrhnns for reviewing this!

I took the liberty of renaming the proposed newsfragment to add the correct Rst extension, as well as repointed it to the issue itself instead of this PR.

PR #2674, opened 2026-07-04
fix(routing): implement proper HEAD support for static routes

### LINKED ISSUE #2337: Proper `HEAD` support for static routes

zain-asif-dev:
## Summary

Closes #2337.

Static routes (`StaticRoute`/`StaticRouteAsync`, as configured via `App.add_static_route()`) didn't have proper `HEAD` support:

- A `HEAD` request was handled exactly like `GET`: the responder always opened the underlying file via `io.open()`/`os.fstat()` and set it as `resp.stream`, even though a `HEAD` response never includes a body. Since the body is discarded later in the response pipeline (`resp.stream` is never iterated for `HEAD`), the opened file handle was never closed — every `HEAD` request to a static route leaked a file descriptor.
- Any other HTTP method (`POST`, `PUT`, `DELETE`, etc.) was also accepted and served the file's contents, instead of being rejected.
- The `Allow`/`Access-Control-Allow-Methods` header returned for `OPTIONS` requests only advertised `GET`.

### Changes

- `HEAD` requests are now served from `os.stat()` directly, without ever opening a file handle, while still setting the same headers a `GET` would (`Content-Length`, `ETag`, `Last-Modified`, `Accept-Ranges`, and `Content-Range`/`206` for ranged requests).
- Added `_compute_head_range()`, which mirrors the offset/length arithmetic of the existing `_set_range()` helper, but operates purely on the file size (no file handle required).
- Requests with methods other than `GET`, `HEAD`, and `OPTIONS` now get a `405 Method Not Allowed` response (via `falcon.HTTPMethodNotAllowed(['GET', 'HEAD'])`) instead of silently being served the file.
- The `OPTIONS` responder now sets `Allow: GET, HEAD` instead of `Allow: GET`.
- Extracted the "resolve file honoring `fallback_filename`" logic into a small `_locate()` helper shared between the `GET` and `HEAD` code paths, to keep `__call__()`'s complexity in check.

## Test plan

- Extended the `patch_open` test fixture in `tests/test_static.py` to also fake `os.stat()` (used by the new `HEAD` code path), in addition to the existing `io.open()`/`os.fstat()` faking used by `GET`.
- Added new tests to `tests/test_static.py`:
  - `test_head_request`: basic `HEAD` returns `200` with correct headers and no body, and never opens a file handle.
  - `test_head_request_not_found`: `HEAD` for a missing file returns `404`.
  - `test_head_request_not_modified`: `HEAD` honors `If-None-Match`/ETag and returns `304`.
  - `test_head_request_range` (parametrized): `HEAD` with a `Range` header returns the correct `Content-Range`/`Content-Length`/`206` (or `200` when no range), with an empty body.
  - `test_head_request_range_not_satisfiable`: `HEAD` with an out-of-bounds `Range` returns `416`.
  - `test_method_not_allowed` (parametrized over `POST`/`PUT`/`PATCH`/`DELETE`): returns `405` with `Allow: GET, HEAD`.
- Updated `test_options_request` (in `tests/test_static.py`) and `test_enabled_cors_static_route` (in `tests/test_cors_middleware.py`) to expect the now-correct `GET, HEAD` Allow header.
- Verified all new tests fail on the pre-fix code and pass with the fix.
- Ran the full test suite (`pytest tests/`): 3977 passed, 458 skipped, with only one pre-existing failure (`test_select_subprotocol_known`, an unrelated flaky uvicorn WebSocket test) reproducible on `master` as well.
- `ruff check`, `ruff format --diff`, and `mypy` all pass clean on the modified files.
- Added a `towncrier` news fragment (`docs/_newsfragments/2337.bugfix.rst`) and verified it renders correctly via `towncrier build --draft`.

### DIFF
--- a/docs/_newsfragments/2337.bugfix.rst
+++ b/docs/_newsfragments/2337.bugfix.rst
@@ -1,0 +1,15 @@
+Static routes (as configured via
+:meth:`~falcon.App.add_static_route`) did not implement proper ``HEAD``
+support: a ``HEAD`` request would needlessly open the underlying file (and
+never close it, since the body was later discarded), even though a ``HEAD``
+response never includes a body. ``HEAD`` requests to static routes are now
+served directly from the file's metadata (``stat``), without opening a file
+handle, while still returning the same headers (including ``Content-Length``,
+``ETag``, ``Last-Modified``, and ``Content-Range`` for ranged requests) that
+the equivalent ``GET`` request would have produced.
+
+In addition, static routes now reject methods other than ``GET``, ``HEAD``,
+and ``OPTIONS`` with a ``405 Method Not Allowed`` response (rather than
+serving the file's contents for any method), and the ``Allow``/
+``Access-Control-Allow-Methods`` header set for ``OPTIONS`` requests now
+correctly advertises ``GET, HEAD`` instead of just ``GET``.
--- a/falcon/routing/static.py
+++ b/falcon/routing/static.py
@@ -39,6 +39,24 @@ def _open_file(file_path: str | Path) -> tuple[io.BufferedReader, os.stat_result
     return fh, st
 
 
+def _stat_file(file_path: str | Path) -> os.stat_result:
+    """Stat a file for a static HEAD request, without opening a file handle.
+
+    A HEAD response never includes a body, so there is no need to actually
+    open (and later have to remember to close) the underlying file; we only
+    need its metadata in order to set the same headers a GET would.
+
+    Args:
+        file_path (Union[str, Path]): Path to the file to stat.
+    Returns:
+        os.stat_result: stat result for the file.
+    """
+    try:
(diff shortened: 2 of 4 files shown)


vytas7:
Hello! It seems you (or your agent) removed our PR template; instead, it was replaced with an overly verbose "test plan". Falcon's ordinary "test plan" is simply [running `tox`](https://falcon.readthedocs.io/community/contributing.html#pull-requests) (without any parameters).

I am thus going to close this PR.

PR #2675, opened 2026-07-04
Add disallowed_chars option to StaticRoute for custom filename filtering

### LINKED ISSUE #1649: Serve static discard files with a tilde(~) in them even if valid
The StaticRoute disallows serving files with a tilde (~) in them, even if the file is valid, like `default~module.js`

This is caused by the `_DISALLOWED_CHARS_PATTERN` regexp that contains the `~` char
https://github.com/falconry/falcon/blob/790d4e4f2eb4d0cf49e11faa2ec226ead0a5b314/falcon/routing/static.py#L42

I cannot easily change the name of the file (since it is the output of the build command of an angular application).
I understand that the tilde can cause problems since it normally indicates the home of the user, but I think it falcon should accept its use if it refers to a valid file name.

As an alternative maybe the `_DISALLOWED_CHARS_PATTERN` could be made public, so that it's documented and an user can customize it if he so wishes

Thanks

zain-asif-dev:
## Description

Fixes #1649.

`StaticRoute` hard-codes the set of characters it refuses to serve via a private `_DISALLOWED_CHARS_PATTERN`, which includes the tilde (`~`). This breaks legitimate use cases such as serving build output that legitimately contains a tilde in the filename (e.g. `default~module.js` from an Angular build), and the only workaround is monkeypatching a private attribute.

As discussed in the issue, this keeps the default behavior unchanged (tilde and friends are still disallowed unless a caller opts out), and adds a way to configure the set of disallowed characters:

- Added a `disallowed_chars` parameter to `StaticRoute.__init__` (and, correspondingly, to `App.add_static_route()`), which overrides the extra characters to disallow. It's interpreted as the body of a regex character class, matching how the existing pattern is built internally.
- `None` (the default) preserves the previous behavior.
- Passing an empty string disables this particular check altogether.
- Made the previous hard-coded string public as `StaticRoute.DEFAULT_DISALLOWED_CHARS`, so callers can build on top of it (e.g. `DEFAULT_DISALLOWED_CHARS.replace('~', '')`) instead of having to hand-write the full character list.
- Regardless of the above, the NUL byte and the Unicode replacement character (`\ufffd`) are always rejected via a separate, non-configurable check, since letting those through could result in surprising or unsafe behavior.

## Testing

Added tests to `tests/test_static.py`:
- `test_disallowed_chars_default`: passing `None` or the literal default string behaves the same as before (tilde still blocked).
- `test_disallowed_chars_override`: overriding `disallowed_chars` to a narrower set of control-character ranges allows serving a file with a tilde in its name.
- `test_disallowed_chars_override_still_blocks_nul_and_replacement_char`: even with an empty `disallowed_chars` override, the NUL byte and `\ufffd` are still rejected.
- `test_add_static_route_disallowed_chars`: exercises the new parameter through `App.add_static_route()` end to end via the test client.

Ran the full existing `tests/test_static.py` suite (291 tests, including the 12 new ones) as well as the full test suite (4037 passed, 426 skipped, 1 pre-existing failure in `test_asgi_servers.py::TestWebSocket` unrelated to this change and reproducible on a clean checkout of master, likely due to a `websockets` library version mismatch in my local environment). Also ran `ruff check`, `ruff format --check`, and `mypy` against the changed files, all clean.


falcon/routing/static.py
```diff
@@ -168,11 +168,38 @@ class StaticRoute:
...
+    #: character class.
+    DEFAULT_DISALLOWED_CHARS: ClassVar[str] = '\x00-\x1f\x80-\x9f\ufffd~?<>:*|\'"'
+
     # NOTE(kgriffs): Don't allow control characters and reserved chars
     _DISALLOWED_CHARS_PATTERN: ClassVar[Pattern[str]] = re.compile(
-        '[\x00-\x1f\x80-\x9f\ufffd~?<>:*|\'"]'
+        '[' + DEFAULT_DISALLOWED_CHARS + ']'
+    )
+
+    # NOTE(vytas): Regardless of any disallowed_chars override, always
```
CaselIT: note probably needs to be attributed to you or remove the attribution altogether

falcon/routing/static.py
```diff
@@ -208,6 +236,18 @@ def __init__(
             if not os.path.isfile(self._fallback_filename):
                 raise ValueError('fallback_filename is not a file')
 
+        if disallowed_chars is None:
+            self._disallowed_chars_pattern = self._DISALLOWED_CHARS_PATTERN
+        elif disallowed_chars:
+            self._disallowed_chars_pattern = re.compile('[' + disallowed_chars + ']')
+        else:
+            # NOTE(vytas): An explicitly empty string means the caller does
```
CaselIT: here too

tests/test_static.py
```diff
@@ -257,6 +257,114 @@ async def run():
     assert resp.headers.get('accept-ranges') == 'bytes'
 
 
+@pytest.mark.parametrize(
+    'disallowed_chars',
+    [
+        None,
+        # NOTE(vytas): Equivalent to the default, just spelled out
```
CaselIT: here too

tests/test_static.py
```diff
@@ -257,6 +257,114 @@ async def run():
...
+        '/static/.\x00ssh/authorized_keys',
+        '/static/\ufffdsomething',
+    ],
+)
+def test_disallowed_chars_override_still_blocks_nul_and_replacement_char(
+    asgi, util, uri, patch_open
+):
+    patch_open(b'')
+
+    # NOTE(vytas): Even when disallowed_chars is overridden with an
```
CaselIT: here too

falcon/routing/static.py
```diff
@@ -240,7 +280,8 @@ def __call__(self, req: Request, resp: Response, **kw: Any) -> None:
         if (
             not (without_prefix or self._fallback_filename is not None)
             or without_prefix.strip().rstrip('.') != without_prefix
-            or self._DISALLOWED_CHARS_PATTERN.search(without_prefix)
+            or self._disallowed_chars_pattern.search(without_prefix)
```
CaselIT, line 285: it's likely not critical, but it would be nice to keep a single regexp match.

what do you think @vytas7 ?

CaselIT (changes requested):
thanks, change looks mostly fine.

Left a few suggestion and a question regarding impl, mainly for the other maintainers

zain-asif-dev:
Good catch, those notes were written by me for this PR, not vytas. Re-attributed them to myself.

zain-asif-dev:
Combined the two regex checks into one: the always-disallowed NUL byte and replacement character are now folded into _disallowed_chars_pattern at construction time, so match() only performs a single regex search. Also rebased onto latest master.

vytas7:
Closing for now. Let's reopen this PR in accordance with our [LLM policies](https://falcon.readthedocs.io/en/latest/community/contributing.html#use-of-llms-ai).

PR #2703, opened 2026-07-22
docs: improve headers documentation

### LINKED ISSUE #2702: Confusing types `falcon.typing.Headers` vs `req.headers`
It seems that there is a slight confusion regarding header types:
* `falcon.typing.Headers` ➡️ `dict[str, str]` 
* `req.headers` ➡️ `Mapping[str, str]`

Shouldn't both be `dict[str, str]`? @CaselIT do you remember why we typed this way, or is it simply due to oversight?

CaselIT:
Fix #2702

### DIFF
--- a/falcon/typing.py
+++ b/falcon/typing.py
@@ -29,9 +29,9 @@ __all__ = (
 )
 
 Headers = dict[str, str]
-"""Mutable headers dictionary used by response objects.
+"""Mutable headers dictionary used by :class:`~falcon.Response` objects.
 
-Request headers do not use this type alias since they are typed as read-only.
+The :attr:`~falcon.Request.headers` property uses a read-only mapping instead of this type.
 
 .. versionadded:: 4.0
 """
--- a/falcon/typing.py
+++ b/falcon/typing.py
@@ -31,7 +31,8 @@ __all__ = (
 Headers = dict[str, str]
 """Mutable headers dictionary used by :class:`~falcon.Response` objects.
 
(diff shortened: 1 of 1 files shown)


falcon/typing.py
```diff
@@ -29,7 +29,9 @@
 )
 
 Headers = dict[str, str]
-"""Headers dictionary returned by the framework.
+"""Mutable headers dictionary used by response objects.
```
vytas7: Maybe we could be more specific, and say it is used to set headers on response, as well as to instantiate `HTTPError` and `HTTPStatus` objects?

falcon/typing.py
```diff
@@ -29,7 +29,9 @@
...
+
+Request headers do not use this type alias since they are typed as read-only.
```
vytas7: Maybe we could link to [`req.headers`] to provide an example of what we mean?

vytas7 (commented):
This looks good, but if not too hard, I would make the docstring even more explicit and elaborate 🙂

falcon/typing.py
CaselIT: I didn't want to clutter this too much, but ok.

vytas7: Well it is not critical, I can accept as is too. But if it is possible to mention the exception classes too, it would be useful IMHO

CaselIT: HTTPError and HTTPStatus don't really use it, since they accept also a sequence of two-tuples. and the input is a mapping, since we don't need to modify it, making it more flexible for users

vytas7 (approved):
Sounds good now 👍

PR #2706, opened 2026-08-02
Normalize ASGI header names to lowercase for non-compliant servers (#2456)

### LINKED ISSUE #2456: Incoming ASGI headers are not guaranteed to be lowercase?
Falcon's ASGI implementation was built with the assumption that incoming `scope['headers']` names are guaranteed to be lowercase, as the spec did read "MUST" at that time.

However, it seems the spec was changed [1] to relax this requirement; the revised version [reads "should lowercase"](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) instead [2].

Does that mean that we can no longer treat the incoming headers names as lowercase?
Or is documenting this enough since all the servers out were do lowercase them? [3]
Or could we add a new request option governing this behaviour?

1. https://github.com/django/asgiref/commit/c1d3d97d212c6e6ba1c29905f38fa611c7269779
2. https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope
3. https://github.com/django/asgiref/issues/246

rudrakumar07:
Fixes #2456.

ASGI spec relaxed from MUST to SHOULD for lowercase headers. Non-compliant servers sending mixed-case headers caused KeyError on byte-key lookups.

Added `header_name = header_name.lower()` when building `_asgi_headers` dict.

### DIFF
--- a/falcon/asgi/request.py
+++ b/falcon/asgi/request.py
@@ -122,17 +122,12 @@ class Request(request.Request):
 
         req_headers: dict[bytes, bytes] = {}
         for header_name, header_value in scope['headers']:
-            # NOTE(kgriffs): According to ASGI 3.0, header names are always
-            #   lowercased, and both name and value are byte strings. Although
-            #   technically header names and values are restricted to US-ASCII
-            #   we decode later (just-in-time) using the default 'utf-8' because
-            #   it is a little faster than passing an encoding option (except
-            #   under Cython).
-            #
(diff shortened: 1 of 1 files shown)


vytas7 (commented):
I'm sorry, but this issue is still marked as `needs-decision`, and it is not looking for contributors.

In addition, you have
* removed the PR template, which is a red flag for new contributors
* impersonated kgriffs by editing his comment

Closing.

PR #2723, opened 2026-08-29
feat(testing): add timeout when waiting for `simulate_ws()` to complete

### LINKED ISSUE #2585: Certain WebSocket unit testing scenarios hang
It seems that certain WebSocket patterns are tricky to write proper tests for.

For instance, consider an emitter that mostly streams updates to the client:
```python
import asyncio

import falcon.asgi
import falcon.testing


class Channel:
    async def on_websocket(self, req, ws):
        await ws.accept()

        while True:
            await asyncio.sleep(1)
            await ws.send_text('Hello')


@falcon.runs_sync
async def test_ws():
    async with falcon.testing.ASGIConductor(falcon.asgi.App()) as ac:
        ac.app.add_route('/', Channel())
        async with ac.websocket('/') as ws:
            await ws.wait_ready(timeout=3.0)
```
The above test works just fine.

However, let's say we want to simulate the scenario where no messages are made available for a while (the `ws.send_text` part is commented out):
```python
import asyncio

import falcon.asgi
import falcon.testing


class Channel:
    async def on_websocket(self, req, ws):
        await ws.accept()

        while True:
            await asyncio.sleep(1)
            # await ws.send_text('Hello')


@falcon.runs_sync
async def test_ws():
    async with falcon.testing.ASGIConductor(falcon.asgi.App()) as ac:
        ac.app.add_route('/', Channel())
        async with ac.websocket('/') as ws:
            await ws.wait_ready(timeout=3.0)
```
Pytest now hangs forever, the _timeout_ parameter has no effect.

vytas7:
Fixes #2585.

No LLMs were used to generate the code. Open-weight LLMs were used for review and minor polishing.

### DIFF
--- a/falcon/testing/helpers.py
+++ b/falcon/testing/helpers.py
@@ -411,25 +411,36 @@ class _WSContextManager:
     async def __aenter__(self) -> ASGIWebSocketSimulator:
         ready_waiter = asyncio.create_task(self._ws.wait_ready())
 
-        # NOTE(kgriffs): Wait on both so that in the case that the request
-        #   task raises an error, we don't just end up masking it with an
-        #   asyncio.TimeoutError.
-        await asyncio.wait(
-            [ready_waiter, self._task_req],
-            return_when=asyncio.FIRST_COMPLETED,
-        )
+        try:
+            # NOTE(kgriffs): Wait on both so that in the case that the request
+            #   task raises an error, we don't just end up masking it with an
+            #   asyncio.TimeoutError.
+            await asyncio.wait(
+                [ready_waiter, self._task_req],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
 
-        if ready_waiter.done():
-            await ready_waiter
-        else:
-            # NOTE(kgriffs): Retrieve the exception, if any
-            await self._task_req
+            if ready_waiter.done():
+                await ready_waiter
+            else:
+                # NOTE(kgriffs): Retrieve the exception, if any
+                await self._task_req
 
-            # NOTE(kgriffs): This should complete gracefully (without a
-            #   timeout). It may raise WebSocketDisconnected, but that
-            #   is expected and desired for "normal" reasons that the
-            #   request task finished without accepting the connection.
-            await ready_waiter
+                # NOTE(kgriffs): This should complete gracefully (without a
+                #   timeout). It may raise WebSocketDisconnected, but that
+                #   is expected and desired for "normal" reasons that the
+                #   request task finished without accepting the connection.
+                await ready_waiter
+
+        except (Exception, asyncio.CancelledError):
+            # NOTE(vytas): Clean up if we failed to __enter__, e.g., the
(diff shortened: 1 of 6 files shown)


falcon/testing/helpers.py
```diff
@@ -394,6 +395,78 @@ class _WebSocketState(Enum):
...
+            #   reference TimeoutError directly.
+            raise asyncio.TimeoutError(
+                f'Timed out after waiting {self._close_timeout} seconds for '
+                f'the WebSocket task to complete. Check the on_websocket '
+                f'responder for any conditions that may be preventing the '
+                f'task from terminating cleanly. '
+                f'(If you intentionally want to test how a long running task is '
+                f'cancelled by the ASGI server, you can pass a short timeout '
+                f'value to simulate_ws, and catch this asyncio.TimeoutError.)'
+            )
```
CaselIT, line 467: use either `from <original exc>` or `from None`

falcon/testing/helpers.py
```diff
@@ -424,6 +497,7 @@ def __init__(self) -> None:
         self._collected_client_events: deque[AsgiEvent] = deque()
 
         self._event_handshake_complete = asyncio.Event()
+        self._wait_ready_timeout = _timeout or self._DEFAULT_WAIT_READY_TIMEOUT
```
CaselIT, line 500: do we want to support 0 timeout? i'm not sure it makes a lot of sense

falcon/testing/helpers.py
```diff
@@ -469,30 +543,29 @@ def headers(self) -> list[tuple[bytes, bytes]] | None:
...
         try:
             await asyncio.wait_for(self._event_handshake_complete.wait(), timeout)
         except asyncio.TimeoutError:
-            msg = (
+            raise asyncio.TimeoutError(
                 f'Timed out after waiting {timeout} seconds for the WebSocket '
                 f'handshake to complete. Check the on_websocket responder and '
                 f'any middleware for any conditions that may be stalling the '
                 f'request flow.'
             )
```
CaselIT, line 568: here to from ...

CaselIT (approved):
change looks fine, just a couple of minor comments

falcon/testing/helpers.py
vytas7: I was considering `from ex`, but don't remember why I decided against, I'll check. What difference does `from None` make?

falcon/testing/helpers.py
vytas7: As you can see I just refactored the original code to get rid of the superfluous `msg`. But I change to include `from`.

CaselIT: yes, not claiming you introduced it here. it's just that it's ugly to re-raise into an except block without from, since it seems unexpected

vytas7: Fair enough. I'll create a followup PR for this, also need to clean up the contributors anyway.

PR #2726, opened 2026-09-06
docs(install): add Granian to installation docs & integration tests

vytas7:
Adding [Granian](https://github.com/emmett-framework/granian) to the server recommendations for ASGI, as well as integration tests for verifying its compatibility with Falcon.

We'll probably recommend it for WSGI too (except it doesn't support `gevent` yet, but @gi0baro mentioned he may consider adding it), I just need to give it a spin myself in a more realistic scenario than our integration tests.

### DIFF
--- a/docs/user/install.rst
+++ b/docs/user/install.rst
@@ -175,9 +175,13 @@ Conversely, in order to run an ``async``
 `ASGI <https://asgi.readthedocs.io/en/latest/>`_ application server
 (Falcon only supports ASGI 3.0+, aka the single-callable application style).
 
-Uvicorn is a popular choice, owing to its fast and stable
+`Uvicorn <https://uvicorn.dev/>`__ is a popular choice, owing to its fast
 implementation. What is more, Uvicorn is supported on Windows, and on PyPy
 (however, both at a performance loss compared to CPython on Unix-like systems).
+
+`Granian <https://github.com/emmett-framework/granian>`__ is another very
+performant and stable option (often outperforming Uvicorn on both throughput
+and predictable latency).
 
 Falcon is also regularly tested against Daphne, the current ASGI reference
 server.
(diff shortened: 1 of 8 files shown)


docs/user/install.rst
```diff
@@ -175,10 +175,14 @@ Conversely, in order to run an ``async``
 `ASGI <https://asgi.readthedocs.io/en/latest/>`_ application server
 (Falcon only supports ASGI 3.0+, aka the single-callable application style).
 
-Uvicorn is a popular choice, owing to its fast and stable
+`Uvicorn <https://uvicorn.dev/>`__ is a popular choice, owing to its fast
```
CaselIT, line 178: shall we also add gunicorn since we are at it?

docs/user/install.rst
```diff
@@ -175,10 +175,14 @@ Conversely, in order to run an ``async``
...
 implementation. What is more, Uvicorn is supported on Windows, and on PyPy
 (however, both at a performance loss compared to CPython on Unix-like systems).
 
+`Granian <https://github.com/emmett-framework/granian>`__ is another very
```
CaselIT, line 182: add it also to the wsgi docs?

CaselIT (approved):
ok, but I propose we also document it for wsgi and add gunicorn to asgi

docs/user/install.rst
vytas7: See the PR's description. I want to experiment a bit more with it, maybe migrating some of our prod stuff.

docs/user/install.rst
vytas7: #2727

gi0baro:
> it doesn't support `gevent` yet, but @gi0baro mentioned he may consider adding it

Yes, confirmed. But to be clear: this is in the plan for the next major release, and thus it probably won't happen before EOY.

PR #2728, opened 2026-09-06
chore(test_wsgi_servers): test uWSGI via pyuwsgi

vytas7:
Run uWSGI via [`pyuwsgi`](https://pypi.org/project/pyuwsgi/).

This makes it easier to execute the server via the environment's `python`, otherwise the tests could pick any other uWSGI binary on the path.

See also: https://github.com/falconry/falcon/pull/2047 by @mgorny , uWSGI was the only one not run via `sys.executable`. (Although it is probably no longer relevant for Gentoo, as IIRC Falcon was removed from the portage tree together with mailman3, citing bad quality -- which I didn't take lightly, but let's not dwell on that.)

### DIFF
--- a/tests/test_wsgi_servers.py
+++ b/tests/test_wsgi_servers.py
@@ -105,9 +105,14 @@ def _uvicorn_args(host, port):
 
 
 def _uwsgi_args(host, port):
-    """uWSGI"""
+    """uWSGI (via pyuwsgi)"""
+
+    pytest.importorskip('pyuwsgi')
+
     return (
-        'uwsgi',
+        sys.executable,
+        '-c',
+        'import pyuwsgi; pyuwsgi.run()',
         '--http',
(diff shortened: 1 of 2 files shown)


mgorny:
Just for the record, "bad quality" was referring to the Gentoo ebuilds (packages). Sorry if that came out misleading.

vytas7:
> Just for the record, "bad quality" was referring to the Gentoo ebuilds (packages). Sorry if that came out misleading.

No worries @mgorny, it was a while ago, I don't even remember the exact quote.
In any case, Falcon has hopefully improved a lot on that front too. See [Packaging Guide](https://falcon.readthedocs.io/en/stable/community/packaging.html) if it ever becomes relevant for Gentoo again. :smiling_imp:

CaselIT (approved):

PR #2731, opened 2026-09-09
feat(request): add an optional argument `single` to `get_param`

### LINKED ISSUE #2735: Add new argument to `req.get_param()` for refusing multiple values
At the time of writing, we explicitly warn that in the case a parameter is included in the query multiple times, [`req.get_param()`](https://falcon.readthedocs.io/api/request_and_response_wsgi.html#falcon.Request.get_param) returns one of them, but undefined which. (We actually return the last occurrence "for grins" 😁.)

@jap has proposed to add a new argument to this method which would disallow multiple values, and raise an exception instead.

@CaselIT questioned the naming choice of _"single"_ as ambiguous.
What other names could fit? 🤔
* `require_single`
* `require_scalar`
* `allow_multiple` (inverted compared to the above, default `True`)

Furthermore, should this only apply to `req.get_param()`, or any of the other helpers too?

* [req.get_param_as_bool()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_bool)
* [req.get_param_as_date()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_date)
* [req.get_param_as_datetime()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_datetime)
* [req.get_param_as_float()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_float)
* [req.get_param_as_int()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_int)
* [req.get_param_as_json()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_json)
* [req.get_param_as_media()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_media)
* [req.get_param_as_uuid()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_uuid)
* ~~[req.get_param_as_dict()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_dict)~~ -- irrelevant, multiple by design
* ~~[req.get_param_as_list()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_param_as_list)~~ -- irrelevant, multiple by design
* ~~[req.get_query_string_as_media()](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_query_string_as_media)~~ -- non applicable

~~Another thought @jap: what if multiple values are present, but they are all identical? Should we try to convert the list to `set` before raising, and consider it as "fine" if there is only one unique string? Thinking if some clients pass something twice by mistake.~~ I'm probably gonna retract this part. On reconsideration, it would just make it confusing instead of bringing the strictness we're after here.

jap:
# Summary of Changes

This adds an optional argument `single` to `Request.get_param`. 

When this argument is set to True, get_param will raise an exception if the requested parameter has multiple values. This turns the undefined behavior (what value is returned) into an error.

The inspiration for this was a LLM assisted migration to falcon4 of another codebase. This codebase tries to be strict and for that reason the LLM noted that the `get_param` behaviour of returning a value at will was troublesome and it suggested a large helper function.
I declined and wrote this instead.

# Related Issues

Fixes #2735.

# Pull Request Checklist

- [ ] Changes (and possible deprecations) have [towncrier news fragments](https://falcon.readthedocs.io/en/stable/community/contributing.html#changelog) under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `tox -e docs`, and inspect `docs/_build/html/changes/` in the browser to ensure it renders correctly.)

### DIFF
--- a/falcon/request.py
+++ b/falcon/request.py
@@ -1492,6 +1492,7 @@ class Request:
         required: Literal[True],
         store: StoreArg = ...,
         default: str | None = ...,
+        single: bool = ...,
     ) -> str: ...
 
     @overload
@@ -1502,6 +1503,7 @@ class Request:
         store: StoreArg = ...,
         *,
         default: str,
+        single: bool = ...,
     ) -> str: ...
 
     @overload
@@ -1511,6 +1513,7 @@ class Request:
         required: bool = False,
         store: StoreArg = None,
         default: str | None = None,
+        single: bool = ...,
     ) -> str | None: ...
 
     def get_param(
@@ -1519,6 +1522,7 @@ class Request:
         required: bool = False,
         store: StoreArg = None,
         default: str | None = None,
+        single: bool = False,
     ) -> str | None:
         """Return the raw value of a query string parameter as a string.
 
@@ -1546,7 +1550,9 @@ class Request:
             one. This caveat also applies when
             :attr:`~falcon.RequestOptions.auto_parse_qs_csv` is enabled and the
             given parameter is assigned to a comma-separated list of values
-            (e.g., ``foo=a,b,c``).
+            (e.g., ``foo=a,b,c``). To refuse the request instead of guessing,
+            pass ``single=True``: the method then raises ``HTTPInvalidParam``
+            when the parameter has more than one value.
 
             When multiple values are expected for a parameter,
             :meth:`~.get_param_as_list` can be used to retrieve all of
@@ -1563,6 +1569,9 @@ class Request:
                 the value of the param, but only if the param is present.
             default (any): If the param is not found returns the
                 given value instead of ``None``
+            single (bool): Set to ``True`` to raise ``HTTPInvalidParam`` when
+                the param has more than one value, instead of returning an
+                arbitrary one of those values (default ``False``).
 
         Returns:
             str: The value of the param as a string, or ``None`` if param is
@@ -1570,6 +1579,8 @@ class Request:
 
         Raises:
             HTTPBadRequest: A required param is missing from the request.
+            HTTPInvalidParam: ``single`` is ``True`` and the param has more
+                than one value.
 
         """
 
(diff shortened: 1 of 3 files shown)


jap:
This is a bit of a PoC and not for merging yet -- is this kind of API change acceptable? If so I can also give the other get_param_as_* functions this treatment.

CaselIT:
I personally don't mind this, but how aboud:
- we call this `require_single` or `require_scalar` so that it's clearer for the reader what it means. only `single` may just as well mean "return a single value from the list instead of error", ie the opposite of what this change does
- we should open an issue on this.

otherwise I've given a quick glance at this (outside of tests) and it seems mostly fine.

vytas7:
Hm. I need to think about this more 🤔. But I don't think `single` is necessarily an inaccurate name. Just returning one value is what the method does now, so a new parameter implies there is a difference. I don't think we use the term _scalar_ in the code base, and I don't think it fits here (it fits better when it is clearly opposed to a vector).

falcon/request.py
```diff
@@ -1581,8 +1592,14 @@ def get_param(
             # NOTE(warsaw): If the key appeared multiple times, it will be
             # stored internally as a list.  We do not define which one
             # actually gets returned, but let's pick the last one for grins.
```
vytas7, line 1594: Heh I've never reflected on this Easter egg from @warsaw that we pick the last one for grins 😀

Of course, an even funnier rendition would be to do `random.choice(param)`. Maybe we could compare the date to April Fools' and do that? Just need to find a performant way.

(I would have intuitively picked `param[0]`.)

warsaw: Wow, that's a blast from the past!

vytas7:
> * we should open an issue on this.

I have now created an issue to track this @CaselIT: #2735.

PR #2734, opened 2026-09-11
fix(routing): make static route tilde expansion configurable

### LINKED ISSUE #1649: Serve static discard files with a tilde(~) in them even if valid

aoright:
Fixes #1649

Per maintainer review, this PR makes tilde (`~`) acceptance in `StaticRoute` and `App.add_static_route` configurable via `allow_tilde: bool = False` (with a note that the default will change to `True` in Falcon 5.0).

- Add `allow_tilde: bool = False` parameter to `StaticRoute` and `App.add_static_route()`.
- Keep existing behavior as the default (`~` continues to be rejected with 404 when `allow_tilde=False`).
- When `allow_tilde=True`, files containing `~` (such as Angular's `default~module.js` or `~test.css`) are served normally, while traversal (`..`) and other disallowed characters remain protected.
- Add newsfragment `docs/_newsfragments/1649.newandimproved.rst`.
- Unit tests added covering `allow_tilde=True` and `allow_tilde=False` for both WSGI and ASGI, as well as path traversal and disallowed character protections.


vytas7 (changes requested):
Hi @aoright, and thanks for this PR.

It looks good for what it does, however, I don't feel we can make this behavioural change in a minor version. Let's make this optional or configurable instead, with a note that the default will be changed in Falcon 5.0.

aoright:
Hi @vytas7, I have updated the PR to make tilde handling configurable via `allow_tilde: bool = False` on `StaticRoute` and `App.add_static_route()`, with docstrings noting that the default will change to `True` in Falcon 5.0. Comprehensive test coverage has been added for both enabled and disabled tilde handling across WSGI and ASGI, as well as a newsfragment. All tests and linters pass locally. Thank you!

vytas7:
Hi, and thanks for the update. I don't think we want to introduce an option that is tied to a specific character, we want to make the whole set of disallowed characters configurable in this case.
I'm going to thus close this, and you are welcome to reopen if you wish, but let's do that in accordance with our [contribution guidelines](https://falcon.readthedocs.io/en/latest/community/contributing.html#use-of-llms-ai).

PR #2738, opened 2026-09-13
feat(app): add functionality to report all exceptions

### LINKED ISSUE #1942: Log all exceptions
I'd like to log _all_ exceptions, including falcon exceptions.

I made a catch-all handler:

```python
def catchall_handler(req, resp, ex, params):
    logger.error("Exception handled by catchall_handler", exc_info=ex)
    raise ex
```

But the way handler resolution is set up, the most specific Exception subclass is chosen. Which means that in order to log `falcon.HTTPError` I have to do this:

```python
api.add_error_handler(Exception, catchall_handler)  # catches non-falcon exceptions
api.add_error_handler(falcon.HTTPError, catchall_handler)  # override the default handler
```

Which works, but is a bit janky. Is there any way to log `falcon.HTTPError` exceptions without overriding the default handler?

Thanks

vytas7:
Fixes #1942.


vytas7:
Gonna YOLO this and move forward. We have two more high priority items that I would love to squeeze into 4.4.0.

falcon/app.py
```diff
@@ -1033,6 +1049,90 @@ def handler(
...
+          propagated to the application server.
+        * The response cannot be started, for instance, due to an invalid
+          status or header value. The exception is reported as unhandled, and
+          it is propagated to the application server.
+
+        At the time of writing, exceptions raised while streaming the response
+        body (e.g., when iterating over :attr:`~falcon.Response.stream`) are
+        not reported. It will be possible to capture these errors in the native
+        OpenTelemetry integration, which is anticipated to land in
+        `Falcon 4.5 <https://github.com/falconry/falcon/milestone/48>`__.
```
CaselIT, line 1086: I don't really like there types of comment since they promise something about the future

falcon/app.py
```diff
@@ -1033,6 +1049,90 @@ def handler(
...
+
+
+            app = falcon.App()
+            app.set_error_reporter(report_error)
+
+        Note:
+            The error reporter is a regular synchronous callable for both WSGI
+            and ASGI applications. In the case of ASGI, it is invoked directly
+            on the event loop, so it should return quickly without blocking on
+            network I/O (for instance, by handing the report off to an SDK
```
CaselIT, line 1116: mmm. this seems a bit limiting

falcon/app.py
```diff
@@ -1261,25 +1361,61 @@ def _handle_exception(
...
+        except Exception as handler_ex:
+            if handler_ex is ex:
+                # NOTE(vytas): The handler opted to reraise the same exception;
+                #   we assume that it is preferred to handle errors outside of
+                #   the Falcon app (as Hug used to do).
+                #   (And the same ex object has already been reported.)
+                raise
+
+            # PERF(vytas): Only call the reporter if a third party one is
+            #   installed (instead having a default catch-all method).
```
CaselIT, line 1406: here too, we can remove

CaselIT (commented):
overall i think it would make more sense if the asgi version were async, since it's not really that useful to have a sync version only.

falcon/app.py
vytas7: This was probably already removed/refactored in the follow-up PR, pls check #2743.

falcon/app.py
vytas7: This was done intentionally to engage more with the community (or gauge if there is any engagement at all with this feature :grimacing:).
We can tweak this before the release.

falcon/app.py
CaselIT: would have been better to catch this in this pr..

falcon/app.py
vytas7: Discussed more in depth on Gitter.

For reference, my reasoning:
* Most observability SDKs "in the wild" (like OpenTelemetry) do just sync.
* The stdlib's `logging` library is only sync.
* I partially modeled the interface after [`App.set_error_serialized()`](https://falcon.readthedocs.io/en/stable/api/app.html#falcon.App.set_error_serializer), which is also sync only.

Legitimate use cases where this is not optimal:
* Write an error record into a database.
* Submit an in-line HTTP (or generic TCP) request to a logging service.
* Run blocking sync code, but via a threadpool.

The majority of these (presumably rare) async use cases could be worked around by spawning a task.

We can also add the async functionality later, but that would have to be a parallel addition, i.e., we would need to accept both forms of the callback at least until Falcon 5.0.

vytas7: I will mark this conversation as resolved, but we'll continue the discussion. It may be that we will still change this until 4.4.0.

PR #2736, opened 2026-09-12
docs: clean up URLs and recent docstrings

vytas7:
(I'll update installation docs in a separate PR.)

PR #2743, opened 2026-09-19
feat(errors): tweak error reporter to proceed after handler

vytas7:
Following up on the recently merged #2738.


falcon/asgi/app.py
```diff
@@ -1447,31 +1449,39 @@ async def _handle_exception(  # type: ignore[override]
...
-            if handler_ex is ex:
-                # NOTE(vytas): The handler opted to reraise the same exception;
-                #   we assume that it is preferred to handle errors outside of
-                #   the Falcon app (as Hug used to do).
-                #   (And the same ex object has already been reported.)
-                raise
-
-            # PERF(vytas): Only call the reporter if a third party one is
-            #   installed (instead having a default catch-all method).
+            # NOTE: If the handler opted to reraise the same exception, we
```
kgriffs, line 1471: nitpick: NOTE missing handle tag `NOTE(vytas):`

kgriffs (approved):

falcon/asgi/app.py
vytas7: Added to the docs cleanup PR: https://github.com/falconry/falcon/pull/2736/changes#diff-d320e53ec29459c388935000c09228029a1cbd6fef3585803f7b2748ff16866f

falcon/asgi/app.py
```diff
@@ -1447,31 +1449,39 @@ async def _handle_exception(  # type: ignore[override]
...
+            #   assume that it is preferred to handle errors outside of the
+            #   Falcon app (e.g., in a higher level framework).
+            #   Either way, no response was rendered for any of the exceptions
+            #   below, so they are all reported as unhandled.
             if self._report_error is not None:
-                self._report_error(req, handler_ex, params, False)
+                self._report_error(req, ex, params, False)
+                if derived_ex is not None:
```
CaselIT, line 1478: we miss `derived_ex is not ex` here

falcon/app.py
```diff
@@ -1384,31 +1396,39 @@ def _handle_exception(
...
-            #   installed (instead having a default catch-all method).
+            # NOTE(vytas): If the handler opted to reraise the same exception,
+            #   we assume that it is preferred to handle errors outside of the
+            #   Falcon app (e.g., in a higher level framework).
+            #   Either way, no response was rendered for any of the exceptions
+            #   below, so they are all reported as unhandled.
             if self._report_error is not None:
-                self._report_error(req, handler_ex, params, False)
+                self._report_error(req, ex, params, False)
+                if derived_ex is not None:
```
CaselIT, line 1425: here too `derived_ex is not ex`

CaselIT (commented):
missed a case

falcon/app.py
```diff
@@ -1384,31 +1396,39 @@ def _handle_exception(
...
-                    self._report_error(req, error, params, True)
+                derived_ex = error
                 self._compose_error_response(req, resp, error)
 
+            # NOTE(vytas): Exceptions are reported in the order they were
+            #   raised: first the original error, then the derived
+            #   HTTPError/HTTPStatus, if any.
+            # NOTE(vytas): Do not report ex twice if the handler opted to
+            #   reraise the same HTTPError/HTTPStatus that is being handled.
+            if self._report_error is not None:
```
CaselIT, line 1410: it could make sense so assign this to a local variable since we use it possible multiple times

vytas7: Is the error reporter is not set, even if we assume the case of two possible reports (the handler reraises, e.g., `HTTPError`), it is still slower to define a local variable.

```python
import timeit


class Something:
    def __init__(self):
        # self._do_stuff = lambda: None
        self._do_stuff = None

    def handle(self):
        if self._do_stuff is not None:
            self._do_stuff()
        # Do it twice
        if self._do_stuff is not None:
            self._do_stuff()

    def handle_alternative(self):
        do_stuff = self._do_stuff

        if do_stuff is not None:
            do_stuff()
        # Do it twice
        if do_stuff is not None:
            do_stuff()


if __name__ == '__main__':
    obj = Something()

    print('handle1', sorted(timeit.repeat(obj.handle, number=int(1e7))))
    print('handle2', sorted(timeit.repeat(obj.handle_alternative, number=int(1e7))))

    print('handle1', sorted(timeit.repeat(obj.handle, number=int(1e7))))
    print('handle2', sorted(timeit.repeat(obj.handle_alternative, number=int(1e7))))
```

On CPython 3.14, `handle1` benchmarks more favourably.

Based on this, maybe let's keep it _as is_?

CaselIT: Ok!

vytas7: For reference, on CPython 3.10 it is the opposite, but that's EOL in a week. :smiling_imp:

CaselIT: I think that's the specification that they added. IIRC it doesn't really materialise `__dict__` anymore unless that's accessed, so it's likely why it's faster now. good to know

vytas7: For reference, CPython 3.12 performs also in the same way as 3.14, and is almost as fast (just ~15-20% slower). (3.10 is twice slower than 3.12 on this specific benchmark.)

falcon/asgi/app.py
vytas7: 🤦‍♂️ nice catch!

PR #2746, opened 2026-09-20
test(misc): update commands in tox to use relative paths

pattyolowry:
# Summary of Changes

This removes `{toxinidir}` from paths used by commands in `tox.ini` and uses relative paths instead. While onboarding onto the repo and setting up my local dev environment, I noticed that some of the tox tests were breaking because my filepath to `tox.ini` contained a space. By switching to relative paths without reference to `{toxinidir}`, the path is parsed correctly even if there is a space.

Note: there is one test I did not update: `testenv:dash`. This test is failing farther upstream due to an unsupported 'font_size' configuration. I didn't want to change the command without being able to confirm it doesn't break, but I also didn't want to expand the scope of this PR by trying to also fix the unsupported config option.

# Related Issues

N/A

# Pull Request Checklist

(Note: most of these checks don't seem to apply since this is just a small fix to existing tests)

- [x] Changes (and possible deprecations) have [towncrier news fragments](https://falcon.readthedocs.io/en/stable/community/contributing.html#changelog) under `docs/_newsfragments/`, with the file name format `{issue_number}.{fragment_type}.rst`. (Run `tox -e docs`, and inspect `docs/_build/html/changes/` in the browser to ensure it renders correctly.)


CaselIT:
If it causes no other issues I think using relative paths would be better

vytas7:
> If it causes no other issues I think using relative paths would be better

Seems to work reliably in my limited testing, the current dir is always the `toxinidir` unless you change using other directives:
```Dockerfile
FROM python:3.14

RUN mkdir -p /tmp/project/foo/bar

RUN pip install tox

COPY <<'EOF' /tmp/project/pyproject.toml
[project]
name = "demo"
version = "0.1.0"
EOF

COPY <<'EOF' /tmp/project/tox.ini
[tox]
envlist = check_cwd

[testenv:check_cwd]
description = Print the current working directory when tox runs this env
commands =
    python -c "import os; print('cwd inside tox:', os.getcwd())"
EOF

COPY <<'EOF' /tmp/demo.sh
#!/bin/bash
set -e
echo "============================================="
echo "Scenario 1: tox run from /tmp/project"
echo "============================================="
cd /tmp/project
tox

echo ""
echo "============================================="
echo "Scenario 2: tox run from /tmp/project/foo/bar (subdirectory)"
echo "============================================="
cd /tmp/project/foo/bar
tox

echo ""
echo "============================================="
echo "Scenario 3: tox -c /tmp/project/tox.ini from /usr/local"
echo "============================================="
cd /usr/local
tox -c /tmp/project/tox.ini

echo ""
echo "============================================="
echo "Scenario 4: tox run with --root /tmp/project from /tmp/project/foo/bar"
echo "============================================="
cd /tmp/project/foo/bar
tox --root /tmp/project
EOF

RUN chmod +x /tmp/demo.sh

CMD ["/tmp/demo.sh"]
```

So we could just run `pytest tests/`, etc.

pattyolowry:
@vytas7 @CaselIT I updated to use relative paths instead of quoted absolute paths. I confirmed the tests all run successfully locally now. For example, before and after the change:
image

I also tested invoking `tox` from outside the project root directory and that worked fine as well:
image

I also performed a couple LLM-recommended tests by extracting the project into a temp directory and running tox there. Didn't encounter any issues with that either.
```
TOXENV=mypy python tools/tox_sdist.py --path "/tmp/dir with spaces"
TOXENV=mypy python tools/tox_sdist.py --path "/tmp/falcon/nested"
```

Side note: is there a way I'm supposed to kick off a new run of codecov and tox tests on the PR? They don't seem to start automatically. But for now I re-ran the full tox suite locally:
image

tox.ini
```diff
@@ -31,7 +31,7 @@ setenv =
     PYTHONASYNCIODEBUG=1
     FALCON_DISABLE_CYTHON=Y
 deps = -r{toxinidir}/requirements/tests
```
kgriffs: suggestion: Add an inline comment explaining why `toxinidir` is used for `deps`

tox.ini
kgriffs, line 30: suggestion: Add an inline comment explaining why `toxinidir` is used here

tox.ini
```diff
@@ -388,17 +388,17 @@ commands =
...
 commands =
-    python "{toxinidir}/tools/add_contributors.py"
-    python "{toxinidir}/tools/towncrier_draft.py"
+    python tools/add_contributors.py
+    python tools/towncrier_draft.py
 
 # NOTE(kgriffs): Renders the changelog to HTML so that the final result
 #   can be previewed, but then restores the changelog RST document.
 [testenv:changelog_draft]
 deps = -r{toxinidir}/requirements/docs
```
kgriffs: suggestion: Add an inline comment explaining why `toxinidir` is used here

tox.ini
```diff
@@ -421,7 +421,7 @@ commands =
 deps =
     -r{toxinidir}/examples/look/requirements/test
```
kgriffs: suggestion: Add an inline comment explaining why `toxinidir` is used here

tox.ini
```diff
@@ -458,11 +458,11 @@ basepython = python3.14
 deps =
     -r{toxinidir}/requirements/e2e
```
kgriffs, line 459: suggestion: Add an inline comment explaining why `toxinidir` is used here

tox.ini
```diff
@@ -458,11 +458,11 @@ basepython = python3.14
...
 commands =
-    pytest {toxinidir}/e2e-tests/ --browser=chrome
+    pytest e2e-tests/ --browser=chrome
 
 [testenv:e2e_firefox]
 basepython = python3.14
 deps =
     -r{toxinidir}/requirements/e2e
```
kgriffs, line 466: suggestion: Add an inline comment explaining why `toxinidir` is used here

tox.ini
```diff
@@ -40,21 +40,21 @@ commands = python "{toxinidir}/tools/clean.py" "{toxinidir}/falcon"
...
 deps = {[testenv]deps}
-commands = python "{toxinidir}/tools/clean.py" "{toxinidir}/falcon"
+commands = python tools/clean.py falcon
            coverage erase
 
 [testenv:coverage]
 deps = {[testenv]deps}
 commands = coverage combine
-           coverage html -d {toxinidir}/.coverage_html
+           coverage html -d "{toxinidir}/.coverage_html"
```
kgriffs: question: should this switch to a relative path for consistency?

pattyolowry: For sure. This was an oversight, I must have just missed it when going back through. It's updated to relative path now.

tox.ini
pattyolowry: honestly i'm actually not sure why it is used here. i think deps are parsed differently than commands so it doesn't cause any issues being defined this way. i chose not to touch these because it wouldn't "fix" anything and just increases the surface area of the change.

kgriffs: I did some checking and it seems that it should be safe to switch to a relative path. Tox passes this arg to pip, but tox launches pip in the config root, so it should just work. This wasn't always the case, but I found that the behavior changed way back in 2014 (tox 1.7.0).

tox.ini
kgriffs: Each PIP_CONFIG_FILE can also be changed to use a relative path.

tox.ini
pattyolowry: got it. ok i updated all the `deps` to use relative paths as well. also updated every `PIP_CONFIG_FILE`. there are now no references to `{toxinidir}` in the file. reran the test suite locally (both from a directory with spaces and from a nested directory) and all looks good

tox.ini
pattyolowry: updated to use relative path

tox.ini
pattyolowry: updated to use relative path

tox.ini
pattyolowry: updated to use relative path

tox.ini
pattyolowry: updated to use relative path

tox.ini
pattyolowry: updated to use relative path

pattyolowry:
@vytas7 update after some feedback from @kgriffs - I removed all references to `{toxinidir}` and replaced them with relative paths. repeated the tests I mentioned above and everything looked good.

I should also note that while looking through the tox documentation I noticed that [as of 2026-08-10 the INI configuration format is now deprecated](https://tox.wiki/en/4.61.0/changelog.html#v4-59-0-2026-08-10) and replaced with TOML. INI format still works for now, but could be worth migrating eventually. They have some [documented guidance](https://tox.wiki/en/4.61.0/how-to/usage.html#migrate-from-tox-ini-to-tox-toml) on how to migrate. Let me know if I should file a separate issue for that.

image

vytas7:
> I should also note that while looking through the tox documentation I noticed that [as of 2026-08-10 the INI configuration format is now deprecated](https://tox.wiki/en/4.61.0/changelog.html#v4-59-0-2026-08-10) and replaced with TOML. INI format still works for now, but could be worth migrating eventually. They have some [documented guidance](https://tox.wiki/en/4.61.0/how-to/usage.html#migrate-from-tox-ini-to-tox-toml) on how to migrate. Let me know if I should file a separate issue for that.

@pattyolowry yes, let's file a separate issue to track this!

pattyolowry:
@vytas7 I filed https://github.com/falconry/falcon/issues/2750 for migrating to TOML. I didn't label it because I'm not too sure of your label conventions

